PR #21, opened 2022-02-25
Prevent non amd64 images during push

djfarrelly:
# Purpose
Prior to the upcoming cli work, it's good to prevent pushing non amd64 images that we cannot run.

### DIFF
--- a/inngest/deploy.go
+++ b/inngest/deploy.go
@@ -91,15 +91,15 @@ func prepareAndPushImage(ctx context.Context, a deployImageOptions) (err error) 
 	}
 
 	if a.image == "" {
-		return fmt.Errorf("no image specified")
+		return errors.New("no image specified")
 	}
 
 	resp, _, err := dkr.ImageInspectWithRaw(ctx, a.image)
 	if err != nil {
-		return fmt.Errorf("could not locate image")
+		return errors.New("could not locate image")
 	}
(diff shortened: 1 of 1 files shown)


tonyhb (approved):

inngest/deploy.go
```diff
@@ -94,6 +94,14 @@ func prepareAndPushImage(ctx context.Context, a deployImageOptions) (err error)
 		return fmt.Errorf("no image specified")
 	}
 
+	resp, _, err := dkr.ImageInspectWithRaw(ctx, a.image)
+	if err != nil {
+		return fmt.Errorf("could not locate image")
```
dramich, line 99:
```suggestion
		return errors.New("could not locate image")
```

inngest/deploy.go
```diff
@@ -91,7 +91,15 @@ func prepareAndPushImage(ctx context.Context, a deployImageOptions) (err error)
 	}
 
 	if a.image == "" {
-		return fmt.Errorf("no image specified")
+		return errors.New("no image specified")
+	}
+
+	resp, _, err := dkr.ImageInspectWithRaw(ctx, a.image)
+	if err != nil {
+		return errors.New("could not locate image")
```
dramich, line 99: This should return the actual error

dramich (approved):

PR #26, opened 2022-03-06
Add initial function file configuration logic as its own package

tonyhb:
This commit introduces a single file which is used to define serverless
functions.  The file is used to define a workflow and its triggers, and
contains extra information for use with the CLI as a higher-kind config:

- The event definition, including its type.  This allows the CLI to
  inspect the event type, generate test data, SDKs, etc.
- Sensible defaults.  We default to a single action within a workflow (a
  traditional serverless function,  not a step function) which is built
  via a "./Dockerfile" in the current working directory.

Implementation specifics:

Given the function contains workflow information and action information,
it derives the canonical `workflow.cue` and `action.cue` files necessary
to deploy the logic and run this in our executor.

It contains types which are a superset of those found in `./inngest/`.
These supersets within function contain additional information necessary
for development of functions.  Types within `/inngest` are used
specifically for deploying and running workflows.

https://user-images.githubusercontent.com/306177/157280186-1b2e5820-7ccf-45aa-a013-85520c9985be.mov


cmd/commands/init.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func NewCmdInit() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "init",
+		Short:   "Create a new serverless function",
+		Example: "inngestctl init",
+		Run:     runInit,
+	}
+	cmd.Flags().StringP("builder", "b", "docker", "Specify the builder to use. Options: docker or podman")
```
dramich, line 21: Not needed, I moved this to top level

PR #29, opened 2022-03-09
Fix: Support scopes defined in actions

djfarrelly:
## Description
Scopes were not defined, so when parsing a local action `cue` file, the scopes were dropped before creating the action version.

### DIFF
--- a/inngest/action.go
+++ b/inngest/action.go
@@ -54,6 +54,9 @@ type ActionVersion struct {
 	// actions run.
 	Response map[string]Response `json:"response"`
 
+	// Scopes defines the permissions that this action requires to execute.
+	Scopes []string `json:"scopes"`
+
 	// Edges define predetermined edges based off of responses for this action.  For example,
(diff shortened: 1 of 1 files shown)


inngest/action.go
```diff
@@ -54,6 +54,9 @@ type ActionVersion struct {
 	// actions run.
 	Response map[string]Response `json:"response"`
 
+	// Scopes defines the permissions that this action requires to execute.
+	Scopes []string `json:"scopes"`
```
dramich, line 58: This should mirror https://github.com/inngest/monorepo/blob/develop/pkg%2Fapplogic%2Factions%2Factions.go#L40

tonyhb (approved):

inngest/action.go
tonyhb: It does:  `Scopes` is an alias for []string, but with some internal functionality needed for the executor only.

PR #34, opened 2022-03-10
Render empty type with no event schemas in typescript.

tonyhb:
This fixes an invalid TS file, by rendering:

```js
export type EventTriggers = {};

export type Args = {
  event: EventTriggers;
  actions: {
    [clientID: string]: any
  };
};
```

Instead of the broken:

```js

export type Args = {
  event: EventTriggers; // undefined.
  actions: {
    [clientID: string]: any
  };
};
```


dramich (approved):

djfarrelly (approved):
LGTM

PR #50, opened 2022-04-14
Allow templating in next steps

tonyhb:
This passes the function ID, name, slug, and dir into postSetup for
templating, allowing us to show `cd into ./foo-bar-baz` post setup.


djfarrelly (commented):
🙌

PR #51, opened 2022-04-26
Fix nil pointer error if no workspace set

jpwilliams:
Following the first `inngest init` run (without logging in), I had a `~/.config/inngest/state` file that didn't contain a `workspace.id` key. This led to `cli.fetchEvents()` assuming a workspace existed and therefore throwing a nil pointer error on the second run of `inngest init`.

Is it sensible to catch this here or should `state.GetState()` be throwing an error in this case?

<details>
  <summary>Error</summary>

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe8e412]

goroutine 16 [running]:
github.com/inngest/inngestctl/pkg/cli.fetchEvents()
        /[redacted_path]/src/github.com/inngest/inngest-cli/pkg/cli/initialize_fn.go:533 +0xf2
github.com/inngest/inngestctl/pkg/cli.(*initModel).Init.func1()
        /[redacted_path]/src/github.com/inngest/inngest-cli/pkg/cli/initialize_fn.go:184 +0x26
github.com/charmbracelet/bubbletea.(*Program).StartReturningModel.func7.1()
        /[redacted_path]/src/github.com/inngest/inngest-cli/vendor/github.com/charmbracelet/bubbletea/tea.go:483 +0x51
created by github.com/charmbracelet/bubbletea.(*Program).StartReturningModel.func7
        /[redacted_path]/src/github.com/inngest/inngest-cli/vendor/github.com/charmbracelet/bubbletea/tea.go:481 +0x19b
exit status 2
```
</details>

<details>
  <summary>~/.config/state</summary>
  
```json
{
  "credentials": null,
  "account": {
    "id": "00000000-0000-0000-0000-000000000000",
    "name": "",
    "test": false,
    "identifier": {
      "dsnPrefix": "",
      "domain": null,
      "verifiedAt": null
    }
  },
  "settings": {
    "ranInit": true
  }
}
```
</details>

### DIFF
--- a/pkg/cli/initialize_fn.go
+++ b/pkg/cli/initialize_fn.go
@@ -525,7 +525,7 @@ func fetchEvents() ([]client.Event, error) {
 	defer done()
 
 	var workspaceID *uuid.UUID
-	if s, err := state.GetState(ctx); err == nil {
+	if s, err := state.GetState(ctx); err == nil && s.SelectedWorkspace != nil {
 		workspaceID = &s.SelectedWorkspace.ID
 	}
 


tonyhb (approved):
> Is it sensible to catch this here or should state.GetState() be throwing an error in this case?

This is probably the preferred way of doing things _for now_.

As in, a root fix involves tidying up how the CLI manages "workspaces".  Some history and context around a workspace:

## What are workspaces and why do they exist?

A workspace is basically a sub-account.  When you sign up to Inngest, we create an account — and also two workspaces.  Each workspace has its own source keys, its own functions, its own historical events, etc:  we organize things in account -> workspace -> (data) hierarchies.

This lets us do test environments nicely:  we have one _test_ workspace and one _prod_ workspace.

You can then do things like promote from test -> prod, etc.

## Workspace state refactor

We probably don't want to be supporting > 2 workspaces for now.  We only want "prod/live" and "test".  If that's the case, we don't need to store the "selected" workspace in state at all.  In fact, it kind of sucks that it's stateful.

A larger refactor would involve removing the selected workspace overall, then defaulting to test (or live) for all commands, forcing the workspace with a (--env-test) or (--env-prod) flag.

Then, happy days, rejoice!  No more stateful workspaces :D

## Wrapping up

State shouldn't be stateful.  At least when it comes to workspaces! 😄 This is something we haven't yet implemented and you've hit on it, so thank you for looking at it!

And, thank you for the PR.  It's very much appreciated! 🎉

PR #58, opened 2022-05-06
Use the workspace environment in published HTTP actions

tonyhb:
This ensures that actions are independent between test and prod;
publishing new test steps doesn't automatically update anything.

### DIFF
--- a/pkg/function/function.go
+++ b/pkg/function/function.go
@@ -193,7 +193,12 @@ func (f Function) Actions(ctx context.Context) ([]inngest.ActionVersion, []innge
 }
 
 func (f Function) action(ctx context.Context, s Step, n int) (inngest.ActionVersion, error) {
-	id := fmt.Sprintf("%s-step-%d", f.ID, n)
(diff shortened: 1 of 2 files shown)


pkg/function/function.go
```diff
@@ -193,7 +193,12 @@ func (f Function) Actions(ctx context.Context) ([]inngest.ActionVersion, []innge
 }
 
 func (f Function) action(ctx context.Context, s Step, n int) (inngest.ActionVersion, error) {
-	id := fmt.Sprintf("%s-step-%d", f.ID, n)
+	suffix := "-test"
+	if state.IsProd() {
+		suffix = "-prod"
+	}
+
+	id := fmt.Sprintf("%s-step-%d-%s", f.ID, n, suffix)
```
djfarrelly, line 201: This looks like the suffix already has `-` so this would be `<id>-step-2--prod` ?

tonyhb: Thanks for catching!  Fixed.

PR #65, opened 2022-05-18
Update dev, run, build for multi-step and multi-fn build

tonyhb:


cmd/commands/deploy.go
```diff
@@ -109,10 +109,12 @@ func deployAction(ctx context.Context, a inngest.ActionVersion) error {
 		// an up-to-date image and checksum for the action.
 		ui, err := cli.NewBuilder(ctx, cli.BuilderUIOpts{
 			QuitOnComplete: true,
-			BuildOpts: dockerdriver.BuildOpts{
-				Path:     ".",
-				Tag:      tag,
-				Platform: "linux/amd64",
+			BuildOpts: []dockerdriver.BuildOpts{
+				dockerdriver.BuildOpts{
+					Path:     ".",
```
djfarrelly, line 114: Not needed for this PR, but when we update the deploy cmd for multi-function deploys or deploying a specific directory, we could pass the path through to this :)

pkg/engine/engine.go
```diff
@@ -40,39 +39,36 @@ func (eng *Engine) Load(ctx context.Context, dir string) error {
...
-	a, _, _ := fn.Actions(ctx)
-	if a[0].Runtime.RuntimeType() != inngest.RuntimeTypeDocker {
-		return nil
+// buildImages builds all images hosted within the engine.  This iterates through all
+// functions discovered during Load.
+func (eng Engine) buildImages(ctx context.Context) error {
+	opts := []dockerdriver.BuildOpts{}
+
+	for _, fn := range eng.Functions {
+		steps, err := dockerdriver.FnBuildOpts(ctx, *fn)
```
djfarrelly, line 59: Much nicer!

pkg/engine/engine.go
```diff
@@ -82,7 +78,8 @@ func (eng Engine) BuildImages(ctx context.Context, fn *function.Function) error
 		fmt.Println("\n" + cli.RenderError(err.Error()) + "\n")
 		os.Exit(1)
 	}
-	return ui.Builder.Error()
+
+	return ui.Error()
```
djfarrelly, line 82: Should we output to the logger that the functions were loaded/built successfully? The TUI seems to show "Exporting image..." as the last message, but some sort of final confirmation could be useful to tell the user it was successful

djfarrelly (approved):

PR #82, opened 2022-05-27
This changes the expression data gen type to use inngest.Edge directly

tonyhb:
We do not need the GraphEdge type;  this makes the function harder to
use and redefine.


djfarrelly (approved):
Looks like a good change to me

Issue #120, opened 2022-06-13
Plan:  Open-source the queue interface for the runner


tonyhb:
In the Inngest cloud, we use our own queueing mechanism to ensure that edges within a function are traversed correctly and that the steps are executed.

Locally, we have a wrapper to the `State` store which provides a basic `Queue` interface.  We need to:

- Create a generic interface for managing the queue of scheduled work
- Which allows for arbitrary implementations (custom implementations, running on Celery)

It must account for:

- Scheduled work (eg. run this particular step at this time, which also includes retries)
- Asynchronous timeouts (eg. wait for an event for up to 10 minutes, and on timeout run this edge).

This is necessary for providing self-hosted solutions.

tonyhb:
https://github.com/inngest/inngest-cli/pull/135 adds a basic queue interface to our project, which can be implemented within SQS, Celery, Faktory, etc.

We'll add some implementations for reference and self hosting :)

djfarrelly:
Completed in #150

PR #121, opened 2022-06-13
[WIP] Redis state store

tonyhb:
This PR is a work-in-progress implementation of a Redis state store.  This Redis state store is usable for all function state;  it does not currently implement queues (see #120).


pkg/execution/state/inmemory/inmemory.go
```diff
@@ -186,67 +207,18 @@ func (m *mem) ConsumePause(ctx context.Context, id uuid.UUID) error {
...
-	// errors stores a map of action errors
-	errors map[string]error
+	return copied
 }
 
-func (s memstate) Identifier() state.Identifier {
-	return state.Identifier{
-		WorkflowID: s.workflowID,
-		RunID:      s.runID,
+func copyMap[K comparable, V any](m map[K]V) map[K]V {
```
djfarrelly, line 218: Generics! 😁

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -0,0 +1,349 @@
...
+		kf: defaultKeyFunc{},
+	}
+
+	for _, opt := range opts {
+		opt(m)
+	}
+
+	if m.r == nil {
+		m.r = redis.NewClient(&redis.Options{
+			Addr:     "localhost:6379",
```
djfarrelly, line 43: We should decide on some standard env variables for this, like `<package>_<var>` and a standard pattern to have defaults either through viper or similar. I could see this being `REDIS_STATE_ADDR` or `STATE_REDIS_ADDR`. The package could be useful as one might use Redis as a queue as well which could be the same instance of Redis or different.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -0,0 +1,349 @@
...
+	r  *redis.Client
+}
+
+func (m mgr) New(ctx context.Context, workflow inngest.Workflow, runID ulid.ULID, input map[string]any) (state.State, error) {
+	id := state.Identifier{
+		WorkflowID: workflow.UUID,
+		RunID:      runID,
+	}
+
+	// TODO: We could probably optimize the commands here by storing the event
```
djfarrelly, line 105: I think this is a good idea for a future optimization

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -0,0 +1,349 @@
...
+	inputJSON, err := json.Marshal(input)
+	if err != nil {
+		return nil, err
+	}
+
+	err = m.r.Watch(ctx, func(tx *redis.Tx) error {
+
+		// Ensure that the workflow exists within the state store.
+		key := m.kf.Workflow(ctx, workflow.UUID, workflow.Version)
+		val, err := tx.Exists(ctx, key).Uint64()
```
djfarrelly, line 125: Could this use [`SETNX`](https://redis.io/commands/setnx/) to combine these steps or is it more performant not to have to marshal the workflow every new run?

pkg/execution/state/state.go
```diff
@@ -95,15 +97,27 @@ type Loader interface {
...
+	New(ctx context.Context, workflow inngest.Workflow, runID ulid.ULID, input map[string]any) (State, error)
 
 	// SaveActionOutput stores output for a single action within a workflow run.
+	//
+	// This should clear any error that exists for the current action, indicating that
+	// the step is a success.
 	SaveActionOutput(ctx context.Context, i Identifier, actionID string, data map[string]interface{}) (State, error)
 
 	// SaveActionError stores an error for a single action within a workflow run.  This is
 	// considered final, as in the action will not be retried.
```
djfarrelly, line 109: This line saying that it's considered final and the action won't be retried seems like it is contrary to the other comments about overwriting the error on a successful action or passing `nil` to this method. Maybe its just a comment we need to update?

djfarrelly (approved):
I did a pretty deep review so far through the Redis state and it all looks solid, I had some thoughts, but invalidated thoughts during the review so I only have a few smaller comments here.

pkg/execution/state/state.go
```diff
@@ -130,7 +143,19 @@ type PauseMutater interface {
 	// leasing or locking mechanism.
```
djfarrelly, line 143: The above comment mentions
> This basic interface does not provide a mechanism for "leasing" a pause temporarily whilst a step is scheduled

But now we have a method to allow it to implemented, should we adjust that comment? I may be missing something

pkg/execution/state/inmemory/inmemory.go
```diff
@@ -133,8 +145,17 @@ func (m *mem) SaveActionOutput(ctx context.Context, i state.Identifier, actionID
...
-	state := s.(*memstate)
-	state.errors[actionID] = err
+	state := s.(memstate)
+
+	// Copy the maps so that any previous state references aren't updated.
+	state.actions = copyMap(state.actions)
+	state.errors = copyMap(state.errors)
+
+	if err == nil {
+		delete(state.errors, actionID)
```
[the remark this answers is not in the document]
tonyhb: Correct.  We'll change the state implementation after to store our own data structure which has a list of errors!

pkg/execution/state/inmemory/inmemory.go
tonyhb: finally! 😂

pkg/execution/state/redis_state/redis_state.go
tonyhb: Yeah, agree.  We need to document how to initialize a new self-hosted server with config stored in eg. `/etc/inngest/config.json` or via a --config path.

tonyhb: https://github.com/inngest/inngest-cli/issues/124

pkg/execution/state/inmemory/inmemory.go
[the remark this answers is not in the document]
tonyhb: https://github.com/inngest/inngest-cli/issues/125

PR #130, opened 2022-06-17
Add ability to specify hostname to bind to when starting dev server

### LINKED ISSUE #113: Allow the dev server to bind to custom IPs

jpwilliams:
Aiming to close #113.

- Adds a new `hostname` flag to `inngest dev`
- Passes through hostname to `http.ListenAndServe()`

### DIFF
--- a/cmd/commands/dev.go
+++ b/cmd/commands/dev.go
@@ -17,6 +17,7 @@ func NewCmdDev() *cobra.Command {
 		Run:     doDev,
 	}
 
+	cmd.Flags().String("hostname", "", "hostname to run the API on")
 	cmd.Flags().StringP("port", "p", "9999", "port to run the API on")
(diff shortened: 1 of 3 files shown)


tonyhb:
Awesome, thanks for the PR!  Looks good!

For brevity, I might make a followup which changes this to `host` for people (with a `-h` flag).  My thought here is fewer characters == easier.

jpwilliams:
> For brevity, I might make a followup which changes this to `host` for people (with a `-h` flag). My thought here is fewer characters == easier.

@tonyhb I can make the change 😃 I did give it the `-h` short flag initially but realised folks usually use that as a shortcut for `--help`. 😄

tonyhb:
Aha good call on the `-h` flag.  We should probably not overwrite that!

PR #134, opened 2022-06-27
Refactor state.Manager interface, add generic Queue interface

tonyhb:
We refactor driver.Response into state.DriverResponse, and move
to single StoreResult within state.Manager interface.

This allows us to unify storing output and errors into one call.
Because the state.DriverResult struct dictates whether the result is
Final(), we also store an incremental counter specifying the number
of function steps that have been scheduled and have been finalized.

These counters lets us calculate whether a function has finished.

This works more like a waitgroup:  to wait for a function to finish we
can wait for the pending count to reach 0.


pkg/execution/state/driver_response.go
```diff
@@ -0,0 +1,119 @@
...
+	status, ok := r.Output["status"]
+	if !ok {
+		// If actions don't return a status, we assume that they're
+		// always retryable.  We prefer that actions respond with a
+		// { "status": xxx, "body": ... } format to disable retries.
+		return true
+	}
+	switch v := status.(type) {
+	case float64:
+		if int(v) >= 499 {
```
djfarrelly, line 74: Do we want `>=500`? I honestly wasn't familiar with `499`'s until now 😁

PR #135, opened 2022-06-27
Generic queue interface

tonyhb:
This commit introduces a basic queue interface.  The interface allows
us to enqueue steps to run at a specific time, and to provide a function
which should run when items are dequeued.

We change the in-memory queue to use this interface, and will provide
other implementations of the interface such as celery, SQS, etc. in a
future PR.

### DIFF
--- a/pkg/execution/queue/queue.go
+++ b/pkg/execution/queue/queue.go
@@ -1,0 +1,25 @@
+package queue
+
+import (
+	"context"
+	"time"
+)
+
+type Queue interface {
+	Producer
+	Consumer
+}
+
+type Producer interface {
+	// Enqueue allows an item to be enqueued ton run at the given time.
+	Enqueue(context.Context, Item, time.Time) error
+}
+
+type Consumer interface {
+	// Run is a blocking function which listens to the queue and executes the
+	// given function each time a new Item becomes available.
+	//
+	// This must only return an error if we fail to subscribe to the queue's
+	// implementation and can no longer process jobs.
+	Run(context.Context, func(context.Context, Item) error) error
+}
--- a/pkg/execution/state/inmemory/inmemory.go
+++ b/pkg/execution/state/inmemory/inmemory.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/inngest/inngest-cli/inngest"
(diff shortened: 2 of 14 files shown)


pkg/execution/queue/queue.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+	"time"
+)
+
+type Queue interface {
+	Enqueuer
+	Worker
+}
+
+type Enqueuer interface {
```
djfarrelly, line 13: Should we call these `Producer` and `Consumer` instead of `Enqueuer` and `Worker` maybe to use more standard queue terms. For code readability I think it would be useful. `Enqueue` could also become `Send`.

pkg/execution/queue/queue.go
```diff
@@ -0,0 +1,25 @@
...
+	Enqueue(context.Context, Item, time.Time) error
+}
+
+type Worker interface {
+	// Run is a blocking function which listens to the queue and executes the
+	// given function each time a new Item becomes available.
+	//
+	// This must only return an error if we fail to subscribe to the queue's
+	// implementation and can no longer process jobs.
+	Run(context.Context, func(context.Context, Item) error) error
```
djfarrelly, line 24: Would other implementations, e.g. RabbitMQ, extend this interface with functions like `Receive`/`Fetch` and `Poll` that passes the data to `Run`? I'm trying to imagine how one might write another implementation.

pkg/execution/state/inmemory/inmemory.go
```diff
@@ -50,20 +45,47 @@ type mem struct {
 
 	lock *sync.RWMutex
 
-	q chan QueueItem
+	q chan queue.Item
 }
 
-func (m *mem) Enqueue(item QueueItem, at time.Time) {
+func (m *mem) Enqueue(ctx context.Context, item queue.Item, at time.Time) error {
```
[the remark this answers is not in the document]
tonyhb: That's a fair point.  In a follow-up PR we should separate the in-memory state store from the queue.

pkg/execution/queue/queue.go
tonyhb: Done - thanks!

PR #145, opened 2022-07-05
Add Server package for long running servers

tonyhb:
This abstracts pre/start/stop across all components of the self-hosted
runner, ie. the API, the events API, and the executor.

We provide the ability for each component to specify its own pre-up
routines which, if errored, prevent the server from running.  This
allows us to eg. load secrets from each service prior to running.

We also handle interrupts and graceful shutdowns of the server.


pkg/api/server.go
```diff
@@ -0,0 +1,43 @@
...
+func (a *apiServer) Name() string {
+	return "api"
+}
+
+func (a *apiServer) Pre(ctx context.Context) error {
+	var err error
+
+	// TODO: Conenct and load ingest keys.
+	a.api, err = NewAPI(Options{
+		Hostname: "localhost",
```
tonyhb, line 26: This is only a demo prior to adding configuration via Cue;  I'll start work on the config package now

PR #147, opened 2022-07-05
Add event flag to specify trigger in multi-trigger functions

djfarrelly:
### Description

Useful for if you have a multi-trigger function and you would like to test a particular function.

Otherwise, the CLI randomly chooses a trigger from the list of triggers.

### DIFF
--- a/cmd/commands/run.go
+++ b/cmd/commands/run.go
@@ -29,6 +29,7 @@ func NewCmdRun() *cobra.Command {
 		Run:     doRun,
 	}
 
+	cmd.Flags().String("event", "", "Specifies the event trigger to use if there are multiple configured")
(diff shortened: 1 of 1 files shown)


cmd/commands/run.go
```diff
@@ -29,6 +29,7 @@ func NewCmdRun() *cobra.Command {
 		Run:     doRun,
 	}
 
+	cmd.Flags().String("event", "", "Specifies the event trigger to use if there are multiple configured")
```
djfarrelly, line 32: Could use `StringP` for a `-e` arg shorthand. I used `event` instead of `trigger` b/c we use `event` for `inngest init`

tonyhb (approved):

tonyhb (approved):

PR #149, opened 2022-07-07
Add idempotency keys to state.Identifier

tonyhb:
This allows us to atomically create workflows when calling New(),
failing if the given idempotency key already exists.

### DIFF
--- a/pkg/execution/state/state.go
+++ b/pkg/execution/state/state.go
@@ -34,7 +34,7 @@ type Identifier struct {
 	RunID      ulid.ULID `json:"runID"`
 	// Key represents a unique idempotency key used to deduplicate this
 	// workflow run amongst other runs for the same workflow.
-	Key string `json:"key:"`
+	Key string `json:"key"`
 }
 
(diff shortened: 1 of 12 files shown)


pkg/execution/state/state.go
```diff
@@ -31,6 +32,17 @@ const (
 type Identifier struct {
 	WorkflowID uuid.UUID `json:"workflowID"`
 	RunID      ulid.ULID `json:"runID"`
+	// Key represents a unique idempotency key used to deduplicate this
+	// workflow run amongst other runs for the same workflow.
+	Key string `json:"key:"`
```
djfarrelly, line 37: Is it correct that `json:"key:"` should contain the semi-colon? Just looked like a potential typo

djfarrelly (approved):
Really like this change w/ `IdempotencyKey` - looks great! Just one comment

pkg/execution/state/state.go
tonyhb: Def a typo — appreciate it!

PR #153, opened 2022-07-11
Create CoreAPI service

djfarrelly:
## Description

A new "core-api" service (`coreapi` package) which acts as a way to manage functions, actions and read configuration of the Inngest system.

### GraphQL

The new GraphQL server implements the following mutations that will be used by the `deploy` command:

```graphql
deployFunction(input: DeployFunctionInput!): FunctionVersion
createActionVersion(input: CreateActionVersionInput!): ActionVersion
updateActionVersion(input: UpdateActionVersionInput!): ActionVersion
```

### Data loaders

There is a new set of API Data Loaders for functions and actions that support writing to the persisted state of the system (`coredata`). These are stubbed out currently for this PR just to ensure the rest of the code works, but the Core API will never share memory with the Executor or Runner services that require it, so it must use a persisted data store like Postgres.

## Next steps

Beyond this PR the next areas we will need to implement:

1. Add Core API configuration to the new config package so more can be configured about the API
2. Create an implementation of the loaders for a persisted data store (PostgreSQL)
3. Update the config query resolver to pull real configuration for the CLI to use
4. Refactor the CLI to use the new endpoints
5. Upstream the adjusted resolvers and mutations to Inngest Cloud


pkg/coreapi/graph/resolvers/config.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	model "github.com/inngest/inngest-cli/pkg/coreapi/graph/models"
+)
+
+func (r *queryResolver) Config(ctx context.Context) (*model.Config, error) {
+	dockerRegistry := "registry.inngest.com"
+	dockerNamespace := "inngest" // TODO - Update w/ account DSN
+
+	config := &model.Config{
```
djfarrelly, line 13: This should read from the `config.go` package, not static values.

pkg/config/config_test.go
```diff
@@ -21,6 +21,10 @@ func defaultConfig() *Config {
 			Port:    8288,
 			MaxSize: 524288,
 		},
+		CoreAPI: CoreAPI{
+			Addr: "0.0.0.0",
+			Port: 8300,
```
djfarrelly, line 26: Any thoughts on a port? This was a random pick.

pkg/coreapi/graph/resolvers/functions.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	fv, err := r.APILoader.CreateFunctionVersion(ctx, *f, *input.Live)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO convert function to cue config
```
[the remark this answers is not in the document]
djfarrelly: That would be great and be useful in the CLI deploy function cmd. Maybe a `Function.MarshalCue` method that could be called like `fv.Function.MarshalCue()`? I may be influenced by the JavaScript patterns of `<object>.toJSON()` and similar.

I also wonder in this method if we should pass the string to the `CreateFunctionVersion` function as the string should be what is saved to the database in most cases - I would like to take guesswork out of implementing drivers (i.e. what data to save or not),

PR #165, opened 2022-07-14
Add PostgreSQL DataStore

djfarrelly:
## Description

To implement a new PostgreSQL DataStore for use by the coredata package and therefore the Executor and Core API. 

### Naming
I was going to use "CoreData" or just "Data" but "Data" was strange as a plural in different places as it's not a noun (e.g. `registeredDataStores`). I thought "data store" was a solid generic term as it could mean in-memory, a file system, a blob store, or a traditional database and not just a database only.

If this feels good, it may make sense to rename `coredata` package `datastore`. 

### Documentation

A README has been added [the in package directory](https://github.com/inngest/inngest/tree/feature/postgres-datastore/pkg/coredata/postgres).

### Postgres Tests
Using [embedded-postgres](https://github.com/fergusstrange/embedded-postgres/) works on CI, but it doesn't work locally on M1/Mx Macs, potentially due to [this issue](https://github.com/zonkyio/embedded-postgres-binaries/issues/49). When running tests locally, Postgres must be run in a container.

### TODO

* [x] Create Postgres package config
* [x] Add function to convert `Function` to formatted cue string and use in `CreateFunctionVersion` (or in GraphQL resolver 🤔) 
  * @tonyhb did this in #170 - One it is merged, we can rebase this branch onto main and use that function instead
* [x] Implement all read and write functions for the Executor and API
* [x] Document Postgres DataStore w/ README w/ basic config and requirements
* [x] Document how to run migrations


pkg/coredata/postgres/migrations/0001_init.up.sql
```diff
@@ -0,0 +1,58 @@
...
+CREATE TABLE function_triggers (
+  function_id character varying(255) NOT NULL,
+  event_name character varying(255),
+  schedule character varying(50),
+  -- only 1 schedule trigger may exist for a given function. those will have blank event_names
+  PRIMARY KEY (function_id, event_name)
+);
+
+ALTER TABLE ONLY function_triggers
+  ADD CONSTRAINT function_triggers_function_id FOREIGN KEY (function_id) REFERENCES functions(function_id) ON DELETE CASCADE;
```
djfarrelly, line 38: We probably need an multi-key index for `function_id` + `schedule` - What do you think?

pkg/coredata/postgres/postgres.go
```diff
@@ -0,0 +1,190 @@
...
+			_, err := tx.ExecContext(ctx, sqlUpdateFunctionVersionValidTo, f.ID, existingVersion, now)
+			if err != nil {
+				return function.FunctionVersion{}, err
+			}
+		}
+
+	}
+
+	// Create the function version
+	config, err := function.MarshalJSON(f)
```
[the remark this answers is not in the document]
djfarrelly: TODO - Use #170 after it's merged.

pkg/coredata/postgres/migrations/0001_init.up.sql
```diff
@@ -0,0 +1,62 @@
...
+  -- valid date range marks when the version is "live"
+  valid_from timestamp without time zone,
+  valid_to timestamp without time zone,
+  created_at timestamp without time zone DEFAULT now() NOT NULL,
+  updated_at timestamp without time zone DEFAULT now() NOT NULL,
+  PRIMARY KEY (function_id, version)
+);
+
+ALTER TABLE ONLY function_versions
+  ADD CONSTRAINT function_versions_function_id FOREIGN KEY (function_id) REFERENCES functions(function_id) ON DELETE CASCADE;
```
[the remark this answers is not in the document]
djfarrelly: Added in e85723bd6ebb727f28785dc3876c7a3ab955b53e

pkg/coredata/postgres/0001_init.down.sql
```diff
@@ -0,0 +1,4 @@
+DROP TABLE action_versions;
```
tonyhb, line 1: I think this file needs to be dropped into the migrations dir?

tonyhb: Hmm, it looks like goose separates things by up && down, not by files.  Maybe this can be removed?

pkg/cuedefs/config/config.cue
```diff
@@ -162,6 +167,22 @@ package config
 	keyPrefix: string | *"inngest:state"
 }
 
+// # DataStore
+//
+// DataStore stores the persisted data for the system.
+#DataStoreService: #Postgres //#InmemStore | #Postgres
```
tonyhb, line 173: Do we need to enable the in-memory item here?

tonyhb (approved):
Looks good!!  Two minor comments and, wow, this is really great to have moved into OS!

pkg/coredata/postgres/0001_init.down.sql
djfarrelly: Ah you're right - I thought I deleted this outdated file - should be gone as of latest commit!

pkg/cuedefs/config/config.cue
djfarrelly: Ah good - catch - I commented this out before I moved the `filesystem_loader.go` to `inmemory/inmemory.go` - Fixing in next commit.

PR #174, opened 2022-07-19
Make cmd/run utilize the new execution serivce & state interfaces

tonyhb:
This commit fixes commands/run and cli/run such that the run UI uses the
new execution service, initializing a new function run via the singleton
in-memory queue.

This replicates the execution environment entirely whilst manually
scheduling the function run (ie. there is no actual runner service).

### DIFF
--- a/cmd/commands/run.go
+++ b/cmd/commands/run.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -12,8 +13,10 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/inngest/inngest-cli/inngest"
 	"github.com/inngest/inngest-cli/pkg/cli"
+	"github.com/inngest/inngest-cli/pkg/event"
 	"github.com/inngest/inngest-cli/pkg/execution/driver/dockerdriver"
 	"github.com/inngest/inngest-cli/pkg/function"
+	"github.com/inngest/inngest-cli/pkg/logger"
 	"github.com/spf13/cobra"
 )
 
@@ -43,13 +46,6 @@ func doRun(cmd *cobra.Command, args []string) {
 	if err != nil {
 		fmt.Println("\n" + cli.RenderError(err.Error()) + "\n")
 		os.Exit(1)
-		return
-	}
-
-	if cmd.Flag("event-only").Value.String() == "true" {
-		evt, _ := fakeEvent(cmd.Context(), *fn, "")
-		out, _ := json.Marshal(evt)
-		fmt.Println(string(out))
 		return
 	}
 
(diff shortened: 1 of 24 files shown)


pkg/execution/executor/service.go
```diff
@@ -253,6 +254,20 @@ func (s *svc) handleQueueItem(ctx context.Context, item queue.Item) error {
...
+		// item increases the scheduled count.
+		//
+		// Hopefully, if the backing implementation is the same (eg. a database which
+		// hosts the queue and the state store), Enqueue increases the pending count
+		// and this is a no-op - things should be atomic where possible.
+		//
+		// TODO: Add a unit test to ensure WG is 0 at the end of execution.
+		if err := s.state.Scheduled(ctx, item.Identifier, next.Outgoing); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
tonyhb: This is up to the implementation of each individual driver.  If we decide to implement an all-in-one postgres backend, we'd need to handle this during that driver's implementation.

cmd/commands/run.go
```diff
@@ -118,21 +124,21 @@ func runFunction(ctx context.Context, fn function.Function, eventName string) er
 	return ui.Error()
 }
 
-// event retrieves the event for use within testing the function.  It first checks stdin
+// generateEvent retrieves the event for use within testing the function.  It first checks stdin
 // to see if we're passed an event, or resorts to generating a fake event based off of
 // the function's event type.
-func event(ctx context.Context, fn function.Function, eventName string) (map[string]interface{}, error) {
+func generateEvent(ctx context.Context, fn function.Function, eventName string) (event.Event, error) {
```
djfarrelly, line 130: Good change! Much better to use `event.Event` here!

pkg/cli/run.go
```diff
@@ -77,37 +86,72 @@ func (r *RunUI) Init() tea.Cmd {
...
+	// service, given the config struct has not been copied.
+	r.sm, r.err = c.State.Service.Concrete.Manager(ctx)
+	if r.err != nil {
+		return
+	}
+
+	// In order to execute the function we need to create a new executor
+	// service to execute the steps of our function.  We'll manually initialize
+	// a new function run.
 	exec := executor.NewService(*c, executor.WithExecutionLoader(el))
```
djfarrelly, line 117: Do we need to run `exec.Stop()` at some point to shut down this service?

djfarrelly (commented):
From what I gather, the key change besides the run command is that we add a `Pending` counter which allows us to record in state if there is a step due to be executed, allowing the executor to check for timeouts (function didn't run). Is that about right?

_Gut checking if I'm understanding this PR change right:_
I think the other changes with using the services directly via the config concrete and keeping the executor as a service makes sense since that Executor does need some more of the state to be running for the step functions. If we didn't use the `executor/service` we'd have to re-implement the edge traversal in the run command which doesn't feel good.

Please correct me if my understanding is wrong above!

PR #178, opened 2022-07-20
Add stub driver to fix import cycles

tonyhb:
Unfortunately, centralizing our dependencies to config means that the fn
package cannot import mockdriver.


pkg/cuedefs/config/config.cue
```diff
@@ -180,11 +180,7 @@ package config
 
 #Postgres: {
 	backend: "postgres"
-<<<<<<< HEAD
```
djfarrelly, line 183: dang - I thought I resolved this w/ my rebase

Issue #52, opened 2022-04-26
inngest run does not wrap function output


djfarrelly:
### Description

When printing a very long line to stdout in the function, the output that the `inngest run` command prints is truncated to a single line and does not wrap.

### Steps to reproduce

1. Create a new function that returns a very long line that will not fit in your terminal
2. Run `inngest run`

### Screenshots / Stack traces

![image](https://user-images.githubusercontent.com/1509457/165405570-fccaa22d-58b2-4e8b-8aac-99a9e3426184.png)

```
inngest run

███████████████████████████████████ 100%
Exporting image...
#12 DONE 26.2s

Running your function using seed 143497
Function complete

Input:
{"data":{"billing_plan":"Spem manna mortuis palpa avertitur en fatemur nimis."},"name":"myapp/user.signup","user":{"account_name":"Prince Aron Kassulke","emai":"Abditis
hi casu donasti tu est fit peccare vae.","external_id":"f26dc065-df83-4b98-a121-0d3f690de8fd"}}

Output:
{"message":"inngest-run.js: payload:pages/api/hello.js","payload":"{\"args_version\":1,\"baggage\":{\"Actions\":{\"0\":{}},\"WorkspaceEvent\":{\"Event\":{\"data\":{\"bil
```

PR #179, opened 2022-07-20
Add `inngest run --replay` flag to replay previous events through your local function

jpwilliams:
## inngest run --replay

Adds an `inngest run --replay` flag (and some related flags) to pull recent events from Inngest and test your local function against them.

```sh
# Replay the latest 10 events
inngest run --replay

# Replay the latest 5 events
inngest run --replay --count 5

# Replay a specific event
# (runs in verbose mode due to single event)
inngest run --replay --event-id 01G8BG3BTBS291YNWHX8G449WQ

# Replay the latest 10 events for a specific trigger
# (required if your function has >1 trigger)
inngest run --replay --trigger test/event.sent
```

<details>
<summary>Replaying with failures (returns 1)</summary>

</details>

<details>
<summary>Replaying events with successes (returns 0)</summary>

</details>

<details>
<summary>Replaying with a generated event (enables verbose mode due to a single run)</summary>

</details>

<details>
<summary>Replaying a specific problem event</summary>

</details>

## Todo

- [x] Error cases/tests (no event found, unauthed, no events found, etc)
- [x] In-code comments
- [x] General tidy (remove dangling comments/logs, etc)


inngest/client/client.go
```diff
@@ -23,6 +24,8 @@ type Client interface {
 
 	AllEvents(ctx context.Context, query *EventQuery) ([]Event, error)
 	Events(ctx context.Context, query *EventQuery, cursor *Cursor) (*PaginatedEvents, error)
+	RecentEvent(ctx context.Context, workspaceID uuid.UUID, eventID ulid.ULID) (*ArchivedEvent, error)
+	RecentEvents(ctx context.Context, workspaceID uuid.UUID, eventName string, count int64) ([]ArchivedEvent, error)
```
jpwilliams, line 28: Is that nomenclature okay? To fit with the ideal these would be `Event` and `Events`, but that'd require renaming `Events` to `EventTriggers`; I didn't want to be too destructive. ;)

pkg/cli/run.go
```diff
@@ -118,40 +137,89 @@ func (r *RunUI) run(ctx context.Context) {
...
+				if run.Metadata().Pending == 0 || hasErrors {
+					if hasErrors {
+						r.err = fmt.Errorf("run failed")
+					}
+
+					*execution.done = true
+
+					return
+				}
+				<-time.After(time.Millisecond * 5)
```
jpwilliams, line 216: This feels clunky; I could use channels instead to ensure it's not needlessly looping.

cmd/commands/root.go
```diff
@@ -50,6 +50,7 @@ func Execute() {
 
 	rootCmd.PersistentFlags().Bool("prod", false, "Use the production environment for the current command.")
 	rootCmd.PersistentFlags().Bool("json", false, "Output logs as JSON.  Set to true if stdout is not a TTY.")
+	rootCmd.PersistentFlags().Bool("verbose", false, "Enable verbose logging.")
```
tonyhb, line 53: 👌

cmd/commands/run.go
```diff
@@ -94,15 +143,67 @@ func buildImg(ctx context.Context, fn function.Function) error {
...
+		if err != nil {
+			return err
+		}
+
+		if evt == nil {
+			return fmt.Errorf("event not found: %s", *opts.fetchEventId)
+		}
+
+		evts = []event.Event{*evt}
+	} else if opts.fetchRecentEvents > 0 {
```
tonyhb, line 166: I was thinking about this and there's something we could do to improve the several if / else blocks chained together.

If we changed `runFunctionOpts` to:

```
type runFunctionOpts struct {
  eventFunc func ([]event.Event, error)
  verbose bool
}
```

When we initialize the flags, we could pass in a function which loads the events.  This `runFunction` code becomes much simpler — it's one single case of calling the evenFunc to return >= 1 event and an error.  Then, each `if` block becomes its own function that we can test individually, if desired.

It should make this area generally cleaner and a little easier to modify in the future.  WDYT?

cmd/commands/run.go
```diff
@@ -147,20 +255,113 @@ func generateEvent(ctx context.Context, fn function.Function, eventName string)
...
+	archivedEvents, err := s.Client.RecentEvents(ctx, ws.ID, triggerName, count)
+	if err != nil {
+		return nil, err
+	}
+
+	events := []event.Event{}
+
+	for _, archivedEvent := range archivedEvents {
+		type evtData struct {
+			Data map[string]interface{}
```
djfarrelly, line 309: Is this missing the `User` `map[string]interface{}` and `v` as well to be parsed or were those intentionally left off? ([related docs](https://www.inngest.com/docs/event-format-and-structure))

cmd/commands/run.go
```diff
@@ -147,20 +255,113 @@ func generateEvent(ctx context.Context, fn function.Function, eventName string)
...
+		evt := &evtData{}
+
+		if err := json.Unmarshal([]byte(archivedEvent.Event), &evt); err != nil {
+			return nil, err
+		}
+
+		events = append(events, event.Event{
+			ID:        archivedEvent.ID,
+			Name:      archivedEvent.Name,
+			Data:      evt.Data,
```
djfarrelly, line 323: See above about `User` & `v`

cmd/commands/run.go
djfarrelly, line 309: Is this missing the `User` `map[string]interface{}` and `v` as well to be parsed or were those intentionally left off? ([related docs](https://www.inngest.com/docs/event-format-and-structure)). This would be the same comment for `fetchEvent` below.

cmd/commands/run.go
djfarrelly, line 323: See above about `User` & `v`

pkg/cli/run.go
```diff
@@ -46,32 +50,66 @@ type RunUI struct {
...
+// A style for the label to show that a function is running.
+var runsStyle = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#ffffff")).Background(lipgloss.Color("#fbbf24")).Padding(0, 1)
+
+// A style for the label to show that all of a function's steps have
+// successfully run.
+var passStyle = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#ffffff")).Background(lipgloss.Color("#84cc16")).Padding(0, 1)
+
+// A style for the label to show that at least one of a function's steps  has
+// failed to run.
+var failStyle = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#ffffff")).Background(lipgloss.Color("#b91c1c")).Padding(0, 1)
```
djfarrelly, line 111: In a future PR it could be cool if we moved these to and re-used the styled in [styles.go](https://github.com/inngest/inngest/blob/1abb0c639c982ed5c1b520c44c1b1e4dadbc319a/pkg/cli/styles.go)

pkg/cli/run.go
```diff
@@ -218,56 +319,77 @@ func (r *RunUI) View() string {
...
-	for id, data := range output {
-		byt, _ := json.Marshal(data)
-		s.WriteString(BoldStyle.Render(fmt.Sprintf("Step '%s'", id)))
-		s.WriteString(": \n")
-		s.WriteString(wrap.String(string(byt), width))
-		s.WriteString("\n")
+	metadata := state.Metadata()
+	done := *run.done
+	passed := done && len(errors) == 0
+	failed := done && len(errors) > 0
```
djfarrelly, line 348: I really like this entire section and refactor!

djfarrelly (commented):
This is great @jpwilliams! Lots of fantastic stuff in here and I really like the way you've displayed the info and added the verbose flag. I did leave a comment on potentially missing fields from the fetched events, but other than that this PR looks fantastic!

inngest/client/client.go
tonyhb: This is totally okay.

pkg/cli/run.go
tonyhb: I understand what you mean and think that we could provide something that allows you to subscribe to function runs from the state store (eg. new state, and/or completion).

For now, we have to poll _somewhere_ to monitor state, so I'm okay with adding this here :)

pkg/cli/run.go
jpwilliams: Aye good shout. I figured for the moment that it was okay for them to be here as I couldn't think of another spot they'd be reused atm.

It might be sensible to put them in there regardless of that though?

cmd/commands/run.go
jpwilliams: Yep good spot! Added.

cmd/commands/run.go
jpwilliams: Yep good spot! Added.

cmd/commands/run.go
jpwilliams: Agreed - would be much nicer!

pkg/cli/run.go
djfarrelly: No need to make that change in this PR! We probably have several areas in the CLI that we could bring more consistency - that can be a future project :)

cmd/commands/run.go
jpwilliams: Have changed this to use a loading function for each scenario now.

jpwilliams:
Appears to also fix #52.

tonyhb (approved):

PR #182, opened 2022-07-21
Fix dev command flags

djfarrelly:
### Description

This fixes the regression with the new config package (9eeb15d92e93ca861b2dd55c1a9340c97e777091).


tonyhb (approved):
Ah, good catch - thank you!

PR #184, opened 2022-07-22
Support deploy command to self-hosted Core API

djfarrelly:
## Description
This forks the client based on parsing if the `INNGEST_API` environment variable. 

This is a temporary change as we should:

1. Update Inngest Cloud to support all the new GraphQL queries and mutations
2. Create a separate `pkg/coreapi/client` package which can re-use code from the `coreapi` package
3. Use a `config` GraphQL edge to allow the given `INNGEST_API` to configure the cli client, enabling config to be set centrally rather than on the per-client basis. This has the added bonus of allowing us to push configuration on a per-account basis via Inngest Cloud's GraphQL API.

### How to use

The user must currently pass their own `INNGEST_API`, `INNGEST_REGISTRY` and `INNGEST_DSN_PREFIX` to configure where to deploy. The registry and the DSN prefix should be set by the `config` edge in the future. Here's an example using a locally running Core API and Docker Hub with the `inngest` Docker Hub org which will create images with the `inngest/<action-image-name>:<action-version>` tag.

```shell
INNGEST_API="http://127.0.0.1:8300" \
  INNGEST_REGISTRY="https://index.docker.io/v1/" \ 
  INNGEST_DSN_PREFIX="inngest" \
  go run ../../cmd/main.go deploy
```

### DIFF
--- a/cmd/commands/deploy.go
+++ b/cmd/commands/deploy.go
@@ -11,6 +11,7 @@ import (
 	"github.com/inngest/inngest-cli/inngest/clistate"
 	"github.com/inngest/inngest-cli/internal/cuedefs"
 	"github.com/inngest/inngest-cli/pkg/cli"
+	"github.com/inngest/inngest-cli/pkg/coredata"
 	"github.com/inngest/inngest-cli/pkg/execution/driver/dockerdriver"
 	"github.com/inngest/inngest-cli/pkg/function"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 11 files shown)


cmd/commands/deploy.go
```diff
@@ -189,10 +214,11 @@ func configureVersionInfo(ctx context.Context, a inngest.ActionVersion) (inngest
 
 	// If we're publishing a specific version, attempt to find it.  Else, load the latest
 	// action version.  This automatically happens depending on whether a.Version is nil.
-	found, err := state.Action(ctx, a.DSN, a.Version)
+	found, err := state.Client.Action(ctx, a.DSN, a.Version)
+
 	// When deploying without specifying an action, allow "action not found"
 	// errors.
-	if a.Version == nil && err != nil && err.Error() == "action not found" {
+	if a.Version == nil && err != nil && (err.Error() == "action not found" || err.Error() == coredata.ErrActionVersionNotFound.Error()) {
```
tonyhb, line 221: Nit for later:  we can use `errors.Is(err, coredata.ErrActionVersionNotFound)` to check for this error (https://pkg.go.dev/errors).

tonyhb (approved):
LGTM.  Very much looking forward to getting the config endpoint which could eg. describe the API version, and as you say the registry host!

PR #185, opened 2022-07-22
Fix merge issue with renamed package

djfarrelly:
## Description

`state` was renamed `clistate` to avoid exact things like this 😆 

- https://github.com/inngest/inngest/pull/179
- https://github.com/inngest/inngest/pull/181


djfarrelly:
Closing w/ same fix in: https://github.com/inngest/inngest/pull/186

PR #188, opened 2022-07-25
Avoid use of `fs.FS` for Windows

### LINKED ISSUE #175: init does not work on Windows
**Describe the bug**
Running `inngest init` on Windows OS fails after selecting the language with the following error:

> There was an error creating the function: open .: invalid argument

**To Reproduce**
1. On Windows, install inngest cli
2. Run `inngest init` and select any options to scaffold a new functions
3. You should see the but

**Expected behavior**
I expect a new function to be created - the new files should all be created successfully in my new directory.

**Stack trace dump**
The error is thrown on this line in `template.go`: https://github.com/inngest/inngest/blob/e9071758252641f939eab2956eab59c3b410d0b0/pkg/scaffold/template.go#L156

It seems there are known issues with using `fs.WalkDir` as it uses `fs.FS`'s `Open()` function:

* https://github.com/golang/go/issues/44166
* https://github.com/golang/go/issues/44279

A potential solution ([see comment](https://github.com/golang/go/issues/44279#issuecomment-955778704)) is for us to not use `fs.FS` and `fs.WalkDir`, but instead re-implement this code (and any other similar `fs.Open`/`fs.WalkDir` usage) using `filepath.WalkDir`and `os.Open`. We also may have to use `filepath.ToSlash()` and/or make other modifications to the windows file paths to not throw issues. 

**System info**
- inngest-cli version: 0.5.0-rc2
- OS: Windows
- Version: 10

**Additional context**
I was not able to confirm if this caused any issues when running `inngest init` in [WSL](https://docs.microsoft.com/en-us/windows/wsl/)

jpwilliams:
Closes #175.

`fs.FS` seems to cause some problems just when opening Windows file paths. Even when utilising something else for the directory walking, I'm still getting errors from `fs.FS` from totally valid Windows paths.

We can get away with not using it here and it seems okay.

There may well be some other places we need to look at that use `fs.FS` as it's likely a problem there too.

inittest

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/inngest/inngestgo v0.4.0
 	github.com/jedib0t/go-pretty/v6 v6.3.0
 	github.com/joho/godotenv v1.3.0
+	github.com/karrick/godirwalk v1.17.0
 	github.com/lib/pq v1.10.6
 	github.com/lnquy/cron v1.1.1
 	github.com/mattn/go-isatty v0.0.14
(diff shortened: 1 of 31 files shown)


djfarrelly (approved):
LGTM!

PR #189, opened 2022-07-25
Add traversal of parent directories to find `inngest.json` or `inngest.cue`

### LINKED ISSUE #139: `inngest run` should check parent directories for `inngest.json`
**Is your feature request related to a problem? Please describe.**
When I'm using the command line, I'm often in a step's directory, e.g. `<my-function-dir>/steps/step-1` and I want to quickly run the function from that directory. The `inngest.json` file is in the root dir for the function itself so it throws an error saying there is not config.

**Describe the solution you'd like**
`inngest run` should traverse parent directories if it does not find `inngest.json` in the cwd

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

jpwilliams:
Closes #139.

Adds `func findFileUp(path string) (string, []byte, error)` and refactors config finding slightly in order to have `inngest run` recurse parent directories until it finds an `inngest.json` or `inngest.cue` file.

<details>
<summary>Happy path, with file in current dir or above</summary>

nested-inngest-run
</details>

<details>
<summary>Error path; searched up to root and no file</summary>

nested-inngest-run-fail
</details>

### DIFF
--- a/pkg/function/config.go
+++ b/pkg/function/config.go
@@ -22,6 +22,11 @@ var (
 	ErrNotFound = fmt.Errorf("No inngest file could be found.")
 )
 
+const (
+	cueConfigName  = "inngest.cue"
+	jsonConfigName = "inngest.json"
+)
+
 // Load loads the inngest function from the given directory.  It searches for both inngest.cue
 // and inngest.json as both are supported.  If neither exist, this returns ErrNotFound.
 func Load(ctx context.Context, dir string) (*Function, error) {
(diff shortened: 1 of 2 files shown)


pkg/function/config.go
```diff
@@ -44,31 +49,65 @@ func Load(ctx context.Context, dir string) (*Function, error) {
...
 
 	// Then attempt to find inngest.cue, the canonical reference.
-	cue := filepath.Join(abs, "inngest.cue")
-	if _, err := os.Stat(cue); err == nil {
-		// The cue file exists.
-		byt, err := os.ReadFile(cue)
-		if err != nil {
-			return nil, err
-		}
+	cue, byt, err := findFileUp(filepath.Join(abs, cueConfigName))
```
djfarrelly, line 52: I wonder if we're doing the traversal in `findFileUp`, should we search for both files at once and return the filename and contents to the `Load` function? Either way, the output is passed to `Unmarshal` so we could reduce duplicate code and directory traversal.

djfarrelly (commented):
Great change! I have one comment on a potential way to reduce the code repeating itself, what do you think?

pkg/function/config.go
jpwilliams: Have refactored this slightly to check for both `.cue` and `.json` files in each dir as it goes.

pkg/function/config.go
```diff
@@ -44,31 +49,65 @@ func Load(ctx context.Context, dir string) (*Function, error) {
...
+}
+
+// Finds a file at the given `path``, iterating up through the directory tree
+// until it finds the file or reaches the root.
+//
+// Returns the final path and the file contents.
+//
+// Will return an error if reading a file errored, but will return empty values
+// if no file could be found.
+func findFileUp(path string) (string, []byte, error) {
```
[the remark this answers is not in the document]
jpwilliams: This is more specific to finding the configs now.

tonyhb (approved):

PR #192, opened 2022-07-25
WIP: Debug mode

tonyhb:
This commit adds a `--debug` flag (`-d`) which enables debug mode for
running functions locally.  Debug mode swaps each edge out for a `debug`
edge, allowing users to "step over" step functions locally.

This is required for eg. step functions which have delays or coordinate
between events.


tonyhb:
While we need a debug mode, I think we're going to work on adding a step-over debugger to a UI.

PR #195, opened 2022-07-28
Add version constraints to steps to specify exact action versions

### LINKED ISSUE #190: Add version constraints to Function steps to specify exact action versions
### **Is your feature request related to a problem? Please describe.**
With the project's shift towards using "Functions" as the core definition of work to be performed by the system, it does not have full feature compatibility with the legacy "Workflows."

Without supporting exact or range version specification within Function's Steps, the Executor will always run the latest version of the action. This means that deployments and rollbacks of functions will not function as expected.

### **Describe the solution you'd like**
The function cue definition for `#Step` ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/cuedefs/v1/function.cue#L64-L84)) should add an optional version range that is currently used by the workflow cue's `#Action` ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/internal/cuedefs/pkg/workflows/workflows.cue#L58-L61)).

During `inngest deploy`, the newly created/deployed actions ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/cmd/commands/deploy.go#L74-L78)) should return the specific action version and the function's steps should be updated prior to deploying the function/workflow.

Other necessary changes / areas of the code to explore:

* The unmarshalling of function cue's ([test](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/function/config_test.go#L73-L75)) should be able to read the exact version of the actions used in each step. This will ensure when a function is loaded from the database (coredata pkg), the executor can find the exact version of an action desired.
    * Similarly, the marshalling (object to cue string) ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/function/function.go#L120-L123)) will need to be checked that it's working correctly with the new config and objects.
* When extracting the actions from a given function ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/function/function.go#L250-L252)), the versions should be included.
* When the `Function` is converted into a `Workflow` for back-compatibility (and current use by the Executor), the action versions should be set correctly ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/function/function.go#L180-L183)).
* Remove the `FunctionVersion`'s setting of action versions ([source](https://github.com/inngest/inngest/blob/7269dd1c9e72d008aece99cf8a4f58a690e2cb46/pkg/function/function_version.go#L30-L48)). This will be outdated and may be able to be fully removed and instead the code should use `FunctionVersion.Function.Actions()` (as mentioned above)

### **Describe alternatives you've considered**
Since function is the format that is displacing workflow, we must have action versions specified.

jpwilliams:
Closes #190.

## Summary

Adds version constraints to function steps derived from action versions upon deployment.

- `function.Step` has an optional `inngest.VersionConstraint` added as `version`, which is also added to the `function.cue` definition
- During deployment and after deploying actions, step version constraints are derived from the upserted actions
- Added and adjusted tests to ensure that step versions are correctly derived and (un)marshalled for when they're sent to the core

## Todo

- [x] Changes summary
- [x] New test in `function_test.go` to assert that `Function.Workflow()` returns step version constraints (and only if available)


pkg/cuedefs/v1/function.cue
```diff
@@ -81,6 +81,11 @@ package v1
 	// If more than one item is supplied in this array, the step will run multiple times after
 	// each preceeding step finishes.
 	after: [...#After]
+
+	version?: {
```
jpwilliams, line 85: This needs a comment

cmd/commands/deploy.go
```diff
@@ -72,10 +72,30 @@ func deploy(cmd *cobra.Command, args []string) error {
...
+
 	for _, a := range actions {
-		if err := deployAction(ctx, a); err != nil {
+		actionVersion, err := deployAction(ctx, a)
+		if err != nil {
 			return err
 		}
+
+		// TODO: Move this to a dedicated function.
+		step, ok := fn.Steps[dsnToKeySteps[actionVersion.DSN]]
```
jpwilliams, line 88: I found another spot in the code that matches a `Step` to an `ActionVersion` via the `DSN`.

I wanted to add a `func (a ActionVersion) Step(steps []Step) *Step` but ran into a circular dependency.

I'd love to quickly spike with someone about the best way to alleviate those issues in the future. My guess is moving packages around, which feels unideal.

tonyhb: Yeah, this is part of our cleanup we need to do with the top-level `inngest` package, plus actions in general.  Definitely agree on tidying this in the future.

cmd/commands/deploy.go
```diff
@@ -235,6 +255,10 @@ func configureVersionInfo(ctx context.Context, a inngest.ActionVersion) (inngest
 		return a, fmt.Errorf("Version %s of the action already exists", a.Version.String())
 	}
 
+	// Set the version to the found action version.
+	// We will update this below if it needs doing.
+	a.Version = found.Version
```
tonyhb, line 260: 👍

tonyhb (approved):

pkg/cuedefs/v1/function.cue
```diff
@@ -81,6 +81,13 @@ package v1
 	// If more than one item is supplied in this array, the step will run multiple times after
 	// each preceeding step finishes.
 	after: [...#After]
+
+	// version is the version constraint for the step when resolving the action to
+	// run.
+	version?: {
+		version?: uint
```
djfarrelly, line 88: Major?

PR #198, opened 2022-08-01
On `npm publish`, include `README.md` from main repo

jpwilliams:
The npm repo [inngest-cli](https://npmjs.com/package/inngest-cli) has no dedicated `README.md`.

We can just steal the main readme for this whenever we publish.

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -4,7 +4,7 @@
   "description": "The event-driven queue for any language.",
   "license": "SEE LICENSE IN LICENSE.md",
   "scripts": {
-    "prepublishOnly": "mkdir bin && touch bin/inngest",
+    "prepublishOnly": "mkdir bin && touch bin/inngest && cp ../README.md .",
     "postinstall": "node postinstall.js",
     "build": "tsc"
   },


npm/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The event-driven queue for any language.",
   "license": "SEE LICENSE IN LICENSE.md",
   "scripts": {
-    "prepublishOnly": "mkdir bin && touch bin/inngest",
+    "prepublishOnly": "mkdir bin && touch bin/inngest && cp ../README.md .",
```
djfarrelly, line 7: Can we add README to the `files` array in the package.json instead or is there something here that I'm not realizing 😄

jpwilliams: @djfarrelly It's just 'cause we're in the `npm` subdirectory here; you can't specify files outside of the root dir (i.e. it would be `../README.md`).

Either we add a package-specific `README.md` or copy across whatever the current root `README.md` is every time we publish.

tonyhb (approved):

PR #199, opened 2022-08-04
Ensure `run` and `dev` use network mode host

tonyhb:
This ensures that locally our functions can access the host network and
any extra services.


djfarrelly (approved):
Solid fix - I like the separation of dev config as well. LGTM!

jpwilliams (approved):

PR #201, opened 2022-08-06
Load all workspace events

### LINKED ISSUE #90: Events are only from test environment during `inngest init`
**Describe the bug**
When I run `inngest init` I expect there to be events listed from production and test modes (or at least production). 

**To Reproduce**
1. Log into your Inngest account via `inngest login`
2. Run `inngest init` 
3. Select event driven - you will only see events from the test environment

**Expected behavior**
I expect all events, test and prod to be shown

**System info**
- inngest-cli version: latest
- OS: macOS

**Additional context**
n/a

### LINKED ISSUE #115: When requesting events for `inngest init`, warn if you're not logged in
When you're not logged in you're unable to query for the events in your current workspace, and can't show custom events in the event browser.

We should warn you that you're not logged in when running `inngest init`.

tonyhb:
This PR changes `init` to return events from all workspaces, then shows
which workspaces the event has been seen within.

In the future we should also show the last seen at date, allowing users
to select recently active events.

Closes #115
Closes #90

### DIFF
--- a/pkg/cli/initialize/initialize.go
+++ b/pkg/cli/initialize/initialize.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math"
 	"os"
-	"sort"
 	"strings"
 	"time"
 
@@ -14,7 +13,6 @@ import (
 	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
-	"github.com/google/uuid"
 	"github.com/inngest/inngest/inngest"
 	"github.com/inngest/inngest/inngest/client"
 	"github.com/inngest/inngest/inngest/clistate"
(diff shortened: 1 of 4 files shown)


pkg/cli/initialize/initialize.go
```diff
@@ -516,24 +512,62 @@ func (f *initModel) renderWelcome() string {
...
+	// Fetch events from every workspace.
+	workspaces, err := clistate.Client(ctx).Workspaces(ctx)
 	if err != nil {
-		// Attempt to fetch unauthed events.
-		return c.AllEvents(ctx, nil)
+		// If this is an unauthorized error we can ignore it.
+		if !strings.Contains(err.Error(), "unauthorized") {
+			return nil, err
+		}
+		err = fmt.Errorf("You're not logged in so we're only showing public events")
```
djfarrelly, line 532: Good add

djfarrelly (approved):

PR #208, opened 2022-08-15
INN-613 Allow custom retry counts per step

### LINKED ISSUE #117: Add custom retry policies to steps
Steps within a function should be able to declare their own retries via configuration.  For example:

```
{
  "name": "My lovely lil function 😍",
  "id": "happy-pigeon-07dde7",
  "triggers": [
    {
      "event": "lol/parent.jokes",
      "definition": {
        "format": "cue",
        "synced": true,
        "def": "file://./events/lol-parent-jokes.cue"
      }
    }
  ],
  "steps": {
    "step-1": {
      "id": "step-1",
      "path": "file://./steps/step-1",
      "name": "my lil func",
      "runtime": {
        "type": "docker"
      }
      "retries": {
        "attempts": 10,
      }
    }
  }
}

### PR DESCRIPTION
**Pull request by jpwilliams**
Closes #117.

Adds `retries.attempts` to each defined step in `inngest.json|cue`, letting the user specify a value between `0` and `20`.

If not provided, the retry count will default to the previous `3`.

```jsonc
// From #117 
{
  "name": "My lovely lil function 😍",
  "id": "happy-pigeon-07dde7",
  "triggers": [
    {
      "event": "lol/parent.jokes",
      "definition": {
        "format": "cue",
        "synced": true,
        "def": "file://./events/lol-parent-jokes.cue"
      }
    }
  ],
  "steps": {
    "step-1": {
      "id": "step-1",
      "path": "file://./steps/step-1",
      "name": "my lil func",
      "runtime": {
        "type": "docker"
      },
      "retries": {
        "attempts": 10
      }
    }
  }
}
```

Adds tests (and a tool for the test harness for later) to check retry counts that are:

- Higher-than-default
- Default (none defined)
- Lower-than-default
- None (`0`)


internal/cuedefs/pkg/workflows/workflows.cue
```diff
@@ -62,6 +62,10 @@ import (
 	// Metadata about how the action will be used.  Each action requires custom
 	// input to work, eg. what data to transform, what email template to use, etc.
 	metadata?: [string]: _
+
+	retries?: {
+		attempts?: int & >=0 & <=20
```
tonyhb, line 67: Should we add the default in Cue, here?

```
retries?: {
  attempts?: int & >=0 & <=20 | *3
}
```

pkg/cuedefs/v1/function.cue
```diff
@@ -88,6 +88,10 @@ package v1
 		major?: uint
 		minor?: uint
 	}
+
+	retries?: {
+		attempts?: int & >=0 & <=20
```
tonyhb, line 93: Should we add the default here to function definitions?

```
retries?: {
  attempts?: int & >=0 & <=20 | *3
}
```

tests/testdsl/testdsl.go
```diff
@@ -141,6 +142,84 @@ func RequireOutputWithin(output string, within time.Duration) Proc {
...
+			backoffTime = uint(10) << i
+
+			fmt.Printf("\t> Checking attempt #%d executes (waiting %d seconds)\n", i+1, backoffTime)
+			if err := timeout(time.Second*time.Duration(backoffTime), func() error {
+				return requireLogFields(ctx, td, map[string]any{
+					"caller":  "executor",
+					"step":    step,
+					"message": "executing step",
+					"attempt": i,
+				})
```
tonyhb, line 161: This is super nice!

tests/testdsl/testdsl.go
```diff
@@ -141,6 +142,84 @@ func RequireOutputWithin(output string, within time.Duration) Proc {
...
+			}); err != nil {
+				return err
+			}
+
+			if i+1 < count {
```
tonyhb, line 166: Maybe revert this to lower the indent count?  I typically prefer to guard data patterns then quit, vs do things in if:

```
if i+1 >= count {
  continue
}
// ...do logic here if i < count

tonyhb (approved):
This is great!  One question on defaults within function.cue and one nit — the tests are a great touch to confirm this works as anticipated 👏

tests/testdsl/testdsl.go
jpwilliams: Yep yep big agree.

pkg/cuedefs/v1/function.cue
jpwilliams: @tonyhb My thought for omitting was so that we can always control/change the default attempt count.

If we later increased/decreased the default, do you think users would want to use our new default if they haven't defined one? Or would they perhaps have assumed it would always be 3?

PR #209, opened 2022-08-15
Add auth via `inngest login --token` and `INNGEST_AUTH_TOKEN` env var

jpwilliams:
## Summary

Adds two new auth methods for the CLI, aimed at providing support for CI at [inngest/setup-inngest](https://github.com/inngest/setup-inngest).

- Logging in using an API key via `inngest login --token [key]`
- Performing an authed command with an env var API key via `INNGEST_API_KEY=key inngest deploy`

## Questions 🤔

Should we:
1. Remove `inngest login --token` and only ever look at `INNGEST_API_KEY`?
2. Remove `INNGEST_API_KEY` and only have `inngest login --token` (or another auth method)?

Both methods seems cluttered.


djfarrelly (approved):
The code looks good @jpwilliams! I'm personally fine w/ just the environment variable, especially if this is just targeting CI, I think that's fine. 😄  

**Naming**
If we want to have both methods, my only comment is that the variable and the flag use different names (token vs. key) - I would maybe then suggest that we converge on a single name. The tricky part is that we also use `API KEY` for our sources. That could get confusing for users. Perhaps we could try to solidify our naming for the 2 types of keys like:

- Source key / API key
- API key / Auth token
- Publish key / API key

What do you all think?

tonyhb (approved):
I think leaving both `--token` and the environment variable is a fine approach.  The environment variable is necessary for CI, and `--token` flows are generally preferred for logging in to the CLI as they're easily created and revoked (vs. password resets, which are possible but annoying).

Also, on Dan's point I think `token` is an _okay_ name given we use `key` for event keys;  token is familiar (auth token) so I'm okay with this overall.

Good work!

jpwilliams:
Thanks for review both @djfarrelly @tonyhb.

@djfarrelly It's a fair point on the use of "token" naming - my thinking was the same as @tonyhb in that it's a familiar term for that purpose, but then that doesn't match up nicely with the env var that is also being used as an access/auth token.

@tonyhb Would you be happy taking @djfarrelly's suggestion and using `INNGEST_TOKEN` for this purpose to ensure it's always different from an "API Key"? Leans in to the access/auth token naming.

tonyhb:
Absolutely, name works fine with me - either that or the more verbose `INNGEST_AUTH_TOKEN` :)

PR #211, opened 2022-08-16
Standardize parsing of cue overlay for windows

tonyhb:
This commit changes cue overlay parsing such that it works on windows.
On windows, Cue requires a `C:\` prefix to all files loaded in the
overlay.  It also requires that path separators be OS-specific;
fs.WalkDir does not respect os-specific separators.

Finally, we change `inngest run` to exit when run errors, instead of
hanging.


cmd/commands/run.go
```diff
@@ -289,6 +289,8 @@ func runFunction(ctx context.Context, fn function.Function, opts runFunctionOpts
 	log := logger.Buffered(buf)
 	ctx = logger.With(ctx, *log)
 
+	fmt.Println("RUNNING")
```
djfarrelly, line 292: Is this left from debugging or is this intentional? Just saw it outside of bubbletea :)

djfarrelly (approved):
These changes look great - excited to get this bug squared away! I left a couple of suggestions/comments

PR #213, opened 2022-08-17
When an example changes, rebuild the website

jpwilliams:
Ensure the website is rebuilt with any new/changed/removed examples as they are edited.

- inngest/website#163
- [x] Add `WEBSITE_CLOUDFLARE_DEPLOY_HOOK` to repo secrets

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,16 @@
+name: Trigger example changes
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'examples/**'
+
+jobs:
+  deploy-website:
+    runs-on: ubuntu-latest
+    steps:
+      - run: curl -X POST "${{ env.WEBSITE_CLOUDFLARE_DEPLOY_HOOK }}" -v
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/examples.yml
```diff
@@ -0,0 +1,16 @@
...
+    paths:
+      - 'examples/**'
+
+jobs:
+  deploy-website:
+    runs-on: ubuntu-latest
+    steps:
+      - run: curl -X POST "${{ env.WEBSITE_CLOUDFLARE_DEPLOY_HOOK }}" -v
+        env:
+          WEBSITE_CLOUDFLARE_DEPLOY_HOOK: ${{ secrets.WEBSITE_CLOUDFLARE_DEPLOY_HOOK }}
```
tonyhb, line 16: This works, and looks good.  I was thinking that we could also trigger an Inngest event here, and we can run our own functions which update the website as well as any internal things we might need to happen — instead of going directly to cloudflare :)

tonyhb (approved):

.github/workflows/examples.yml
jpwilliams: Yes! [INN-623](https://linear.app/inngest/issue/INN-623) would add `inngest/send-event` for GitHub Actions - it just needs to be a cURL which would be super fast.

PR #215, opened 2022-08-19
Add support for custom Dockerfile

djfarrelly:
## Description

To enable folks with monorepos to use Dockerfiles that are outside of the path of the `inngest.json` file so files outside of the `inngest.json` context can be added. 

### Example

Example project structure
```
/functions/my-fn-a/inngest.json
/functions/my-fn-a/index.js
/src/api.js
/src/someLibrary.js
Dockerfile.my-fn
Dockerfile.api
```

`index.js`:
```js
const something = require("../../src/someLibrary")

const args = JSON.parse(process.argv.pop())
const output = something(args.data)
console.log(output)
```

`inngest.json` file:
```json
{
  "name": "Backfill User Data",
  "id": "something-cool-08kj53",
  "triggers": [
    {
      "event": "retool/backfill.requested"
    }
  ],
  "steps": {
    "my-fn": {
      "id": "my-fn",
      "path": "file://.",
      "name": "Run some code",
      "runtime": {
        "type": "docker",
        "dockerfile": "../../Dockerfile.my-fn"
      }
    }
  }
}

```

### TODO

* [ ] Add tests

### DIFF
--- a/cmd/commands/deploy.go
+++ b/cmd/commands/deploy.go
@@ -208,7 +208,18 @@ func deployAction(ctx context.Context, a inngest.ActionVersion) (inngest.ActionV
 		return a, fmt.Errorf("error preparing action: %w", err)
 	}
 
-	config, err := cuedefs.FormatAction(a)
+	// XXX: Remove the dockerfile field from the runtime struct b/c the action shouldn't need this info
+	clean := a
+	rt, ok := clean.Runtime.Runtime.(inngest.RuntimeDocker)
+	if !ok {
+		return a, fmt.Errorf("failed to parse runtime")
+	}
+	clean.Runtime.Runtime = inngest.RuntimeDocker{
+		Entrypoint: rt.Entrypoint,
+		Memory:     rt.Memory,
+	}
+
+	config, err := cuedefs.FormatAction(clean)
 	if err != nil {
 		return a, err
 	}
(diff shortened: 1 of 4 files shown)


cmd/commands/deploy.go
```diff
@@ -208,7 +208,18 @@ func deployAction(ctx context.Context, a inngest.ActionVersion) (inngest.ActionV
 		return a, fmt.Errorf("error preparing action: %w", err)
 	}
 
-	config, err := cuedefs.FormatAction(a)
+	// XXX: Remove the dockerfile field from the runtime struct b/c the action shouldn't need this info
```
djfarrelly, line 211: I'm unsure of the cleanest way to do this - Some fields may be useful in step's `runtime` locally for building, etc. but are just not needed or relevant to the action when as the image is already build. Would love suggestions on this - I thought this work around could be alright for now until we sort out a better way to decide how to do this - perhaps have 2 `runtime` structs and/or convert a `Step` struct to the action here.

tonyhb: Yeah, this is _definitely_ some tech debt we accrued from the prototyping stage.  We should only need to dictate that the action is of type docker for now.  The "XXX" is a great comment so that we know we need to come back and clean this up — I'm assuming we'll get to that as we near v1 and we figure out the self hosting flows too.

pkg/execution/driver/dockerdriver/build.go
```diff
@@ -55,9 +66,10 @@ func FnBuildOpts(ctx context.Context, f function.Function, args ...string) ([]Bu
 
 // BuildOpts represents options for a single docker build, using buildx.
 type BuildOpts struct {
-	Path string
-	Tag  string
-	Args []string
+	Path       string
+	Tag        string
+	Dockerfile string
```
tonyhb, line 71: Good call

tonyhb (approved):
Let's ship!

PR #218, opened 2022-08-22
Add support for .env files in local development

tonyhb:
This commit introduces support for reading .env files for each function
when running locally via `inngest run` or `inngest dev`.  We do this by
creating a new `env` package which reads .env files for functions.

We extend the Docker driver with the ability to read from this env
reader, and we parse and set the reader when creating the execution
engine and drivers during `inngest run` and `inngest dev`.

Because .env files locally produce plaintext we do not reference
"secrets", here.  We'll extend the concept of secrets with eg. SOPS and
Vault integrations for encryption in the near future.


jpwilliams (approved):
LGTM.

One additional test it might be nice to cover would be a more complex env var with line breaks like a private key or a lengthy JSON object (e.g. a Google Service Account); lots of platforms have different rules for the best way to add these, so it'd be good to have a test example of ours.

tonyhb:
Good idea!  Done.  Thank you :)

PR #219, opened 2022-08-23
Add `TEMPLATE.md` for quick starts

jpwilliams:
An example `README.md` for quick-starts, based on [inngest/example-drip-campaign](https://github.com/inngest/example-drip-campaign).


jpwilliams:
Will push this in seeing as inngest/website#163 is now merged. We can adjust as we go!

PR #222, opened 2022-08-24
Alter `id` and `name` of functions created using `inngest init --template`

jpwilliams:
### Summary

A post-PR change from #214 to adjust the `name` and `id` of functions created using `inngest init --template`.

Alters JSON in-place rather than unmarshalling/marshalling to preserve as much of the structure and order of the config file as possible; it's confusing for the user if they use a template and then the config file looks different to what they expect.

### Related

- #214


pkg/function/config.go
```diff
@@ -20,7 +20,7 @@ var (
 
 const (
 	cueConfigName  = "inngest.cue"
```
tonyhb, line 22: Okay, absolutely tiny nit, but should we export this fella too?

tonyhb (approved):
Looks good to me, my dude.  Good use of sjson!

We'll figure out JSON vs Cue at some point when we hit v1 :)

PR #230, opened 2022-08-25
Ensure function deploy errors are more specific

jpwilliams:
When running `inngest deploy ./functions` and encountering an error, it was unclear which function was triggering the error; let's sort that out here.


cmd/commands/deploy.go
```diff
@@ -78,7 +78,7 @@ func deploy(cmd *cobra.Command, args []string) error {
 		// Build all steps.
 		buildOpts, err := dockerdriver.FnBuildOpts(ctx, *fn, "--platform", "linux/amd64")
 		if err != nil {
-			return err
+			return fmt.Errorf("Failed to deploy function \"%s\": %s", fn.Name, err)
```
djfarrelly, line 81: There is a convention for Go errors to be lowercase (my linter yells at me for that), but I'm not sure how strict we stay to that so maybe Tony has more thoughts on that 😁

jpwilliams: Aye mine does do the same - I ignored here so that the CLI response starts with capitalisation, but you're right that this might not be very "Go-ish".

tonyhb: Really, we probably want a `publicerr` style package that we have in our cloud which can wrap and document errors with a public message.

We only show the public errors by default.  If you're running in verbose mode (`-v`) we can print all errors and the backtrace.

tonyhb (approved):

PR #231, opened 2022-08-25
Add static `schema.json` for Inngest config JSON Schema typing

jpwilliams:
## Summary

Adds a static `schema.json` in the root directory to be used to provide type hinting and validation to users while creating/editing Inngest config files.

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/inngest/inngest/main/schema.json",
  // other usual props...
}
```

We could also refer to specific versions of the schema, though I think it makes sense to target main so we're looking at the very latest valid version.

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/inngest/inngest/v0.5.5/schema.json",
  // other usual props...
}
```

I'd love to generate this directly from [pkg/cuedefs/v1/function.cue](https://github.com/inngest/inngest/blob/main/pkg/cuedefs/v1/function.cue), but our path to conversion doesn't yet support comments - see cue-lang/cue#1180. The comments are crucial to fulfil the purpose of teaching the user through tooling, hence the static version.

We could potentially introduce a very simple [Dangerfile](https://danger.systems/js/) to prompt in PRs if `function.cue` is changed and `schema.json` is not to ensure they're kept in sync until cue-lang/cue#1180 is resolved.

##  Next

- cue-lang/cue#1180
- Add to all [inngest/scaffolds](https://github.com/inngest/scaffolds)

### DIFF
--- a/examples/prisma-typescript-function/inngest.json
+++ b/examples/prisma-typescript-function/inngest.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://raw.githubusercontent.com/inngest/inngest/main/schema.json",
   "name": "Update your Prisma database from Stripe events",
   "description": "Create a record in Prisma when a Stripe webhook is received.",
   "tags": ["typescript", "prisma", "stripe"],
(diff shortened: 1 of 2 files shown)


tonyhb (approved):
This is pretty amazing!  Definitely a huge fan of adding these kinds of type hints.

PR #233, opened 2022-08-26
Add quick-start for processing images with Google Cloud Vision API and Sharp

jpwilliams:
Adds a new quick-start for processing images using the [Google Cloud Vision API](https://cloud.google.com/vision) and [Sharp](https://www.npmjs.com/package/sharp).

It covers:

- Simple multi-step flow (happy path, not-so-happy path)
- Using streams to process images
- Google API usage with service accounts
- Using secrets within the Inngest platform


tonyhb (approved):
This is great dude — very useful example to get started.

Should we bundle in a .env for an example google API key?

PR #234, opened 2022-08-26
Refactor init into form builder for reusability with `inngest step add`

tonyhb:


jpwilliams:
Ah awesome refactor! This is super exciting if we can get this feature out fast!

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/spf13/cobra"
+)
+
+var ()
+
+func NewCmdSteps() *cobra.Command {
+	root := &cobra.Command{
+		Use:    "step",
+		Short:  "",
+		Hidden: true,
```
jpwilliams, line 26: Why the choice to hide it?

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	byt, err := function.MarshalJSON(*fn)
+	if err != nil {
+		fmt.Println(cli.RenderError(fmt.Sprintf("Step '%s' already exists in this function", id)))
+		os.Exit(1)
+	}
+
+	if err := os.WriteFile(filepath.Join(fn.Dir(), "inngest.json"), byt, 0644); err != nil {
```
jpwilliams, line 170: Can't remember the exact export name - maybe this is right - might as well use it though :D

```suggestion
	if err := os.WriteFile(filepath.Join(fn.Dir(), function.JsonConfigName), byt, 0644); err != nil {
```

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/inngest/inngest/inngest"
+	"github.com/inngest/inngest/pkg/cli"
+	"github.com/inngest/inngest/pkg/cli/form"
+	"github.com/inngest/inngest/pkg/cli/form/formutil"
+	"github.com/inngest/inngest/pkg/function"
+	"github.com/inngest/inngest/pkg/scaffold"
+	"github.com/spf13/cobra"
+)
+
+var ()
```
jpwilliams, line 20: Dangling

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+	Language string
+}
+
+func addStep(cmd *cobra.Command, args []string) {
+	ctx := cmd.Context()
+	model := &stepModel{}
+
+	fn, err := function.Load(ctx, ".")
+	if err != nil {
+		fmt.Println(cli.RenderError(err.Error()))
```
jpwilliams, line 50: Should this be prefixed with something e.g. "_Failed to load funtion:_" so that we get context if there's an internalish error?

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+			),
+		},
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// Render interactive UI to ask questions.
+	if err := stepForm.Ask(); err != nil {
+		fmt.Println(cli.RenderError(err.Error()))
```
jpwilliams, line 132: Should this be prefixed with something e.g. "_Failed to ask question:_" so that we get context if there's an internalish error?

cmd/commands/step.go
```diff
@@ -0,0 +1,177 @@
...
+		Path: function.FilePrefix + path.Join("steps", id),
+		Runtime: inngest.RuntimeWrapper{
+			Runtime: inngest.RuntimeDocker{},
+		},
+	}
+	fn.Steps[id] = step
+	if model.Language != "Another language" {
+		mapping, err := scaffold.Parse(ctx)
+		if err != nil {
+			fmt.Println(cli.RenderError(err.Error()))
```
jpwilliams, line 154: Should this be prefixed with something e.g. "_Failed to parse scaffold:_" so that we get context if there's an internalish error?

pkg/scaffold/template.go
```diff
@@ -76,20 +80,89 @@ func (t Template) Render(f function.Function, step function.Step) error {
...
+	}
+
+	// Once complete, state should contain everything we need to create our
+	// function file.
+	byt, err := function.MarshalJSON(f)
+	if err != nil {
+		return fmt.Errorf("error creating JSON: %s", err)
+	}
+
+	if err := os.WriteFile(filepath.Join(root, "inngest.json"), byt, 0644); err != nil {
```
jpwilliams, line 99:
```suggestion
	if err := os.WriteFile(filepath.Join(root, function.JsonConfigName), byt, 0644); err != nil {
```

pkg/scaffold/template.go
```diff
@@ -76,20 +80,89 @@ func (t Template) Render(f function.Function, step function.Step) error {
...
+		if trigger.EventTrigger == nil {
+			continue
+		}
+
+		if trigger.EventTrigger.Definition == nil || trigger.EventTrigger.Definition.Def == "" {
+			// Use an empty event format.
+			trigger.EventTrigger.Definition = &function.EventDefinition{
+				Format: function.FormatCue,
+				Synced: false,
+				Def:    fmt.Sprintf(evtDefinition, strconv.Quote(trigger.Event)),
```
jpwilliams, line 124: Is this the piece that's embedding the event definition?

pkg/scaffold/template.go
```diff
@@ -76,20 +80,89 @@ func (t Template) Render(f function.Function, step function.Step) error {
...
+		}
+
+		cue, err := trigger.Definition.Cue(ctx)
+		if err != nil {
+			// XXX: We would like to log this as a warning.
+			continue
+		}
+
+		name := fmt.Sprintf("%s.cue", eventFilename(trigger.Event))
+		path := filepath.Join(root, "events", name)
```
jpwilliams, line 135: Can we move this `"events"` string to be a `const` or `var` somewhere? It's a nit, but I'm thinking that the higher it is, the more it becomes a standard.

pkg/scaffold/template.go
```diff
@@ -76,20 +80,89 @@ func (t Template) Render(f function.Function, step function.Step) error {
...
+		if err := os.WriteFile(path, []byte(cue), 0644); err != nil {
+			return fmt.Errorf("error writing event definition: %w", err)
+		}
+		f.Triggers[n].Definition.Def = fmt.Sprintf("file://./events/%s", name)
```
jpwilliams, line 139: (Then using the same new `const`/`var` here)

jpwilliams (commented):
Basically all nits and questions 😄

cmd/commands/step.go
tonyhb: Load takes care of human errors for us.

cmd/commands/step.go
tonyhb: This was for internal use during the last week

cmd/commands/step.go
tonyhb: This is taken care of in the scaffold package now :)

jpwilliams (approved):

PR #235, opened 2022-08-26
Add account federation example

tonyhb:
This adds an account federation example, running functions to add the user to external systems immediately after signup.

To do:
- [x] Add Readme
- [x] Add intercom step
- [x] Add slack step
- [x] Add Close.io step


examples/account-federation-step-function/steps/add-to-close-io/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY package*.json *.lock tsconfig.json .
+RUN npm install
+COPY . .
+RUN npm run build
+
+# This is the output image, which contains only the prebuilt JS.
+FROM root
+WORKDIR /opt/
+COPY package*.json /opt/
+RUN npm install --production
```
jpwilliams, line 19: All of these examples are using Yarn (`yarn.lock`), but then install deps using `npm`. I think this will result in us creating a new `package-lock.json` at build time which means our deps could shift.

We could either:

- Keep `yarn.lock` and add `yarn install --frozen-lockfile` here (we might also need to install Yarn)
- Replace `yarn.lock` with `package-lock.json` and add `npm ci` here

Personal preference would be to stick to npm as it ships with Node so is the assumed default, but I don't actually know npm vs. Yarn usage numbers.

examples/account-federation-step-function/steps/add-to-close-io/src/index.ts
```diff
@@ -0,0 +1,41 @@
+import type { Args } from "./types";
+import fetch from "node-fetch";
+
+export async function run({ event }: Args) {
+
+  const auth = btoa(`${process.env.CLOSE_API_KEY}:`);
```
jpwilliams, line 6: Ooh base64ing the env vars! Does this need documenting?

tonyhb: No, this is a close specific basic auth need :)

jpwilliams (approved):

PR #241, opened 2022-08-31
Allow `response` in wait field expressions

tonyhb:
EG this will use `waitFor` from the response of the current step as the
wait duration in seconds:

```
after: [{
  step: "foo",
  wait: "response.waitFor"
}]
```


jpwilliams (approved):
Oh that is such a cool idea.

Is the "current" step synonymous with the "parent" step? I see that we mention `response` is the "_output of the parent step_" in the docs: https://github.com/inngest/website/blob/main/pages/docs/_docs/3-functions/85-expressions.md?plain=1#L151

PR #252, opened 2022-09-06
Make `inngest run ./dir` recursive, like `inngest deploy ./dir`

jpwilliams:
## Summary

Replicates the functionality of `inngest deploy ./dir` (recursively deploying all functions found within `./dir`) to `inngest run`. It refactors out of the loading functionality of `inngest deploy` so we can use it in both places.

The drawback is that we don't have a reasonable way to support `--snapshot` or `--event-id` flags if multiple functions have been found.

The change is a small one, but I'm not sure that it complements the DX; it hamstrings some functionality around replay and data generation that doesn't feel good. What do you think, @tonyhb?

### DIFF
--- a/cmd/commands/deploy.go
+++ b/cmd/commands/deploy.go
@@ -49,21 +49,12 @@ func deploy(cmd *cobra.Command, args []string) error {
 		givenDir = args[0]
 	}
 
-	var fns []*function.Function
-	var err error
-
-	if givenDir != "" {
-		fns, err = function.LoadRecursive(ctx, givenDir)
-		if err != nil {
-			return err
-		}
-	} else {
-		fn, err := function.Load(ctx, ".")
-		if err != nil {
-			return err
-		}
-
-		fns = []*function.Function{fn}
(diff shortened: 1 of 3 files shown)


tonyhb:
A reasonable solution to snapshot / stdin / replay may be to treat `inngest run` as an ephemeral dev server, and only run the functions that are triggered by those specific events being used.

For example:

- Given you run `inngest run --replay .` within a monorepo containing 20 functions, each with different event triggers;
- We generate a new seed;
- This seed randomly shuffles the triggers and fetches N events for the specific triggers (one trigger or many)
- We check all of the functions and only trigger / run the appropriate functions.

I think that this is expected behaviour if you pipe in via `stdin` or use snapshot files.  The only thing is, it could be surprising if you `inngest run --replay .` and expect _all_ functions to run with appropriate events.  Maybe we need to automatically set --number to the number of found triggers, and take 1 event from each trigger?

That said, we have an issue already in which if your trigger has an expression to conditionally run, `--replay` might pull an event that doesn't match the function :)

tonyhb:
Think we can close this one and move on :)

PR #256, opened 2022-09-22
Add `inngest types typescript|ts` command

jpwilliams:
Adds a new command: `inngest types`, used to generate types from Inngest Cloud events.

inngest_types_test

```
Generate types from your Inngest Cloud account.

Usage:
  inngest types [command]

Examples:
inngest types typescript

Available Commands:
  typescript  Generate TypeScript types in a .ts file

Flags:
  -c, --check           Compare found types with an existing generated file defined by --output, failing if they're different. Useful for CI to check you are up to date
  -h, --help            help for types
  -o, --output string   Specify the location of the generated .ts file (default "./__generated__/inngest.ts")

Global Flags:
      --json      Output logs as JSON.  Set to true if stdout is not a TTY.
      --prod      Use the production environment for the current command.
  -v, --verbose   Enable verbose logging.

Use "inngest types [command] --help" for more information about a command.
```

Intended use for the [SDK](https://github.com/inngest/inngest-js) is via `npx` for ease of use:
```sh
npx inngest-cli types ts
```

For SDK dev, this might be added to a `predev` script, so when running `yarn|npm run dev`, we'll automatically pull Inngest Cloud types first:
```json
{
  "scripts": {
    "dev": "next dev",
    "predev": "npx inngest-cli types ts"
  }
}
```

You can customise the output file, which defaults to `./__generated__/inngest.ts`:
```sh
npx inngest-cli types ts --output foo.ts
```

You can also add a CI check to ensure that your local types are up-to-date with Inngest Cloud; useful for protecting deploys from breaking changes. It will find your `inngest.ts` file (either at the default path or specified via `--output`), download types from Inngest Cloud, and compare the two. It'll return `0` if they're the same and `1` if not.
```sh
npx inngest-cli types ts --check
```
You could put this in your deployment script that's run in CI:
```json
{
  "scripts": {
    "build": "next build",
    "deploy": "npx inngest-cli types ts --check && npm run build"
  }
}
```

### DIFF
--- a/cmd/commands/login.go
+++ b/cmd/commands/login.go
@@ -41,7 +41,7 @@ func doLogin(cmd *cobra.Command, args []string) {
 	} else if token != "" {
 		tokenAuth(ctx)
 	} else {
-		deviceAuth(ctx)
+		DeviceAuth(ctx)
 	}
 
 	if err := fetchAccount(cmd.Context()); err != nil {
@@ -63,7 +63,7 @@ func tokenAuth(ctx context.Context) {
 	}
 }
 
-func deviceAuth(ctx context.Context) {
+func DeviceAuth(ctx context.Context) {
 	state := clistate.RequireState(ctx)
 	start, err := state.Client.StartDeviceLogin(ctx, state.ClientID)
 	if err != nil {
--- a/cmd/commands/types.go
+++ b/cmd/commands/types.go
@@ -166,11 +166,11 @@ func typescript(cmd *cobra.Command, args []string) (string, error) {
 	}
 
 	// Create a catch-all type that we'll use when creating clients
-	b.WriteString("export type Events = {\n")
+	b.WriteString("export type Events = Readonly<{\n")
 	for eventId, tsEventName := range eventNames {
-		b.WriteString(fmt.Sprintf("  \"%s\": %s;\n", eventId, tsEventName))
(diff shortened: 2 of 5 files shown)


cmd/commands/login.go
```diff
@@ -63,7 +63,7 @@ func tokenAuth(ctx context.Context) {
 	}
 }
 
-func deviceAuth(ctx context.Context) {
+func DeviceAuth(ctx context.Context) {
```
jpwilliams, line 66: I kinda want to allow any type of automatic auth here - including passing in a token - but it feels odd to duplicate the flags.

It just supports `DeviceAuth` right now for local dev; we should probably look at allowing `--token` and `--username` at a later date.

cmd/commands/types.go
```diff
@@ -0,0 +1,176 @@
...
+		if !strings.Contains(err.Error(), "not logged in") {
+			return "", err
+		}
+
+		// If we got an error because we're not logged in, try logging in
+		// and attempt once more
+		DeviceAuth(ctx)
+		events, err = initialize.FetchEvents()
+		if err != nil {
+			return "", err
```
tonyhb, line 116: Hmm, what if you only want to get public events here?  Should we force a login? 🤔

tonyhb (approved):

PR #259, opened 2022-09-29
Add SDK auto-discovery to the dev server

tonyhb:
This commit introduces a new devserver service which autodiscovers SDKs hosted on the local machine via supported default ports and paths.

Any SDKs found are added to a list, and the dev server polls the SDK endpoint to auto-register functions every second.


pkg/devserver/discovery.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// openPorts simultaneously checks all supported localhost ports to see if
+// any are open. This allows us to filter the default ports to only those
+// that are serving connections prior to making HTTP requests.
+func openPorts(ctx context.Context) []int {
+	results := []int{}
+	// Create a buffered channel with the number of ports, letting us push
+	// valid ports without reading from the channel at the same time.
+	found := make(chan int, len(Ports))
```
djfarrelly, line 70: Should we also check for a `PORT` environment variable? That could be an enhancement in future versions.

pkg/devserver/service.go
```diff
@@ -0,0 +1,175 @@
...
+// This lets the dev server start and wait for the SDK server to come up at
+// any point.
+func (d *devserver) autodiscover(ctx context.Context) {
+	logger.From(ctx).Info().Msg("autodiscovering locally hosted SDKs")
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+		d.urlLock.Lock()
+		d.urls = Autodiscover(ctx)
```
djfarrelly, line 127: If a user passes urls via the dev server options, does this overwrite those URLs? Should this merge them instead?

djfarrelly (commented):
Looking good Tony! I had some questions and one potential issue, but otherwise it looks awesome - the autodiscover is really great!

pkg/devserver/api.go
```diff
@@ -0,0 +1,156 @@
...
+		return http.HandlerFunc(fn)
+	})
+
+	a.Get("/", a.UI)
+	a.Get("/dev", a.Info)
+	a.Post("/fn/register", a.Register)
+}
+
+func (a devapi) UI(w http.ResponseWriter, r *http.Request) {
+	// TODO: Respond with a UI here.
```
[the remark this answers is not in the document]
tonyhb: It won't cause an error: it'll be a blank page :)

pkg/devserver/discovery.go
tonyhb: In this case I think you'd use `--sdk-url` / `-u` to specify the canonical URL you'd like to hit (which is usually 127.0.0.1:whatever!) :)

pkg/devserver/service.go
tonyhb: This only runs if URLs aren't supplied;  when URLs are supplied we don't run autodiscovery.  We could probably change that, but for now figured it was safe 🤔 

WDYT?

pkg/devserver/discovery.go
djfarrelly: Yeah - I was just thinking for the sake of auto-discovery it would be a cool automatic setup. If we extended #218 to also make environment variables available in the Dev Server code, devs who use Next.js and set `PORT` would just be able to run `inngest dev` in their project directory and it would just pick it up automatically not having to specify the URL arg. With how smart the auto-discovery already is, it could be a nice addition - not blocking merging this PR.

pkg/devserver/service.go
djfarrelly: Oh I see now - I didn't traverse to where the `Autodiscover` option was set in the parsing of args. This makes sense then!

pkg/devserver/discovery.go
tonyhb: Added: https://github.com/inngest/inngest/pull/259/files#diff-f33add88ab91a28b9194b39a63ae00da34d0c7928432bf7aa0d8eec7cb0a9eb4R30-R36.  Thanks1

pkg/devserver/discovery.go
```diff
@@ -0,0 +1,105 @@
...
+	"net/http"
+	"strconv"
+	"sync"
+	"time"
+)
+
+var (
+	// Ports indicate the default ports that we attempt to scan on localhost
+	// when discovering SDK-based endpoints
+	Ports = []int{80, 3000, 3001, 3002, 3003, 3004, 3005, 5000, 8000, 8080, 8081, 8888}
```
jpwilliams, line 17: If my process was running on a port not specified here, what would be my action?

Either we could add the ability to specify extra ports to scan, or we could put the onus on the SDK UI to find the dev server and register itself?

pkg/sdk/sdk.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+var (
+	ErrNoFunctions = fmt.Errorf("No functions registered within your app")
+)
+
+// RegisterRequest represents a new deploy request from SDK-based functions.
+// This lets us know that a new deploy has started and that we need to
+// upsert function definitions from the newly deployed endpoint.
+type RegisterRequest struct {
```
jpwilliams, line 19: **Thinking out loud** - this is the number one place where the comms could fail, where we mismatch expected data shapes.

Is there anything we could do programmatically or otherwise to ensure consistency across the various repos that'll use this?

jpwilliams (approved):

pkg/sdk/sdk.go
tonyhb: Hmm, a few things come to mind:

- We added the `v` tag specifically so that we can version this payload, allowing us to release new versions of this while maintaining backcompat.  This should minimize mismatched data once you've got a working implementation — this will only crop up when you're building a new SDK for the first time.
- We can & do report errors to the SDK via HTTP!

pkg/devserver/discovery.go
tonyhb: I think @djfarrelly had the same question and recommended the PORT env var, which we've added.  This will add your PORT to this list for autodiscovery.

PR #261, opened 2022-10-18
INN-706 `inngest dev` panic with no Docker

jpwilliams:
Fixes `inngest dev` always requiring Docker, even if there are no local functions to load. It will now assume that it requires Docker if any local functions are found, else it'll not load the Docker driver.

It also fixes a small nil pointer bug where `API#Stop()` would panic if the server doesn't exist.


djfarrelly (approved):

tonyhb (approved):
LGTM.  Definitely safest to do that here.

PR #262, opened 2022-10-19
Add CORS headers to the event API

tonyhb:
This allows arbitrary sites to post events, as expected.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/cors"
 	"github.com/inngest/inngest/pkg/config"
 	"github.com/inngest/inngest/pkg/event"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


pkg/api/api.go
```diff
@@ -42,6 +43,18 @@ func NewAPI(o Options) (chi.Router, error) {
...
 
+	cors := cors.New(cors.Options{
+		// AllowedOrigins is used instead of AllowOriginFunc as it's faster and we don't need
+		// to allow credentials, so an asterisk works.
+		AllowOriginFunc:  func(r *http.Request, origin string) bool { return true },
+		AllowedMethods:   []string{"GET", "POST", "PUT", "DELETE", "OPTIONS"},
+		AllowedHeaders:   []string{"Accept", "Authorization", "Content-Type"},
+		ExposedHeaders:   []string{"Link"},
+		AllowCredentials: false,
+		MaxAge:           60 * 60, // 1 hour
```
jpwilliams, line 54: Do we need this set to >0? Not sure what we want cached atm

pkg/api/api.go
```diff
@@ -42,6 +43,18 @@ func NewAPI(o Options) (chi.Router, error) {
 		log:     &logger,
 	}
 
+	cors := cors.New(cors.Options{
+		// AllowedOrigins is used instead of AllowOriginFunc as it's faster and we don't need
+		// to allow credentials, so an asterisk works.
+		AllowOriginFunc:  func(r *http.Request, origin string) bool { return true },
+		AllowedMethods:   []string{"GET", "POST", "PUT", "DELETE", "OPTIONS"},
```
djfarrelly, line 50: Do we need anything other than POST?

pkg/api/api.go
tonyhb: It doesn't hurt — if you're sending > 1 event you'll stop doing another OPTIONS request which is good.

pkg/api/api.go
tonyhb: Not yet, though even if they're unused I'm assuming that we're going to add more capabilities to manipulate the dev server from a UI so it's not harmful :)

PR #263, opened 2022-10-19
Prevent "functions registered" log spam

tonyhb:
The SDK sends over a hash during each registration request (https://github.com/inngest/inngest-js/pull/27), which we use to prevent duplicate registrations and therefore duplicate logs.


djfarrelly (approved):
LGTM

PR #287, opened 2022-11-22
Add state history asserts

tonyhb:
This PR adds assertions to state mutations for logs.


pkg/execution/state/inmemory/inmemory.go
```diff
@@ -278,25 +293,30 @@ func (m *mem) SaveResponse(ctx context.Context, i state.Identifier, r state.Driv
...
+				Attempt: attempt,
 			},
 		})
 	} else {
 		instance.errors[r.Step.ID] = r.Err
 
+		typ := enums.HistoryTypeStepErrored
+		if r.Final() {
+			typ = enums.HistoryTypeStepFailed
+		}
```
jpwilliams, line 309: Good catch!

pkg/execution/state/redis_state/key_generator.go
```diff
@@ -53,8 +53,8 @@ type KeyGenerator interface {
 	// PauseStep returns the key used to store a pause ID by the run ID and step ID.
 	PauseStep(context.Context, state.Identifier, string) string
 
-	// Log returns the key used to store a log entry for run hisotry
-	Log(context.Context, state.Identifier) string
+	// History returns the key used to store a log entry for run hisotry
+	History(context.Context, state.Identifier) string
```
jpwilliams, line 57: 🙌

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -380,59 +382,127 @@ func (m mgr) Load(ctx context.Context, id state.Identifier) (state.State, error)
...
-			return nil, err
+	} else {
+		typ = enums.HistoryTypeStepErrored
+		data = r.Err.Error()
+		if r.Final() {
+			typ = enums.HistoryTypeStepFailed
+			funcFailHistory = state.History{
+				Type:       enums.HistoryTypeFunctionFailed,
+				Identifier: i,
+				CreatedAt:  now,
```
jpwilliams, line 407: Will this store the time in microseconds? Is this okay?

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -757,7 +823,7 @@ type runMetadata struct {
 func (r runMetadata) Map() map[string]any {
 	return map[string]any{
 		"version":       r.Version,
-		"status":        r.Status,
+		"status":        int(r.Status), // Always store this as an int
```
jpwilliams, line 826: I'm guessing with these enums, then, that we should always add new values to the end of the list, else we risk skewing the data? 😬

jpwilliams (approved):

pkg/execution/state/redis_state/redis_state.go
tonyhb: Yep, definitely.  New enums go at the end otherwise the values for those enums change.

pkg/execution/state/redis_state/redis_state.go
tonyhb: This is encoded by the binary encoder which uses JSON marshalling with gzip under the hood, so we're okay.

PR #292, opened 2022-11-30
Add JSON highlighting to UI

jpwilliams:
## Summary

Adds syntax highlighting to existing `<CodeBlock>` elements. It's a tiny Nord theme with some mods (mainly removing background colours).

@EdPoole, I feel like pink and green are the wrong colours. 😅 Should be able to modify `highlight.min.css` (and use rules like `@apply bg-slate-800`) to get something more reasonable. Is this an alright foundation to work off of?

## Example

image


tonyhb (approved):
LGTM!

PR #294, opened 2022-11-30
Event stream animations

jpwilliams:
## Summary

Adds simple animations to the event stream and function logs using [@formkit/auto-animate](https://www.npmjs.com/package/@formkit/auto-animate) - great find, @EdPoole!

## Example

animationsoo


tonyhb (approved):

tonyhb:
This is so nice!

PR #300, opened 2022-12-02
Squash timeline

tonyhb:
Squashes the function log timeline to hide noisy events, eg. "scheduled", "started" if the step has finished.

### DIFF
--- a/cmd/commands/dev.go
+++ b/cmd/commands/dev.go
@@ -22,6 +22,7 @@ func NewCmdDev() *cobra.Command {
 	cmd.Flags().StringP("port", "p", "8288", "port to run the API on")
 	cmd.Flags().String("dir", ".", "directory to load functions from")
 	cmd.Flags().StringSliceP("sdk-url", "u", []string{}, "SDK URLs to load functions from")
+	cmd.Flags().Bool("docker", false, "enable docker-based steps")
 
 	return cmd
 }
@@ -47,10 +48,12 @@ func doDev(cmd *cobra.Command, args []string) {
 	}
 
 	urls, _ := cmd.Flags().GetStringSlice("sdk-url")
+	docker, _ := strconv.ParseBool(cmd.Flag("docker").Value.String())
 	opts := devserver.StartOpts{
 		Config:       *conf,
 		RootDir:      cmd.Flag("dir").Value.String(),
 		URLs:         urls,
+		Docker:       docker,
 		Autodiscover: len(urls) == 0,
 	}
 
--- a/pkg/coreapi/graph/resolvers/events.go
+++ b/pkg/coreapi/graph/resolvers/events.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/inngest/inngest/pkg/coreapi/graph/models"
+	"github.com/oklog/ulid/v2"
 )
 
 func (r *queryResolver) Event(ctx context.Context, query models.EventQuery) (*models.Event, error) {
@@ -20,7 +21,14 @@ func (r *queryResolver) Event(ctx context.Context, query models.EventQuery) (*mo
 	}
 
 	evt := evts[0]
+
 	createdAt := time.UnixMilli(evt.Timestamp)
+	if evt.Timestamp == 0 {
+		if id, err := ulid.Parse(evt.ID); err == nil {
+			createdAt = ulid.Time(id.Time())
+		}
+	}
+
 	payloadByt, err := json.Marshal(evt.Data)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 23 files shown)


pkg/devserver/devserver.go
```diff
@@ -45,12 +46,15 @@ func New(ctx context.Context, opts StartOpts) error {
...
 	// service starting.
-	el, err := prepareDockerImages(ctx, opts.RootDir)
-	if err != nil {
-		return err
+	el, _ := inmemorydatastore.NewFSLoader(ctx, opts.RootDir)
+	if opts.Docker {
+		var err error
+		if el, err = prepareDockerImages(ctx, opts.RootDir); err != nil {
+			return err
+		}
```
jpwilliams, line 54: 🙌 Awesome!

pkg/coreapi/graph/resolvers/events.go
```diff
@@ -20,7 +21,14 @@ func (r *queryResolver) Event(ctx context.Context, query models.EventQuery) (*mo
 	}
 
 	evt := evts[0]
+
 	createdAt := time.UnixMilli(evt.Timestamp)
+	if evt.Timestamp == 0 {
```
jpwilliams, line 26: Is it still a bug that `evt.Timestamp` isn't defined here?

cmd/commands/dev.go
```diff
@@ -22,6 +22,7 @@ func NewCmdDev() *cobra.Command {
 	cmd.Flags().StringP("port", "p", "8288", "port to run the API on")
 	cmd.Flags().String("dir", ".", "directory to load functions from")
 	cmd.Flags().StringSliceP("sdk-url", "u", []string{}, "SDK URLs to load functions from")
+	cmd.Flags().Bool("docker", false, "enable docker-based steps")
```
jpwilliams, line 25: Do we need to alter any docs for this?

pkg/execution/executor/service.go
```diff
@@ -358,7 +358,7 @@ func (s *svc) handleQueueItem(ctx context.Context, item queue.Item) error {
 		// and this is a no-op - things should be atomic where possible.
 		//
 		// TODO: Add a unit test to ensure WG is 0 at the end of execution.
-		if err := s.state.Scheduled(ctx, item.Identifier, next.Incoming, 0, nil); err != nil {
+		if err := s.state.Scheduled(ctx, item.Identifier, next.Incoming, 0, &at); err != nil {
```
jpwilliams, line 361: Ah - good catch

ui/src/components/Event/Stream.tsx
```diff
@@ -3,6 +3,7 @@ import { selectEvent } from "../../store/global";
 import { useAppDispatch, useAppSelector } from "../../store/hooks";
 import TimelineFeedContent from "../Timeline/TimelineFeedContent";
 import TimelineRow from "../Timeline/TimelineRow";
+import { decodeTime } from 'ulid'
```
jpwilliams, line 6: Dangling import?

ui/src/components/Function/Stream.tsx
```diff
@@ -1,4 +1,5 @@
 import {
+    FunctionRunEvent,
```
jpwilliams, line 2: Dangling import?

ui/src/components/Time.tsx
```diff
@@ -1,23 +1,27 @@
...
 
-  return <TimeAgo date={date} formatter={formatter} title={tooltip} />;
+  // Refresh each second such that this component re-renders times appropriately.
+  const [time, setTime] = useState(new Date().valueOf());
+  useEffect(() => {
+    const interval = setInterval(() => { setTime(new Date().valueOf()) }, 1000);
+    return () => { clearInterval(interval); };
+  }, []);
+
+  return <TimeAgo date={date} formatter={formatter} title={tooltip} key={time} />;
```
jpwilliams, line 26: [react-timeago](https://www.npmjs.com/package/react-timeago) should already do this

jpwilliams (commented):
Awesome! We have [INN-768](https://linear.app/inngest/issue/INN-768); this feels like a sensible mid point to make the timelines easy to view. 🙌

Only a couple of notes/nits

pkg/coreapi/graph/resolvers/events.go
tonyhb: It is - that needs to be fixed with additional unit tests in a followup PR

cmd/commands/dev.go
tonyhb: This is legacy and isn't supported for `inngest dev`, so we shouldn't!

ui/src/components/Event/Stream.tsx
tonyhb: thanks! :derp:

ui/src/components/Time.tsx
tonyhb: It doesn't, for some reason :/

pkg/devserver/devserver.go
djfarrelly: Big step here! 🔥

djfarrelly (approved):

PR #314, opened 2022-12-16
Add `make dev` command for command local dev build

jpwilliams:
I keep adding this to my `make build` command and @djfarrelly needed the same this week.

Might as well add it.


tonyhb (approved):

tonyhb:
Rebased and pushed on main;  this should most definitely be green!

PR #317, opened 2022-12-16
Order generated event IDs during catch-all type generation for TS

### LINKED ISSUE #316: `inngest types ts` generates different ordering of types every time it is run
**Describe the bug**

When running `inngest types ts` at different times without changing the event schemas on Inngest Cloud, the ordering of the event keys in the generated code changes randomly. This produces unnecessary diffs when working on unrelated code changes.

**To Reproduce**

1. Create a Next.js project using Yarn
2. Add `inngest` and `inngest-cli` as dependencies
3. Add a `predev` script to package.json as described in [this doc](https://www.inngest.com/docs/typescript)
4. Run `yarn dev`
5. Commit the changes
6. Run `yarn dev` again
7. See that the ordering of the event keys changes producing a diff

**Expected behavior**

I would expect that the code generation would be deterministic with respect to the ordering given no additions or deletions. Ideally it is ordered alphabetically, so as to respect the event prefixes.

CleanShot 2022-12-16 at 11 27 05@2x

CleanShot 2022-12-16 at 11 05 53@2x

**System info**
- inngest-cli version: 0.8.3
- OS: macOS
- Version 13.0.1 (Ventura)

**Additional context**

I also tried not installing the `inngest-cli` as a project dependency, and using `npx` directly and I had the same behaviour.

jpwilliams:
## Summary

During TS type generation, ensure event IDs are sorted before writing file to avoid pointless diffs as seen in #316.

The types themselves were correctly ordered in this fashion, though not the catch-all exported `Events` type.

## Related

- Fixes #316

### DIFF
--- a/cmd/commands/types.go
+++ b/cmd/commands/types.go
@@ -183,8 +183,13 @@ func typescript(cmd *cobra.Command, args []string) (string, error) {
 
 	// Create a catch-all type that we'll use when creating clients
 	b.WriteString("type GeneratedEvents = Readonly<{\n")
-	for eventId, tsEventName := range eventNames {
-		b.WriteString(fmt.Sprintf("  \"%s\": Readonly<%s>;\n", eventId, tsEventName))
+	unorderedEventIds := make([]string, 0, len(eventNames))
+	for eventId := range eventNames {
+		unorderedEventIds = append(unorderedEventIds, eventId)
+	}
+	sort.Strings(unorderedEventIds)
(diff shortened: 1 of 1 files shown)


tonyhb:
This looks good!  We're also going to need to order all fields - probably alphabetically 😭 

I can follow up with a PR in addition to this :)

tonyhb (approved):

djfarrelly (approved):

PR #322, opened 2022-12-19
Use current hostname for GraphQL requests

### LINKED ISSUE #303: Running `inngest-cli dev` on remote server
I'm running `npx inngest-cli dev --host server_ip` on a remote Ubuntu server. However when I go to the CLI page `http://server_ip:8288` ... the page tries to make requests to `http://localhost:8300/gql`

image

Is there a way to change the `coreapi` URL so it uses my remote server's IP instead of localhost? There doesn't appear to be a way to change the coreapi `host` via the CLI options.

Running with `--host` changes the `api` host just fine... etc `http://server_ip:8288` but it keeps the `coreapi` host as `0.0.0.0:8300` ... which seems like a bug to me?

```
root@remote_server:~# npx inngest-cli dev --host server_ip
3:03PM INF runner > starting event stream backend=inmemory
3:03PM INF runner > starting queue backend=inmemory
3:03PM INF runner > initializing scheduled messages len=0
3:03PM INF devserver > autodiscovering locally hosted SDKs
3:03PM INF devserver > service starting
3:03PM INF executor > starting queue backend=inmemory
3:03PM INF api > starting server addr=server_ip:8288
3:03PM INF executor > service starting
3:03PM INF executor > subscribing to function queue
3:03PM INF runner > service starting
3:03PM INF runner > subscribing to events topic=events
3:03PM INF coreapi > service starting
3:03PM INF coreapi > starting server addr=0.0.0.0:8300
```

djfarrelly:
## Description

If the user decides to run Inngest on another ip or host, the UI should make requests to the core api on that same host.

Fixes #303.


jpwilliams (approved):
Awesome!

I'll clean this up by removing the `index.html` diff and merge!

PR #331, opened 2023-01-04
Add tally metrics to queue

tonyhb:
This PR adds basic tally metrics to the Redis queue.  The following is instrumented:

- Counter for items enqueued, started, finished, and errored (individually)
- Gauge for item latency EWMA
- Histogram for item latency
- Counter for number of times sequential leases are claimed

We also move to a better EWMA package.


pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -42,6 +57,10 @@ func (q *queue) Enqueue(ctx context.Context, item osqueue.Item, at time.Time) er
 		queueName = &name
 	}
 
+	go q.scope.Tagged(map[string]string{
+		"kind": item.Kind,
+	}).Counter("queue_items_enqueued_total").Inc(1)
```
jpwilliams, line 62: Nit: Can all of these strings be consts?

jpwilliams (approved):

pkg/execution/state/redis_state/queue_processor.go
tonyhb: Yep, good idea!  I'll push out a commit to do that.

PR #341, opened 2023-01-11
Disable and error on dev server redirects to the Inngest handler.

### LINKED ISSUE #340: Log an error if the SDK redirects from /api/inngest during dev registration
**Describe the bug**

As a user, I might set up auth for my API.  If the auth layer redirects `/api/inngest` requests we should warn within the dev server's console (and UI?) so that we know how to fix the SDK & routing.

tonyhb:
Should fix #340 - but will require tests.

### DIFF
--- a/pkg/devserver/discovery.go
+++ b/pkg/devserver/discovery.go
@@ -43,6 +43,9 @@ var (
 
 	hc = http.Client{
 		Timeout: timeout,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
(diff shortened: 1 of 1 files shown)


jpwilliams:
@tonyhb I'm thinking some redirects might be expected, though it's only a problem if we don't eventually receive comms with the SDK.

Could we perhaps change this to _detect_ a redirect and then in the case where we fail to see the SDK we can throw a redirection warning, else all is fine?

PR #351, opened 2023-01-16
[WIP] Step parallelism

tonyhb:
This PR introduces step parallelism by:

- Adding support for > 1 generator opcode from SDKs
- Adding stacks to function state
- Passing stack index pointers within drivers


pkg/execution/executor/service.go
```diff
@@ -478,12 +479,13 @@ func (s *svc) scheduleGeneratorResponse(ctx context.Context, item queue.Item, r
 				return s.queue.Enqueue(ctx, item, time.Now())
 			case enums.OpcodeStep:
 				// Re-enqueue the exact same edge to run now.
-				if err := s.state.Scheduled(ctx, item.Identifier, edge.Edge.Incoming, 0, nil); err != nil {
+				if err := s.state.Scheduled(ctx, item.Identifier, "step", 0, nil); err != nil {
```
tonyhb, line 482: Hmm, why is this hard-coded?  This is a generator edge, which means "replay the current edge" - edge.Edge.Incoming should be of type "step".

pkg/execution/runner/runner.go
```diff
@@ -447,9 +447,20 @@ func (s *svc) pauses(ctx context.Context, evt event.Event) error {
 			return err
 		}
 
+		logger.From(ctx).Debug().
+			Str("pause_id", pause.ID.String()).
+			Str("run_id", pause.Identifier.RunID.String()).
+			Msg("consuming pause")
+
+		idx, err := s.state.ConsumePause(ctx, pause.ID, evtMap)
```
tonyhb, line 455: Moving `consumePause` to above enqueue adds a failure condition which means the function may never be ran:

- If enqueueing fails, the pause is already consumed.
- This means that no new events will trigger the pause, but the step doesn't run.

Can we move this back?  Any particular reason this was moved?

tonyhb: If we need the index, shouldn't this be a part of the pause itself vs "whatever the index is now"?  The index is mutated by any ongoing parallel work while this step is paused.

jpwilliams: Mm was thinking this as we had a similar problem with enqueuing the next step outside of pauses.

The stack only contains resolved work, so the index needs to be the length of the stack as a result of inserting the resolved pause; we can't enqueue the next task before we've mutated the stack. 🤔

pkg/execution/executor/service.go
jpwilliams: 👍 This can likely be removed - most probably a remnant of me throwing changes at it to see what breaks. 😄

PR #373, opened 2023-01-24
Use request body on redirect

tonyhb:


djfarrelly (approved):
Looks good to me - this is a good change that I wouldn't have considered before, but I like how simple it is to handle it!

PR #379, opened 2023-02-01
Move to React from Preact

jpwilliams:
## Summary

We've been seeing OOM errors when using Monaco with Preact. Using React seems to fix this.

We were using Preact to reduce bundle size, but for the CLI it doesn't matter quite as much as the SDK which will be bundled with apps.


ui/src/components/Event/SendEventModal.tsx
```diff
@@ -175,8 +169,8 @@ export const SendEventModal = () => {
             }}
             options={{
               fixedOverflowWidgets: true,
-              formatOnPaste: true,
-              formatOnType: true,
+              formatOnPaste: false,
+              formatOnType: false,
```
djfarrelly, line 173: Oh interesting - I only noticed this last night when I removed `input` from `value={}`. It wasn't working before me, I don't mind this either way though.

djfarrelly (approved):
LGTM @jpwilliams.

Are some of the re-formatting prettier-on-save or something? Just wondering if we should include an npm run script that does that for folks that may not have that setup.

tonyhb (approved):

PR #399, opened 2023-02-23
Add ability to send logs in SDK generator steps

jpwilliams:
## Summary

Logging is a common need which we can meet with step tooling. This PR supports inngest/inngest-js#109 by adding support for `OpcodeLog` to the executor, state store, and dev server UI.

See inngest/inngest-js#109 for end-user usage.

## Related

- [x] inngest/inngest-js#96
- [x] inngest/inngest-js#109
- [ ] Handle additional log op codes when running a step


tonyhb:
Closing this as our implementation is moving SDK side.

PR #402, opened 2023-02-28
Cconcurrency limits

tonyhb:
This PR introduces concurrency limits into our queueing system.  Concurrency limits are typically handled at the worker level, but we handle it at the queue level.

There are three separate concurrency toggles:

- Partition concurrency limits;  these are limits for a specific function/queue in the queue
- Global concurrency limits;  these are limits for a specific tenant
- Custom concurrency limits;  these are limits for a specific key, such as a step.

Because we lease partitions, we check the partition concurrency limits at the start of job processing when leasing, and store the counter in memory.  Global and custom limits are checked each time we lease an item.


pkg/execution/state/redis_state/key_generator.go
```diff
@@ -139,8 +139,19 @@ type QueueKeyGenerator interface {
 	// Sequential returns the key which allows a worker to claim sequential processing
 	// of the partitions.
 	Sequential() string
+	// Scavenger returns the key which allows a worker to claim scavenger processing
+	// of the partitions for lost jobs
+	Scavenger() string
 	// Idempotency stores the map for storing idempotency keys in redis
 	Idempotency(key string) string
+	// ConcurrencyKey returns a key for a given concurrency string.  This stores an ordered
```
djfarrelly, line 147:
```suggestion
	// Concurrency returns a key for a given concurrency string.  This stores an ordered
```

PR #409, opened 2023-03-09
Add support for external concurrency services

tonyhb:
This PR allows out-of-queue concurrency limiting via any external service.  The queue adds items to the concurrency queue, but external code is expected to remove items from the queue;  this doesn't happen automatically.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -22,10 +22,10 @@ jobs:
         run: |
           curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.50.1
           ./bin/golangci-lint run --verbose
-  test:
+  test-linux-race:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -428,7 +442,7 @@ func (q *queue) processPartition(ctx context.Context, p *QueuePartition, f osque
 			case ErrPartitionConcurrencyLimit, ErrConcurrencyLimit:
 				q.scope.Counter(counterConcurrencyLimit).Inc(1)
 				q.sem.Release(1)
-				return nil
+				return err
```
goodoldneon, line 445:
```suggestion
				// Since the queue is at capacity, return the error so that we
				// don't keep hammering with "does the queue have room?" logic.
				return err
```

.github/workflows/go.yaml
```diff
@@ -35,3 +35,16 @@ jobs:
           go-version: 1.18
       - name: Test
         run: go test $(go list ./... | grep -v tests) -race -count=1
+  test-windows:
```
jpwilliams, line 38: Why separated, out of interest?

jpwilliams (approved):

.github/workflows/go.yaml
tonyhb: We use the race condition checker on linux, and on windows run the usual tests.  Running the race checker on windows is slower than linux, which makes some tests flaky as we sometimes assert latency & timing on tests :(

PR #410, opened 2023-03-13
Add `inngest/function.failed` internal event for fn failures

jpwilliams:
## Summary

Adds the sending of an internal `inngest/function.failed` event when a function fails for the final time.

Unsure about the required use of `EventStream` here; it could be better.

## Todo

- [x] Create `WithFailureHandler()` for when instantiating an executor - takes a function to call when a run has failed for the final time
- [x] Ensure executor has no knowledge of required pubsub/eventstream/etc
- [ ] Tests

## Related

- inngest/inngest-js#133

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -2,23 +2,18 @@ package executor
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/inngest/inngest/inngest"
 	"github.com/inngest/inngest/pkg/config"
 	"github.com/inngest/inngest/pkg/coredata"
 	"github.com/inngest/inngest/pkg/enums"
-	"github.com/inngest/inngest/pkg/event"
 	"github.com/inngest/inngest/pkg/execution/driver"
 	"github.com/inngest/inngest/pkg/execution/queue"
 	"github.com/inngest/inngest/pkg/execution/state"
 	"github.com/inngest/inngest/pkg/logger"
-	"github.com/inngest/inngest/pkg/pubsub"
-	"github.com/oklog/ulid/v2"
 	"github.com/rs/zerolog"
 )
 
@@ -116,10 +111,6 @@ func NewExecutor(opts ...ExecutorOpt) (Executor, error) {
 		return nil, ErrNoActionLoader
 	}
 
-	if m.pb == nil {
-		return nil, ErrNoPublisher
-	}
-
 	return m, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/execution/executor/executor.go
```diff
@@ -146,6 +146,13 @@ func WithConfig(c config.Execution) ExecutorOpt {
 	}
 }
 
+func WithFailureHandler(f *func(context.Context, state.Identifier, state.State, state.DriverResponse) error) ExecutorOpt {
```
tonyhb, line 149: We could type this handler for readability/ease of use:  you can then use this type here and in the struct definition.


```
type FailureHandler func(context.Context, state.Identifier, state.State, state.DriverResponse) error

func WithFailureHandler(f FailureHandler) ExecutorOpt {
```

pkg/execution/executor/executor.go
```diff
@@ -173,6 +180,7 @@ type executor struct {
 	sm             state.Manager
 	al             coredata.ExecutionActionLoader
 	runtimeDrivers map[string]driver.Driver
+	failureHandler *func(context.Context, state.Identifier, state.State, state.DriverResponse) error
```
tonyhb, line 183: This doesn't need to be a pointer, esp. when we convert this to a type :)

pkg/execution/executor/executor.go
```diff
@@ -306,6 +314,12 @@ func (e *executor) run(ctx context.Context, w inngest.Workflow, id state.Identif
 		return nil, idx, err
 	}
 	if response.Err != nil {
+		if response.Final() && e.failureHandler != nil {
+			if err := (*e.failureHandler)(ctx, id, s, *response); err != nil {
```
tonyhb, line 318: When this isn't a pointer you can literally just `if err := e.failureHandler(ctx, id, s, *response)` :)

pkg/execution/state/driver_response.go
```diff
@@ -236,3 +237,46 @@ func (r DriverResponse) Error() string {
...
+	}
+
+	var output UserError
+
+	if err := json.Unmarshal([]byte(s), &output); err != nil {
+		return nil
+	}
+
+	return &output
+}
```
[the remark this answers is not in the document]
tonyhb: I hate the way we do errors so this isn't on you.  Let's standardize next quarter.  We could pair on a unit test on Monday!

tonyhb (changes requested):

pkg/execution/executor/executor.go
jpwilliams: Lovely!

pkg/execution/executor/executor.go
jpwilliams: @tonyhb Figured that this `WithX` pattern is for adding items that aren't required; wouldn't making this a value mean that it's always needed?

Just a meta question; I'm aware we always want it in this particular case!

jpwilliams:
Looking more reasonable, @tonyhb?

tonyhb (approved):

tonyhb:
Absolutely, this looks great!

PR #412, opened 2023-03-15
Add group IDs to queue items and history

tonyhb:
This PR adjusts our history and queue to allow for "group IDs".  A group ID is a string that's shared across all queue items and history items for a _single_ step, letting us group and correlate disparate jobs + history items together.


tonyhb:
Superseded by #413

PR #415, opened 2023-03-16
Reserve `inngest/*` event names for internal use only

jpwilliams:
## Summary

Stops clients sending `inngest/*` events to the API; they should be reserved for internal use only.


tonyhb (approved):

tonyhb:
Good idea!

PR #423, opened 2023-03-31
Add basic function list view

djfarrelly:
## Description

This adds a new "Function List" view on the side bar from which a user can view all registered functions. The purpose is to help users detect issues with their configuration.

While there could be a lot more work done here, this is a feature that has been requested and should greatly help with users debugging their local setup.

### Function card

Re-used from the rest of the UI. Shows the URL where the function is being served from and the trigger. All functions show "✅ Registered"

image

### View with functions

image

### Blank slate view

image

## Notes

* This adds a new view and GraphQL edge to query for registered functions. 
* This also merges the loader on the dev server so the `FSLoader`/`ReadWriter` that registers the functions is the same as the one that is used to fetch data in the API ensuring the memory is shared.
* This also drops some legacy options for Docker that are no longer used. I don't believe this should break anything as this is just removing them from the dev server package only.


jpwilliams:
INN-760 🥳

cmd/commands/dev.go
```diff
@@ -13,16 +13,14 @@ import (
 func NewCmdDev() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "dev",
-		Short:   "Run the inngest dev server",
-		Example: "inngest dev -p 1234",
```
djfarrelly, line 17: This shows the recommended default that we want people to use now

pkg/coreapi/graph/resolvers/functions.go
```diff
@@ -6,12 +6,62 @@ import (
...
 
+func (r *queryResolver) Functions(ctx context.Context) ([]*models.Function, error) {
+	fns, err := r.APIReadWriter.Functions(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var functions []*models.Function
+	for _, fn := range fns {
+		id := string(fn.ID)
```
djfarrelly, line 24: There were some go-isms that I tried to do in a nicer way, but found to have issues with references so I emulated what was done in other resolvers. Not super pretty, but it works consistently.

ui/src/views/FunctionList.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <div className="flex-1">
+        {isFetching ? (
+          <>Loading...</>
+        ) : functions?.length ? (
+          <div className="flex flex-col items-center basis-[36px] gap-4 max-w-xl">
+            {functions.map((f, idx) => {
+              const triggers = f?.triggers
+                ?.map((t) => t.event || t.cron)
+                .join(", ");
+              const cleanUrl = new URL(f.url || "");
```
jpwilliams, line 30: In the unlikely event that `f.url` is falsey, `new URL("")` will throw. Probably worth extracting this into a `useMemo()` call the conditionally setting the `contextualBar`.

cmd/commands/dev.go
tonyhb: Yeah great idea.

tonyhb (approved):

pkg/coreapi/schema.graphql
```diff
@@ -87,6 +87,19 @@ enum EventStatus {
...
+  id: String
+  concurrency: Int
+  triggers: [FunctionTrigger!]
+  url: String
+}
+
+type FunctionTrigger {
+  event: String
+  cron: String
+}
```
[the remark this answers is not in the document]
djfarrelly: Yeah - I agree that this should probably be a union here and in `monorepo` too. let me look into making that happen 🤔

ui/src/icons/StatusNoFn.tsx
```diff
@@ -4,8 +4,8 @@ export function IconStatusNoFn() {
       <path
         d="M2.4857 6.9238h1.586v-.3965c0-.6835.253-1.039 1.1484-1.5585.9297-.5538 1.4492-1.2852 1.4492-2.3243V2.631c0-1.4971-1.251-2.5772-3.124-2.5772C1.515.0537.4008 1.2295.353 2.8291v.0137l1.5928-.0069.0137-.0068c.0478-.8408.6152-1.3809 1.497-1.3809.8614 0 1.4356.5264 1.4356 1.2442v.0137c0 .6699-.2871 1.039-1.1348 1.5517-.9433.5537-1.3398 1.169-1.2783 2.1807l.0068.4853Zm.8682 3.2471c.6973 0 1.1348-.4238 1.1348-1.0732 0-.6563-.4375-1.0801-1.1348-1.0801-.6836 0-1.1348.4238-1.1348 1.08 0 .6495.4512 1.0733 1.1348 1.0733Z"
         fill="#94A3B8"
-        fill-rule="nonzero"
+        fillRule="nonzero"
```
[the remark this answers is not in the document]
djfarrelly: No - not broken, just React warnings since they only like camelCase attributes 😄

ui/src/store/baseApi.ts
```diff
@@ -2,7 +2,7 @@ import { createApi } from "@reduxjs/toolkit/query/react";
 import { graphqlRequestBaseQuery } from "@rtk-query/graphql-request-base-query";
 import { GraphQLClient } from "graphql-request";
 
-const hostname = window.location.host;
+const hostname = import.meta.env.DEV ? "localhost:8288" : window.location.host;
```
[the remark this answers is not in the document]
djfarrelly: Yep - New to me too - felt like the right way to go for development 😄

ui/src/utils/statusStyles.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { EventStatus, FunctionRunStatus } from "../store/generated";
 
 export default function statusStyles(
-  status: EventStatus | FunctionRunStatus | null
+  status: EventStatus | FunctionRunStatus | "Registered" | null
```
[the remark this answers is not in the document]
djfarrelly: Yeah - I dislike this too, maybe I'll move to a type instead of a string, but it's kind of a temp hack to get this status thing working for a function itself that doesn't return a status from the server.

ui/src/views/FunctionList.tsx
djfarrelly: I changed the API so that URL cannot be empty from the server response. I think that should cover it in this case, but if you think not, I can try/catch this one.

PR #429, opened 2023-04-21
INN-1098 Allow the executor to check status inside streamed responses

jpwilliams:
## Summary INN-1098

We can't use status codes to signify functionality for streamed responses, as we'd be sending it before we understand what we're doing or what has happened.

In these cases, the SDK (inngest/inngest-js#163) embeds the final status inside the returned body.

This small shim catches this occurence.

## Related

- inngest/inngest-js#163


tonyhb (approved):

tonyhb:
Looks good!  Exciting stuff :)

PR #434, opened 2023-04-25
Refactor function configuration

tonyhb:
This removes cue, and refactors function configuration into an "all-in-one" approach.  We previously stored steps and functions individually, and had different structs for functions vs workflows. This has all been simplified into a single struct.


goodoldneon (approved):

darwin67:
Is there something blocking this code to be merged?

tonyhb:
Yes.  This is a general refactor that needs to be upstreamed into cloud;  we cannot merge until the refactors and upstreams have been prepped.

PR #435, opened 2023-04-26
INN-1255: Refactored the UserError function to handle arbitrary response

darwin67:
# Summary

Update the code to handle errors that doesn't follow the previously provided format.
e.g. 50x errors

### DIFF
--- a/pkg/execution/state/driver_response_test.go
+++ b/pkg/execution/state/driver_response_test.go
@@ -218,6 +218,55 @@ func TestDriverResponseFinal(t *testing.T) {
 	}
 }
 
+func TestDriverResponseUserError(t *testing.T) {
+	tests := []struct {
+		name     string
+		r        DriverResponse
+		expected map[string]any
+	}{
+		{
+			name: "with no Output and Err",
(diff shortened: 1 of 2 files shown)


pkg/execution/state/driver_response_test.go
```diff
@@ -218,6 +218,63 @@ func TestDriverResponseFinal(t *testing.T) {
...
+			r:    DriverResponse{Output: nil, Err: fmt.Errorf("something went wrong")},
+			expected: map[string]any{
+				"error": fmt.Errorf("something went wrong"),
+			},
+		},
+		{
+			name: "with no Output and no Err",
+			r:    DriverResponse{Output: nil},
+			expected: map[string]any{
+				"error": "Unknown error running SDK",
```
darwin67, line 238: Is there ever a case where `Output` and `Err` will both be null?

tonyhb: Yes, but in that case we won't have an error and this shouldn't hit!

tonyhb (approved):

PR #436, opened 2023-04-27
Add V0 cancellation API

tonyhb:
This adds an API for cancelling functions using a specific function run ID.

To do:
- [x] Implement API
- [x] Implement public error wrapping
- [x] Add integration tests

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -2,19 +2,21 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/cors"
 	"github.com/inngest/inngest/pkg/config"
+	"github.com/inngest/inngest/pkg/coreapi/apiutil"
 	"github.com/inngest/inngest/pkg/event"
+	"github.com/inngest/inngest/pkg/eventstream"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
 )
 
-type EventHandler func(context.Context, *event.Event) error
+type EventHandler func(context.Context, *event.Event) (string, error)
 
 type Options struct {
 	Config config.Config
@@ -99,15 +101,8 @@ func (a API) HealthCheck(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
 	defer r.Body.Close()
-
-	if r.ContentLength > int64(a.config.EventAPI.MaxSize) {
-		a.writeResponse(w, apiResponse{
-			StatusCode: http.StatusRequestEntityTooLarge,
-			Error:      "Payload larger than maximum allowed",
-		})
-		return
-	}
 
 	key := chi.URLParam(r, "key")
 	if key == "" {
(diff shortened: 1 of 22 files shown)


pkg/publicerr/publicerr.go
```diff
@@ -0,0 +1,94 @@
...
+	if pe, ok := e.(Error); ok {
+		w.WriteHeader(pe.Status)
+	} else if pe, ok := e.(*Error); ok && pe != nil {
+		w.WriteHeader(pe.Status)
+	}
+	return json.NewEncoder(w).Encode(e)
+}
+
+// Error fulfils the builtin go error interface.  This returns the original root cause
+// for debugging via logging - all logging interfaces & tracing packages use Error() to
```
darwin67, line 84: huh, good to know.

pkg/api/api.go
```diff
@@ -118,47 +113,59 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
...
+		return nil
+	})
 
 	if err := eg.Wait(); err != nil {
-		a.writeResponse(w, apiResponse{
-			StatusCode: http.StatusBadRequest,
-			Error:      err.Error(),
+		w.WriteHeader(400)
+		_ = json.NewEncoder(w).Encode(apiutil.EventAPIResponse{
+			IDs:    ids,
```
darwin67, line 159: the `ids` here mean that we're returning the IDs that have been successfully processed?

pkg/eventstream/eventstream.go
```diff
@@ -0,0 +1,98 @@
...
+			if i == MaxEvents {
+				return fmt.Errorf("maximum events parsed within a batch: %d", MaxEvents)
+			}
+
+			jsonEvt := json.RawMessage{}
+			if err := d.Decode(&jsonEvt); err != nil {
+				return err
+			}
+			if len(jsonEvt) > maxSize {
+				return fmt.Errorf("%w: Max %d bytes / Size %d bytes", ErrEventTooLarge, maxSize, len(jsonEvt))
```
darwin67, line 89: is `MaxSize` something users are aware of?
if not, might be good to have a separate issue to add it to the docs so people can be aware of it.

pkg/api/api.go
```diff
@@ -118,47 +113,59 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		a.writeResponse(w, apiResponse{
-			StatusCode: http.StatusBadRequest,
-			Error:      "Could not read event payload",
-		})
-		return
-	}
+	// Create a new channel which receives a stream of events from the incoming HTTP request
+	byteStream := make(chan json.RawMessage)
+	eg := errgroup.Group{}
```
darwin67, line 118: didn't know about this, nice.

pkg/eventstream/eventstream.go
darwin67: also, should the error message be similar to the deleted one?

https://github.com/inngest/inngest/pull/436/files#diff-67a6e36014be865053c43d38b2091bd2d3f17a91b117d97c0f5244cd7d71127dL107

pkg/coreapi/coreapi.go
```diff
@@ -73,3 +90,71 @@ func (a *CoreAPI) Start(ctx context.Context) error {
 func (a CoreAPI) Stop(ctx context.Context) error {
 	return a.server.Shutdown(ctx)
 }
+
+func (a CoreAPI) EventRuns(w http.ResponseWriter, r *http.Request) {
+	if a.tracker == nil {
+		_ = publicerr.WriteHTTP(w, publicerr.Error{
+			Status:  500,
+			Message: "No tracker",
```
darwin67, line 98: nit: not quite sure how actionable this error message is

pkg/coreapi/apiutil/apiutil.go
```diff
@@ -0,0 +1,54 @@
...
+type EventAPIResponse struct {
+	IDs    []string `json:"ids"`
+	Status int      `json:"status"`
+	Error  error    `json:"error,omitempty"`
+}
+
+// CancelRun cancels a run for a given run ID, returning consistent errors for public APIs
+func CancelRun(ctx context.Context, sm state.Manager, runID ulid.ULID) error {
+	if sm == nil {
+		return fmt.Errorf("no state manager supplied to cancel run")
```
darwin67, line 27: does this error message consider an abstration leak?
in other words, if I get this error, what do I as the user can do to fix it?

pkg/eventstream/eventstream.go
```diff
@@ -0,0 +1,98 @@
...
+
+	delim, ok := token.(json.Delim)
+	if !ok {
+		// Invalid type
+		return ErrInvalidRequestBody
+	}
+
+	i := 0
+
+	switch delim {
```
darwin67, line 57: this is an interesting way of parsing JSON :thinking:

pkg/eventstream/eventstream.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+
+	delim, ok := token.(json.Delim)
+	if !ok {
+		// Invalid type
+		return ErrInvalidRequestBody
+	}
+
+	i := 0
```
darwin67, line 55: nit: `i` could probably be in the `[` case right before the `for` loop if it's not being used else where.

darwin67 (approved):
looks good!
posted a couple questions

pkg/publicerr/publicerr.go
```diff
@@ -0,0 +1,94 @@
...
+// in `gql.go` we create an ErrorPresenter middleware item which checks to see if the error
+// is a publicerr.Error and, if so, shows the friendly error directly.
+type Error struct {
+	// Message represents the message to display
+	Message string `json:"error"`
+	// Status represents the HTTP status code to use when responding via HTTP
+	Status int `json:"status"`
+	// Err is the original err which represents the cause of the issue.  This is
+	// the internal error used for debugging.
+	Err error `json:"-"`
```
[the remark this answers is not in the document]
tonyhb: They need to be public to be encoded into JSON.  Private fields are not accessible by the JSON package and are not serialized.

pkg/api/api.go
tonyhb: Correct!

pkg/publicerr/publicerr.go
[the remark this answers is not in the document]
darwin67: Ah! gotcha.

pkg/coreapi/coreapi.go
tonyhb: Agreed, it's not — kind of by design?  This should never happen, and if we see it we'll get a report and know something is extremely broken.  It's an internal error for our testing and can't happen to a user.

pkg/eventstream/eventstream.go
tonyhb: A few things:
- We want to parse both arrays and objects.  This is hard with Go's crappy typing.
- We also want to parse as a stream, without allocating the entire thing up front, so we can parse each individual object of an array separately.

These requirements make this parsing... _unique_, to say the least :D

pkg/coreapi/apiutil/apiutil.go
tonyhb: The idea is that we use this error to catch issues in integration tests, and that users _never_ see it.  It's an internal error, hence no `publicerr.Error`

pkg/eventstream/eventstream.go
tonyhb: great catch, thank you!

PR #438, opened 2023-05-03
Queue change: prioritization of older function run jobs

tonyhb:
This change ensures that older function run jobs are prioritized over new function run jobs, meaning that older functions should finish before new functions run in the case of a high-volume function.

### DIFF
--- a/pkg/execution/state/redis_state/queue.go
+++ b/pkg/execution/state/redis_state/queue.go
@@ -63,6 +63,10 @@ const (
 	PriorityDefault uint = 5
 	PriorityMin     uint = 9
 
+	// FunctionStartScoreBufferTime is the grace period used to compare function start
+	// times to edg enqueue times.
+	FunctionStartScoreBufferTime = 10 * time.Second
+
 	defaultNumWorkers           = 100
 	defaultPollTick             = 10 * time.Millisecond
 	defaultIdempotencyTTL       = 12 * time.Hour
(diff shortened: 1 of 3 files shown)


pkg/execution/state/redis_state/queue.go
```diff
@@ -63,6 +63,10 @@ const (
 	PriorityDefault uint = 5
 	PriorityMin     uint = 9
 
+	// FunctionStartScoreBufferTime is the grace period used to compare function start
+	// times to edg enqueue times.
```
goodoldneon, line 67:
```suggestion
	// times to edge enqueue times.
```

pkg/execution/state/redis_state/queue.go
```diff
@@ -377,6 +381,36 @@ type QueueItem struct {
...
 
+// Score returns the score (time that the item should run) for the queue item.
+//
+// NOTE: In order to prioritize finishing older function runs with a busy function
+// queue, we sometimes use the function run's "started at" time to enqueue edges which
+// run steps.  This lets us push older function steps to the beginning of the queue,
+// ensuring they run before other newer function runs.
+//
+// We can ONLY do this for the first attempt, and we can ONLY do this for edges that
+// are not sleeps (eg. immediate runs)
```
darwin67, line 392: why only the first attempt?

tonyhb: Because attempts >= 1 are scheduled in the future.

PR #448, opened 2023-05-19
Add env to ctx

tonyhb:
This allows the SDK to fetch env names when running.

Note:  open-source uses a static "local" env;  we'll want to configure this as and when we do direct integrations with dev server <> cloud.

### DIFF
--- a/pkg/execution/driver/dockerdriver/dockerdriver.go
+++ b/pkg/execution/driver/dockerdriver/dockerdriver.go
@@ -152,7 +152,7 @@ func (d *dockerExec) start(ctx context.Context, state state.State, wa inngest.St
 }
 
 func (d *dockerExec) startOpts(ctx context.Context, state state.State, wa inngest.Step, idx int) (docker.CreateContainerOptions, error) {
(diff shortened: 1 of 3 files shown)


pkg/execution/driver/dockerdriver/dockerdriver.go
```diff
@@ -152,7 +152,7 @@ func (d *dockerExec) start(ctx context.Context, state state.State, wa inngest.St
 }
 
 func (d *dockerExec) startOpts(ctx context.Context, state state.State, wa inngest.Step, idx int) (docker.CreateContainerOptions, error) {
-	marshalled, err := driver.MarshalV1(ctx, state, wa, idx)
+	marshalled, err := driver.MarshalV1(ctx, state, wa, idx, "local")
```
darwin67, line 155: do we still need this driver?

darwin67 (approved):

pkg/execution/driver/dockerdriver/dockerdriver.go
tonyhb: No, though that's out of scope for this PR and is ripped out in the open refactor

PR #450, opened 2023-05-23
INN-1400: Allow `Event` list in state input for batching

darwin67:
## Summary

Update expected `Input` and `Interface` to accept a list of events for batching purposes.
Executor will accept a list of events as input, and invoke the SDK with it.

From the SDK's perspective, there will be redundant data from the executor in order to keep backward compatibility.
Until the lastest SDK version with single event input gets deprecated, this redundancy will need to coexist for a while.

Data passed to the SDK will be
* `event`
* `events`

For single item invocations, this will be true `[event] == events`, which is what this PR is expected to do.


pkg/execution/state/redis_state/redis_state.go
```diff
@@ -463,6 +471,17 @@ func (m mgr) Load(ctx context.Context, runID ulid.ULID) (state.State, error) {
 		return nil, fmt.Errorf("failed to unmarshal event; %w", err)
 	}
 
+	// Load the batch of events
+	cmd = m.r.B().Get().Key(m.kf.Events(ctx, id)).Build()
+	byt, err = m.r.Do(ctx, cmd).AsBytes()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get batch; %w", err)
```
tonyhb, line 478: We need to check if `err` is a redis nil error;  old function states without batches will error here, but that is actually safe.

This code needs to handle two compat modes:

1. `batches` are not present, use event as the batch
2. Only load batches, and use the first item for the batch

darwin67: Follow up with adding a Version to metadata, in order to mitigate backward incompatibility issues in the executor.

pkg/execution/driver/driver.go
```diff
@@ -31,6 +31,7 @@ type FunctionStack struct {
 func MarshalV1(ctx context.Context, s state.State, step inngest.Step, stackIndex int, env string) ([]byte, error) {
 	data := map[string]interface{}{
 		"event": s.Event(),
+		// "events": s.Events(), disabling this for now so we don't send unnecessary data to customer infra
```
darwin67, line 34: per discussion in Discord, don't send list of events to SDK for now since it will reduce the quota for amount of data that can be sent by half.
the appropriate way to do it is to determine the fields to send over based on SDK version.

which I'll do as a follow up.

tonyhb (approved):

PR #454, opened 2023-05-26
Remove unused reference `RunFunc`

darwin67:
`f RunFunc` is passed down but never used, so removing it all together for scanning.
Plus some small config changes.


tonyhb (approved):
Nice!  Not sure how I left that in there, thanks for catching my cruft haha!

PR #460, opened 2023-06-07
Add indexes to event stream parsing

tonyhb:
This ensures that the response for event IDs contains the same order as incoming events, fixing any race conditions.

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -33,6 +33,8 @@ const (
 
 	MaxBatchSize    = 100
 	MaxBatchTimeout = 60 * time.Second
+	// MaxEvents is the maximum number of events we can parse in a single batch.
+	MaxEvents = 5_000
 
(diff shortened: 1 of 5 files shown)


pkg/consts/consts.go
```diff
@@ -15,4 +15,6 @@ const (
 
 	MaxBatchSize    = 100
 	MaxBatchTimeout = 60 * time.Second
+	// MaxEvents is the maximum number of events we can parse in a single batch.
+	MaxEvents = 5_000
```
darwin67, line 19: there's already a value defined here.
https://github.com/inngest/inngest/blob/d67b9e3a0cc170484066ba27c656ca1f38f2f50a/pkg/eventstream/eventstream.go#L15-L16

we should consolidate so we don't have 2 values.

darwin67 (approved):
lgtm once `MaxEvents` const is consolidated so we don't have 2 of them.

PR #463, opened 2023-06-09
Add headers to DriverResponse

goodoldneon:
# Changes

Add `DriverResponse.Headers` and set it after execution

# Motivation

We'd like to log the execution response headers within Worker, since that'll help us understand whether execution failures are happening at the SDK or app level


darwin67:
Is this ready for review? tests and linter is failing.

goodoldneon:
@jpwilliams, I think I created this PR because you wanted some extra info

PR #464, opened 2023-06-09
Add rate limiting to OSS dev server

tonyhb:
This PR introduces rate limiting to the OSS as we have in the cloud.  It uses GCRA to ensure that functions are only called as specified in the rate limit config


pkg/execution/runner/runner.go
```diff
@@ -534,6 +543,22 @@ func (s *svc) pauses(ctx context.Context, evt event.Event) error {
...
+		if err != nil {
+			return err
+		}
+		limited, err := s.rl.RateLimit(ctx, key, *fn.RateLimit)
+		if err != nil {
+			return err
+		}
+		if limited {
+			// Do nothing.
+			return nil
```
darwin67, line 558: probably should still return an error, bubbling up to the user that execution has been rate limited.
people will think it's a bug when there's no reaction even thought it's the expected outcome for us.

and ignore that error in logging.

honestly it's best to be a warning instead of an error, but seems like it needs more changes to do that with the current structure.

pkg/execution/ratelimit/rueidis.go
```diff
@@ -0,0 +1,158 @@
...
+local v = redis.call('get', KEYS[1])
+if v == false then
+  return redis.error_reply("key does not exist")
+end
+if v ~= ARGV[1] then
+  return 0
+end
+redis.call('setex', KEYS[1], ARGV[3], ARGV[2])
+return 1
+`
```
darwin67, line 26: nit:
make it an actual lua script as a follow up.

pkg/execution/ratelimit/rueidis.go
```diff
@@ -0,0 +1,158 @@
...
+	err = r.r.Do(ctx, r.r.B().Expire().Key(key).Seconds(int64(ttl.Seconds())).Build()).Error()
+	return updated == 1, err
+}
+
+// CompareAndSwapWithTTL atomically compares the value at key to the
+// old value. If it matches, it sets it to the new value and returns
+// true. Otherwise, it returns false. If the key does not exist in the
+// store, it returns false with no error. If the swap succeeds, the
+// ttl for the key is updated atomically.
+func (r *rueidisStore) CompareAndSwapWithTTL(ctx context.Context, key string, old, new int64, ttl time.Duration) (bool, error) {
```
darwin67, line 127: it's not used? is it gonna be in the cloud code?

pkg/execution/ratelimit/rueidis.go
```diff
@@ -0,0 +1,158 @@
...
+
+	now := time.Unix(int64(secs), int64(msecs)*int64(time.Microsecond))
+	return now, nil
+}
+
+// SetIfNotExistsWithTTL sets the value of key only if it is not
+// already set in the store it returns whether a new value was set.
+// If a new value was set, the ttl in the key is also set, though this
+// operation is not performed atomically.
+func (r *rueidisStore) SetIfNotExistsWithTTL(ctx context.Context, key string, value int64, ttl time.Duration) (bool, error) {
```
darwin67, line 100: it's not used? is it gonna be in the cloud code?

pkg/execution/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,66 @@
...
+	ErrRateLimitExceeded = fmt.Errorf("rate limit exceeded")
+)
+
+type RateLimiter interface {
+	RateLimit(ctx context.Context, key string, c inngest.RateLimit) (bool, error)
+}
+
+// RateLimitKey returns the rate limiting key given a function ID, rate limit config,
+// and incoming event data.
+func RateLimitKey(ctx context.Context, id uuid.UUID, c inngest.RateLimit, evt map[string]any) (string, error) {
```
darwin67, line 27: should this be part of the key generator?

darwin67 (approved):
Approved with some comments

pkg/execution/ratelimit/rueidis.go
tonyhb: It is used.  It's used by the underlying throttle package.

pkg/execution/ratelimit/rueidis.go
tonyhb: Used by underlying throttle package again.

pkg/execution/ratelimit/ratelimit.go
tonyhb: The key generator is for the state store which is a separate package.  Rate limiting lives independently of that package.

pkg/execution/ratelimit/rueidis.go
tonyhb: lol ya, i was being particularly lazy for this small script;  didnt want to have to deal with the tediousness of `fs.Embed` given there's such a small single script.

pkg/execution/runner/runner.go
tonyhb: Yeah, I think we should probably tackle this UI side here and add something that shows rate limiting directly into the dev server UI.  It's a great shout, and something we can loop Ed into.

PR #478, opened 2023-06-26
build(ui): switch to pnpm

IGassmann:
## Description

This switches the package manager of the Dev Server from Yarn 1 to [pnpm](https://pnpm.io/).

### Why pnpm?

- We've just [migrated the Dashhboard to pnpm](https://github.com/inngest/ui/pull/411) and are looking to also migrate the SDK to it. That way, we can use the same package manager everywhere.
- Yarn 1 is in maintenance mode and [doesn't receive any more functional improvements](https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md)
- [Performs often faster than the alternatives](https://pnpm.io/benchmarks)
- Automatically installs/handles needed peer dependencies—no more a long list of peer dependency warnings on every install.
- Comes with Lerna-like monorepo features such as multi-package task execution
- Protects against [bugs caused by importing packages that aren't declared in the app's `package.json`](https://www.kochan.io/nodejs/pnpms-strictness-helps-to-avoid-silly-bugs.html), e.g., we were using `graphql-request` without declaring it in our `package.json`.
- [Recommended package manager to use with Turborepo](https://turbo.build/repo/docs/getting-started/create-new#which-package-manager-do-you-want-to-use)

**Pro Tip**: tired of having to use a different command whenever you switch to a project that uses a different package manager? Use [`ni`](https://github.com/antfu/ni)! It auto-detects which package manager to use and has super short commands, e.g., `ni` over `yarn` or `pnpm i`.


ui/.gitignore
```diff
@@ -1,12 +1,3 @@
-# Logs
-logs
-*.log
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-pnpm-debug.log*
-lerna-debug.log*
-
```
IGassmann, line 9: pnpm place its debug log file within `node_modules/`, which is already ignored.

ui/package.json
```diff
@@ -3,12 +3,13 @@
...
-    "dev": "yarn && concurrently --prefix-colors \"green.inverse,magenta.inverse,blue.inverse\" yarn:dev:*",
+    "dev": "pnpm install && concurrently --prefix-colors \"green.inverse,magenta.inverse,blue.inverse\" pnpm:dev:*",
     "dev:vite": "vite",
-    "dev:codegen": "nodemon -w ../pkg/coreapi -e graphql -x 'yarn codegen'",
+    "dev:codegen": "nodemon -w ../pkg/coreapi -e graphql -x 'pnpm codegen'",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "codegen": "graphql-codegen --config codegen.ts"
+    "codegen": "graphql-codegen --config codegen.ts",
+    "preinstall": "npx only-allow pnpm"
```
IGassmann, line 12: Ensures one doesn't accidentally run `yarn install` or `npm install`.

PR #512, opened 2023-07-14
Fix: Upsert functions upon update

djfarrelly:
## Description

When a user added or removed functions for a given app, the checksum doesn't match, but the first duplicate function that is inserted (based on the deterministic function id) fails to be inserted and returns and error, preventing any further functions from being created.

This aims to change the insert to a upsert-like behavior which should be able to gracefully handle functions that already exist, just updating those documents.

NOTES

* We also need to soft delete functions that have been removed so we need to also iterate through any existing ones and remove them if they weren't inserted/updated.

- [ ] Docs

## Checklist


djfarrelly:
Replaced by #513

PR #525, opened 2023-07-19
Add specific function invocation via events

tonyhb:
This commit introduces an API to invoke specific functions by slug.  We do this via events.  In short:

- We create a new internal event: `inngest/function.invoked`.  This can invoke any function by slug in the current workspace (via signing keys).
- POST data to this invoke API is sent as `event.data` to the functon
- We can only ever invoke a single function
- Scheduled functions (eg. crons) can be invoked.

In the future, we can add a blocking ?async query parameter which lets the frontend await on function output.

This is _primarily_ needed to test scheduled functions via our new UI.  However, it has broad implications and should get first class support within our SDK.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,13 +5,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/cors"
 	"github.com/inngest/inngest/pkg/config"
+	"github.com/inngest/inngest/pkg/consts"
 	"github.com/inngest/inngest/pkg/coreapi/apiutil"
 	"github.com/inngest/inngest/pkg/event"
 	"github.com/inngest/inngest/pkg/eventstream"
+	"github.com/inngest/inngest/pkg/publicerr"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
 )
@@ -57,6 +61,7 @@ func NewAPI(o Options) (chi.Router, error) {
 
 	api.Get("/health", api.HealthCheck)
 	api.Post("/e/{key}", api.ReceiveEvent)
+	api.Post("/invoke/{slug}", api.Invoke)
 
 	return api, nil
 }
@@ -134,12 +139,17 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 
 	// Process those incoming events
 	eg.Go(func() error {
-		// TODO: Iterate through event stream and process event.
 		for byt := range byteStream {
 			evt := event.Event{}
 			if err := json.Unmarshal(byt, &evt); err != nil {
 				return err
 			}
+
+			if strings.HasPrefix(strings.ToLower(evt.Name), "inngest/") {
+				err := fmt.Errorf("event name is reserved for internal use: %s", evt.Name)
+				return err
+			}
+
 			id, err := a.handler(r.Context(), &evt)
 			if err != nil {
 				a.log.Error().Str("event", evt.Name).Err(err).Msg("error handling event")
@@ -168,4 +178,37 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 		IDs:    ids,
 		Status: 200,
 	})
+}
+
+// Invoke creates an event to invoke a specific function.
+func (a API) Invoke(w http.ResponseWriter, r *http.Request) {
+	// XXX: In OSS self hosting, check signing keys here.
+
+	// Get the function slug from the route parameter.   This is the function
+	// we'll invoke.  Any request is passed as the event data to the function.
+	slug := chi.URLParam(r, "slug")
+
+	data := map[string]any{}
+	if err := json.NewDecoder(r.Body).Decode(&data); err != nil {
+		_ = publicerr.WriteHTTP(w, publicerr.Wrap(err, 400, "Unable to read post data request"))
+		return
+	}
+
+	data[consts.InvokeSlugKey] = slug
+
+	now := time.Now().UnixMilli()
+	evt := event.Event{
+		Name:      consts.InvokeEventName,
+		Data:      data,
+		Timestamp: now,
+	}
+
+	evtID, err := a.handler(r.Context(), &evt)
+	if err != nil {
+		_ = publicerr.WriteHTTP(w, publicerr.Wrapf(err, 500, "Unable to create invocation event: %s", err))
+		return
+	}
+
+	// TODO: If await is true as a query parameter, await the data from the function.
+	_, _ = w.Write([]byte(evtID))
 }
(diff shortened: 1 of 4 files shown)


pkg/api/service.go
```diff
@@ -95,7 +96,7 @@ func (a *apiServer) handleEvent(ctx context.Context, e *event.Event) (string, er
 
 	l.Debug().Str("event", e.Name).Msg("handling event")
 
-	if strings.HasPrefix(strings.ToLower(e.Name), "inngest/") {
+	if strings.HasPrefix(strings.ToLower(e.Name), "inngest/") && e.Name != consts.InvokeEventName {
```
jpwilliams, line 99: This makes sense, though breaks the easy contract of "_you can never send `inngest/*` events_."

My initial thought was to send which functions should be invoked in the event itself in addition to the `name`, `data`, and `user` properties, but this obviously wouldn't work for crons.

What's the shape of the `inngest/function.invoked` event?

tonyhb: You can post any data to the invoke endpoint as a JSON object, which we copy into `event.data`.  We add a specific `_inngest_fn` field which stores the slug of the function to call.

We could block the regular event ingestion endpoint from receiving any `inngest/*` functions.  That means users can never send random `inngest/*` events _at all_ - the only way to generate the `inngest/function.invoked` event is via the invoke API

pkg/api/api.go
```diff
@@ -169,3 +173,36 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 		Status: 200,
 	})
 }
+
+// Invoke creates an event to invoke a specific function.
+func (a API) Invoke(w http.ResponseWriter, r *http.Request) {
+	// XXX: In OSS self hosting, check signing keys here.
```
darwin67, line 179: for future discussions:
this should really be API keys at this point. using signing keys for authn / authz is expanding the use case too much.

pkg/execution/runner/runner.go
```diff
@@ -310,13 +311,67 @@ func (s *svc) handleMessage(ctx context.Context, m pubsub.Message) error {
 	return errs
 }
 
+// InvokeFunction is a helper method which loads all available functions, checks
+// the incoming event and returns the function to be invoked via the RPC invoke event,
+// or nil if a function is not being invoked.
+func InvokeFunction(ctx context.Context, evt event.Event, fl cqrs.ExecutionLoader) (*inngest.Function, error) {
```
darwin67, line 317: nit:
this sounds more like a `FindFunction` then `Invoke`.
it's not actually invoking anything per se.

darwin67:
I didn't see the API handler being assigned to a router. is that intended?

pkg/api/api.go
```diff
@@ -169,3 +173,36 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
...
+
+	// Get the function slug from the route parameter.   This is the function
+	// we'll invoke.  Any request is passed as the event data to the function.
+	slug := chi.URLParam(r, "slug")
```
jpwilliams, line 183: Could we add the ability to specify multiple functions here via `?slug[]=id&slug[]=id1`? Looks like `net` just expects [repeated values without the `[]`](https://pkg.go.dev/net/url#Values); not sure about chi.

tonyhb:
> I didn't see the API handler being assigned to a router. is that intended?

The handler is being assigned here: https://github.com/inngest/inngest/pull/525/files#diff-67a6e36014be865053c43d38b2091bd2d3f17a91b117d97c0f5244cd7d71127dR63

pkg/execution/runner/runner.go
tonyhb: Let's call this `FindInvokedFunction` - it's really only finding functions if the event is an invoke event.  Thanks for the feedback!

pkg/api/api.go
tonyhb: It's a solid idea, though for this iteration let's keep the scope small.  The current plan:

1. Release an API that allows invoking functions via the runner
2. Implement invoking functions in the SDK
3. Implement `await` as an optional parameter to await on the function response

Each of these are individual deliverables.  Once this is done, we can assess usage, gather feedback, and jam on new ideas.

Some initial thoughts:  if you're invoking multiple functions you should probably be using `inngest.send`.

darwin67 (approved):

pkg/api/service.go
tonyhb: Done!

tonyhb:
Let's not merge this until we have a discussion on the proposed SDK APIs.  I'm (slightly) concerned that this has no room for our `user` field.

pkg/execution/runner/runner.go
```diff
@@ -310,13 +311,67 @@ func (s *svc) handleMessage(ctx context.Context, m pubsub.Message) error {
...
+	if evt.Name != consts.InvokeEventName {
+		return nil, nil
+	}
+
+	fns, err := fl.Functions(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	name := evt.Data[consts.InvokeSlugKey]
```
goodoldneon, line 327: What's the reason for putting the slug key in `event.data` instead of a new top-level property? I don't think that's a blocker for this PR, but it feels icky to pollute the event data. Maybe that causes problems with some data validation libraries that don't allow excess properties? Or maybe that complicates event replay (when we add that feature)?

goodoldneon:
We already chatted about this, Tony, but I wonder if we should restrict specific invocation to functions with user-defined slugs. Slugs derived from names usually (but not always!) change when the function name changes. And since "name" doesn't scream "I'm a durable value", I can see some issues with specifically invoking functions with derived slugs.

I don't think that's a blocker for this PR tho

jpwilliams:
> @tonyhb: Let's not merge this until we have a discussion on the proposed SDK APIs. I'm (slightly) concerned that this has no room for our `user` field.

If this is primarily an API the SDK will use, we can change this to be more explicit about function to invoke, data, and user without trying to send in a sort-of-event and transforming it.

I'd love to get this merged so we can trigger crons in the dev server and we can go ahead with monorepo changes.

> @goodoldneon: I wonder if we should restrict specific invocation to functions with user-defined slugs. Slugs derived from names usually (but not always!) change when the function name changes. And since "name" doesn't scream "I'm a durable value", I can see some issues with specifically invoking functions with derived slugs.

This might be a requirement for the SDK's side anyway, yeah.

pkg/api/api.go
tonyhb: I'm fairly certain that signing keys will become environment-scoped API keys, but of course open to feedback and ideas.  Having three keys seems excessive, plus APIs like this are always scoped by environment to "invoke" a function within your env.

pkg/execution/runner/runner.go
tonyhb: We don't want to add top-level event properties for specific features.  The format we have is small and concise;  changing it impacts internals deeply.  If anything, we should add a single `ctx` map to the event data which allows for baggage and other event metadata to be included maybe.  This can be a future change.

One of the key aspects is that Inngest controls the creation of this event.  We can adjust this at any time without having to worry about SDK rollouts.

PR #530, opened 2023-07-21
INN-1808: Add telemetry setup

darwin67:
## Description

Add otel tracer and meter setup.
Provides the option to use Jaeger, stdout or OTLP.

- [ ] Docs

## Checklist


cmd/commands/serve.go
```diff
@@ -59,24 +60,40 @@ func serve(cmd *cobra.Command, args []string) {
...
 		}
 	}
 
-	if err := service.StartAll(ctx, svc...); err != nil {
+	// NOTE: what to do with the returned shutdown function?
+	_, err = tracing.Setup(svcName, "io")
+	if err != nil {
+		fmt.Println(err.Error())
+		os.Exit(1)
+	}
```
goodoldneon, line 94: IIUC, `service.StartAll` will block until the services shutdown. So you can prob do:

```suggestion
	// NOTE: what to do with the returned shutdown function?
	close, err = tracing.Setup(svcName, "io")
	if err != nil {
		fmt.Println(err.Error())
		os.Exit(1)
	}
	defer close()
```

pkg/execution/queue/item.go
```diff
@@ -63,6 +63,10 @@ type Item struct {
 	// Payload stores item-specific data for use when processing the item.  For example,
 	// this may contain the function's edge for running a step.
 	Payload any `json:"payload,omitempty"`
+
+	// Metadata is used for storing additional metadata related to the queue item.
+	// e.g. tracing data
+	Metadata map[string]string
```
goodoldneon, line 69: Does this mean that all trace attributes must be strings?

darwin67: Yes, because the `MapCarrier` type for span context and baggage are always `map[string]string`

https://pkg.go.dev/go.opentelemetry.io/otel@v1.16.0/propagation#MapCarrier

cmd/commands/serve.go
darwin67: thanks! updated

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,154 @@
...
+			propagation.Baggage{},
+		),
+	)
+
+	return func() { tracer.Shutdown() }, nil
+}
+
+func NewTracerProvider(ctx context.Context, svc, ttype string) (*tracer, error) {
+	switch ttype {
+	case "otlp":
```
tonyhb, line 52: should ttype be a const or enum?

darwin67: maybe enum?

darwin67: actually, I'll go with const for now since there are no much difference

darwin67:
Added otel `MeterProvider` as well while at it.

pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,164 @@
...
+
+func TracerSetup(svc string, ttype TracerType) (func(), error) {
+	ctx := context.Background()
+
+	var (
+		tracer *tracer
+		err    error
+	)
+
+	tracer, err = NewTracerProvider(ctx, svc, ttype)
```
goodoldneon, line 42: Is this separate declaration needed?

```suggestion
	tracer, err := NewTracerProvider(ctx, svc, ttype)
```

darwin67: nope. there was a reason but don't remember why.

tonyhb (approved):

PR #570, opened 2023-08-14
Add cancellation, event resuming to executor interface

tonyhb:
This centralizes logic into one place, and allows us to start using the executor interface to manage history directly.

In the future, the state store and queue should be entirely abstracted by the executor.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -1,8 +1,10 @@
 package event
 
 import (
+	"crypto/rand"
 	"encoding/json"
-	"sync"
+
+	"github.com/oklog/ulid/v2"
 )
 
 const (
@@ -10,75 +12,9 @@ const (
 	FnFailedName      = "inngest/function.failed"
 )
 
-type Manager struct {
-	events map[string]Event
-	l      *sync.RWMutex
-}
-
-func NewManager() Manager {
-	return Manager{
-		events: make(map[string]Event),
-		l:      &sync.RWMutex{},
-	}
-}
-
-// Fetch an individual event by its ID.
-func (e Manager) EventById(id string) *Event {
-	e.l.RLock()
-	defer e.l.RUnlock()
-
-	evt, ok := e.events[id]
-	if !ok {
-		return nil
-	}
-
-	return &evt
-}
-
-// Fetch all events with a given name.
-func (e Manager) EventsByName(name string) []Event {
-	e.l.RLock()
-	defer e.l.RUnlock()
-
-	events := []Event{}
-
-	for _, evt := range e.events {
-		if evt.Name == name {
-			events = append(events, evt)
-		}
-	}
-
-	return events
-
-}
-
-// Fetch all events.
-func (e Manager) Events() []Event {
-	e.l.RLock()
-	defer e.l.RUnlock()
-
-	events := []Event{}
-
-	for _, evt := range e.events {
-		events = append(events, evt)
-	}
-
-	return events
-}
-
-// Parse and create a new event, adding it to the in-memory map as we go.
-func (e Manager) NewEvent(data string) (*Event, error) {
-	e.l.Lock()
-	defer e.l.Unlock()
-
-	evt, err := NewEvent(data)
-	if err != nil {
-		return nil, err
-	}
-
-	e.events[evt.ID] = *evt
-
-	return evt, err
+type TrackedEvent interface {
+	InternalID() ulid.ULID
+	Event() Event
 }
 
 func NewEvent(data string) (*Event, error) {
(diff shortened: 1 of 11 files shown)


pkg/execution/executor/executor.go
```diff
@@ -80,6 +92,29 @@ type Executor interface {
...
 	) (*state.DriverResponse, int, error)
+
+	// HandlePauses handles pauses loaded from an incoming event.  This delegates to Cancel and
+	// Resume where necessary, depending on pauses that have been loaded and matched.
+	HandlePauses(ctx context.Context, iter state.PauseIterator, event event.TrackedEvent) error
+	// Cancel cancels an in-progress function, preventing any enqueued or future steps from running.
+	Cancel(ctx context.Context, id state.Identifier, r CancelRequest) error
+	// Resume resumes an in-progress function from the given waitForEvent pause.
+	Resume(ctx context.Context, p state.Pause, r ResumeRequest) error
+	// SetFailureHandler sets the failure handler, called when a function permanently fails.
```
goodoldneon, line 103: Super duper nits

```suggestion
	// Cancel cancels an in-progress function run, preventing any enqueued or future steps from running.
	Cancel(ctx context.Context, id state.Identifier, r CancelRequest) error
	// Resume resumes an in-progress function run from the given waitForEvent pause.
	Resume(ctx context.Context, p state.Pause, r ResumeRequest) error
	// SetFailureHandler sets the failure handler, called when a function run permanently fails.
```

pkg/execution/executor/executor.go
```diff
@@ -80,6 +92,29 @@ type Executor interface {
...
+// CancelRequest stores information about the incoming cancellation request within
+// history.
+type CancelRequest struct {
+	EventID    *ulid.ULID
+	Expression *string
+	UserID     *uuid.UUID
+}
+
+type ResumeRequest struct {
+	With    any
```
goodoldneon, line 116: What does `With` do?

pkg/execution/executor/executor.go
```diff
@@ -450,6 +498,205 @@ func (e *executor) executeStep(ctx context.Context, id state.Identifier, step *i
...
+			break
+		}
+
+		pause := iter.Val(ctx)
+
+		// Block until we have capacity
+		_ = sem.Acquire(ctx, 1)
+
+		wg.Add(1)
+		go func() {
```
darwin67, line 530: nice, this will speed things up.

pkg/execution/executor/executor.go
```diff
@@ -450,6 +498,205 @@ func (e *executor) executeStep(ctx context.Context, id state.Identifier, step *i
...
+			}
+
+			// Ensure that we store the group ID for this pause, letting us properly track cancellation
+			// or continuation history
+			ctx = state.WithGroupID(ctx, pause.GroupID)
+
+			// Run an expression if this exists.
+			if pause.Expression != nil {
+				// Precompute the expression data once, as a value (not pointer)
+				data := expressions.NewData(map[string]interface{}{
```
darwin67, line 557: consistency with other changes.

```suggestion
				data := expressions.NewData(map[string]any{
```

pkg/execution/executor/executor.go
```diff
@@ -450,6 +498,205 @@ func (e *executor) executeStep(ctx context.Context, id state.Identifier, step *i
...
+				// cancellation
+				return
+			}
+
+			// NOTE: Some pauses may be nil or expired, as the iterator may take
+			// time to process.  We handle that here and assume that the event
+			// did not occur in time.
+			if pause == nil || pause.Expires.Time().Before(time.Now()) {
+				// Consume this pause to remove it entirely
+				_ = e.sm.ConsumePause(context.Background(), pause.ID, nil)
```
darwin67, line 546: `ConsumePause` seems like it always need to called before a return.
Could it be wrapped in a `defer` to reduce explicit calls everywhere?

pkg/execution/executor/executor.go
darwin67: Seems like just assigning some arbitrary data.
https://github.com/inngest/inngest/pull/570/files#diff-ccd0679a9f24c84fef4f5d4d5fe2c32651d23886d85af8448bb968921c7b8a95R605

darwin67 (approved):
lgtm with some comments and questions

pkg/event/event.go
```diff
@@ -133,3 +69,27 @@ func (evt Event) Map() map[string]interface{} {
 
 	return data
 }
+
+func NewOSSTrackedEvent(e Event) TrackedEvent {
```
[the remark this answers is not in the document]
tonyhb: We already have a similar struct in the cloud

pkg/event/mgr_legacy.go
```diff
@@ -0,0 +1,74 @@
+package event
+
```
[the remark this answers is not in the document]
tonyhb: Yep.  This exists but will be yanked out and replaced later this week.

pkg/execution/executor/executor.go
tonyhb: Yeah.  This assigns arbitrary data to the step, allowing us to use that data in the SDK's function.

goodoldneon (approved):

pkg/execution/executor/executor.go
tonyhb: Almost always, though sometimes we can error eg. if creating a new expression evaluator, or if there's somehow an error in the state store cancelling the function.  In these cases we don't want to consume the pause - we'd hope that the error is transient and will retry.

PR #612, opened 2023-09-10
Add debounce support

tonyhb:
This PR introduces debounce support for functions as a first class configuration parameter.

Debounce requires two configuration parameters:

```
{
  key: string,
  period: time.Duration
}
```

- `key` represents a unique key for the function debounce.  This is an expression, and can use event data:  `event.data.name` returns the `name` field from `event.data` to use as the debounce key. Keys are scoped per-function.  If two separate functions use the same debounce key they do not collide.
- `period` is the period to debounce.

# What is debounce?

Debounce delays functions for the `period` specified.  If an event is sent, the function will not run until *at least* `period` has elapsed.

If any new events are received that match the same debounce key, the function is reshceduled for another `period` delay, and the triggering event is replaced with the latest event received.

For example:

```
debounce: 5s
time:     [0 1 2 3 4 5 6 7 8 9]
timeline: [* - - - - ^        ]
           |         |
	   event received
	             |
                     function runs

debounce: 5s
time:     [0 1 2 3 4 5 6 7 8 9]
timeline: [* - - * - - - - ^  ]
           |     |         |
	   event received  |
	                   |
                           function runs
```

In order to facilitate debounce, several new components and things have been added:

- The ability to easily reschedule jobs within the queue with only an ID (an alternative API to an existing Reschedule API)
- Debounce-specific managers, for creating an dupdating debounce keys

Note that extra care has been taken into consideration to resolve debounce-related race conditions.  This will be modelled with TLA+ in the future.

## To do
- [x] Add debounce transactions to state store
- [x] Ensure debounce reschedules function run
- [x] Delete debounce item after function run
- [ ] Ensure Golang integration tests run on CI

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/devserver/devserver.go
+++ b/pkg/devserver/devserver.go
@@ -13,6 +13,7 @@ import (
 	"github.com/inngest/inngest/pkg/deploy"
 	"github.com/inngest/inngest/pkg/enums"
 	"github.com/inngest/inngest/pkg/event"
+	"github.com/inngest/inngest/pkg/execution/debounce"
 	"github.com/inngest/inngest/pkg/execution/driver"
 	"github.com/inngest/inngest/pkg/execution/driver/httpdriver"
 	"github.com/inngest/inngest/pkg/execution/executor"
@@ -100,14 +101,15 @@ func start(ctx context.Context, opts StartOpts) error {
 		return err
 	}
 
+	queueKG := &redis_state.DefaultQueueKeyGenerator{
+		Prefix: "{queue}",
+	}
 	queue := redis_state.NewQueue(
 		rc,
 		redis_state.WithIdempotencyTTL(time.Hour),
 		redis_state.WithNumWorkers(100),
 		redis_state.WithPollTick(150*time.Millisecond),
-		redis_state.WithQueueKeyGenerator(&redis_state.DefaultQueueKeyGenerator{
-			Prefix: "{queue}",
-		}),
+		redis_state.WithQueueKeyGenerator(queueKG),
 		redis_state.WithPartitionConcurrencyKeyGenerator(func(ctx context.Context, p redis_state.QueuePartition) (string, int) {
 			// Ensure that we return the correct concurrency values per
 			// partition.
@@ -129,6 +131,8 @@ func start(ctx context.Context, opts StartOpts) error {
 	)
 
 	rl := ratelimit.New(ctx, rc, "{ratelimit}:")
+
+	debouncer := debounce.NewRedisDebouncer(rc, queueKG, queue)
 
 	var drivers = []driver.Driver{}
 	for _, driverConfig := range opts.Config.Execution.Drivers {
@@ -154,6 +158,7 @@ func start(ctx context.Context, opts StartOpts) error {
 			),
 		),
 		executor.WithStepLimits(consts.DefaultMaxStepLimit),
+		executor.WithDebouncer(debouncer),
 	)
 	if err != nil {
 		return err
@@ -166,6 +171,7 @@ func start(ctx context.Context, opts StartOpts) error {
 		executor.WithState(sm),
 		executor.WithServiceQueue(queue),
 		executor.WithServiceExecutor(exec),
+		executor.WithServiceDebouncer(debouncer),
 	)
 
 	runner := runner.NewService(
--- a/pkg/execution/debounce/debounce.go
+++ b/pkg/execution/debounce/debounce.go
@@ -98,9 +98,12 @@ type DebouncePayload struct {
 	FunctionID uuid.UUID `json:"fnID"`
 }
 
+// Debouncer represents an implementation-agnostic function debouncer, delaying function runs
+// until a specific time period passes when no more events matching a key are received.
 type Debouncer interface {
 	Debounce(ctx context.Context, d DebounceItem, fn inngest.Function) error
 	GetDebounceItem(ctx context.Context, debounceID ulid.ULID) (*DebounceItem, error)
+	DeleteDebounceItem(ctx context.Context, debounceID ulid.ULID) error
 }
 
 func NewRedisDebouncer(r rueidis.Client, k redis_state.DebounceKeyGenerator, q redis_state.QueueManager) Debouncer {
@@ -117,6 +120,18 @@ type debouncer struct {
 	q redis_state.QueueManager
 }
 
+// DeleteDebounceItem returns a DebounceItem given a debounce ID.
+func (d debouncer) DeleteDebounceItem(ctx context.Context, debounceID ulid.ULID) error {
+	keyDbc := d.k.Debounce(ctx)
+	cmd := d.r.B().Hdel().Key(keyDbc).Field(debounceID.String()).Build()
+	err := d.r.Do(ctx, cmd).Error()
+	if rueidis.IsRedisNil(err) {
+		return nil
+	}
+	return fmt.Errorf("error removing debounce: %w", err)
+}
+
+// GetDebounceItem returns a DebounceItem given a debounce ID.
 func (d debouncer) GetDebounceItem(ctx context.Context, debounceID ulid.ULID) (*DebounceItem, error) {
 	keyDbc := d.k.Debounce(ctx)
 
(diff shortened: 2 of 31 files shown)


pkg/execution/executor/executor.go
```diff
@@ -179,7 +189,25 @@ func (e *executor) AddLifecycleListener(l execution.LifecycleListener) {
...
+			WorkspaceID: req.WorkspaceID,
+			FunctionID:  req.Function.ID,
+			EventID:     req.Events[0].GetInternalID(),
+			Event:       req.Events[0].GetEvent(),
+		}, req.Function)
+		if err != nil {
+			return nil, err
+		}
+		e.log.Info().Interface("event", req.Events[0].GetEvent()).Msg("debouncing function")
+		return nil, ErrFunctionDebounced
```
tonyhb, line 208: Delete the debounce item if this schedules successfully.

pkg/execution/debounce/debounce.go
```diff
@@ -0,0 +1,362 @@
...
+
+// Debounce debounces a given function with the given DebounceItem.
+func (d debouncer) Debounce(ctx context.Context, di DebounceItem, fn inngest.Function) error {
+	if fn.Debounce == nil {
+		return fmt.Errorf("fn has no debounce config")
+	}
+	ttl, err := str2duration.ParseDuration(fn.Debounce.Period)
+	if err != nil {
+		return fmt.Errorf("invalid debounce duration: %w", err)
+	}
```
jpwilliams, line 164: Do we want any maximums here? Thinking that a debounce period higher than a particular limit should probably fail at the time of registration.

jpwilliams:
@tonyhb Does debounce have any exclusivity with other config options? Are we okay to also be using rate limiting alongside debounce?

tonyhb:
> @tonyhb Does debounce have any exclusivity with other config options? Are we okay to also be using rate limiting alongside debounce?

It conflicts with batching, as batching changes the setup from 1:1 -> N:1, while debouncing is _always_ 1:1.  We validate this in the function config.

I think rate limiting is actually okay;  if you have 1000 events/second you might still want to rate limit a debounce (for some reason)?

pkg/execution/debounce/debounce.go
tonyhb: Good idea, yeah, I'll add some consts.

pkg/execution/executor/executor.go
```diff
@@ -182,7 +192,25 @@ func (e *executor) AddLifecycleListener(l execution.LifecycleListener) {
...
+			WorkspaceID:     req.WorkspaceID,
+			FunctionID:      req.Function.ID,
+			FunctionVersion: req.Function.FunctionVersion,
+			EventID:         req.Events[0].GetInternalID(),
+			Event:           req.Events[0].GetEvent(),
+		}, req.Function)
+		if err != nil {
+			return nil, err
+		}
+		return nil, ErrFunctionDebounced
```
darwin67, line 211: Non-blocker and follow up:
Probably better to make it _not_ return an error since this is intended behavior.

pkg/execution/state/redis_state/queue.go
```diff
@@ -519,11 +535,14 @@ func (q QueuePartition) MarshalBinary() ([]byte, error) {
 // The queue score must be added in milliseconds to process sub-second items in order.
 func (q *queue) EnqueueItem(ctx context.Context, i QueueItem, at time.Time) (QueueItem, error) {
 	if len(i.ID) == 0 {
-		i.ID = ulid.MustNew(ulid.Now(), rnd).String()
+		i.SetID(ctx, ulid.MustNew(ulid.Now(), rnd).String())
+	} else {
+		// Hash the ID.
+		// TODO: What if this is already hashed?
```
darwin67, line 541: how do we know if there's a collision?

pkg/execution/state/redis_state/queue_test.go
```diff
@@ -1158,6 +1158,243 @@ func TestQueuePartitionReprioritize(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, 1, len(partitions))
+
+			// Lease
+			lid, err := q.Lease(ctx, *partitions[0], item, time.Second*10)
+			require.NoError(t, err)
+			require.NotNil(t, lid)
+
+			err = q.RequeueByJobID(ctx, wsB.String(), jid, time.Now().Add(5*time.Second))
+			require.NotNil(t, err)
```
darwin67, line 1249: non-blocking:
testing the actual message matches would be good in the future so we know exactly which error was hit, and have more confidence in the error handling.

the current code just mean any error, and is not necessarily accurate for testing specific failure scenarios.
same for the other error matches as well.

pkg/inngest/function.go
```diff
@@ -75,6 +77,11 @@ type Function struct {
 	Edges []Edge `json:"edges,omitempty"`
 }
 
+type Debounce struct {
+	Key    *string `json:"key"`
```
darwin67, line 81: No key will just mean debounce will apply to all function runs for this function, correct?

pkg/devserver/devserver.go
```diff
@@ -166,6 +171,7 @@ func start(ctx context.Context, opts StartOpts) error {
 		executor.WithState(sm),
 		executor.WithServiceQueue(queue),
 		executor.WithServiceExecutor(exec),
+		executor.WithServiceDebouncer(debouncer),
```
darwin67, line 174: why is `debouncer` needed for both the executor instance and the executor service?

pkg/execution/debounce/debounce.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return "", fmt.Errorf("invalid debounce expression: %w", err)
+	}
+	if str, ok := out.(string); ok {
+		return str, nil
+	}
+	return fmt.Sprintf("%v", out), nil
+}
+
+func readRedisScripts(path string, entries []fs.DirEntry) {
```
darwin67, line 327: non-blocking:

this probably should be a utility function or something.
there's probably like 3 copies of this code in the entire code base at this point.

pkg/execution/debounce/debounce.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != ErrDebounceExists {
+		// There was an unkown error creating the debounce.
+		return err
+	}
+	if debounceID == nil {
+		return fmt.Errorf("expected debounce ID when debounce exists")
+	}
+
+	// A debounce must already exist for this fn.  Update it.
+	err = d.updateDebounce(ctx, di, fn, ttl, *debounceID)
```
darwin67, line 185: are there potential race conditions here since they're separate calls?

goodoldneon:
If a run takes longer than the debounce period than 2 runs could be concurrent, right? Is that expected and fine? Not sure if debounce users will expect "1 run at a time" behavior

darwin67 (approved):
Some comments and questions but otherwise looks good.

darwin67:
> If a run takes longer than the debounce period than 2 runs could be concurrent, right? Is that expected and fine? Not sure if debounce users will expect "1 run at a time" behavior

As long as the next run starts after the 1st debounce period expired, that seems fine to me.
We can update the docs to clarify on it, but debounce just means "run after some time from the last occurrence", it doesn't necessarily care about how many things are running at the same time.

If folks need only X running at any time, they should combine it with concurrency limits (though that's on steps, not functions).

pkg/execution/state/redis_state/queue.go
tonyhb: This is a followup for later.  We're going to ignore it for now;  it's a QoL issue as we expand on the queue.

pkg/inngest/function.go
tonyhb: Yep

pkg/execution/debounce/debounce.go
tonyhb: Nope.  Updating a debounce, if a debounce doesn't exist, will fail, re-invoking this function.  Debounce minimums are 1 second, so we're okay.

pkg/execution/executor/executor.go
tonyhb: Yeah.  I'm not the biggest fan of errors as status identifiers — ideally we'd have a way of responding with a status alongside an error.

PR #623, opened 2023-09-19
Add request version to metadata

tonyhb:
This PR supports request versions per function run.  The request version is discovered on the first SDK call with generator responses.  This allows us to store the 'version' that an SDK uses for hashing/request response styles.

We then use this version for the lifetime of the function.  Without this, incompatible changes released are never backwards compatible - breaking any in-progress functions.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/driver/httpdriver/httpdriver.go
+++ b/pkg/execution/driver/httpdriver/httpdriver.go
@@ -154,7 +154,7 @@ func (e executor) Execute(ctx context.Context, s state.State, item queue.Item, e
 		return nil, fmt.Errorf("Unable to use HTTP executor for non-HTTP runtime")
 	}
 
-	input, err := driver.MarshalV1(ctx, s, step, idx, "local", attempt)
+	input, err := driver.MarshalV1(ctx, s, step, idx, "", attempt)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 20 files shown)


pkg/execution/driver/httpdriver/httpdriver.go
```diff
@@ -272,6 +274,8 @@ func (e executor) do(ctx context.Context, url string, input []byte) (*response,
 		return nil, fmt.Errorf("error reading response body: %w", err)
 	}
 
+	hv, _ := strconv.Atoi(resp.Header.Get("x-inngest-req-version"))
```
goodoldneon, line 277: Can we log when their's an `atoi` error?

pkg/execution/executor/executor.go
```diff
@@ -829,31 +829,51 @@ func (e *executor) Resume(ctx context.Context, pause state.Pause, r execution.Re
...
+	if md.RequestVersion == -1 {
+		update = &state.MetadataUpdate{
+			Context:                   md.Context,
+			Debugger:                  md.Debugger,
+			DisableImmediateExecution: md.DisableImmediateExecution,
+			RequestVersion:            resp.RequestVersion,
 		}
+	}
 
+	if len(resp.Generator) > 1 {
```
goodoldneon, line 850: What does `len(resp.Generator) > 1` mean? That at least 1 generator step was executed?

pkg/execution/driver/httpdriver/httpdriver.go
tonyhb: Nope.  If there's no version, this means "0" which is explicitly different to unset.

pkg/execution/executor/executor.go
tonyhb: Parallel execution, as there's more than one thing to run.

PR #651, opened 2023-09-28
INN-2090: Replace MaxBatchSize with DefaultBatchSize to make it more configurable

darwin67:
## Description

The maximum batch size is currently hardcoded into the system.
Removing this limit and make it more configurable for the user. The limit can be adjusted on the service.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/inngest/batch.go
+++ b/pkg/inngest/batch.go
@@ -23,8 +23,8 @@ func NewEventBatchConfig(conf map[string]any) (*EventBatchConfig, error) {
 		return nil, fmt.Errorf("failed to decode batch config: %v", err)
 	}
 
-	if config.MaxSize > consts.MaxBatchSize {
-		config.MaxSize = consts.MaxBatchSize
+	if config.MaxSize <= 0 {
+		config.MaxSize = consts.DefaultBatchSize
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/inngest/batch.go
```diff
@@ -23,8 +23,8 @@ func NewEventBatchConfig(conf map[string]any) (*EventBatchConfig, error) {
 		return nil, fmt.Errorf("failed to decode batch config: %v", err)
 	}
 
-	if config.MaxSize > consts.MaxBatchSize {
-		config.MaxSize = consts.MaxBatchSize
+	if config.MaxSize <= 0 {
```
tonyhb, line 26: Can we still add a max size?  We probably don't want to increase above 100 yet, and if we do it might be for new billing plans.

This specifiically let a customer cut costs by 100x (on inngest AND Vercel) lol

darwin67: I'm going to make changes on the cloud side to make it easy to override.
This part of the code is the early validation when a function registers, and if we override here, we won't know what the user intended vs what is the actual restriction, so this provides better visibility as well.

e.g.
Oh, I see you set 100000 in your function config, but your account only allows 100. that's why it's only 100.

tonyhb (approved):

PR #662, opened 2023-10-04
INN-1091: Add concurrency limit hook to lifecycle interface

darwin67:
## Description

Add a placeholder hook when concurrency limit is hit.
This is currently not called since queues doesn't have lifecycles registered.

This hook can be used as reference as we improve concurrency instrumentation, and how it should be exposed to other areas of the codebase.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/history/lifecycle.go
+++ b/pkg/execution/history/lifecycle.go
@@ -526,6 +526,14 @@ func (l lifecycle) OnSleep(
 	}
 }
 
+// Do nothing
+func (lifecycle) OnConcurrencyLimit(
+	context.Context,
+	state.Identifier,
+	queue.Item,
+) {
+}
+
 func applyResponse(
 	h *History,
 	resp *state.DriverResponse,
--- a/pkg/execution/lifecycle.go
+++ b/pkg/execution/lifecycle.go
@@ -9,7 +9,7 @@ import (
 	"github.com/inngest/inngest/pkg/inngest"
 )
 
-var _ LifecycleListener = (*NoopLifecyceListener)(nil)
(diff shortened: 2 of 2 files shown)


pkg/execution/lifecycle.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/inngest/inngest/pkg/inngest"
 )
 
-var _ LifecycleListener = (*NoopLifecyceListener)(nil)
+var _ LifecycleListener = (*NoopLifecycleListener)(nil)
```
darwin67, line 12: just fixing a typo here.

pkg/execution/history/lifecycle.go
```diff
@@ -526,6 +526,14 @@ func (l lifecycle) OnSleep(
 	}
 }
 
+// Do nothing
+func (lifecycle) OnConcurrencyLimit(
+	context.Context,
```
goodoldneon, line 531: Are these params enough for the consumer to know which level of concurrency was hit? Like whether it was account or workflow concurrency, and eventually workspace concurrency

darwin67: I thought a little about it, but metrics only tell what happens, not why.
Which concurrency being hit should have a separate indicator, because that's the why.
This is good enough, at least for now until we learn otherwise.

tonyhb (changes requested):
This is hard to hit.  Concurrency limits are held within the queue and aren't exposed to the executor.  I'm not sure that this is viable and I dont think this is part of an execution lifecycle.  If we want to hook into concurrency limits we need to do this on a queue level, outside of this lifecycle interface.

tonyhb:
Superseded by #726

PR #691, opened 2023-10-11
refactor: clean up Tailwind CSS configuration

IGassmann:
## Description

This cleans up the Tailwind CSS configuration in preparation for the component library.

We must share a conflict-free configuration between packages to support a shared component library with Tailwind CSS.

**For a detailed list of changes, check out the commit history of this PR.**

### Key Changes

**Use complete class names**

Complete class names are safer to use with Tailwind CSS.

Tailwind can't find dynamically constructed class names and, therefore, will fail to generate the corresponding CSS classes unless we always remember to add all possible permutations to the Tailwind config file in the [`safelist` option](https://tailwindcss.com/docs/content-configuration#safelisting-classes).

If we change some code that affects the dynamically constructed class names and forget to update the `safelist` option, the CSS class names will be missing in the generated CSS file.

If we forget to remove a class name in the config file, the class name will be unnecessarily added to the CSS file, even if it's not used anywhere.

See https://tailwindcss.com/docs/content-configuration#dynamic-class-names for more.

**Don't override Tailwind CSS's default configuration**

The current configuration was overriding certain default class names in Tailwind CSS. While overriding class names can provide more customization options, it requires ejecting from the default configuration. Ejecting from it makes it more difficult to use various tools in the Tailwind ecosystem since many of them follow that default config. This includes tools we already use, such as the [Tailwind CSS Raycast extension](https://www.raycast.com/vimtor/tailwindcss).

I've adjusted the configuration so that we don't override the default values but instead extend it when possible. In the cases where that wasn't possible, I replaced it with the closest available values, which resulted in pretty minor visual changes:

**Stream (Before/After)**
before-1
after-1

**Trigger (Before/After)**

before-2
after-2

- [ ] Docs

## Checklist

### DIFF
--- a/dev-server-ui/src/app/(dashboard)/apps/page.tsx
+++ b/dev-server-ui/src/app/(dashboard)/apps/page.tsx
@@ -40,7 +40,7 @@ export default function AppList() {
         </div>
       </header>
       <div className="flex items-center gap-3 mb-4">
-        <IconWindow className="icon-xl" />
+        <IconWindow className="h-5 w-5" />
         <p className="text-slate-200">
           {numberOfConnectedApps} / {apps.length} Apps Connected
         </p>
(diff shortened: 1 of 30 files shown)


goodoldneon (approved):

anafilipadealmeida (approved):
Great job there!

The only thing probably missing is adjusting or removing this function that was setting icons height and width based on fontSizes that no longer exist:
https://github.com/inngest/inngest/blob/main/dev-server-ui/tailwind.config.cjs#L152-L160

PR #772, opened 2023-10-24
Fix GraphQL default values not working

goodoldneon:
## Description

Fix GraphQL default values not working. We had disabled optionals in GraphQL->TypeScript generated types, which breaks default values since they rely on `undefined` values.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/settings/integrations/vercel/useUpdateVercelIntegration.ts
+++ b/ui/apps/dashboard/src/app/(dashboard)/settings/integrations/vercel/useUpdateVercelIntegration.ts
@@ -68,7 +68,7 @@ export default function useUpdateVercelIntegration(initialVercelIntegration: Ver
     const createVercelAppPromises = projectsToCreate.map((project) =>
       createVercelApp({
         input: {
-          path: project.servePath ?? null,
+          path: project.servePath,
           projectID: project.id,
           workspaceID: environment!.id,
         },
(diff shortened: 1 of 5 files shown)


tonyhb (approved):
whoa weird

ui/apps/dashboard/src/app/(dashboard)/settings/integrations/vercel/useUpdateVercelIntegration.ts
```diff
@@ -68,7 +68,7 @@ export default function useUpdateVercelIntegration(initialVercelIntegration: Ver
     const createVercelAppPromises = projectsToCreate.map((project) =>
       createVercelApp({
         input: {
-          path: project.servePath ?? null,
```
jpwilliams, line 71: Ah, so is this that `servePath` was nullish because we wanted the default, but the GraphQL endpoint is then using that as a value as it has been defined?

Would it be reasonable to ensure the GraphQL API handles this instance better?

jpwilliams (approved):

ui/apps/dashboard/src/app/(dashboard)/settings/integrations/vercel/useUpdateVercelIntegration.ts
goodoldneon: Like make the Go code in GraphQL handle defaults instead of the schema?

PR #782, opened 2023-10-30
Add `step.invoke()` support for SDKs

jpwilliams:
## Description

Adds support for `step.invoke()` to SDKs.

- Adds new `inngest/function.completed` internal event
- Adds new `StepInvokingFunction` op code and generator handler
- Adds new `OnInvokeFunction` and `OnInvokeFunctionResumed` lifecycle methods
- Adds new `InvokeFunction` and `InvokeFunctionResult` fields to history
- Refactors `event.data._inngest_fn` to be `event.data._inngest.fn_id` to make room for other such Inngest data

```mermaid
sequenceDiagram
autonumber

participant SDK
participant E as Executor
participant EI as Event PubSub

E->>SDK: Executes function
SDK-->>+E: 206 - StepInvokingFunction
E->>E: Create waitForEvent-style pause + timeout job<br/>for inngest/function.completed
E->>-EI: Send inngest/function.invoked event
EI-)+E: inngest/function.invoked
E->>E: Find function to invoke from ID in event
E->>-E: Initialize function...
note over SDK,EI: On function completion...
E->>+E: OnFunctionFinished lifecycle
E->>-EI: inngest/function.completed
EI-)+E: inngest/function.completed
E->>-E: Consume pause for inngest/function.completed,<br/>set data to result (in event body)
E->>SDK: Continue function
```

## Todo

- [x] Add toggle for internal `inngest/*` events that haven't triggered anything on dev server
- [x] Fix UI not displaying payloads in some cases
- [x] Failure cases + tests
  - [x] Function not found
  - [x] Function failed
- [x] TODOs

- [ ] Docs

## Checklist

## Related

- inngest/inngest-js#368
- inngest/monorepo#1988

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/event-search/EventSearch.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/event-search/EventSearch.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { useCallback, useState } from 'react';
+import type { Route } from 'next';
 import { Button } from '@inngest/components/Button';
 import { Link } from '@inngest/components/Link';
 import type { NavigateToRunFn } from '@inngest/components/Timeline';
@@ -100,9 +101,11 @@ export function EventSearch({ environmentSlug }: Props) {
       return (
         <Link
           internalNavigation
-          href={`/env/${encodeURIComponent(environment.slug)}/functions/${encodeURIComponent(
-            opts.fnID
-          )}/logs/${opts.runID}`}
+          href={
+            `/env/${encodeURIComponent(environment.slug)}/functions/${encodeURIComponent(
+              opts.fnID
+            )}/logs/${opts.runID}` as Route
+          }
         >
           Go to run
         </Link>
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/logs/[runId]/StreamDetails.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/logs/[runId]/StreamDetails.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { useMemo } from 'react';
+import type { Route } from 'next';
 import { EventDetails } from '@inngest/components/EventDetails';
 import { Link } from '@inngest/components/Link';
 import { RunDetails } from '@inngest/components/RunDetails';
@@ -60,9 +61,11 @@ export function StreamDetails({
     return (
       <Link
         internalNavigation
-        href={`/env/${environment.slug}/functions/${encodeURIComponent(opts.fnID)}/logs/${
-          opts.runID
-        }`}
+        href={
+          `/env/${environment.slug}/functions/${encodeURIComponent(opts.fnID)}/logs/${
+            opts.runID
+          }` as Route
+        }
       >
         Go to run
       </Link>
--- a/ui/packages/components/src/Link/Link.tsx
+++ b/ui/packages/components/src/Link/Link.tsx
@@ -18,7 +18,7 @@ const defaultLinkStyles =
 export function Link({ href, children, className, internalNavigation = false }: LinkProps) {
   if (internalNavigation) {
     return (
-      <NextLink href={href as Route} className={classNames(className, defaultLinkStyles)}>
+      <NextLink href={href} className={classNames(className, defaultLinkStyles)}>
         {children}
         <IconArrowRight className="h-3 w-3 -translate-x-3 text-indigo-400 opacity-0 transition-all group-hover:translate-x-0 group-hover:opacity-100" />
       </NextLink>


pkg/consts/consts.go
```diff
@@ -64,12 +64,14 @@ const (
 	// MaxEvents is the maximum number of events we can parse in a single batch.
 	MaxEvents = 5_000
 
+	InngestEventDataPrefix = "_inngest"
```
jpwilliams, line 67: We already seemed to need two variables here, so wanted to roll them up. Either this or the `inngest/function.invoked` event namespaces `data` instead of us adding some `_inngest*` keys to it.

pkg/devserver/lifecycle.go
```diff
@@ -45,3 +53,60 @@ func (l lifecycle) OnFunctionScheduled(
...
+	} else {
+		data["result"] = resp.Output
+	}
+
+	evt := event.Event{
+		ID:        ulid.MustNew(uint64(now.UnixMilli()), rand.Reader).String(),
+		Name:      event.FnCompletedName,
+		Timestamp: now.UnixMilli(),
+		Data:      data,
+	}
```
jpwilliams, line 93: Should this use the same `user` as the original event? To get a trail of all events stemming from a particular user this seems reasonable, though this isn't _triggered_ by the user.

pkg/devserver/lifecycle.go
```diff
@@ -45,3 +53,60 @@ func (l lifecycle) OnFunctionScheduled(
...
+
+	_ = l.pb.Publish(
+		ctx,
+		l.eventTopic,
+		pubsub.Message{
+			Name:      event.EventReceivedName,
+			Data:      string(byt),
+			Timestamp: now,
+		},
+	)
```
jpwilliams, line 111: Sending this is pretty critical, though lifecycle methods probably shouldn't contain crticial flows. Thinking this should move outside of a lifecycle method.

pkg/event/event.go
```diff
@@ -7,12 +7,14 @@ import (
...
 	"time"
 
+	"github.com/inngest/inngest/pkg/consts"
 	"github.com/oklog/ulid/v2"
 )
 
 const (
 	EventReceivedName = "event/event.received"
 	FnFailedName      = "inngest/function.failed"
+	FnCompletedName   = "inngest/function.completed"
```
jpwilliams, line 17: This should probably be named `inngest/function.finished` to be inkeeping with other terminology.

This also doesn't fit nicely with `inngest/function.failed`. The downside is that we'd have to create two pauses when waiting for invocation results if we had the two events being sent. My lean is that we ditch `inngest/function.failed` and consolidate everything into the `inngest/function.finished` event, though this would likely break some user flows if they have manually targetted that event.

pkg/execution/runner/runner.go
```diff
@@ -352,7 +352,15 @@ func FindInvokedFunction(ctx context.Context, tracked event.TrackedEvent, fl cqr
...
-	name := evt.Data[consts.InvokeSlugKey]
+	name := ""
+	if inngestObj, ok := evt.Data[consts.InngestEventDataPrefix].(map[string]any); ok {
+		if slug, ok := inngestObj[consts.InvokeFnID].(string); ok {
+			logger.From(ctx).Debug().Str("event", evt.Name).Str("FunctionID", slug).Msg("invoking function by event; it was an invoke event")
+			name = slug
+		} else {
+			return nil, fmt.Errorf("invalid slug type")
+		}
+	}
```
jpwilliams, line 363: 🤮 I'm guessing it's better to create a struct for this instead of manually stepping over `map[string]any`s.

pkg/consts/consts.go
```diff
@@ -64,12 +64,14 @@ const (
...
 	// InvokeEventName is the event name used to invoke specific functions via an
 	// API.  Note that invoking functions still sends an event in the usual manner.
 	InvokeEventName = "inngest/function.invoked"
 	// InvokeSlugKey is the data key used to store the fn name when invoking a function
 	// via an RPC-like call, abstracting event-driven fanout.
-	InvokeSlugKey = "_inngest_fn"
+	InvokeFnID          = "fn_id"
+	InvokeCorrelationId = "correlation_id"
```
tonyhb, line 74: Good ol' MQ.  It's good to stick to the standards — great idea.

pkg/devserver/lifecycle.go
```diff
@@ -45,3 +53,60 @@ func (l lifecycle) OnFunctionScheduled(
...
+
+func (l lifecycle) OnFunctionFinished(
+	ctx context.Context,
+	id state.Identifier,
+	item queue.Item,
+	resp state.DriverResponse,
+	s state.State,
+) {
+	now := time.Now()
+	data := map[string]interface{}{
```
tonyhb, line 65: FYI, and we do _not_ need to change it, Go 1.20 adds `any` as a nicer keyword for `interface{}`:  `map[string]any`.

pkg/devserver/lifecycle.go
tonyhb: This is a really good question.  To be honest, I quite like the idea of being able to `POST` the entire event, excluding name:
```
{
  id: "whatever", // for idempotency,
  data: {},
  user: {},
}
```

WDYT?  We might want to be able to invoke with idempotency.  User could be a good idea to add for our future plans.  I don't want to complicate the SDK's API, though.

pkg/devserver/lifecycle.go
tonyhb: We definitely need to log a hard error here.  Realistically, the one component that _cannot_ go down are our message streams.  Without those, things are in dire straights (as they are for every large company).

pkg/event/event.go
tonyhb: 100% on `function.finished`.  Ditching `inngest/function.failed` is backcompat difficult.  A migration plan:

1. We start sending one event now: `inngest/function.completed` with a status of success/error/cancel, etc. while keeping `inngest/function.failed.
2. We update the SDK's function handlers to listen to these events with an expression
3. Eventually we deprecate `inngest/function.failed`

That said, failure handling is _so common_ that I don't hate special casing it.

pkg/execution/history/lifecycle.go
```diff
@@ -482,6 +483,102 @@ func (l lifecycle) OnWaitForEventResumed(
 	}
 }
 
+// OnInvokeFunction is called when a function is invoked from a step.
+func (l lifecycle) OnInvokeFunction(
```
tonyhb, line 487: change to `OnInvokeScheduled` to fit the rest of the naming params?

pkg/execution/history/lifecycle.go
```diff
@@ -482,6 +483,102 @@ func (l lifecycle) OnWaitForEventResumed(
...
+		if err := d.Write(context.WithoutCancel(ctx), h); err != nil {
+			l.log.Error("execution lifecycle error", "lifecycle", "onInvokeFunction", "error", err)
+		}
+	}
+}
+
+// OnInvokeFunctionResumed is called when a function is resumed from an
+// invoke function step. This happens when the invoked function has
+// completed or the step timed out whilst waiting.
+func (l lifecycle) OnInvokeFunctionResumed(
```
tonyhb, line 537: `OnInvokeCompleted`?

pkg/execution/runner/runner.go
tonyhb: 100%, the problem being we can't get that from a plain ol' event.

ui/packages/components/src/utils/historyParser/types.ts
```diff
@@ -37,7 +37,8 @@ export type HistoryType =
   | 'StepScheduled'
   | 'StepSleeping'
   | 'StepStarted'
-  | 'StepWaiting';
+  | 'StepWaiting'
+  | 'StepInvokingFunction';
```
tonyhb, line 41: `StepInvoked`?  shorter == better all round.  At scale (and we do hundreds of millions a month) bandwidth costs go up for us and our users.

tonyhb (commented):
This is looking really good.

pkg/devserver/lifecycle.go
jpwilliams: We have support for `data`, `user`, and `v` right now, but `id` is omitted from any event sending in TS. Sounds like we can add that to all entry points (`.invoke()`, `.send()`, `.sendEvent()`) in a follow-up.

pkg/devserver/lifecycle.go
jpwilliams: Yeah. Let's keep the lifecycle contract the same and I'll pull it out into another function.

pkg/consts/consts.go
jpwilliams: Mostly as we can have many invocations from one run, so anything we have in the triggering event (event ID, run ID) still needs more data adding to it to be meaningful.

pkg/execution/history/lifecycle.go
jpwilliams: Matching `OnWaitForEvent`, where we seem to have `On[Tool]`:
```diff
OnWaitForEvent
OnWaitForEventResumed
OnSleep
+ OnInvokeFunction
+ OnInvokeFunctionResumed
```

pkg/execution/history/lifecycle.go
jpwilliams: Matching `OnWaitForEvent`, where we seem to have `On[Tool]`:
```diff
OnWaitForEvent
OnWaitForEventResumed
OnSleep
+ OnInvokeFunction
+ OnInvokeFunctionResumed
```

ui/packages/components/src/utils/historyParser/types.ts
jpwilliams: I can change the enum: https://github.com/inngest/inngest/blob/571accdd55e046fa18cf311c1e2c2e77742970c1/pkg/enums/history_type.go#L7-L26
Honestly I figured they were serialized as integers. 😅

pkg/execution/driver/driver.go
```diff
@@ -64,8 +73,34 @@ func MarshalV1(
...
+		}
+
+		if len(fn.Triggers) != 1 {
+			return nil, fmt.Errorf("invocation event found, but function %s has %d triggers; could not fill in invocation event name automatically", fn.Slug, len(fn.Triggers))
+		}
+
+		trigger := fn.Triggers[0]
+		// Crons can keep the original event name
+		if trigger.EventTrigger != nil && trigger.Event != "" {
+			evt["name"] = trigger.Event
```
tonyhb, line 102: So, I get it but I'm wondering why this is necessary for the SDK?

A few things that could go wrong:

- It might make external logging lie.  We didn't actually receive this event name.  On the user's side, they'll see many more events of this type than they're actually sending.
- The events received from the SDK won't correlate with our dashboard

jpwilliams: @tonyhb Two paths here: we can do this or make `event.name` in the SDK be a union of `"the-actual-event" | "inngest/function.invoked"`.

I don't have a strong opinion on this either way - the latter might feel a touch confusing if a user doesn't know what invocation is, but it also provides a discovery path.

I'll delete this code and the SDK can declare it.

pkg/execution/history/lifecycle.go
```diff
@@ -486,6 +487,126 @@ func (l lifecycle) OnWaitForEventResumed(
...
+		Type:            enums.HistoryTypeStepInvokingFunction.String(),
+		Attempt:         int64(item.Attempt),
+		IdempotencyKey:  id.IdempotencyKey(),
+		EventID:         id.EventID,
+		BatchID:         id.BatchID,
+		StepName:        &op.Name,
+		StepID:          &op.ID,
+		InvokeFunction: &InvokeFunction{
+			EventID:       eventID,
+			FunctionID:    fnID,
```
goodoldneon, line 549: `fnId` and `expiry` could have their zero values here. Should we set `InvokeFunction` to `nil` in that case? Or should `InvokeFunction.FunctionID` and `InvokeFunction.Timeout` be nullable?

pkg/execution/history/lifecycle.go
```diff
@@ -486,6 +487,126 @@ func (l lifecycle) OnWaitForEventResumed(
...
+			l.log.Error("error parsing invoke function options expiry", "error", err)
+		} else {
+			expiry = optsExp
+		}
+	} else {
+		l.log.Error("invoke function options are nil")
+	}
+
+	h := History{
+		ID:              ulid.MustNew(ulid.Now(), rand.Reader),
```
goodoldneon, line 532: Nit: Can you alphabetize the fields? It makes it easier to see if we're missing one, which is tough when there are tons of them. I need to put up a PR to alphabetize the preexisting ones too

pkg/execution/history/lifecycle.go
jpwilliams: @goodoldneon Mm, I guess if `fnID` and `expiry` are zero values here then we should just return early instead of logging an error and continuing; they're required to be defined if we're calling this lifecycle hook.

pkg/execution/history/lifecycle.go
jpwilliams: @goodoldneon Is there something that does that for me? 😬

goodoldneon: I use vim's `:sort` but VS Code has an equivalent!
image

pkg/execution/history/lifecycle.go
goodoldneon: Maybe we should still write but set `InvokeFunction: nil` if we don't have all the data for `InvokeFunction`. Seeing the field as `null` in BigQuery would help troubleshooting

jpwilliams: @goodoldneon I'll add a check that we have all of the data before writing the log.

For the future, is it better for history in these cases to provide only everything/nothing, or is partial data also helpful? It'd lower the quality of the data overall, requiring readers to do more work, though in terms of helping debugging it might provide more threads to pull.

pkg/devserver/devserver.go
```diff
@@ -252,3 +267,84 @@ func createInmemoryRedis(ctx context.Context) (rueidis.Client, error) {
...
+					logger.From(ctx).Debug().Str("data_value_str", dataValue).Msg("data_value")
+					data[consts.InvokeCorrelationId] = dataValue
+				}
+			}
+		}
+
+		if opts.Err != nil {
+			data["error"] = opts.Err
+		} else {
+			data["result"] = opts.Result
```
tonyhb, line 317: This seems like it's repeated here and on our cloud.  Is there a way we can normalize this into one function which we use both here and imported into our cloud repo?

pkg/execution/executor/executor.go
```diff
@@ -806,9 +838,33 @@ func (e *executor) HandlePauses(ctx context.Context, iter state.PauseIterator, e
...
 			}
 
+			withEvt := evt.GetEvent()
+			with := withEvt.Map()
+			var runID *ulid.ULID
+
+			isInvokeFunctionOpcode := pause.Opcode != nil && *pause.Opcode == enums.OpcodeInvokeFunction.String()
+			isFnFinishedEvent := withEvt.Name == event.FnFinishedName
+
+			if isInvokeFunctionOpcode && isFnFinishedEvent {
```
tonyhb, line 848: Can we add a comment or extract this into a function so that it's clearer _why_ we need to do this?

On reading the code, it's not obvious that we need to do this specifically because "function invocation should resume with the function's returned data, not with the event wrapper".

pkg/execution/executor/executor.go
```diff
@@ -1168,6 +1232,101 @@ func (e *executor) handleGeneratorSleep(ctx context.Context, gen state.Generator
...
+	// Always create an invocation event.
+	evt := event.NewInvocationEvent(event.NewInvocationEventOpts{
+		Event:         *opts.Payload,
+		FnID:          opts.FunctionID,
+		CorrelationID: &correlationID,
+	})
+
+	logger.From(ctx).Debug().Interface("evt", evt).Str("gen.ID", gen.ID).Msg("created invocation event")
+
+	if e.handleSendingEvent == nil {
```
tonyhb, line 1313: We should check this at the top before starting this function.  Fail early before any mutations.

pkg/execution/executor/executor.go
```diff
@@ -1229,7 +1390,9 @@ func (e *executor) handleGeneratorWaitForEvent(ctx context.Context, gen state.Ge
 	// the pause so this race will conclude by calling the function once, as only
 	// one thread can lease and consume a pause;  the other will find that the
 	// pause is no longer available and return.
+	jobID := fmt.Sprintf("%s-%s-%s", item.Identifier.IdempotencyKey(), gen.ID, "wait")
 	err = e.queue.Enqueue(ctx, queue.Item{
+		JobID:       &jobID,
```
tonyhb, line 1395: Why was this added?  I understand it, but an curious given waitForEvent has existed since day zero.

pkg/execution/executor/service.go
```diff
@@ -114,47 +116,110 @@ func (s *svc) Executor() execution.Executor {
...
+				if err != nil {
+					return fmt.Errorf("error publishing failure event: %w", err)
+				}
+
+				return nil
+			})
 		}
 
-		return nil
+		// send inngest/function.finished
```
tonyhb, line 168: Are we doing this on _every_ function finish now?  Or only if there's a correlation ID?

Additionally, should the "function finished" event generation be bundled into the executor itself?  In the cloud we obviously have a wrapper such, but the actual generation of the `event.Event` is the same, right?

tonyhb (changes requested):

pkg/execution/executor/service.go
jpwilliams: @tonyhb That's the plan atm from https://github.com/inngest/inngest/pull/782#discussion_r1376256525. We can limit, though an `inngest/function.finished` event is pretty cool to be able to listen to across the system.

It doubles the number of events, of course...

pkg/execution/executor/executor.go
jpwilliams: Hm I think this might have been creating some duplicate calls after a `waitForEvent` after I'd added `invoke`. https://github.com/inngest/inngest/commit/30c7b8db4bc98914b6d7374b9409c3868c0f8d1b

I can pull this out and debug a touch - a comment would be good if it's the case.

jpwilliams: @tonyhb atm, because `waitForEvent` has no `JobID`, it looks like it uses an empty string `""` for idempotency. When that happens, it looks like we instead use a generated ID for idempotency, which is obviously different every time.

If two tools lack setting a `JobID` then parallelism can cause us to re-enqueue the same steps again and again because idempotency is basically removed.

In fact, while this is mostly a fix for adding a new tool, this would definitely cause problems with parallelism and `waitForEvent` anyway. Setting this now won't cause any more damage than it being unset currently. I'll raise a separate PR to get that fix in.

jpwilliams: @tonyhb Have added #859 for this.

pkg/devserver/devserver.go
tonyhb: I still see this logic - specifically from line 305 -> 325 - duplicated between here and the cloud.  Ideally we'd extract this into a function such that the actual _event generation itself_ is written once and imported.

pkg/execution/state/pause.go
```diff
@@ -160,3 +166,39 @@ func (p Pause) Edge() inngest.Edge {
 		Incoming: p.Incoming,
 	}
 }
+
+type ResumeData struct {
+	// RunID is the
```
tonyhb, line 171: Im dying with anticipation :D

PR #843, opened 2023-11-14
Handle returning step errors inside ops

jpwilliams:
## Description

Handles errors during running a step being returned within that step's regular op instead of as a `500 Internal Server Error`.

image

- [ ] Docs

## Checklist
- [ ] Confirm that attempt counts still work with this change

## Related

- inngest/inngest-js#411


goodoldneon (approved):

pkg/execution/state/driver_response.go
```diff
@@ -381,6 +401,24 @@ func (r *DriverResponse) Final() bool {
...
+	if r.Generator == nil || len(r.Generator) != 1 {
+		return nil
+	}
+	return r.Generator[0]
+}
+
+// IsSingleStepError returns whether this response is an error from running a
+// step.
+func (r *DriverResponse) IsSingleStepError() bool {
+	if step := r.SingleStep(); step != nil && isJsonNonNullish(step.Error) {
```
tonyhb, line 416: Why do we have to call` isJsonNonNullish` every time?  Shouldn't we be able to track whether this is a step error by looking at the opcode?

PR #866, opened 2023-11-22
Components for run list

anafilipadealmeida:
## Description

- Adjusts Link component to new design and fixes types.
- Adds new Slideover shared component already using radix instead of headlessUI

https://github.com/inngest/inngest/assets/16758464/5eaa8841-420c-42f8-803d-a8b96690b6ef

https://github.com/inngest/inngest/assets/16758464/11c17404-6897-420c-baf5-d7ae1011b0d3

- [X] Chore (refactors, upgrades, etc.)
- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/Link/Link.tsx
+++ b/ui/packages/components/src/Link/Link.tsx
@@ -1,6 +1,6 @@
 import NextLink from 'next/link';
+import { IconArrowRight } from '@inngest/components/icons/ArrowRight';
 import { IconArrowTopRightOnSquare } from '@inngest/components/icons/ArrowTopRightOnSquare';
-import { IconChevron } from '@inngest/components/icons/Chevron';
 import { classNames } from '@inngest/components/utils/classNames';
 
 interface LinkProps {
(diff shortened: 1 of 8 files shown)


ui/packages/components/src/Metadata/MetadataItem.tsx
```diff
@@ -25,8 +25,11 @@ export function MetadataItem({ className, value, title, label, type, tooltip }:
         {tooltip && (
           <TooltipProvider>
             <Tooltip>
-              <TooltipTrigger>
-                <IconInfo className="h-4 w-4 text-slate-400" />
+              <TooltipTrigger asChild>
+                {/* Temporarily breaks accessibility https://github.com/radix-ui/primitives/discussions/560 */}
```
anafilipadealmeida, line 29: I'll look for a long term solution after I finish the rest of the work with the runs feature

ui/packages/components/src/Link/Link.tsx
```diff
@@ -1,31 +1,44 @@
...
 import { classNames } from '@inngest/components/utils/classNames';
 
-interface LinkProps {
+type LinkProps<PassedHref extends string> = {
   internalNavigation?: boolean;
   children: React.ReactNode;
   className?: string;
-  href: string;
-}
+  href?: Route<PassedHref> | UrlObject;
```
goodoldneon, line 12: What's the reason for the `Route<PassedHref>` instead of `string`? We end up passing values of type `string` to the `href` property

anafilipadealmeida: It's required for the <NextLink>. Since the <Link> component is either a Next js link or an anchor, without it we get a `Type error: Type 'string' is not assignable to type 'UrlObject | RouteImpl<string>'.`

goodoldneon: What does the generic do? Should we use `Route` instead of `Route<PassedHref>` since it looks like we're passing a normal string to it in the `BlankSlate` component?

anafilipadealmeida: To be honest I don't think we'll need more control over the href prop, so using just the `Route` is good enough for me

darwin67 (approved):

jpwilliams (approved):

PR #872, opened 2023-11-24
Fix some `waitForEvent` snagging

jpwilliams:
## Description

Fixes some oddities when using `waitForEvent` with expressions.

- Process `waitForEvent` opcodes first, ensuring we don't enqueue future work if expressions fail to parse
- Mark the function as failed if a `waitForEvent` expression fails to parse, solving an issue where the function is seen as successfully completed in history even if the expression has failed and canceled the function
- Cancel any other parallel steps running when expressions fail

There are still some outstanding fixes that can land in another PR:

- [ ] Neither the `waitForEvent` step nor the function itself shows the error resulting from failing expressions
- [ ] Reported steps that are not reached are never shown in the timeline; we could show these as "Canceled" or "Skipped"
- [ ] Steps that are not executed following a function failing are shown as hanging; we could show these as "Canceled" or "Skipped"
- [ ] ~~Duplicate pauses are created on every report (from @goodoldneon)~~ Fixed in #906

- [ ] Docs

## Checklist


pkg/execution/executor/util.go
```diff
@@ -82,3 +82,30 @@ func sortOps(opcodes []*state.GeneratorOpcode) {
 		return opcodes[i].Op < opcodes[j].Op
 	})
 }
+
+// opGroups groups opcodes by their type, ensuring we run `waitForEvent` opcodes
+// first. This is used to ensure that we save wait triggers as soon as possible,
+// as well as capturing expression errors early.
+func opGroups(opcodes []*state.GeneratorOpcode) [][]*state.GeneratorOpcode {
+	var waitForEventGroup []*state.GeneratorOpcode
+	var otherGroup []*state.GeneratorOpcode
```
darwin67, line 91: could you make these as a map or a struct?
nested lists are not something to be considered comprehensive.

also this seems like a simple input -> output function, so would be great to have some unit tests on it.

darwin67: Logic here can probably be better encapsulated if a struct is used as well.
https://github.com/inngest/inngest/pull/872/files#diff-ccd0679a9f24c84fef4f5d4d5fe2c32651d23886d85af8448bb968921c7b8a95R962-R980

darwin67 (approved):

PR #891, opened 2023-12-01
Fix finish handler panicking if state is missing

jpwilliams:
## Description

Fixes a panic if state is missing when attempting to run a run finish handler.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -937,10 +937,11 @@ func (e *executor) HandlePauses(ctx context.Context, iter state.PauseIterator, e
 
 // Cancel cancels an in-progress function.
 func (e *executor) Cancel(ctx context.Context, runID ulid.ULID, r execution.CancelRequest) error {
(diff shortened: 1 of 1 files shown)


pkg/execution/executor/executor.go
```diff
@@ -937,7 +946,18 @@ func (e *executor) Cancel(ctx context.Context, runID ulid.ULID, r execution.Canc
 
 	// TODO: Load all pauses for the function and remove, once we index pauses.
 
-	s, _ := e.sm.Load(ctx, runID)
+	s, err := e.sm.Load(ctx, runID)
+	if err != nil {
```
tonyhb, line 950: We call `e.sm.Metadata` at the beginning of this function, so we may as well call `e.sm.Load` at the top and get Metadata out of the loaded state.  This reduces the number of calls to the state store.

jpwilliams: Yeah, also weird that we load state after canceling.

tonyhb (approved):

PR #928, opened 2023-12-07
fix(dashboard): webhooks don't save

IGassmann:
## Description

This fixes an issue with the webhook page. Currently, we're not able to save webhook changes on the Dashboard due to a conflict between Monaco and Clerk. See https://github.com/clerk/javascript/issues/1643 for context.

Fix source: https://clerk.com/docs/troubleshooting/script-loading#script-loading

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/[keyID]/TransformEvent.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/[keyID]/TransformEvent.tsx
@@ -3,6 +3,7 @@
 import { useContext, useEffect, useState } from 'react';
 import type { Route } from 'next';
 import { usePathname, useRouter } from 'next/navigation';
+import { useClerk } from '@clerk/nextjs';
 import { Button } from '@inngest/components/Button';
 import { CodeBlock } from '@inngest/components/CodeBlock';
 import { toast } from 'sonner';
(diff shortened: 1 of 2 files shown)


djfarrelly (commented):
@IGassmann should we include this clerk hook within the `CodeBlock` component instead so the logic isn't duplicated here and there and would help us where codeblock is used in other places?

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/[keyID]/TransformEvent.tsx
```diff
@@ -170,48 +172,60 @@ export default function TransformEvents({ keyID, metadata, keyName }: FilterEven
...
+                {
+                  label: 'Payload',
+                  content: incoming,
+                  readOnly: false,
+                  language: 'json',
+                  handleChange: handleIncomingCodeChange,
+                },
+              ]}
+            />
+          )}
```
Comment (IGassmann) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/[keyID]/TransformEvent.tsx:211]: Git diff is hallucinating here. I only added:

```ts
{/* This is needed because Monaco editor collides with Clerk which causes bugs.
See https://github.com/clerk/javascript/issues/1643 */}
{clerk.loaded && (
// ...
```

IGassmann:
> @IGassmann should we include this clerk hook within the `CodeBlock` component instead so the logic isn't duplicated here and there and would help us where codeblock is used in other places?

We can't include it there because it's in the component library and used by the Dev Server UI, which doesn't use Clerk. I can, however, make a Dashboard-specific component (`DashboardCodeBlock`) that wraps the component and does the check.

djfarrelly (approved):
We could add a nice loading placeholder in the future before clerk is loaded, but let's ship it!

PR #937, opened 2023-12-14
Add useSkippableGraphQLQuery

goodoldneon:
## Description

Split `useGraphQLQuery` into `useSkippableGraphQLQuery` and `useGraphQLQuery`. This is used in the new apps page, since its GraphQL queries aren't skippable.

- [ ] Docs

## Checklist


djfarrelly (approved):
The change looks sound to improve types. I'm happy for this to be merged.

Separate from this change, I'm wasn't exactly sure why we're wrapping `useQuery` at all here with some of the internal features, but Aaron explained about the issues with types returned from urql's useQuery - this makes sense. Before understanding this, it felt like this pattern could be a code smell as it seemed we may be replicating included functionality (e.g. urql's `pause` and changing it to `skip` - extra cognitive load on changing standard arguments, similar to replacing `onClick` with something like `handleAction` in a React component).

PR #938, opened 2023-12-14
Add "Send to Dev Server" button to forward events

djfarrelly:
## Description

The goal of this is to make it easier to test webhook events or debug production issues by forwarding events from the cloud dashboard to the dev server (when running).

This adds a new button behind a feature flip that allows the user to quickly send the event to their dev server running on the default port on localhost. This saves the user from having to copy and paste the event and can speed up local replication of webhook events or events for debugging bugs using production events.

This also uses the new codeblock component which adds a necessary Copy button to the UI.

### Open questions

* Sending the event as is copies the timestamp to the dev server so if the dev server has events that are newer, the forwarded event will show up later in the stream which is not intuitive, should we omit the timestmap? 
* Should we link directly to the event in the local dev server using the correct URL?
* Should the final copy be "Replay in Dev Server" to be inline with our widely used Replay term across cloud and dev server?
* Should there be a dev server icon instead of a lightning bolt? (prob OK while this on a feature flip)

### Full workflow

https://github.com/inngest/inngest/assets/1509457/ccf97500-db99-4f9d-b483-cdc9730f1a6e

### Disabled state
Screen Shot 2023-12-14 at 11 34 42 AM

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
@@ -1,7 +1,12 @@
+'use client';
+
+import { IconLightningBolt } from '@inngest/components/icons/LightningBolt';
 import { type JsonValue } from 'type-fest';
 
-import SyntaxHighlighter from '@/components/SyntaxHighlighter';
+import DashboardCodeBlock from '@/components/DashboardCodeBlock/DashboardCodeBlock';
+import { useBooleanFlag } from '@/components/FeatureFlags/hooks';
 import { getFragmentData, graphql, type FragmentType } from '@/gql';
+import { useDevServer } from '@/utils/useDevServer';
 
 const EventPayloadFragment = graphql(`
   fragment EventPayload on ArchivedEvent {
@@ -13,8 +18,10 @@ type EventPayloadProps = {
   event: FragmentType<typeof EventPayloadFragment>;
 };
 
-export default async function EventPayload({ event }: EventPayloadProps) {
+export default function EventPayload({ event }: EventPayloadProps) {
   const { payload } = getFragmentData(EventPayloadFragment, event);
+  const { isRunning, send } = useDevServer();
+  const { value: isSendToDevServerEnabled } = useBooleanFlag('send-to-dev-server', false);
 
   let parsedPayload: string | JsonValue = '';
   if (typeof payload === 'string') {
@@ -29,15 +36,30 @@ export default async function EventPayload({ event }: EventPayloadProps) {
   const formattedPayload = JSON.stringify(parsedPayload, null, 2);
 
   return (
-    <div className="flex h-full flex-col space-y-1.5 rounded-xl bg-slate-900 text-white">
-      <header className="bg-slate-910 flex items-center justify-between rounded-t-xl p-1.5">
-        <h3 className="px-6 py-2.5">Payload</h3>
-      </header>
-      <div className="flex-1 overflow-auto">
-        <div className="p-6">
-          <SyntaxHighlighter language="json">{formattedPayload}</SyntaxHighlighter>
-        </div>
-      </div>
-    </div>
+    <DashboardCodeBlock
+      tabs={[
+        {
+          label: 'Payload',
+          content: formattedPayload,
+          language: 'json',
+          readOnly: true,
+        },
+      ]}
+      actions={
+        isSendToDevServerEnabled
+          ? [
+              {
+                label: 'Send to Dev Server',
+                title: isRunning
+                  ? 'Send event payload to running Dev Server'
+                  : 'Dev Server is not running',
+                icon: <IconLightningBolt />,
+                onClick: () => send(payload),
+                disabled: !isRunning,
+              },
+            ]
+          : []
+      }
+    />
   );
 }
--- a/ui/packages/components/src/CodeBlock/CodeBlock.tsx
+++ b/ui/packages/components/src/CodeBlock/CodeBlock.tsx
@@ -40,9 +40,16 @@ interface CodeBlockProps {
     language?: string;
     handleChange?: (value: string) => void;
   }[];
+  actions?: {
+    label: string;
+    title?: string;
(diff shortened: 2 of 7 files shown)


ui/packages/components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -313,7 +320,19 @@ export function CodeBlock({ header, tabs }: CodeBlockProps) {
               })}
             </div>
             {!isOutputTooLarge && (
-              <div className="mr-2 flex items-center gap-2 py-2">
+              <div className="dark mr-2 flex items-center gap-2 py-2">
```
djfarrelly, line 323: "dark" was needed to have the button's disabled state look right

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
```diff
@@ -29,15 +36,30 @@ export default async function EventPayload({ event }: EventPayloadProps) {
...
+      ]}
+      actions={
+        isSendToDevServerEnabled
+          ? [
+              {
+                label: 'Send to Dev Server',
+                title: isRunning
+                  ? 'Send event payload to running Dev Server'
+                  : 'Dev Server is not running',
+                icon: <IconLightningBolt />,
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:56]: Nit: Some sort of down arrow icon seems more intuitive. Maybe a down arrow inside a cloud, since the event is coming from Cloud?
image

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
```diff
@@ -29,15 +36,30 @@ export default async function EventPayload({ event }: EventPayloadProps) {
...
+                onClick: () => send(payload),
+                disabled: !isRunning,
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:58]: What are your thoughts on polling? Otherwise the user needs to refresh the page after starting the Dev Server

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
```diff
@@ -29,15 +36,30 @@ export default async function EventPayload({ event }: EventPayloadProps) {
...
+        },
+      ]}
+      actions={
+        isSendToDevServerEnabled
+          ? [
+              {
+                label: 'Send to Dev Server',
+                title: isRunning
+                  ? 'Send event payload to running Dev Server'
+                  : 'Dev Server is not running',
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:55]: Should we mention that we only check port `8288`? Users might be running the Dev Server on a different port

goodoldneon:
> Sending the event as is copies the timestamp to the dev server so if the dev server has events that are newer, the forwarded event will show up later in the stream which is not intuitive, should we omit the timestamp?

I think it'd be good to omit the timestamp for the reason you mentioned. Maybe we add an "Include original timestamp" checkbox if users want the original timestamp

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
Comment (djfarrelly) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:56]: Good idea!

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
Comment (djfarrelly) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:55]: Good idea here too!

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx
Comment (djfarrelly) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/events/[eventName]/logs/[eventId]/EventPayload.tsx:58]: Yeah, that's fair. I thought about adding that, but also didn't want to make it too heavy weight. Perhaps we should add it at a reasonable interval. Realistically, it would be cool to make this a setting to disable as well. During the feature flip testing I think it may be ok without polling and see what user feedback that we get.

djfarrelly:
> I think it'd be good to omit the timestamp for the reason you mentioned. Maybe we add an "Include original timestamp" checkbox if users want the original timestamp

@goodoldneon, I changed it to use the Event ID, so it links directly to the event potentially mitigating this issue a little bit. I think a checkbox might be cumbersome at this point so I'd love to make a call on a good default and see what user feedback that we get. Maybe for a final version we might want a modal which has a checkbox option or some explainer of what will happen 🤔  

The trade off in my brain ends up being if it's better to omit the `ts` and have the event at the top of the stream OR if someone is using the `ts` field in their code, will changing it break their replication? I may be leaning personally toward omitting the `ts` and if someone truly needs it, they can copy and paste instead.

goodoldneon:
> > I think it'd be good to omit the timestamp for the reason you mentioned. Maybe we add an "Include original timestamp" checkbox if users want the original timestamp
> 
> @goodoldneon, I changed it to use the Event ID, so it links directly to the event potentially mitigating this issue a little bit. I think a checkbox might be cumbersome at this point so I'd love to make a call on a good default and see what user feedback that we get. Maybe for a final version we might want a modal which has a checkbox option or some explainer of what will happen 🤔
> 
> The trade off in my brain ends up being if it's better to omit the `ts` and have the event at the top of the stream OR if someone is using the `ts` field in their code, will changing it break their replication? I may be leaning personally toward omitting the `ts` and if someone truly needs it, they can copy and paste instead. What do you think?

I agree. I meant the checkbox as something we might add if users start saying "I want to include the original timestamp". The vast majority of users won't care about the original timestamp

djfarrelly:
Thanks for the input @goodoldneon - made that change!

goodoldneon (approved):

PR #945, opened 2023-12-18
feat(dashboard): rename View All Logs button to View All Runs

IGassmann:
## Description

This:
- [rename Function Runs chart to Ended Function Runs](https://github.com/inngest/inngest/commit/d70f91d94e7cc187538df60c6974a68d35801e90)
- [rename View All Logs button to View All Runs](https://github.com/inngest/inngest/commit/671eecb59efffab3f626996bcb7a45b437fac2e7)

...in the Function Dashboard page.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx
@@ -99,7 +99,7 @@ export default function FunctionRunsChart({
 
   return (
     <StackedBarChart
-      title="Function Runs"
+      title="Ended Function Runs"
       data={metrics}
       legend={[
         { name: 'Completed', dataKey: 'completed', color: colors.teal['500'] },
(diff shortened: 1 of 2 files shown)


ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx
```diff
@@ -99,7 +99,7 @@ export default function FunctionRunsChart({
 
   return (
     <StackedBarChart
-      title="Function Runs"
+      title="Ended Function Runs"
```
Comment (djfarrelly) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx:102]: We don't need to change the title of this chart, they key point of confusion with this chart is not whether they have "ended" or not, it's the x-axis being the run start time and not the end time, which we need to change with the query. The current chart shows by start time, which means that amounts in the past can increase which is non-intuitive - it makes the chart look like it's broken when there is a backlog. If we make the change to the query, I believe it's all fixed.

The legend says "Completed" "Failed" and "Canceled" which are all past-tense, so "Ended" is assumed here.

djfarrelly (changes requested):
Hey @IGassmann - great initiative with this, but I have some notes on the change of title that we shouldn't change now 😄

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx
Comment (IGassmann) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/FunctionRunsChart.tsx]: Makes sense. I reverted that commit to only leaving the View All Logs button change.

djfarrelly (approved):

PR #947, opened 2023-12-18
Persist time range selected in URL state

djfarrelly:
## Description

Move state to the URL search params so the range is persisted on refreshes, with the back button, and is sharable.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx
+++ b/ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx
@@ -15,44 +15,57 @@ const currentTime = new Date();
 
 const timeRanges: TimeRange[] = [
   {
+    key: '30m',
     start: new Date(currentTime.valueOf() - 30 * 60 * 1_000), // 30 Minutes
     end: currentTime,
   },
   {
+    key: '60m',
     start: new Date(currentTime.valueOf() - 60 * 60 * 1_000), // 60 Minutes
     end: currentTime,
   },
   {
+    key: '6h',
     start: new Date(currentTime.valueOf() - 6 * 60 * 60 * 1_000), // 6 Hours
     end: currentTime,
   },
   {
+    key: '12h',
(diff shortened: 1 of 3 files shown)


tonyhb (approved):

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx
```diff
@@ -15,45 +15,58 @@ const currentTime = new Date();
 
 const timeRanges: TimeRange[] = [
   {
+    key: '30m',
     start: new Date(currentTime.valueOf() - 30 * 60 * 1_000), // 30 Minutes
     end: currentTime,
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx:20]: Out of scope: I just realized that `currentTime` never changes after page load, which means that users need to refresh the page to get newer data. Not a great UX 🤮

goodoldneon:
LGTM!

We'll eventually redo time filtering on this page, since "last X days" isn't fine grained enough (e.g. "I know my run was 2 days ago so why do I need to scroll through yesterday's runs to get to it?"). But that's a problem for another day

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx
Comment (djfarrelly) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx:20]: Yeah - I saw that too. I was tempted to refactor all of that to be created  by a function on the fly instead every time the request is sent. In my opinion, the timestamp should be calculated when the request is being sent, not in the component itself in case we added an "auto-refresh" option. I think this component should instead just specify time ranges, e.g. `{ key: '30m', value: 30 * 60 * 1_000 }` then the component can decide to use the default end time (now) or some other time selected by the user and calculate the start or end time on the fly before the request.

We also had duplicated and related code that is ripe for that refactor and code re-sharing as well:

https://github.com/inngest/inngest/blob/74c51fb4d2cdf80dd1f6bef8e451448a959859a6/ui/apps/dashboard/src/app/(dashboard)/env/%5BenvironmentSlug%5D/functions/%5Bslug%5D/logs/TimeRangeFilter.tsx#L32-L68

PR #959, opened 2023-12-22
feat(dashboard): sentry user identification

IGassmann:
## Description

This sets up user identification for Sentry so that we can correlate errors and transactions to users.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/SentryUserIdentification.tsx
+++ b/ui/apps/dashboard/src/app/SentryUserIdentification.tsx
@@ -1,0 +1,21 @@
+'use client';
+
+import { useEffect } from 'react';
+import { useUser } from '@clerk/nextjs';
+import * as Sentry from '@sentry/nextjs';
+
+export default function SentryUserIdentification() {
+  const { user } = useUser();
+
+  useEffect(() => {
+    if (!user?.externalId) return;
+
+    Sentry.setUser({
+      id: user.externalId,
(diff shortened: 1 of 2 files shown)


ui/apps/dashboard/src/app/SentryUserIdentification.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { useEffect } from 'react';
+import { useUser } from '@clerk/nextjs';
+import * as Sentry from '@sentry/nextjs';
+
+export default function SentryUserIdentification() {
+  const { user } = useUser();
+
+  useEffect(() => {
+    if (!user?.externalId) return;
```
goodoldneon, line 11: `useUser` returns a discriminated union, so we could narrow `user` to a non-nullish type by handling `isLoaded` and `isSignedIn`. I think that'd be a little clearer, since `user.externalId` is nullable.

It seems like we should do some alerting when `user.externalId` is null, so we may not want to conflate it with the "loading" and "not signed in" states

IGassmann: Good point. I've updated the code to use `isSignedn` to not conflate. However, I haven't added an alert for when `user.externalId` is null because it's expected to happen for a short moment when the user signs up.

jpwilliams (approved):
Nice!

PR #967, opened 2024-01-02
Add batching support to OSS

darwin67:
## Description

The batching feature is current only available on the Cloud.
This change attempts to bring that into OSS and DevServer so users are able to test batching locally.

### Note

This is mainly the backend changes. Showing the batch events on function runs will be a follow up.

- [ ] Docs

## Checklist


pkg/coreapi/gql.schema.graphql
```diff
@@ -183,6 +183,8 @@ type FunctionRun {
   function: Function
   workspace: Workspace
   event: Event
+  events: [Event]!
+  batchID: ULID # if the run is a batched run
```
darwin67, line 187: I'll add this resolver as a follow up. Not necessarily needed right now.

pkg/consts/consts.go
```diff
@@ -52,6 +52,9 @@ const (
 	// MaxConcurrencyLimits limits the max concurrency constraints for a specific function.
 	MaxConcurrencyLimits = 2
 
+	// MaxBatchTTL represents the maximum amount of duration the batch key will last
+	MaxBatchTTL = 10 * time.Minute
```
goodoldneon, line 56: The docs say the max batch timeout is 1 minute. Should this value match that?

darwin67: This is the redis TTL used for expiring the keys after batch started.

pkg/coreapi/gql.schema.graphql
```diff
@@ -183,6 +183,8 @@ type FunctionRun {
   function: Function
   workspace: Workspace
   event: Event
+  events: [Event]!
```
goodoldneon, line 186:
```suggestion
  events: [Event!]!
```

This change reflects the fact that none of the array elements can be null

pkg/devserver/lifecycle.go
```diff
@@ -33,4 +33,20 @@ func (l lifecycle) OnFunctionScheduled(
...
+			cqrs.WithEventBatchAccountID(id.AccountID),
+			cqrs.WithEventBatchWorkspaceID(id.WorkspaceID),
+			cqrs.WithEventBatchAppID(id.AppID),
+			cqrs.WithEventBatchFunctionID(id.WorkflowID),
+			cqrs.WithEventBatchRunID(id.RunID),
+			cqrs.WithEventBatchEventIDs(id.EventIDs),
+		)
+
+		if batch.IsMulti() {
+			_ = l.cqrs.InsertEventBatch(ctx, *batch)
```
tonyhb, line 49: This looks good.  I wonder if we end up making batching a lifecycle and then move InsertEventBatch to another interface, which lets us normalize dev server & cloud.  Obvs not a blocker and would be a followup

pkg/execution/batch/lua/helpers.lua
```diff
@@ -0,0 +1,30 @@
...
+local function is_status_empty(key)
+  return is_meta_empty(key, "status")
+end
+
+local function set_batch_status(key, value)
+  redis.call("HSET", key, "status", value)
+end
+
+local function get_batch_status(key)
+  return redis.call("HGET", key, "status")
```
tonyhb, line 29: i really like making redis calls functions for clarity!

tonyhb (approved):

pkg/devserver/lifecycle.go
darwin67: sounds like a plan

PR #990, opened 2024-01-09
Instant mass backlog cancellation

tonyhb:
While we already have naive mass cancellation, this PR introduces greater control over backlog management and cancellation.

WIth this PR users can cancel functions in the backlog with the following data:

```
struct {
	From       *time.Time `json:"from"`
	To         time.Time  `json:"to"`
	If         *string    `json:"if,omitempty"`
}
```

This lets you cancel runs started between a given time range, and optionally cancel runs that have an event which matches an expression (to cancel a part of the backlog based off of data).

In this PR, we implement:

- [x] A state store interface and implementation, storing cancellations in the backing state store
- [x] A redis interface to query for cancellations at a specific point in time for a fn
- [x] A checker, which conditionally checks cancellation amongst runs + event data
- [x] The executor implementation

A followup PR for the dev server will include sqlite migrations and SQL storage to implement the APIs defined here.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/apiv1/apiv1.go
+++ b/pkg/api/apiv1/apiv1.go
@@ -29,10 +29,14 @@ type Opts struct {
 	Executor execution.Executor
 	// EventReader allows reading of events from storage.
 	EventReader EventReader
+	// FunctionReader reads functions from a backing store.
+	FunctionReader cqrs.FunctionReader
 	// FunctionRunReader reads function runs, history, etc. from backing storage
 	FunctionRunReader cqrs.APIV1FunctionRunReader
 	// JobQueueReader reads information around a function run's job queues.
 	JobQueueReader queue.JobQueueReader
+	// CancellationReadWriter reads and writes cancellations to/from a backing store.
+	CancellationReadWriter cqrs.CancellationReadWriter
 }
 
 // AddRoutes adds a new API handler to the given router.
@@ -71,8 +75,11 @@ func (a *api) setup() {
 		r.Get("/runs/{runID}", a.GetFunctionRun)
 		r.Delete("/runs/{runID}", a.CancelFunctionRun)
 		r.Get("/runs/{runID}/jobs", a.GetFunctionRunJobs)
-		// r.Get("/functions", a.GetFunctions)
-		// r.Get("/functions/{id}", a.GetFunction)
+
+		r.Get("/apps/{appName}/functions", a.GetAppFunctions) // Returns an app and all of its functions.
+
+		r.Post("/cancellations", a.CreateCancellation)
+		r.Get("/cancellations", a.GetCancellations)
 	})
 }
 
--- a/pkg/api/apiv1/cancellation.go
+++ b/pkg/api/apiv1/cancellation.go
@@ -31,9 +31,9 @@ func (a api) GetCancellations(w http.ResponseWriter, r *http.Request) {
 
 type CreateCancellationBody struct {
 	// AppID is the client ID specified via the SDK in the app that defines the function.
-	AppID string `json:"appID"`
+	AppID string `json:"app_id"`
 	// FunctionID is the function ID string specified in configuration via the SDK.
-	FunctionID string     `json:"functionID"`
+	FunctionID string     `json:"function_id"`
 	From       *time.Time `json:"from"`
 	To         time.Time  `json:"to"`
 	If         *string    `json:"if,omitempty"`
--- a/pkg/api/apiv1/cancellation.go
+++ b/pkg/api/apiv1/cancellation.go
@@ -1,0 +1,82 @@
+package apiv1
+
+import (
+	"crypto/rand"
+	"encoding/json"
+	"net/http"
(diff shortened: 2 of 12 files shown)


pkg/api/apiv1/apiv1.go
```diff
@@ -71,8 +75,11 @@ func (a *api) setup() {
 		r.Get("/runs/{runID}", a.GetFunctionRun)
 		r.Delete("/runs/{runID}", a.CancelFunctionRun)
 		r.Get("/runs/{runID}/jobs", a.GetFunctionRunJobs)
-		// r.Get("/functions", a.GetFunctions)
-		// r.Get("/functions/{id}", a.GetFunction)
+
+		r.Get("/apps/{appName}/functions", a.GetAppFunctions) // Returns an app and all of its functions.
```
djfarrelly, line 79: **suggestion**: Should this be `appID` to match what field is in the response and how it looks like in the Inngest client's id param?
```suggestion
		r.Get("/apps/{appID}/functions", a.GetAppFunctions) // Returns an app and all of its functions.
```

pkg/api/apiv1/functions.go
```diff
@@ -1,11 +1,31 @@
...
+	auth, err := a.opts.AuthFinder(ctx)
+	if err != nil {
+		_ = publicerr.WriteHTTP(w, publicerr.Wrap(err, 401, "No auth found"))
+		return
+	}
+
+	fns, err := a.opts.FunctionReader.GetFunctionsByAppExternalID(
+		ctx,
+		auth.WorkspaceID(),
+		chi.URLParam(r, "appName"),
```
djfarrelly, line 23: See other suggestion for using `appID` to match other use of app id in responses, clients, etc.

djfarrelly (changes requested):
@tonyhb some requested changes for consistency and DX with existing endpoints

pkg/cqrs/cancellations.go
```diff
@@ -0,0 +1,42 @@
...
+	CreateCancellation(ctx context.Context, c Cancellation) error
+}
+
+// Cancellation represents a cancellation of many function runs during the time specified.
+type Cancellation struct {
+	ID          ulid.ULID  `json:"id"`
+	WorkspaceID uuid.UUID  `json:"workspaceID"`
+	FunctionID  uuid.UUID  `json:"functionID"`
+	From        *time.Time `json:"from"`
+	To          time.Time  `json:"to"`
```
djfarrelly, line 38: **suggestion**: Should we use ISO8601 time strings here which could support millisecond precision? We often default to returning data in ISO8601 strings in other parts of the REST API instead of dealing with unix timestamp.s

djfarrelly: **question (dx)**: Is this the function start range? Perhaps it could be a little more obvious using params like:
```
started_after
started_before
```
These types of params could carry over to other endpoints that allow querying in the future like querying all runs by a time range of start time vs. end time (`ended_after`, `ended_before`). Obviously, with this endpoint, the ended would not be supported, but it would be consistent.

pkg/api/apiv1/cancellation.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		_ = publicerr.WriteHTTP(w, publicerr.Wrap(err, 404, "Function not found"))
+		return
+	}
+
+	// Create a new cancellation for the given function ID
+	cancel := cqrs.Cancellation{
+		ID:          ulid.MustNew(ulid.Now(), rand.Reader),
+		WorkspaceID: auth.WorkspaceID(),
+		FunctionID:  fn.ID,
```
djfarrelly, line 71: **issue (dx)**: I believe this is the function's internal `uuid`, which is confusing to return as the payload accepts the same named param (`function_id`), but it's a slug there and a uuid here. We should make this the slug in the response. Ideally, it would also contain `app_id`

djfarrelly (changes requested):

pkg/api/apiv1/apiv1.go
tonyhb: These are for internal usage only, and I prefer `appName` because ID is ambiguous.

pkg/cqrs/cancellations.go
tonyhb: `time.Time` in Golang is nanosecond precision by default, and when serialized to strings is sub-second precision: https://pkg.go.dev/time#Time.MarshalJSON.

I'll make the switch on `started_after` etc.

djfarrelly: Ok! I couldn't remember in my brain if this was serialized to an ISO string of a unix timestamp/int

djfarrelly (approved):

pkg/api/apiv1/cancellation.go
tonyhb: Changed, thanks.  We need new names for referring to internal vs external IDs.

PR #993, opened 2024-01-09
Use App.functions in GraphQL

goodoldneon:
> :warning: Depends on https://github.com/inngest/monorepo/pull/2162

## Description

Use new `App.functions` field for the app list and app info pages. This field gives the non-archived functions for an app regardless of the latest sync

- [ ] Docs

## Checklist


ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts
```diff
@@ -45,20 +44,15 @@ export function useApps({ envID, isArchived }: { envID: string; isArchived: bool
...
-        const functionCount =
-          latestSync?.syncedFunctions.filter((fn) => {
-            return !fn.isArchived;
-          }).length || 0;
-
         return {
           ...app,
           latestSync,
-          functionCount,
+          functionCount: app.functions.length,
```
Comment (djfarrelly) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts:50]: Would it be faster to just have `functionsCount` on the resolver and handle this on the backend? The backend would also just be able to run a count query greater than the existing 200 limit 🤔  https://github.com/inngest/monorepo/pull/2162/files#diff-44bdfdd8fa91cc5626a104f0c1031c739dbe4217bcdac2b085d41a728ba4a717R68

Comment (goodoldneon) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts:50]: We'll eventually have an expandable list of functions in the app list page so we'll still need `App.functions`. But since that will eventually be paginated, you're right that we'll also need `App.functionCount`. I'll add that

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts
```diff
@@ -7,6 +7,7 @@ const query = graphql(`
       apps {
         id
         externalID
+        functionCount
```
Comment (djfarrelly) [inline ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts:10]: I don't see this in the new resolver changes in: https://github.com/inngest/monorepo/pull/2162

Is that in another PR?

djfarrelly (commented):
Everything looks good, but key question about the `functionsCount` existing in the resolver yet. Once that's addressed, this can be approved!

ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts
Comment (goodoldneon) [reply ui/apps/dashboard/src/app/(dashboard)/env/[environmentSlug]/apps/useApps.ts:10]: It'd probably help if I committed and pushed that change 😭. It's there now

PR #1013, opened 2024-01-13
Use version from `package.json` instead of an env variable in NPM `postinstall` script

MonsterDeveloper:
## Description

I was trying to get `inngest-cli` working with `bun x`, and after some debugging I encountered this error:
CleanShot 2024-01-13 at 15 28 16

Although not an issue with `inngest-cli` directly, I looked at the source code of `postinstall.ts`, and found out that [it was using](https://github.com/inngest/inngest/blob/90d09775242d9a759d7a8854be7ab6d38056d496/npm/postinstall.ts#L55C29-L55C48) `npm_package_version` env variable to get the package version.

NPM is [already ditching this variable](https://github.com/npm/rfcs/blob/main/implemented/0021-reduce-lifecycle-script-environment.md#detailed-explanation), but it continues to work with NPM due to compatibility reasons. Bun, however, doesn't set this env variable during the execution of the postinstall script, making `inngest-cli` unable to download the correct binary.

This PR implements a more robust package version fetching — directly from the `package.json`. 

## Checklist

### DIFF
--- a/npm/postinstall.ts
+++ b/npm/postinstall.ts
@@ -5,6 +5,7 @@ import fetch, { Response } from "node-fetch";
 import path from "path";
 import tar from "tar";
 import { URL } from "url";
+import packageJson from "./package.json";
 
 const rootDebug = Debug("inngest:cli");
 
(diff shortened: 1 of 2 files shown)


tonyhb (approved):
Thanks for the changes, @MonsterDeveloper.  Merging for the next release.

PR #1081, opened 2024-01-30
feat(dashboard): organization list page

IGassmann:
## Description

This adds the organization list page, which will later be used to allow users to switch between organizations.

image

- [ ] Docs

## Checklist


ui/apps/dashboard/src/queries/URQLProvider.tsx
```diff
@@ -60,7 +60,8 @@ export default function URQLProvider({ children }: { children: React.ReactNode }
         fetchExchange,
       ],
     });
-  }, [getToken]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- we want to re-create the client whenever the user signs out or switches organizations
+  }, [getToken, isSignedIn, orgId]);
```
goodoldneon, line 64: Is there a way we can avoid this ESLint ignore? Right now it's OK because we're ignoring excessive deps (i.e. rerunning the callback when `isSignedIn` or `orgId` change even though they aren't used in the callback), but it makes this hook vulnerable to an issue in the future. Like if we used a new dep in the callback then ESLint wouldn't warn us.

Maybe the callback should return `undefined` when `!isSignedIn || `orgId`. And then put this at the bottom:
```tsx
if (!urqlClient) {
  return children
}

<Provider value={urqlClient}>{children}</Provider>
```

IGassmann: Good point! I changed this to use a component key to reset the client. That way, we can still use URQL on unauthenticated pages.

jpwilliams (approved):

PR #1087, opened 2024-02-01
feat(dashboard): force users to have an active organization

IGassmann:
## Description

This forces users to have an active organization by redirecting them to the "Organization List" page.

On the `account-setup` page, we auto-select the newly created organization for them.

This is needed since we'll be extracting the account ID from the user's active organization.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
+++ b/ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
@@ -23,6 +23,7 @@ export default function AccountSetupPage() {
 
     // We need to refresh the token before redirecting so that the token contains the account ID
     getToken({ skipCache: true }).then(() => {
+      // We use `replace` so that the user doesn't get redirected back to this page if they click the back button
       router.replace(process.env.NEXT_PUBLIC_HOME_PATH as Route);
     });
   }, [getToken, isAccountSetup, router]);
--- a/ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
+++ b/ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
@@ -66,22 +66,44 @@ export default function AccountSetupPage() {
  */
 function useForceActiveOrganization() {
   const { value: isOrganizationsEnabled } = useBooleanFlag('organizations');
-  const { isLoaded, userMemberships, setActive } = useOrganizationList({
+  const { isLoaded, userMemberships } = useOrganizationList({
     userMemberships: {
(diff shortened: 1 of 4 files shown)


ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
```diff
@@ -3,38 +3,29 @@
...
-    };
-  }, [isAccountSetup, user]);
+  useForceActiveOrganization();
+  const isAccountSetup = useIsAccountSetup();
 
   // Redirect to the home page once the account is set up
   useEffect(() => {
     if (!isAccountSetup) return;
 
-    // We use `replace` so that the user doesn't get redirected back to this page if they click the back button
```
jpwilliams, line 35: Nit: This is a super useful comment, can we keep it?

ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
```diff
@@ -68,6 +59,74 @@ export default function AccountSetupPage() {
...
+
+  useEffect(() => {
+    if (!isOrganizationsEnabled && isLoaded && userMemberships.data[0] && !organization) {
+      setActive({ organization: userMemberships.data[0].organization });
+    }
+  }, [isLoaded, isOrganizationsEnabled, organization, setActive, userMemberships.data]);
+
+  // Reload memberships if the user doesn't have an active organization
+  useEffect(() => {
+    if (isOrganizationsEnabled || !isLoaded || userMemberships.data[0] || organization) return;
```
jpwilliams, line 84: Should this have a negation?
```suggestion
    if (isOrganizationsEnabled || !isLoaded || !userMemberships.data[0] || organization) return;
```
Would be good to roll this up into a boolean with a `useMemo()` or something which these `useEffect()` hooks then call, as it's hard to tell what we're checking with so many conditions being listed.

jpwilliams (approved):

ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
IGassmann: Removed by accident. Re-adding it

ui/apps/dashboard/src/app/(auth)/sign-up/account-setup/page.tsx
IGassmann: It shouldn't be a negation. We want to return early if a user membership is already present. We don't want to keep reloading memberships if we already have one.

PR #1106, opened 2024-02-06
S

tonyhb:
This adds shard claiming to workers, ensuring that shards are claimed according to their minimum capacities.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue_test.go
```diff
@@ -1905,6 +1905,129 @@ func TestSharding(t *testing.T) {
...
+	sf := func(ctx context.Context, wsID uuid.UUID) *QueueShard {
+		return &QueueShard{
+			Name:               wsID.String(),
+			Priority:           0,
+			GuaranteedCapacity: 1,
+		}
+	}
+	q := NewQueue(rc, WithShardFinder(sf))
+
+	t.Run("Leasing a non-existent shard fails", func(t *testing.T) {
```
darwin67, line 1927: would this block be flaky since it's run concurrently?
just thinking the lines below could end up running before this, causing the test to fail.

pkg/execution/state/redis_state/queue_test.go
```diff
@@ -1905,6 +1905,129 @@ func TestSharding(t *testing.T) {
...
+		t.Run("Leasing an index with an expired lease works", func(t *testing.T) {
+			// In this test, we have two leases — but one expires with the wait.  This first lease
+			// is no longer valid, so leasing with an index of (1) should succeed.
+			<-time.After(2 * time.Second) // Wait a few seconds so that time.Now() in the call works.
+			r.FastForward(2 * time.Second)
+			leaseID, err := q.leaseShard(ctx, shard, 10*time.Second, 1)
+			require.NotNil(t, leaseID)
+			require.Nil(t, err)
+
+			// This leases us with two valid leases.
```
darwin67, line 1978:
```suggestion
			// This leaves us with two valid leases.
```

pkg/execution/state/redis_state/queue.go
```diff
@@ -1784,6 +1800,109 @@ func (q *queue) ConfigLease(ctx context.Context, key string, duration time.Durat
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error fetching shards: %w", err)
+	}
+	shards := map[string]*QueueShard{}
+	for k, v := range m {
+		shard := &QueueShard{}
+		if err := v.DecodeJSON(shard); err != nil {
+			return nil, fmt.Errorf("error decoding shards: %w", err)
```
darwin67, line 1815: should this quit all together if it fails to decode one shard?
or should it just move on, but log the error?

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -228,6 +237,171 @@ LOOP:
...
+		default:
+			return true, err
+		}
+	}
+
+	return retry, nil
+}
+
+func (q *queue) addLeasedShard(ctx context.Context, shard *QueueShard, lease ulid.ULID) {
+	for i, n := range q.shardLeases {
```
darwin67, line 347: do these operations need locking?
I guess in other words, will tick in `scanShards` ever cause >1 go routine to run?

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -228,6 +237,171 @@ LOOP:
...
+// and priority shuffles shards to lease in a non-deterministic (but prioritized) order.
+//
+// The returned shards are safe to be leased, and should be attempted in-order.
+func (q *queue) filterShards(ctx context.Context, shards map[string]*QueueShard) ([]*QueueShard, error) {
+	if len(shards) == 0 {
+		return nil, nil
+	}
+
+	for _, v := range q.shardLeases {
+		delete(shards, v.Shard.Name)
```
darwin67, line 375: same here regarding locking.

pkg/execution/state/redis_state/queue_test.go
tonyhb: We don't mark this as `t.Parallel()` so nothing here runs concurrently.  In our cloud integration tests you're totally right, it'd cause issues.

pkg/execution/state/redis_state/queue.go
tonyhb: This indicates a pretty serious issue so I'm thinking we should quit overall.  We should never really find ourselves in a state where shards are malformed, and if we do find ourselves there I doubt the integrity of the data overall.  WDYT?

pkg/execution/state/redis_state/queue_processor.go
tonyhb: Correct, let me add that.  Thanks.

pkg/execution/state/redis_state/queue_processor.go
tonyhb: This one is lock-free - it's passed in data from a single goroutine.

darwin67 (approved):

pkg/execution/state/redis_state/queue.go
tonyhb: SGTM

PR #1129, opened 2024-02-13
INN-2769 On `step.invoke()` timeouts, set an error in state stores

jpwilliams:
## Description

When handling timeouts for `step.invoke()`, ensure an `error` is set in the state store upon failure to result in the step throwing in user code.

Currently, timeouts are very large and will be unmet; this is a change in behaviour, but the initial behaviour hasn't yet been seen.

This area could be refactored slightly to stop everything being `map[string]any`.

DX for handling this specific error for a user would be to check that the `name` of the error is `InvokeTimeoutError`. Two things here:

1. Naming is a bit meh and is a magic string
3. Highlights the need for more info on `StepError`s in the docs, specifically how to tell which step threw and what kind of error it was (runtime, type error, timeout?): INN-2774

- [ ] Docs

## Checklist

## Related

- INN-2769
- inngest/inngest-js#484
- inngest/website#676


pkg/execution/executor/service.go
```diff
@@ -305,7 +305,18 @@ func (s *svc) handlePauseTimeout(ctx context.Context, item queue.Item) error {
...
 	}
 
-	return s.exec.Resume(ctx, *pause, execution.ResumeRequest{})
+	r := execution.ResumeRequest{}
+
+	// If the pause timeout is for an invocation, store an error to cause the
+	// step to fail.
+	if pause.Opcode != nil && *pause.Opcode == enums.OpcodeInvokeFunction.String() {
+		r.SetError(
+			"InvokeTimeoutError",
```
jpwilliams, line 314: I think this should be more explicit that it's also an error generated by Inngest.
```suggestion
			"InngestInvokeTimeoutError",
```

djfarrelly (approved):

PR #1149, opened 2024-02-15
fix(dashboard): can't invite members during second organization creation

IGassmann:
## Description

This fixes a bug where the form for inviting a member would be skipped when one would create a second organization through the "Create Organization" page. This was happening because the user would be redirected by the middleware to the "Account Setup" page as soon as he had created that organization.

## Motivation
I'm currently hunting bugs around the "Organizations" feature.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/middleware.ts
+++ b/ui/apps/dashboard/src/middleware.ts
@@ -26,9 +26,11 @@ export default authMiddleware({
 
     if (
       isSignedIn &&
+      hasActiveOrganization &&
       !isAccountSetup &&
-      request.nextUrl.pathname !== '/sign-up/account-setup' &&
-      request.nextUrl.pathname !== '/organization-list'
+      request.nextUrl.pathname !== '/organization-list' &&
+      request.nextUrl.pathname !== '/create-organization' &&
+      request.nextUrl.pathname !== '/sign-up/account-setup'
     ) {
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/middleware.ts
```diff
@@ -26,9 +26,11 @@ export default authMiddleware({
 
     if (
       isSignedIn &&
+      hasActiveOrganization &&
       !isAccountSetup &&
-      request.nextUrl.pathname !== '/sign-up/account-setup' &&
-      request.nextUrl.pathname !== '/organization-list'
+      request.nextUrl.pathname !== '/organization-list' &&
+      request.nextUrl.pathname !== '/create-organization' &&
+      request.nextUrl.pathname !== '/sign-up/account-setup'
```
jpwilliams, line 33: Worth putting these in a `const` somewhere? I can see this needing to expand is all.

Maybe `const nonRedirectablePaths = ["/organization-list", ...];`.

jpwilliams (approved):

ui/apps/dashboard/src/middleware.ts
IGassmann: They aren't technically non-redirectable because we redirect to the sign-in page when the user is not signed in. I'm not sure what to call them. I'll try to think of something next when I revisit this middleware after switching to the sync flow.

PR #1158, opened 2024-02-17
Clean up initial edge execution path

tonyhb:
There are still some vestiges of our DAG-based days lurking around.  This code simplifies the initial executor path, consolidating on the single-step approach we've been rolling with for a couple years — which won't change.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -463,7 +463,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 	//
 	// This enures that we only ever enqueue the start job for this function once.
 	queueKey := fmt.Sprintf("%s:%s", req.Function.ID, key)
-	retries := req.Function.GetRetries() + 1
+	attempts := req.Function.GetRetries() + 1
 	item := queue.Item{
 		JobID:       &queueKey,
 		GroupID:     uuid.New().String(),
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -53,14 +53,6 @@ var (
 	PauseHandleConcurrency = 100
 )
 
-var (
-	// SourceEdgeRetries represents the number of times we'll retry running a source edge.
-	// Each edge gets their own set of retries in our execution engine, embedded directly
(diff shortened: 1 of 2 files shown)


pkg/execution/executor/executor.go
```diff
@@ -578,26 +571,6 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 	// we automatically enqueue all children of the dag from the root node.
 	// This can be cleaned up.
```
goodoldneon, line 572: Does the comment need updating?

pkg/execution/executor/executor.go
```diff
@@ -471,14 +463,15 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 	//
 	// This enures that we only ever enqueue the start job for this function once.
 	queueKey := fmt.Sprintf("%s:%s", req.Function.ID, key)
+	retries := req.Function.GetRetries() + 1
```
jpwilliams, line 466: This feels a bit confusing, as we're setting `MaxAttempts` to `retries + 1`; should this be called `attempts` instead?
```suggestion
	attempts := req.Function.GetRetries() + 1
```

pkg/inngest/function.go
```diff
@@ -122,6 +122,15 @@ func (f Function) ConcurrencyLimit() int {
...
 }
 
+// GetRetries returns the number of retries for the function.
+func (f Function) GetRetries() int {
+	// This uses retries in steps; a holdover from when steps were used in our DAG based approach.
+	if len(f.Steps) == 0 || f.Steps[0].Retries == nil {
+		return consts.DefaultRetryCount
+	}
+	return *f.Steps[0].Retries
+}
```
jpwilliams, line 132: Related to the comment on `retries -> attempts`, it might be better to encapsulate the `+ 1` inside this function and call it `GetAttempts`, as we can clean up the retries/attempts mess when we add `function.retries` in the config. After that:
- If SDK sends `step.retries.attempts` (legacy), `+ 1`
- If SDK sends `function.retries` (future), do not `+ 1`

PR #1167, opened 2024-02-21
Separate RegisterRequest normalization

goodoldneon:
## Description
- Separate `RegisterRequest` normalization into its own method
- Create `FromReadCloser` function to abstract away the `RegisterRequest` creation implementation details (normalization, env, and platform)
- Strip `deployId` search param during URL normalization
- Optionally force HTTPS during URL normalization. Cloud needs this

## Motivation
Separating normalization and validation is important because we want to normalize before any processing. This solves problems like:
- We're getting the checksum before normalizing
- We're doing some validation before getting/creating the `deploys` row, which means some sync failure scenarios don't save an actionable info in the `deploys` table

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/util/normalize.go
+++ b/pkg/util/normalize.go
@@ -4,32 +4,52 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 )
 
 // NormalizeAppURL normalizes localhost and 127.0.0.1 as the same string.  This
 // ensures that we don't add duplicate apps.
-func NormalizeAppURL(u string) string {
+func NormalizeAppURL(u string, forceHTTPS bool) string {
 	parsed, err := url.Parse(u)
 	if err != nil {
 		return u
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/util/normalize.go
```diff
@@ -4,32 +4,52 @@ import (
...
+		host, port, err := net.SplitHostPort(parsed.Host)
+		if err != nil {
+			return u
+		}
+
+		// this shouldn't be valid: https://api.example.com:80/api/inngest
+		if parsed.Scheme == "https" && port != "" {
+			parsed.Host = host
+			return parsed.String()
+		}
```
jpwilliams, line 36: It's unconventional, but it's valid to have HTTPS running on a custom port. Maybe we only strip the port here if `port == "443"`?

goodoldneon: Sounds reasonable to me but I'll let @tonyhb weigh in since he wrote that code

jpwilliams (approved):

PR #1201, opened 2024-02-29
Fix UI crash when event ID is custom

goodoldneon:
## Description
Fix a UI crash that happens when the event ID is custom. This was caused by some logic that expected a ULID but was given a custom non-ULID event ID.

The solution involves changing our logic to always expect our internal ULID, regardless of whether there's an external ID

## Motivation
https://linear.app/inngest/issue/INN-2831

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/coreapi/gql.query.graphql
+++ b/pkg/coreapi/gql.query.graphql
@@ -6,17 +6,11 @@ type Query {
   # Get an individual event
   event(query: EventQuery!): Event
 
-  # Get all events sent
-  events(query: EventsQuery!): [Event!]
-
   # Get all functions registered
   functions: [Function!]
 
   # Get an individual function run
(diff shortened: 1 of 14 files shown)


pkg/coreapi/gql.query.graphql
```diff
@@ -6,17 +6,11 @@ type Query {
   # Get an individual event
   event(query: EventQuery!): Event
 
-  # Get all events sent
-  events(query: EventsQuery!): [Event!]
```
jpwilliams, line 10: TS SDK uses this for integration tests atm 😔

I've added INN-2842 to move to the REST API, though the tests still need detailed state by fetching the timeline 🤔

IGassmann (approved):
In the future, I think it would be helpful to reference event IDs in the code as either "Internal Event ID" or "External Event ID" to remove the ambiguity.

PR #1210, opened 2024-03-04
Revert default run status change

goodoldneon:
## Description
Revert the change to the default run status when creating a run in the state store.

This change is no longer needed since the UI doesn't use the state store to determine run status anymore. It was a flawed change anyway since the status isn't updated to `Running` after the run starts 😄.

## Testing
Verified that runs show as scheduled and then running in the UI

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/lua/cancel.lua
+++ b/pkg/execution/state/redis_state/lua/cancel.lua
@@ -11,9 +11,7 @@ Output:
 local metadataKey = KEYS[1]
 
 local value = tonumber(redis.call("HGET", metadataKey, "status"))
-
--- If run has ended (completed, failed, etc.)
-if value ~= 0 and value ~= 5 then
+if value ~= 0 then
 	-- Return the function status as an error
 	return value;
 end
--- a/pkg/execution/state/redis_state/redis_state.go
+++ b/pkg/execution/state/redis_state/redis_state.go
@@ -264,7 +264,6 @@ func (m mgr) New(ctx context.Context, input state.Input) (state.State, error) {
 		Version:        currentVersion,
 		RequestVersion: consts.RequestVersionUnknown, // Always use -1 to indicate unset hash version until first request.
 		Context:        input.Context,
-		Status:         enums.RunStatusScheduled,
 	}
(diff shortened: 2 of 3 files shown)


tonyhb (changes requested):
We can't revert this as the dev server uses the state store for UI.

tonyhb (approved):

pkg/execution/state/redis_state/lua/cancel.lua
```diff
@@ -11,9 +11,7 @@ Output:
 local metadataKey = KEYS[1]
 
 local value = tonumber(redis.call("HGET", metadataKey, "status"))
-
--- If run has ended (completed, failed, etc.)
-if value ~= 0 and value ~= 5 then
+if value ~= 0 then
```
tonyhb, line 14: If we're adding queued and running, should we check for both statuses here?

tonyhb: Update:  it looks like this won't work for scheduled statuses?

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -264,7 +264,6 @@ func (m mgr) New(ctx context.Context, input state.Input) (state.State, error) {
 		Version:        currentVersion,
 		RequestVersion: consts.RequestVersionUnknown, // Always use -1 to indicate unset hash version until first request.
 		Context:        input.Context,
-		Status:         enums.RunStatusScheduled,
```
tonyhb, line 267: Why have we removed this?

tonyhb (commented):
Some questions about cancel.lua and redis_state.go changes to be honest 🤔

pkg/execution/state/redis_state/lua/cancel.lua
goodoldneon: The state store doesn't differentiate between scheduled and running statuses. Right now we have to either:
- Have a scheduled status but never transition to running (the behavior in the `main` branch right now)
- Have a running status that's never preceded by the scheduled status (the behavior before the PR I'm reverting)

PR #1217, opened 2024-03-06
Redirect to stream on event/run fetch error

goodoldneon:
## Description
If there's an error fetching the event or run in the stream details, then redirect back to the stream (i.e. close the slide out).

Before this PR, the UI would completely crash

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dev-server-ui/src/app/(dashboard)/stream/StreamDetails.tsx
+++ b/ui/apps/dev-server-ui/src/app/(dashboard)/stream/StreamDetails.tsx
@@ -22,12 +22,12 @@ export default function StreamDetails() {
 
   const eventResult = useEvent(eventID);
   if (eventResult.error) {
-    throw eventResult.error;
+    console.error(eventResult.error);
   }
 
   const runResult = useRun(runID);
   if (runResult.error) {
-    throw runResult.error;
+    console.error(runResult.error);
   }
 
   const getHistoryItemOutput = useGetHistoryItemOutput(runID);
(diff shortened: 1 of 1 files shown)


ui/apps/dev-server-ui/src/app/(dashboard)/stream/StreamDetails.tsx
```diff
@@ -53,6 +53,14 @@ export default function StreamDetails() {
...
   }, [eventResult.data?.payload]);
 
+  if (eventResult.error || runResult.error) {
+    // If there's an error fetching the event or run, we should redirect to the
+    // stream. This happens a lot, since restarting the Dev Server will clear
+    // all data
+    router.push('/stream');
+    return null;
+  }
+
```
djfarrelly, line 63: If there is a real error, this ends up hiding it from the user right? I think this could be a good hotfix, but for long term, how could we implement this:

* An event is missing, redirect back to stream, w/ a toast that says "Event not found"
* An event has some other type of error response (5xx?), show the error in the event panel
* A run is missing, but the event is valid, show the error in the run panel
* A run fails to load, show the error in the run panel
* If both fail to load, redirect to stream and show a toast that says "Failed to load run"

I think it's more ideal DX to have more granular feedback for the user as some of these combinations are valid and could happen if there is some sort of failure in the backend. For now, I think this might be OK if we add a toast before `router.push()` that has copy like: 

```js
toast.message('Failed to load run data', {
  description: 'Please check the browser console for more info',
})
```

djfarrelly (approved):
Functionally, this looks good. I think it's lacking UX wise, so I left suggested improvements, but given this is a bug fix I think it's acceptable to ship this if we can add a toast message so the user at least sees something if they truly encounter this issue, otherwise it isn't obvious if there is a valid error.

PR #1224, opened 2024-03-08
Hide replay buttons for internal events

jpwilliams:
## Description

When viewing an internal event run such as `inngest/function.invoked`, hide the replay buttons, as they will not work.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/EventDetails/EventDetails.tsx
+++ b/ui/packages/components/src/EventDetails/EventDetails.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { Badge } from '@inngest/components/Badge/Badge';
 import { Button } from '@inngest/components/Button';
 import { CodeBlock, type CodeBlockAction } from '@inngest/components/CodeBlock';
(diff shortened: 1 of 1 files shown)


goodoldneon:
Ugh, I forgot how gnarly `EventDetails` is now that it considers batches. We should eventually make a batch a first-class GraphQL field instead of relying on a combination of `batchID` and `events`

ui/packages/components/src/EventDetails/EventDetails.tsx
```diff
@@ -70,6 +71,10 @@ export function EventDetails({
     batch = events;
   }
 
+  const isInternalEvent = useMemo(() => {
+    return Boolean(singleEvent?.name?.startsWith('inngest/'));
+  }, [singleEvent]);
+
```
goodoldneon, line 77: Nit: `useMemo` is overkill here

```suggestion
  const isInternalEvent = Boolean(singleEvent?.name?.startsWith('inngest/'));
```

goodoldneon (approved):

PR #1235, opened 2024-03-14
Add run cancel button

goodoldneon:
> :warning: Depends on #1233

## Description
Add a "Cancel" button to the run details in both Cloud and the Dev Server. This button opens an alert dialog that asks the user to confirm before cancellation.

Other changes:
- Change `AlertDialog` to always show No/Yes buttons. Previously, "No" was "Cancel" and "Yes" was configurable. We can't use "Cancel" anymore because that's very confusing when the action is cancellation
- Add a loading prop to `AlertDialog`. This allows us to show a spinner and disable buttons while the action (e.g. cancellation) is occurring

## Testing

https://github.com/inngest/inngest/assets/20100586/2cf18a78-7647-4c45-9bb9-c4e4d4fe1053

https://github.com/inngest/inngest/assets/20100586/4e22f5dd-6694-4be1-8e94-c408b65caee7

- [ ] Docs

## Checklist


ui/packages/components/src/CancelRunButton/CancelRunModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <AlertModal
+      className="w-1/3"
+      isLoading={isLoading}
+      isOpen={isOpen}
+      onClose={onClose}
+      onSubmit={handleSubmit}
+      title="Are you sure you want to cancel this function run?"
+    >
+      <p className="pt-4 dark:text-white">
+        The function run will end early and its status will be "cancelled". This action cannot be
```
anafilipadealmeida, line 43: Not for now, but this makes it very evident that we should address the UK vs US en issue with cancelled. It just looks weird

anafilipadealmeida (approved):
Left a comment but approving this already. A small note: I think it would still be useful to display a success toast when the run is canceled, as it's a pattern across the apps and users might not realise that the status changed

PR #1238, opened 2024-03-18
INN-2879 Fix Dev Server ingestion returning external IDs

jpwilliams:
## Description

The return value of the ingestion endpoint on the Dev Server contains an `ids` property, which is an array of IDs for events created as a result of the request.

Following #1213, which correctly fixes UIs not using internal IDs, if an `id` is specified in an event, it will be the same ID returned from the endpoint. This is incorrect, as we should always return the internal IDs from ingestion.

| ID given | Before PR | After PR |
| -------- | --------- | -------- |
| None | Internal | Internal |
| Populated string | Custom | Internal |

This PR moves the creation of the internal ID to the point of ingestion, allowing the endpoint to correctly return the internal ID after it's been pushed to pubsub. This also matches Cloud.

- [ ] Docs

## Checklist

## Related

- INN-2879
- #1213
- Unblocks inngest/inngest-js#518

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"net/http"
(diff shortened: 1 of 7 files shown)


djfarrelly (approved):
This is great - I was working on a demo project this morning and I needed this exact fix!!

PR #1244, opened 2024-03-21
Hotfix: Remove inaccurate error return

djfarrelly:
## Description

This no-data code path was hit incorrectly and would produce false errors when data was loading. I debugged this a bit and didn't get to the bottom of this path, but I believe it's due to the asynchronous request within the hook and some sort of in-between state where the request should be sent, but the return path with !data was hit first preventing further execution.

## Motivation
Issues in prod when a user navigates from another page to the Vercel integration management page, but did not happen when landing directly on the URL.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/utils/useRestAPIRequest.ts
+++ b/ui/apps/dashboard/src/utils/useRestAPIRequest.ts
@@ -3,6 +3,7 @@
 import { useEffect, useState } from 'react';
 import { useAuth } from '@clerk/nextjs';
 import {
+  baseFetchSkipped,
   baseFetchSucceeded,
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/utils/useRestAPIRequest.ts
```diff
@@ -63,11 +66,9 @@ export function useRestAPIRequest<T>({
     };
   }
 
-  if (!data) {
-    // Should be unreachable.
```
djfarrelly, line 67: This indeed was reachable. It was producing false errors.

PR #1248, opened 2024-03-23
Create our own data structure for otel spans to workaround otel spec limitations

darwin67:
## Description

The otel library is built on an assumption that span instrumentation API should be within a piece of local resource. In other words, you should not be able to reconstruct a remote span and modify it on another resource over the wire.

While that's a very reasonable and solid argument in most cases, abiding by that usage will make our trace representation incomprehensible for our users.
The goal of Inngest is to abstract the complexity of the inner workings of an event-driven system and make it as close as possible to a "normal" function call on the stack.

In order to achieve that objective for tracing, we'll need to propagate spans across network boundaries in a way that from the data's perspective, as if there were none. Therefore, we need a data structure that can
1. the propagation of span across pub/sub, queue systems
2. override certain span parameters to make it look as it is still the same span
3. keep the inner workings as transparent as possible while allowing multiple submission of the same span

The necessity for `#2` is mainly as a function run progresses, we need to continuously update the function run span's end time / duration. And by allowing multiple submission of the same span with different end times, we can dedup on our end to make sure the last one is always the correct one to stay.

That way, as the function run progresses, the data will also reflect the run in a way that will simulate a stack function call.
This data will also be the source to power the new UI coming up to allow better visibility of what's happening with user's workload.

### Assumptions
Dedupping of similar spans is out of scope, and is the responsibility of either the storage engine (e.g. Clickhouse) or something that's below the application layer.

- [ ] Docs

## Checklist


pkg/api/api.go
```diff
@@ -129,8 +128,9 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 		Start(
 			ctx,
 			"event-ingestion",
+			trace.WithNewRoot(),
```
darwin67, line 131: we can just force a new root this way without needing to utilize a separate context.

pkg/execution/executor/executor.go
```diff
@@ -288,23 +288,11 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
...
-			attribute.Bool(consts.OtelUserTraceFilterKey, true),
-			attribute.String(consts.OtelSysAccountID, req.AccountID.String()),
-			attribute.String(consts.OtelSysWorkspaceID, req.WorkspaceID.String()),
-			attribute.String(consts.OtelSysAppID, req.AppID.String()),
-			attribute.String(consts.OtelSysFunctionID, req.Function.ID.String()),
-			attribute.String(consts.OtelSysFunctionSlug, req.Function.GetSlug()),
-			attribute.Int(consts.OtelSysFunctionVersion, req.Function.FunctionVersion),
-		))
-	defer span.End()
-
```
darwin67, line 303: these are moved to the `Execute` section to align with actual step execution.

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,90 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
 			attribute.String(consts.OtelSysAccountID, id.AccountID.String()),
 			attribute.String(consts.OtelSysWorkspaceID, id.WorkspaceID.String()),
 			attribute.String(consts.OtelSysAppID, id.AppID.String()),
 			attribute.String(consts.OtelSysFunctionID, id.WorkflowID.String()),
 			attribute.Int(consts.OtelSysFunctionVersion, id.WorkflowVersion),
 			attribute.String(consts.OtelAttrSDKRunID, id.RunID.String()),
-		))
-	defer span.End()
+			attribute.StringSlice(consts.OtelSysEventIDs, evtIDs),
+			attribute.String(consts.OtelSysIdempotencyKey, id.IdempotencyKey()),
```
darwin67, line 580: function span construction has been moved here to align with step execution.
https://github.com/inngest/inngest/pull/1248/files#diff-ccd0679a9f24c84fef4f5d4d5fe2c32651d23886d85af8448bb968921c7b8a95L291-L302

pkg/execution/executor/executor.go
```diff
@@ -700,30 +762,30 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+			fnstatus := attribute.String(consts.OtelSysFunctionStatus, enums.RunStatusCompleted.String())
+
 			if resp.StatusCode != 200 {
 				spanName = "function error"
+				fnstatus = attribute.String(consts.OtelSysFunctionStatus, enums.RunStatusFailed.String())
 				span.SetStatus(codes.Error, resp.Error())
 			}
 
-			span.SetName(spanName)
+			fnSpan.SetAttributes(fnstatus)
```
darwin67, line 773: annotating the function span with the status.

pkg/execution/executor/service.go
```diff
@@ -164,11 +164,13 @@ func (s *svc) getFinishHandler(ctx context.Context) (func(context.Context, state
 
 func (s *svc) Run(ctx context.Context) error {
 	logger.From(ctx).Info().Msg("subscribing to function queue")
-	return s.queue.Run(ctx, func(ctx context.Context, _ queue.RunInfo, item queue.Item) error {
+	return s.queue.Run(ctx, func(ctx context.Context, info queue.RunInfo, item queue.Item) error {
 		// Don't stop the service on errors.
 		s.wg.Add(1)
 		defer s.wg.Done()
 
+		item.RunInfo = &info
```
darwin67, line 172: embed queue info to pass on information like latency so we can annotate span latency

pkg/execution/executor/executor.go
```diff
@@ -652,6 +690,29 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 
 		// Only just starting:  run lifecycles on first attempt.
 		if item.Attempt == 0 {
+			// NOTE:
+			// annotate the step as the first step of the function run.
+			// this way the delay associated with this run is directly correlated to the delay of the
+			// function run itself.
+			span.SetAttributes(attribute.Bool(consts.OtelSysStepFirst, true))
```
darwin67, line 697: annotate the step as the first step so we can infer that the latency this step is the same as the function run latency.

pkg/execution/state/redis_state/lua/updateMetadata.lua
```diff
@@ -6,10 +6,39 @@ local ctx      = ARGV[1]
...
+
+-- only update the duration if it's the default value
+if is_field_empty("sysdelay", "-1") then
+  redis.call("HSET", keyMetadata, "sysdelay", sysdelay)
+end
+
+-- only update the duration if it's the default value
+if is_field_empty("sojdelay", "-1") then
+  redis.call("HSET", keyMetadata, "sojdelay", sojdelay)
+end
```
darwin67, line 42: update the lua script to set span related metadata for the function run.
these are only set when they're not using the default values, so update should only happen once, which is on the start of the function run.

pkg/telemetry/span.go
```diff
@@ -0,0 +1,439 @@
...
+func (s *Span) TracerProvider() trace.TracerProvider {
+	return UserTracer().Provider()
+}
+
+func newSpanIDGenerator() *spanIDGenerator {
+	var seed int64
+	_ = binary.Read(crand.Reader, binary.LittleEndian, &seed)
+
+	return &spanIDGenerator{
+		randSrc: rand.New(rand.NewSource(seed)),
```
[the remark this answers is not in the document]
darwin67: This is irrelevant. It follows the similar implementation as the official otel ID generator. I'll dismiss it when the PR is approved and ready to merge, otherwise it keeps popping up.

pkg/telemetry/span.go
```diff
@@ -0,0 +1,555 @@
...
+
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+//
+//	FOOTGUN ALERT
+//
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// Cancel will mark the span as cancelled and it will not be sent out when it ends.
+// This will reset the context so if there are spans that will be created after this,
+// it doesn't create a dangling pointer.
+func (s *Span) Cancel(ctx context.Context) context.Context {
```
darwin67, line 310: Need a better way to highlight this is a footgun that we need to be careful with.

pkg/api/api.go
darwin67: @jpwilliams are there cases we don't want it to be a root span?
if the source is from the event API, I think it's fine they're all root spans atm?

pkg/execution/executor/executor.go
jpwilliams: Does this mean we can only infer latency for the very first step?

I'm wondering how we would infer latency for parallel steps?

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,84 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
 
+	// Retrieve the start time of the function
+	fnStartTime := time.Now()
+	if md.GetStartTime() != nil {
+		fnStartTime = *md.GetStartTime()
+	}
+
 	// Store the metadata in context for future use and propagate trace
 	// context. This can be used to reduce reads in the future.
 	ctx = e.extractTraceCtx(WithContextMetadata(ctx, md), id, &item)
```
jpwilliams, line 538: We already inject and then extract the tracing context from either the run's metadata or the queue item here.

Why do we now also add a separate `SpanID`?

pkg/execution/executor/executor.go
```diff
@@ -652,6 +690,29 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+			// which is enforced on the Lua script.
+			update := state.MetadataUpdate{
+				Context:                   md.Context,
+				DisableImmediateExecution: md.DisableImmediateExecution,
+				SpanID:                    fnSpanID.String(),
+				StartedAt:                 &fnStartTime,
+			}
+			if item.RunInfo != nil {
+				// store in metadata as nano seconds for easier conversion later
+				update.SystemDelay = item.RunInfo.Latency.Nanoseconds()
```
jpwilliams, line 710: Is system delay what we're inferring anyway? I'm wondering if we need to include it here.

pkg/execution/executor/executor.go
```diff
@@ -688,8 +749,9 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 			spanName := op.UserDefinedName()
 			span.SetName(spanName)
 
+			fnSpan.SetAttributes(attribute.String(consts.OtelSysFunctionStatus, enums.RunStatusRunning.String()))
```
jpwilliams, line 752: Nice!

jpwilliams (commented):
Traces are looking mega duplicated now. I guess the duplicated run spans are expected, but also every step is being duplicated.

Suspiciously, the end time of completely different steps are consistently exactly the same and I believe we're now showing discovery executions (seeing "running") and duplicate function ends.

image

How much of this is expected if we don't have the right engine consuming the traces?

pkg/execution/executor/executor.go
jpwilliams: We need to reconstruct to make it a duplicate. 👍

pkg/execution/executor/executor.go
jpwilliams: System/sojourn delay can help us with parallel step latency.

pkg/execution/executor/executor.go
jpwilliams: We're not inferring this.

pkg/execution/executor/executor.go
darwin67: ah I didn't add it in this repo but it's in the cloud. let me see if I can add it here as well.
basically it's the `trigger` span we talked about earlier.

darwin67:
> Traces are looking mega duplicated now. I guess the duplicated run spans are expected, but also every step is being duplicated.
> 
> Suspiciously, the end time of completely different steps are consistently exactly the same and I believe we're now showing discovery executions (seeing "running") and duplicate function ends.
> 
> image
> 
> How much of this is expected if we don't have the right engine consuming the traces?

This is all expected right now. although everything shouldn't be ending around the same time if it's not parallelism, so I can take a quick look at that.

the reliance on the backend to dedup is out of scope for the time being, but the idea is we should still be expecting there's something downstream that will be handling the deduplication so the instrumentation itself don't need to be bothered with those details.

CH is the option right now that's the easiest to leverage, but we can definitely build some kind of adapter for deduplication in the future.
e.g. CH -> just passthrough, Some other database -> dedup prior to/after storing

though that'll raise the question of, "can we still show in progress execution?".
it's a little hard to say for sure, but considering there are no good alternatives without doing a lot of the dedup work ourselves, I'd say let's stick to the current backend assumption, and we'll figure things out when it arises.

pkg/api/api.go
```diff
@@ -180,10 +163,22 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
...
 			}
 
+			ts := time.Now()
 			if evt.Timestamp == 0 {
-				evt.Timestamp = time.Now().UnixMilli()
+				evt.Timestamp = ts.UnixMilli()
 			}
 
+			ctx, span := telemetry.UserTracer().Provider().
+				Tracer(consts.OtelScopeEvent).
```
darwin67, line 172: moved the event span down here as discussed

pkg/execution/runner/runner.go
```diff
@@ -496,10 +514,28 @@ func (s *svc) functions(ctx context.Context, tracked event.TrackedEvent) error {
 					}
 				}
 
+				cancelSpan := false
+				_, span := telemetry.NewSpan(ctx,
+					telemetry.WithScope(consts.OtelScopeTrigger),
+					telemetry.WithName(consts.OtelSpanTrigger),
```
darwin67, line 520: added trigger span here as mentioned.

pkg/consts/otel.go
```diff
@@ -18,32 +22,39 @@ const (
...
-	OtelScopeStep           = "step.function.app.env.inngest"
-	OtelScopeExecution      = "execution.step.function.app.env.inngest"
+	OtelScopeEvent     = "event.inngest"
+	OtelScopeTrigger   = "trigger.inngest"
+	OtelScopeCron      = "cron.inngest"
+	OtelScopeEnv       = "env.inngest"
+	OtelScopeApp       = "app.env.inngest"
+	OtelScopeFunction  = "function.app.env.inngest"
+	OtelScopeStep      = "step.function.app.env.inngest"
+	OtelScopeExecution = "execution.function.app.env.inngest"
```
darwin67, line 57: remove the `step` from here since execution doesn't always mean it's a step

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,84 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+			attribute.Int(consts.OtelSysFunctionVersion, id.WorkflowVersion),
+			attribute.String(consts.OtelAttrSDKRunID, id.RunID.String()),
+			attribute.Int64(consts.OtelSysDelaySystem, sysDelay.Milliseconds()),
+			attribute.Int64(consts.OtelSysDelaySojourn, sojDelay.Milliseconds()),
+		),
+	)
+	defer func() {
+		fnSpan.End()
+		span.End()
+	}()
```
jpwilliams, line 605: Bit of a nit, but would be great if this also set the span's statuses etc if there's an error.

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,84 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+			attribute.String(consts.OtelSysIdempotencyKey, id.IdempotencyKey()),
+		),
+	)
+	if len(id.EventIDs) > 1 {
+		fnSpan.SetAttributes(attribute.String(consts.OtelSysBatchID, id.BatchID.String()))
+	}
+
+	ctx, span := telemetry.NewSpan(ctx,
+		telemetry.WithScope(consts.OtelScopeExecution),
+		telemetry.WithName("execute"),
```
jpwilliams, line 589: For steps that report (first parallel step and some following parallel steps), this span will linger. It sounded like the plan was to remove them entirely?

We likely need to call `span.Cancel()` further down the line for this, then?

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,84 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+			attribute.Int64(consts.OtelSysDelaySystem, sysDelay.Milliseconds()),
+			attribute.Int64(consts.OtelSysDelaySojourn, sojDelay.Milliseconds()),
+		),
+	)
+	defer func() {
+		fnSpan.End()
+		span.End()
+	}()
+	// send early here to help show the span has started and is in-progress
+	span.Send()
```
jpwilliams, line 607: We send the span early here, but then may later attempt to use `span.Cancel()` to filter out discovery requests (or anything else we have deemed not suitable to track).

This means we get dangling spans (or strange intermediary states that never resolve), like lots of `execute` spans in this case. Parallel examples seem to make a lot of these.

jpwilliams (commented):
A note from testing, trying to look at the odd span extension that happens sometimes.

Also, I believe any queued step now has no related span until they are executed, meaning the UI has no way to know if we have received and queued steps until they start executing.

If a user queues 100 steps in parallel and doesn't see them even recognized on the UI, that will definitely be confusing.

pkg/execution/executor/executor.go
darwin67: The status should be set where the error occurs.

pkg/execution/executor/executor.go
darwin67: Is there a way to know if it's a discovery or not early on?

darwin67:
> A couple of notes from testing, trying to look at the odd span extension that happens sometimes.
> 
> Also, I believe any queued step now has no related span until they are executed, meaning the UI has no way to know if we have received and queued steps until they start executing.
> 
> If a user queues 100 steps in parallel and doesn't see them even recognized on the UI, that will definitely be confusing.

@jpwilliams We should be able to infer this information from the response of the step.
e.g. status code == 206 || status code == 500, set queue time for next one.

or we can annotate the step with the queue time of the next step before it ends.

pkg/execution/state/state.go
```diff
@@ -163,13 +164,52 @@ type Metadata struct {
...
+	// SpanID is the spanID used for this function run.
+	SpanID string `json:"sid"`
+	// QueuedAt shows the time the function was queued
+	QueuedAt int64 `json:"queuedAt"`
+	// StartedAt shows the time the function started execution
+	StartedAt int64 `json:"startedAt,omitempty"`
+
+	// The delay duration for the function to start caused by the system
+	// nano second precision
+	SystemDelay int64 `json:"sysdelay"`
```
tonyhb, line 177: Why are these stored in a state's metadata?  Shouldn't this be per-step and per-queue item?

darwin67: This was needed prior because I was trying to annotate the function's delay before it starts.
It has changed to annotating the first step as the start of the function, so it's no longer necessary. I'll delete it.

pkg/execution/executor/executor.go
darwin67: Changing the approach slightly to annotate the span as something to be deleted downstream, instead of returning early.

[0b7d57a](https://github.com/inngest/inngest/pull/1248/commits/0b7d57a4b25065cd33e890a8b492c203d5006c03)

Ref: https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree#is_deleted

pkg/telemetry/span.go
```diff
@@ -0,0 +1,556 @@
...
+	return s.kind
+}
+
+func (s *Span) StartTime() time.Time {
+	return s.start
+}
+
+func (s *Span) EndTime() time.Time {
+	return time.Now()
+}
```
jpwilliams, line 343: Can't find any references to this function in this PR, so I'm guessing this is called internally by the exporter and is part of our matching the otel interface.

If so, is this correct? I'm wondering if `time.Now()` will mean that the end time of the span is set to the moment it is exported rather than when we ended it with `.End()` or `.Send()`. For batched spans especially, this could cause some weird timings.

pkg/execution/executor/executor.go
```diff
@@ -551,22 +527,77 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+		),
+	)
+	if len(id.EventIDs) > 1 {
+		fnSpan.SetAttributes(attribute.String(consts.OtelSysBatchID, id.BatchID.String()))
+	}
+
+	ctx, span := telemetry.NewSpan(ctx,
+		telemetry.WithScope(consts.OtelScopeExecution),
+		telemetry.WithName("execute"),
+		telemetry.WithSpanAttributes(
```
jpwilliams, line 579: We may need some more information here in order to group this execution with related ones in the UI. Without something to group with, we don't know which step this in-progress execution belongs to.
- If we know the step name (parallelism), we can use that
- If we don't know the step name (sequential), we need another identifier

pkg/api/api.go
jpwilliams: With this now the root of the initial trace:

- Should we add a span event to this span that is the event's payload?

- Does invoke need one of these spans above the run?
  
  Currently it's Run -> Invoke Step -> Run
  Should it be Run -> Invoke Step -> Event -> Run?

jpwilliams (commented):
A few more notes from the changes and something I noticed when testing again.

pkg/execution/executor/executor.go
```diff
@@ -1820,10 +1828,14 @@ func (e *executor) handleStepError(ctx context.Context, gen state.GeneratorOpcod
 		MaxAttempts: item.MaxAttempts,
 		Payload:     queue.PayloadEdge{Edge: nextEdge},
 	}
-	err = e.queue.Enqueue(ctx, nextItem, time.Now())
+	err = e.queue.Enqueue(ctx, nextItem, now)
 	if err == redis_state.ErrQueueItemExists {
 		return nil
 	}
+	span.SetAttributes(
+		attribute.Bool(consts.OtelSysStepRetry, true),
```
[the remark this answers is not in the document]
darwin67: Will move this to `Execute` on span creation and use the attempts instead.

pkg/execution/executor/executor.go
darwin67: @jpwilliams let's check with Tony and Aaron to see if `GroupID` can be reused.

pkg/api/api.go
darwin67: > Should we add a span event to this span that is the event's payload?
yes, it's added on the cloud side. but can be added for consistency.

> Does invoke need one of these spans above the run?
Keep the event span, but embed the event data into function run span as discussed.

darwin67: Change storing events into `Schedule`.

pkg/telemetry/span.go
darwin67: addressed at [49ec2da](https://github.com/inngest/inngest/pull/1248/commits/49ec2daf7873519944b10cdeb50d5e1c7ffe5340)

pkg/execution/executor/executor.go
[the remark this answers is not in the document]
darwin67: done at [f9b4b59](https://github.com/inngest/inngest/pull/1248/commits/f9b4b59e10a3f5f7084204e877a046d17867d5b3)

pkg/execution/executor/executor.go
darwin67: Done at [fd606cf](https://github.com/inngest/inngest/pull/1248/commits/fd606cf2565cd0c85ea89ab17dbac5aebbebaac7)

pkg/api/api.go
darwin67: Changing to store the Event's reference instead in separate discussion.

darwin67: Done at [ba2cdc7](https://github.com/inngest/inngest/pull/1248/commits/ba2cdc7a0c688addd79461cd2f9ad2e80d533862)

pkg/execution/executor/executor.go
```diff
@@ -659,47 +747,34 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
-
-	if err == nil && item.Kind == queue.KindStart {
-		md := s.Metadata()
-		// This is the first successful step, so mark the fn as started.
-		_ = e.sm.UpdateMetadata(ctx, s.RunID(), state.MetadataUpdate{
-			Debugger:                  md.Debugger,
-			Context:                   md.Context,
-			DisableImmediateExecution: md.DisableImmediateExecution,
-			RequestVersion:            md.RequestVersion,
-			StartedAt:                 start,
```
darwin67, line 699: move this metadata update here instead since we need it to have a consistent start time.
https://github.com/inngest/inngest/pull/1248/files#diff-ccd0679a9f24c84fef4f5d4d5fe2c32651d23886d85af8448bb968921c7b8a95R687-R698

jpwilliams (approved):

PR #1249, opened 2024-03-25
INN-2888 INN-2876 INN-2887 Trace fan-out, invoke, propagation

jpwilliams:
## Description

Adds propagation for fan-out and invoke and intentionally includes no tracking of any step concepts; only executions.

image

- [ ] Docs

## Checklist

## Related

- INN-2888
- INN-2876
- INN-2887

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -18,7 +18,11 @@ import (
 	"github.com/inngest/inngest/pkg/eventstream"
 	"github.com/inngest/inngest/pkg/headers"
 	"github.com/inngest/inngest/pkg/publicerr"
+	"github.com/inngest/inngest/pkg/telemetry"
 	"github.com/rs/zerolog"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -112,8 +116,29 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Create a new channel which receives a stream of events from the incoming HTTP request
 	ctx, cancel := context.WithCancel(ctx)
+
+	// Create a new trace that may have a link to a previous one
+	fanoutCtx := context.Background()
+	fanoutCtx = telemetry.UserTracer().Propagator().Extract(fanoutCtx, propagation.HeaderCarrier(r.Header))
+
+	ctx, span := telemetry.
+		UserTracer().
+		Provider().
+		Tracer(consts.OtelScopeEventIngestion).
+		Start(
+			ctx,
+			"event-ingestion",
+			trace.WithLinks(
+				trace.LinkFromContext(fanoutCtx),
+			),
+			trace.WithAttributes(
+				attribute.Bool(consts.OtelUserTraceFilterKey, true),
+			),
+		)
+	defer span.End()
+
+	// Create a new channel which receives a stream of events from the incoming HTTP request
 	stream := make(chan eventstream.StreamItem)
 	eg, ctx := errgroup.WithContext(ctx)
 	eg.Go(func() error {
--- a/pkg/consts/otel.go
+++ b/pkg/consts/otel.go
@@ -28,6 +28,9 @@ const (
 	OtelSysStepOutput          = "sys.step.output"
 	OtelSysStepOutputSizeBytes = "sys.step.output.size.bytes"
 
+	// flow attributes
+	OtelSysIgnored = "sys.ignored"
+
 	// SDK attributes
 	OtelAttrSDKServiceName = "sdk.app.id"
 	OtelAttrSDKRunID       = "sdk.run.id"
--- a/pkg/consts/otel.go
+++ b/pkg/consts/otel.go
@@ -40,6 +40,7 @@ const (
 	OtelScopeApp            = "app.env.inngest"
 	OtelScopeFunction       = "function.app.env.inngest"
 	OtelScopeStep           = "step.function.app.env.inngest"
+	OtelScopeExecution      = "execution.step.function.app.env.inngest"
 
 	// otel collector filter keys
 	OtelUserTraceFilterKey = "inngest.user"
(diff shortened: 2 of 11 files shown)


pkg/telemetry/tracing.go
```diff
@@ -220,3 +231,40 @@ func newTextMapPropagator() propagation.TextMapPropagator {
...
+// processor if they have the "ignored" attribute set to true.
+//
+// This can be useful for if a span's creation was required for propagation
+// purposes, but turned out not to be used, e.g. if idempotency filtered out a
+// queued item.
+type IgnoredSpanProcessor struct {
+	next trace.SpanProcessor
+}
+
+func NewIgnoredSpanProcessor(next trace.SpanProcessor) *IgnoredSpanProcessor {
```
darwin67, line 245: smart.

would there be a case where the ignored span will have child spans, resulting in disruption of the connections between spans because the one in the middle was taken out?

darwin67 (changes requested):
Overall look solid with some change requests.

pkg/telemetry/tracing.go
```diff
@@ -220,3 +231,40 @@ func newTextMapPropagator() propagation.TextMapPropagator {
...
 	)
 }
+
+// IgnoredSpanProcessor processes spans and does not send them to the next
+// processor if they have the "ignored" attribute set to true.
+//
+// This can be useful for if a span's creation was required for propagation
+// purposes, but turned out not to be used, e.g. if idempotency filtered out a
+// queued item.
+type IgnoredSpanProcessor struct {
```
darwin67, line 241: I might want to build on top of this for span exporting, so would be nice to name it more generic.
Or we allow a list of span processors and just iterate through them.

darwin67: following your pattern here: https://github.com/inngest/inngest/pull/1248/files#diff-64030f762860cd37b119fd2737b96794faa0c08b5493a0223a1d740a1c01ada2R13-R45

jpwilliams: Ah, I created a `newSpanProcessor` function so that we can have one spot to build a stack of span processors common for each provider.

https://github.com/inngest/inngest/pull/1249/files/81f5620c59106b14d96d55384b14649d7b09f53d#diff-7eafc162c0f1d5cc4d869885af2f577b3bc7ea123a559f134e6102fda1cd366fR91-R95

Right now it's just:
- `IgnoredSpanProcessor`
- `BatchSpanProcessor`

But we can just add anything we need there.

pkg/consts/otel.go
```diff
@@ -40,6 +43,7 @@ const (
 	OtelScopeApp            = "app.env.inngest"
 	OtelScopeFunction       = "function.app.env.inngest"
 	OtelScopeStep           = "step.function.app.env.inngest"
+	OtelScopeExecution      = "execution.step.function.app.env.inngest"
```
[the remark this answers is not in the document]
jpwilliams: There's still a distinction between a span that represents a step (e.g. a `step.sleep()`) and a span that represents an execution (e.g. a `step.run()` attempt).

I'd like to keep this distinction present so it'd be easier to change in the future if we rearrange anything.

pkg/execution/executor/executor.go
```diff
@@ -546,14 +546,22 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
 	ctx = WithContextMetadata(ctx, md)
 
 	// Propagate trace context
-	if md.Context != nil {
+	if trace, ok := item.Metadata[consts.OtelPropagationKey]; ok {
+		carrier := telemetry.NewTraceCarrier()
+		if err := carrier.Unmarshal(trace); err == nil {
+			ctx = telemetry.UserTracer().Propagator().Extract(ctx, propagation.MapCarrier(carrier.Context))
+		}
+	} else if md.Context != nil {
```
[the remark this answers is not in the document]
jpwilliams: Yeah, absolutely. Lots of this was dumping the logic in the right spots before cleaning up. Cleaned.

pkg/telemetry/tracing.go
jpwilliams: Absolutely. The ignored spans atm are always unpropagated spans, so won't have children, but this is brittle, and if a span loses its parent that it'll just use the root.

Right now this is used to remove executions from the trace in order to explicitly not show `StepPlanned` spans as we discussed.

pkg/api/api.go
```diff
@@ -112,8 +116,29 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
...
+	// Create a new trace that may have a link to a previous one
+	fanoutCtx := context.Background()
+	fanoutCtx = telemetry.UserTracer().Propagator().Extract(fanoutCtx, propagation.HeaderCarrier(r.Header))
+
+	ctx, span := telemetry.
+		UserTracer().
+		Provider().
+		Tracer(consts.OtelScopeEventIngestion).
+		Start(
+			ctx,
```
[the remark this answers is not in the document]
jpwilliams: No - the idea here is that the event ingestion creates a fresh trace (no parent spans), so makes sure that it uses to parents here.

We then use `fanoutCtx` to add a link to the execution that caused the fanout just below:

https://github.com/inngest/inngest/pull/1249/files/81f5620c59106b14d96d55384b14649d7b09f53d#diff-67a6e36014be865053c43d38b2091bd2d3f17a91b117d97c0f5244cd7d71127dR132-R134

darwin67 (approved):
Thanks for clarifying. Added some additional nice to have notes, but otherwise looks good.

pkg/execution/state/driver_response.go
```diff
@@ -431,6 +431,34 @@ func (r *DriverResponse) HistoryVisibleStep() *GeneratorOpcode {
...
+// TraceVisibleStepExecution returns a single generator op if this response
+// should be visible in a trace, otherwise nil. If returning nil, the response
+// may still be considered to be a function response, in which case it likely
+// also needs to be tracked.
+func (r *DriverResponse) TraceVisibleStepExecution() *GeneratorOpcode {
+	if r.Generator == nil {
+		return nil
+	}
+
+	if len(r.Generator) != 1 {
```
darwin67, line 443: probably could just combine this with the one on top and make it an `||`

jpwilliams: Added some comments instead, as the checks are nice to have separate.

pkg/telemetry/tracing.go
jpwilliams: Discussed this in relation to #1248.

PR #1262, opened 2024-04-08
Update SDK spec for signing key rotation

goodoldneon:
## Description
Update the SDK spec to support signing key rotation. The key changes are:
- When something with the signing key fails, then retry with the fallback signing key if it exists
- The introspection endpoint needs a schema since we'll rely on it for checking that signing key env vars are properly setup

- [x] Docs

## Checklist

### DIFF
--- a/docs/SDK_SPEC.md
+++ b/docs/SDK_SPEC.md
@@ -188,6 +188,8 @@ An SDK MUST support these environment variables and use them to fill the appropr
 Used to specify an Event Key to be used when sending Events to an Inngest Server. Always recommend that a Developer specifies an Event Key as an environment variable to avoid checking secrets into source control.
 - `INNGEST_SIGNING_KEY`
 Used to specify a Signing Key to be used when contacting the Inngest API and verifying the integrity of requests from an Inngest Server. Always recommend that a Developer specifies a Signing Key as an environment variable to avoid checking secrets into source control.
+- `INNGEST_SIGNING_KEY_FALLBACK`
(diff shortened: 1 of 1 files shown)


tonyhb (approved):

docs/SDK_SPEC.md
```diff
@@ -898,9 +900,31 @@ The execution of a Function, the returned data, and the tooling available to a D
 
 ## 4.5. Introspection Requests
 
-An SDK MAY offer the ability to respond to an Introspection Request, a `GET` request intended to be used as a health check, ensuring that an HTTP request can be correctly handled by an SDK.
+An SDK MUST respond to an Introspection Request, a `GET` request intended to be used as a health check, ensuring that an HTTP request can be correctly handled by an SDK.
```
jpwilliams, line 903: Noting for posterity that Go needs this adding.

docs/SDK_SPEC.md
```diff
@@ -898,9 +900,31 @@ The execution of a Function, the returned data, and the tooling available to a D
...
+If the request signature validation fails, then the SDK MUST NOT return any sensitive data. The response schema MUST be:
+
+```ts
+{
+	function_count: number
+	has_event_key: boolean
+	has_signing_key: boolean
+  mode: "cloud" | "dev"
+}
+```
```
jpwilliams, line 914: Can we rephrase this to say that it must contain at least these fields? For example, as you know in TS it's useful to also see if the mode is explicit or was inferred.

Actually, perhaps we namespace any custom keys to make sure an SDK doesn't take a name we want to later use generically?
```suggestion
If the request signature validation fails, then the SDK MUST NOT return any sensitive data. The response schema MUST be:

```ts
{
  function_count: number;
  has_event_key: boolean;
  has_signing_key: boolean;
  mode: "cloud" | "dev";
  metadata?: Record<string, any>;
}
```
```

PR #1263, opened 2024-04-08
Add GCRA to leasing queue items directly

tonyhb:
This commit adds GCRA to queue items.  It's possible to specify GCRA/throttling parameters directly when enqueueing an item.  If a queue item contains GCRA paremters, the queue will apply GCRA when attempting to lease the queue item.

This ensures that the queue atomically, but selectively, applies throttling as it attempts to claim items to be processed.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/queue/item.go
+++ b/pkg/execution/queue/item.go
@@ -75,6 +75,20 @@ type Item struct {
 	// QueueName allows control over the queue name.  If not provided, this falls
 	// back to the queue mapping defined on the queue or the workflow ID of the fn.
 	QueueName *string `json:"qn,omitempty"`
+
+	Throttle *Throttle `json:"throttle,omitempty"`
+}
+
+type Throttle struct {
+	// Key is the unique throttling key that's used to group queue items when
+	// processing rate limiting/throttling.
+	Key string `json:"k"`
+	// Limit is the actual rate limit
+	Limit int `json:"l"`
+	// Burst is the busrsable capacity of the rate limit
+	Burst int `json:"b"`
+	// Period is the rate limit period, in seconds
+	Period int `json:"p"`
 }
 
 // GetPriorityFactor returns the priority factor for the queue item.  This fudges the job item's
--- a/pkg/execution/state/redis_state/lua/includes/gcra.lua
+++ b/pkg/execution/state/redis_state/lua/includes/gcra.lua
@@ -1,0 +1,36 @@
+-- performs gcra rate limiting for a given key.
+-- 
+-- Returns true on success, false if the key has been rate limited.
+local function gcra(key, now_ms, period_ms, limit, burst)
+	-- Calculate the basic variables for GCRA.
+
+	local cost = 1                            -- everything counts as a single rqeuest
+	local emission  = period_ms / limit       -- how frequently we can admit new requests
+	local increment = emission * cost         -- this request's time delta
+	local variance  = period_ms * (burst + 1) -- variance takes into account bursts
+
+	-- fetch the theoretical arrival time for equally spaced requests
+	-- at exactly the rate limit
+	local tat = redis.call("GET", key)
+	if not tat then
+		tat = now_ms
+	else
+		tat = tonumber(tat)
+	end
+
+	local new_tat = math.max(tat, now_ms) + increment -- add the request's cost to the theoretical arrival time.
+	local allow_at_ms = new_tat - variance         -- handle bursts.
+
+	local diff_ms = now_ms - allow_at_ms
+
+	if diff_ms < 0 then
+		-- return {false, tat, new_tat, allow_at_ms, diff_ms}
+		return false
+	end
+
+	local expiry = math.ceil(now_ms + period_ms / 1000)
+	redis.call("SETEX", key, new_tat, expiry)
+
+	-- return {true, tat, new_tat, allow_at_ms, diff_ms}
+	return true
+end
--- a/pkg/execution/state/redis_state/lua/includes/gcra.lua
+++ b/pkg/execution/state/redis_state/lua/includes/gcra.lua
@@ -3,11 +3,10 @@
 -- Returns true on success, false if the key has been rate limited.
 local function gcra(key, now_ms, period_ms, limit, burst)
 	-- Calculate the basic variables for GCRA.
-
 	local cost = 1                            -- everything counts as a single rqeuest
 	local emission  = period_ms / limit       -- how frequently we can admit new requests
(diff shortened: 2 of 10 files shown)


pkg/execution/state/redis_state/lua/queue/partitionPeek.lua
```diff
@@ -7,10 +7,22 @@ Peek returns partition items from the queue in order via their index.
...
-local items = redis.call("ZRANGE", partitionIndex, "-inf", peekUntil, "BYSCORE", "LIMIT", 0, limit)
+local peekUntil    = math.ceil(peekUntilMS / 1000)
+
+local count = redis.call("ZCOUNT", partitionIndex, "-inf", peekUntil)
+local offset = 0
+
+if count > limit and sequential == 0 then
+	math.randomseed(peekUntilMS);
+	-- We have to +1 then -1 to ensure that we have 0 as a valid random offset.
+	offset = math.random((count-limit)+1) - 1
```
darwin67, line 22: nice :+1:

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -716,20 +713,39 @@ func (q *queue) processPartition(ctx context.Context, p *QueuePartition, shard *
...
+	// queue items later in the array may be processed before queue items earlier in
+	// the array depending on eg. a rate limit becoming available half way through
+	// iteration.
+	staticTime := getNow()
+
 ProcessLoop:
 	for _, qi := range queue {
+		// TODO: Create an in-memory mapping of rate limit keys that have been hit,
+		//       and don't bother to process if the queue item has a limited key.  This
+		//       lessens work done in the queue, as we can `continue` immediately.
```
darwin67, line 747: why not store the map in redis instead with a TTL for those keys?
don't remember if you can expire items in a HSET though.

if you can, that can be shared across all executors so each don't have to recreate the map themselves.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -716,20 +713,39 @@ func (q *queue) processPartition(ctx context.Context, p *QueuePartition, shard *
...
 
-	var processErr error
-	var concurrencyLimitReached bool
+	var (
+		processErr error
+
+		// These flags are used to handle partition rqeueueing.
+		ctrSuccess     int32
+		ctrConcurrency int32
+		ctrRateLimit   int32
```
darwin67, line 728: nit: non-blocker

probably good to wrap these variables into a struct if it's gonna get more complicated than this. easier to create helper functions as well instead of doing things like
```
if ctrConcurrency > 0 || (ctrRateLimit > 0 && ctrConcurrency == 0 && ctrSuccess == 0)
```
which is not intuitive on first sight, and require comments to clarify intent.

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -0,0 +1,36 @@
...
+
+	local diff_ms = now_ms - allow_at_ms
+
+	if diff_ms < 0 then
+		-- return {false, tat, new_tat, allow_at_ms, diff_ms}
+		return false
+	end
+
+	local expiry = math.ceil(now_ms + period_ms / 1000)
+	redis.call("SETEX", key, new_tat, expiry)
```
darwin67, line 32: FYI: `SETEX` is deprecated.
https://redis.io/docs/latest/commands/setex/

pkg/execution/queue/item.go
```diff
@@ -75,6 +75,20 @@ type Item struct {
...
+}
+
+type Throttle struct {
+	// Key is the unique throttling key that's used to group queue items when
+	// processing rate limiting/throttling.
+	Key string `json:"k"`
+	// Limit is the actual rate limit
+	Limit int `json:"l"`
+	// Burst is the busrsable capacity of the rate limit
+	Burst int `json:"b"`
```
darwin67, line 89: what exactly is `Burst` in this context?
the additional amount of times we let items through? or is it something else?
if so, should we be doing some accounting on this? like how many bursts were used.

pkg/execution/state/redis_state/queue_test.go
```diff
@@ -2028,6 +2080,80 @@ func TestShardLease(t *testing.T) {
...
+	q := NewQueue(rc)
+
+	idA, idB := uuid.New(), uuid.New()
+
+	_ = idB
+
+	t.Run("Without bursts", func(t *testing.T) {
+		throttle := &osqueue.Throttle{
+			Key:    "some-key",
+			Limit:  1,
```
darwin67, line 2102: could you add another test here that's
```
Throttle{
  Key: "some-other-key",
  Limit: 1,
  Period: 5,
  Burst: 2
}
```
Would be good to verify that bursting is working as expected (in this case 3 should pass), since the current test has no bursts.

darwin67 (approved):
Looks good with 1 request for an additional test to verify burst works as expected.

pkg/execution/queue/item.go
tonyhb: It's accounted for in GCRA :)

pkg/execution/state/redis_state/queue_processor.go
tonyhb: That's not really any different from checking a rate limit, which is a single get and set :)

This would prevent us from accessing the backend overall, reducing _any_ round trips as a small optimization in this core loop.

pkg/execution/state/redis_state/lua/includes/gcra.lua
tonyhb: Resolved, thanks!

pkg/execution/state/redis_state/queue_test.go
tonyhb: Added, thanks!

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -193,17 +193,14 @@ LOOP:
...
 			if q.capacity() < minWorkersFree {
-				q.seqLeaseLock.RUnlock()
 				// Wait until we have more workers free.  This stops us from
 				// claiming a partition to work on a single job, ensuring we
 				// have capacity to run at least MinWorkersFree concurrent
 				// QueueItems.  This reduces latency of enqueued items when
 				// there are lots of enqueued and available jobs.
 				continue
 			}
-			q.seqLeaseLock.RUnlock()
```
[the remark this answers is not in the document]
tonyhb: It is, indeed, not doing anything!

PR #1265, opened 2024-04-09
Base for moving invokes away from pauses

jpwilliams:
## Motivation

Invokes currently contribute many pauses with expressions to the system. These pauses are attributed to the exceedingly common `inngest/function.finished` event, which is sent after the conclusion of each individual run.

Because of this, every run triggers us to check the incoming event against all pending invocation pauses. If there are many, pulling all of these pauses from Redis to then run CEL expressions against them can be very, very slow.

## Summary

Invokes, unlike userland `step.waitForEvent()` calls, are not arbitrary expressions. We piggybacked on the expressions system to build it, but in reality every invoke pause could be found with a single O(1) lookup, as we are searching for one string.

Because of this, we can pull `step.invoke()` pauses away from `step.waitForEvent()` pauses and optimize their performance without using expressions at all.

This PR:

1. Adds a new `CorrelationID` field to pauses
https://github.com/inngest/inngest/blob/073f71f8165ffe356fedb58f02d144d27911655f/pkg/execution/state/pause.go#L155-L156
2. When saving the pause, this is set in an `invoke` hash in Redis, with a value of the pause's ID
https://github.com/inngest/inngest/blob/073f71f8165ffe356fedb58f02d144d27911655f/pkg/execution/state/redis_state/lua/savePause.lua#L41-L43
3. If this correlation ID is set, we also do not touch the `pause-events` key when saving the promise, meaning regular pause processing will never see this pause and never pull it from Redis
https://github.com/inngest/inngest/blob/073f71f8165ffe356fedb58f02d144d27911655f/pkg/execution/state/redis_state/redis_state.go#L571-L578
4. When receiving an event, we check if the event is an `inngest/function.finished` event with a correlation ID
https://github.com/inngest/inngest/blob/073f71f8165ffe356fedb58f02d144d27911655f/pkg/execution/runner/runner.go#L363-L375
5. If yes, try get the pause ID from the `invoke` hash in Redis. If we find it, consume the pause and resume the invoking function
https://github.com/inngest/inngest/blob/073f71f8165ffe356fedb58f02d144d27911655f/pkg/execution/executor/executor.go#L1422-L1423

This ensures a few things:

- Invokes continue to use pauses and much of their existing logic, so this is a small code change
- Invokes are now completely removed from regular pause processing
- Invokes no longer use expressions and are always O(1) lookups

- [ ] Docs

## Checklist

## Related

- #1263
This PR removes a tonne of pressure off of the `waitForEvent` CEL expressions by taking invokes out of the equation entirely. The linked PR also addresses very large numbers of non-invoke waits.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -1426,9 +1426,6 @@ func (e *executor) HandleInvokeFinish(ctx context.Context, evt event.TrackedEven
 
 	// find the pause with correlationID
 	pause, err := e.sm.PauseByInvokeCorrelationID(ctx, correlationID)
-	if err == state.ErrInvokePauseNotFound || err == state.ErrPauseNotFound {
-		return nil
-	}
 	if err != nil {
 		return err
 	}
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -1890,42 +1890,38 @@ func (e *executor) handleGeneratorInvokeFunction(ctx context.Context, gen state.
 		return fmt.Errorf("unable to parse invoke function expires: %w", err)
 	}
 
-	e.sm.SaveInvokePause()
+	eventName := event.FnFinishedName
+	correlationID := item.Identifier.RunID.String() + "." + gen.ID
 
-	// eventName := event.FnFinishedName
-	// correlationID := item.Identifier.RunID.String() + "." + gen.ID
-	// strExpr := fmt.Sprintf("async.data.%s == %s", consts.InvokeCorrelationId, strconv.Quote(correlationID))
-	// _, err = e.newExpressionEvaluator(ctx, strExpr)
-	// if err != nil {
-	// 	return execError{err: fmt.Errorf("failed to create expression to wait for invoked function completion: %w", err)}
-	// }
+	logger.From(ctx).Info().Interface("opts", opts).Time("expires", expires).Str("event", eventName).Msg("parsed invoke function opts")
 
-	// logger.From(ctx).Info().Interface("opts", opts).Time("expires", expires).Str("event", eventName).Str("expr", strExpr).Msg("parsed invoke function opts")
+	pauseID := uuid.NewSHA1(
+		uuid.NameSpaceOID,
+		[]byte(item.Identifier.RunID.String()+gen.ID),
+	)
 
-	// pauseID := uuid.NewSHA1(
-	// 	uuid.NameSpaceOID,
-	// 	[]byte(item.Identifier.RunID.String()+gen.ID),
-	// )
-
-	// opcode := gen.Op.String()
-	// err = e.sm.SavePause(ctx, state.Pause{
-	// 	ID:          pauseID,
-	// 	WorkspaceID: item.WorkspaceID,
-	// 	Identifier:  item.Identifier,
-	// 	GroupID:     item.GroupID,
-	// 	Outgoing:    gen.ID,
-	// 	Incoming:    edge.Edge.Incoming,
-	// 	StepName:    gen.UserDefinedName(),
-	// 	Opcode:      &opcode,
-	// 	Expires:     state.Time(expires),
-	// 	Event:       &eventName,
-	// 	// Expression:  &strExpr,
-	// 	InvokeCorrelationID: "",
-	// 	DataKey:     gen.ID,
-	// })
-	// if err == state.ErrPauseAlreadyExists {
-	// 	return nil
-	// }
+	opcode := gen.Op.String()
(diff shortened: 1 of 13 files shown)


pkg/execution/state/redis_state/key_generator.go
```diff
@@ -136,6 +139,10 @@ func (d DefaultKeyFunc) PauseIndex(ctx context.Context, kind string, wsID uuid.U
 	return fmt.Sprintf("%s:pause-idx:%s:%s:%s", d.Prefix, kind, wsID, event)
 }
 
+func (d DefaultKeyFunc) Invoke(ctx context.Context) string {
+	return fmt.Sprintf("%s:invoke", d.Prefix)
```
darwin67, line 143: shouldn't need anymore suffix since the correlationID should be guaranteed to be unique, and collusion shouldn't happen

pkg/execution/runner/runner.go
```diff
@@ -360,9 +360,24 @@ func (s *svc) handleMessage(ctx context.Context, m pubsub.Message) error {
 		}
 	}()
 
+	// check if this is an "inngest/function.finished" event
+	// triggered by invoke
+	corrId := tracked.GetEvent().CorrelationID()
+	if tracked.GetEvent().Name == event.FnFinishedName && corrId != "" {
```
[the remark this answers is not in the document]
darwin67: I've seen this a couple of times now. probably worth making it a function on event.
```
tracked.GetEvent().Name == event.FnFinishedName
```

pkg/execution/executor/executor.go
```diff
@@ -1893,44 +1948,41 @@ func (e *executor) handleGeneratorInvokeFunction(ctx context.Context, gen state.
...
+		Event:               &eventName,
+		DataKey:             gen.ID,
+		InvokeCorrelationID: &correlationID,
 	})
 	if err == state.ErrPauseAlreadyExists {
 		return nil
 	}
+
 	if err != nil {
+		// TODO: should we delete both keys if failed?
```
darwin67, line 1980: what do we want to do here?
should we attempt to clean up? otherwise there might be trash data sitting around.

jpwilliams: Ah, this isn't relevant any more - it's an old TODO from when this had a waitgroup.

darwin67 (approved):
LGTM

pkg/execution/executor/executor.go
```diff
@@ -1908,18 +1963,18 @@ func (e *executor) handleGeneratorInvokeFunction(ctx context.Context, gen state.
...
+		WorkspaceID:         item.WorkspaceID,
+		Identifier:          item.Identifier,
+		GroupID:             item.GroupID,
+		Outgoing:            gen.ID,
+		Incoming:            edge.Edge.Incoming,
+		StepName:            gen.UserDefinedName(),
+		Opcode:              &opcode,
+		Expires:             state.Time(expires),
+		Event:               &eventName,
+		DataKey:             gen.ID,
```
tonyhb, line 1976: Can we keep the expression around?  We're still saving pauses, and without this backcompat is broken.  This means that on rollbacks / rollouts with blue-green deploys we'll match wrong pauses.

pkg/execution/state/redis_state/key_generator.go
tonyhb: Should we create maps by function ID?  In the event of account deletions/cleanups/etc this will provide better namespacing for easy cleanup.

pkg/execution/executor/executor.go
darwin67: Sure, we can add that back.

pkg/execution/executor/executor.go
```diff
@@ -1411,6 +1410,67 @@ func (e *executor) handleAggregatePauses(ctx context.Context, evt event.TrackedE
...
+
+	correlationID := evt.GetEvent().CorrelationID()
+	if correlationID == "" {
+		return fmt.Errorf("no correlation ID found in event when trying to handle finish")
+	}
+
+	// find the pause with correlationID
+	pause, err := e.sm.PauseByInvokeCorrelationID(ctx, correlationID)
+	if err == state.ErrInvokePauseNotFound || err == state.ErrPauseNotFound {
+		return nil
```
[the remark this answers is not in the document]
darwin67: will do.

pkg/execution/state/redis_state/key_generator.go
darwin67: Done at [2776a77](https://github.com/inngest/inngest/pull/1265/commits/2776a77df72979e9cd00443fa0954f42dfe9fc86)

PR #1273, opened 2024-04-11
Ana/inn 2921 Update storybook

anafilipadealmeida:
## Description
- Simplify Storybook logic and remove auto-generated docs
- Remove the storybook from Cloud and render those stories in the components storybook
- Append radix Portal components as Modals and Popovers to the `<div id="modals">` instead of `<body>`, so we have more control over their styles

Screenshot 2024-04-11 at 22 31 01

## Motivation
- We have too many storybook and configs, which makes it difficult to upgrade them or fix issues.
- We had a ´<div id="modals" />` in the preview.tsx, but the portals were being appended to <body> , and the font styles were being defined lower in the tree
- The storybook switch between dark/light theme was not changing the story background entirely

- [X] Chore (refactors, upgrades, etc.)
- [ ] Docs

## Checklist


ui/apps/dashboard/src/app/baseWrapper.tsx
```diff
@@ -9,7 +9,10 @@ export function BaseWrapper({ children }: { children: React.ReactNode }) {
       lang="en"
       className={cn('h-full bg-white accent-indigo-500', interTight.variable, robotoMono.variable)}
     >
-      <body className="h-full overflow-hidden">{children}</body>
+      <body className="h-full overflow-hidden">
```
anafilipadealmeida, line 12: At some point we should think if it's worth replacing this with the shared `<AppRoot>` component, but I don't wanna mess with layouts now

darwin67 (approved):
Always happy to approve more deletions!
image

ui/packages/components/.storybook/preview.tsx
```diff
@@ -8,13 +9,13 @@ const preview: Preview = {
...
       return (
-        <div
-          className={`${interTight.variable} ${robotoMono.variable} dark:bg-slate-940 bg-white font-sans`}
-        >
-          <div id="app" />
-          <div id="modals" />
-          <Story />
-        </div>
+        <TooltipProvider>
+          <div className={`${interTight.variable} ${robotoMono.variable} font-sans`}>
```
goodoldneon, line 13: Do we need `w-screen` to avoid squashing stories to the min width?
image

PR #1279, opened 2024-04-16
Add unmarhsalling for TS SDK of throttle

tonyhb:
Ensures that string values from non-Go SDKs are unmarshalled correctly.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/inngest/function.go
+++ b/pkg/inngest/function.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/inngest/function.go
```diff
@@ -109,6 +110,37 @@ type Throttle struct {
...
+	t.Key = input.Key
+	t.Period, err = str2duration.ParseDuration(input.Period)
+
+	// Normalization
+	if t.Limit == 0 {
+		t.Limit = 1
+	}
+	if t.Burst == 0 {
+		t.Burst = 1
+	}
```
jpwilliams, line 140: Can we catch negative values here too?
```suggestion
	// Normalization
	if t.Limit <= 0 {
		t.Limit = 1
	}
	if t.Burst <= 0 {
		t.Burst = 1
	}
```

goodoldneon: Should negative value handling be in validation or normalization?

tonyhb: These are uints, so it's already handled.

PR #1280, opened 2024-04-16
INN-2912: Minor tweaks for span annotations

darwin67:
## Description

Small tweaks to span annotations to collect relevant information related to the function run.

The `StringSlice` type is not ideal since the slice will be serialized to a string, and it makes it harder to deserialize it back to an `Array(String)` type.
Instead, make it a comma delimited string, and let the database (or whatever downstream to) handle the string splitting.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -672,6 +672,7 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 			attribute.Int(consts.OtelSysFunctionVersion, id.WorkflowVersion),
 			attribute.String(consts.OtelAttrSDKRunID, id.RunID.String()),
 			attribute.Int(consts.OtelSysStepAttempt, item.Attempt),
+			attribute.Int(consts.OtelSysStepMaxAttempt, item.GetMaxAttempts()),
 			attribute.String(consts.OtelSysStepGroupID, item.GroupID),
 		),
 	)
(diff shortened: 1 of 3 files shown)


pkg/execution/executor/executor.go
```diff
@@ -342,7 +358,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 		eventIDsStr = append(eventIDsStr, id.String())
 	}
 	spanID := telemetry.NewSpanID(ctx)
-	span.SetAttributes(attribute.StringSlice(consts.OtelSysEventIDs, eventIDsStr))
+	span.SetAttributes(attribute.String(consts.OtelSysEventIDs, strings.Join(eventIDsStr, ",")))
```
jpwilliams, line 361: I'm confused by this change - `StringSlice` was an issue because it serialized the value as a string, so instead we're just making a string?

darwin67: it'll serialize to this in the DB `'["<ULID>"]'`, and doesn't store it as an array.
it's pretty hard to convert it back to `Array(String)` from this format, and instead making it like `<ULID>,<ULID>` is a lot easier because we can do string splitting.

darwin67: it'll serialize to this in the DB `'["<ULID>"]'`, and doesn't store it as an array.
it's pretty hard to convert it back to `Array(String)` from this format, and instead making it like `<ULID>,<ULID>` is a lot easier because we can do string splitting.

darwin67: Discussed in person and agreed it's easier with , delimited string.

jpwilliams (approved):

PR #1285, opened 2024-04-17
make sure batching and debounce are recorded correctly

darwin67:
## Description

Make sure to record batching and debounce since those are not event ingestion triggers

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -311,22 +311,17 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		),
 	)
 	defer span.End()
-	if len(req.Events) > 1 {
+	if req.BatchID != nil {
 		span.SetAttributes(attribute.String(consts.OtelSysBatchID, req.BatchID.String()))
+	}
+	if req.PreventDebounce {
+		span.SetAttributes(attribute.Bool(consts.OtelSysDebounceTimeout, true))
 	}
 	if req.Context != nil {
 		if val, ok := req.Context[consts.OtelPropagationLinkKey]; ok {
 			if link, ok := val.(string); ok {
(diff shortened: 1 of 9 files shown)


pkg/execution/executor/executor.go
```diff
@@ -324,9 +325,11 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 	for _, e := range req.Events {
 		evt := e.GetEvent()
 		if byt, err := json.Marshal(evt); err == nil {
-			span.AddEvent(string(byt), trace.WithAttributes(
-				attribute.Bool(consts.OtelSysEventData, true),
-			))
+			if b, err := jcs.Transform(byt); err == nil {
```
tonyhb, line 328: Why do we need to transform the data before annotating?

pkg/execution/executor/executor.go
```diff
@@ -648,14 +649,16 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
-	if len(id.EventIDs) > 1 {
+	if id.BatchID != nil {
 		fnSpan.SetAttributes(attribute.String(consts.OtelSysBatchID, id.BatchID.String()))
 	}
 	for _, e := range s.Events() {
 		if byt, err := json.Marshal(e); err == nil {
-			fnSpan.AddEvent(string(byt), trace.WithAttributes(
-				attribute.Bool(consts.OtelSysEventData, true),
-			))
+			if b, err := jcs.Transform(byt); err == nil {
```
tonyhb, line 657: And similarly, here, why are we transforming, too?  We've also already transformed on the Schedule call, so this seems to be extra work, CPU, latency on every step?

darwin67: removing this as discussed

tonyhb (approved):

PR #1304, opened 2024-04-25
Allow for increased steps based off of identifiers

tonyhb:
This allows customisation of step limits dependent on various factors in the identifier.  Note that defaults do not change

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -168,11 +168,8 @@ func WithLifecycleListeners(l ...execution.LifecycleListener) ExecutorOpt {
 	}
 }
 
-func WithStepLimits(limit uint) ExecutorOpt {
+func WithStepLimits(limit func(id state.Identifier) int) ExecutorOpt {
 	return func(e execution.Executor) error {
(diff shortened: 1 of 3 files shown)


pkg/execution/executor/executor.go
```diff
@@ -168,11 +168,8 @@ func WithLifecycleListeners(l ...execution.LifecycleListener) ExecutorOpt {
 	}
 }
 
-func WithStepLimits(limit uint) ExecutorOpt {
+func WithStepLimits(limit func(id state.Identifier) int) ExecutorOpt {
```
darwin67, line 171: shouldn't this be `uint` as well if the previous was an `uint`?

darwin67 (approved):

pkg/execution/executor/executor.go
tonyhb: its mainly for types, so that we don't need to continue to typecast everywhere.  we actively check for <= 0 in the validator so it's okay being an int, and is (i suppose) an easier type to work with.

PR #1307, opened 2024-04-26
Dev server trace skeleton

darwin67:
## Description

The generated otel traces need to be stored in the Dev server in order to be queriable for support of the new function run UI.

This change will add the following:
- simple trace endpoint that will accept OTLP traces in HTTP format
- add DB schema for storing traces and runs
- add cqrs interfaces for writing and reading traces/runs

Implementation of the storage and retrieval will be follow up PRs.

- [ ] Docs

## Checklist


pkg/api/apiv1/apiv1.go
```diff
@@ -92,6 +92,9 @@ func (a *router) setup() {
 		r.Post("/cancellations", a.createCancellation)
 		r.Get("/cancellations", a.getCancellations)
 		r.Delete("/cancellations/{id}", a.deleteCancellation)
+
+		// NOTE: this is for dev server only, and not an actual v1 API
+		r.Post("/traces", a.OTLPTrace)
```
tonyhb, line 97: If this is not a V1 API, should we move it to eg /dev/traces out of the v1 package?  The V1 API is specifically meant to be shared between OSS and cloud

tonyhb (approved):

pkg/api/apiv1/apiv1.go
darwin67: I can do that.

darwin67: Done

PR #1320, opened 2024-05-03
INN 2988 add relative time filter to runs

anafilipadealmeida:
## Description

- Add Relative time filter (to be changed when we introduce absolute time filter)
- Wire timeField filter, now the API is released
- Change structure of Select options, to allow disabled options

https://github.com/inngest/inngest/assets/16758464/a7013820-4275-435b-9893-6b813c13a01a

## Motivation
- Relative time url params have to be days, rather than start and end times, as when we refresh the page, we want to keep a relative date.
- Once we add the filter for absolute dates, we need to improve the validation experience (what happens when we want to filter by a range that the current plan doesn't support)

- [ ] Docs

## Checklist


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/TimeFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [{ data }] = useQuery({
+    query: GetBillingPlanDocument,
+  });
+
+  // Since "features" is a map, we can't be 100% sure that there's a log
+  // retention value. So default to 7 days.
+  let logRetention = 7;
+  if (typeof data?.account.plan?.features.log_retention === 'number') {
+    logRetention = data.account.plan.features.log_retention;
+  }
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/TimeFilter.tsx:41]: Can we make a React context for this? There's a handful of places that need to know the plan's history

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx
```diff
@@ -51,14 +65,35 @@ export default function RunsPage() {
...
+  useEffect(() => {
+    if (lastDays) {
+      setStartTime(
+        getTimestampDaysAgo({
+          currentDate: new Date(),
+          days: parseInt(lastDays),
+        })
+      );
+    } else {
+      setStartTime(TimeRangeFilterDefault.startTime);
```
Comment (goodoldneon) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx:86]: This is gonna get stale after the page loads. Probably an esoteric edge case, but it's still cleaner to avoid this `else` block by defaulting `lastDays` if it isn't set:

```ts
const [lastDays = "3", setLastDays] = useSearchParam('last')
```

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/TimeFilter.tsx
Comment (anafilipadealmeida) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/TimeFilter.tsx:41]: Created a [Linear ticket ](https://linear.app/inngest/issue/INN-3005/add-plans-validation-as-context) for that one

Comment (djfarrelly) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/TimeFilter.tsx:41]: Yeah - an "Entitlements" context would be ideal!

ui/packages/components/src/Select/Select.tsx
```diff
@@ -7,42 +7,53 @@ type SelectProps = {
...
 export function Select({
   defaultValue,
   label,
   isLabelVisible = true,
   children,
   onChange,
   multiple,
-}: SelectProps & (MultiProps | SingleProps)) {
+  className,
+}: Props) {
```
djfarrelly, line 41: **suggestion (non-blocking)**: This can be for future, but it might be nice to allow `Option` or it's `id` to be a generic type that can be passed to the `Select` component (ex. `Select<T>(...`). That way there could be less type checking in components that use Select and pass an onChange handler. We might then be able to drop the `if (isFunctionTimeField(option.id)) {...` checking in the other components.

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx
```diff
@@ -51,18 +56,32 @@ export default function RunsPage() {
...
+  const [lastDays = '3', setLastDays] = useSearchParam('last');
+
+  const timeField = toTimeField(rawTimeField ?? '') ?? FunctionRunTimeFieldV2.QueuedAt;
+
+  /* TODO: Time params for absolute time filter */
+  // const [fromTime, setFromTime] = useSearchParam('from');
+  // const [untilTime, setUntilTime] = useSearchParam('until');
+
+  /* TODO: When we have absolute time, the start date will be either coming from the date picker or the relative time */
+  const [startTime, setStartTime] = useState<Date>(new Date());
```
Comment (djfarrelly) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx:68]: **suggestion (non-blocking)**: Probably doesn't matter with the TODO here, but will this cause issues b/c `new Date()` will return a new value every time this component renders? I don't know if it needs to be memo'd or we can just forget it for now until the date picker is there?

djfarrelly (approved):
Looks good - only real suggestion for the future is to add a generic to `Select` 👍

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx
Comment (goodoldneon) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx:68]: I think `startTime` should be derived (and memoized, like you're suggesting) from the URL state. If the URL has something like `relative=5h` then `startTime` would be 5 hours ago. If the URL has something like `startTime=2024-05-07T11:22:33Z` then it would be `new Date("2024-05-07T11:22:33Z")`.

Also, should we rename `startTime`? It's kinda confusing since it reads like it's related to "started at" when it's really the lower time bound for whichever time field is used (queued, started, or ended)

Comment (anafilipadealmeida) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/runs/page.tsx:69]: @goodoldneon was making a good point about always inferring the value from the URL, even if they are the initial defaults that users didn't select. That means changing the behaviour so that the defaults in filters are now also part of the URL, and drop the need for this state. I'll make that change either with pagination or the absolute time work

PR #1325, opened 2024-05-07
Updated state interfaces

tonyhb:
- [ ] Docs

## Checklist


pkg/execution/executor/executor.go
```diff
@@ -1528,55 +1528,63 @@ func (e *executor) HandleInvokeFinish(ctx context.Context, evt event.TrackedEven
 }
 
 // Cancel cancels an in-progress function.
-func (e *executor) Cancel(ctx context.Context, runID ulid.ULID, r execution.CancelRequest) error {
-	s, err := e.sm.Load(ctx, runID)
+func (e *executor) Cancel(ctx context.Context, id sv2.ID, r execution.CancelRequest) error {
+	md, err := e.smv2.LoadMetadata(ctx, id)
+	if err == sv2.ErrMetadataNotFound {
+		logger.From(ctx).Warn().Err(err).Msg("cancelling function already removed")
```
darwin67, line 1534: do we need this log? it's not logging any tenant identifiers so it's kinda useless if it's gonna be used for identifying runs.
if no additional metadata is gonna be added, let's just remove it.

pkg/execution/executor/executor.go
```diff
@@ -2128,32 +2126,34 @@ func (e *executor) handleGeneratorWaitForEvent(ctx context.Context, gen state.Ge
...
+	// 	// TODO: Only need events here.
+	// 	run, err := e.sm.Load(ctx, item.Identifier.RunID)
+	// 	if err != nil {
+	// 		return execError{err: fmt.Errorf("unable to load run after execution: %w", err)}
+	// 	}
+
+	// 	// Take the data for expressions based off of state.
+	// 	ed := expressions.NewData(state.ExpressionData(ctx, run))
+	// 	data = expr.FilteredAttributes(ctx, ed).Map()
+	// }
```
darwin67, line 2152: if it's not needed, let's delete it.

pkg/devserver/lifecycle.go
```diff
@@ -6,51 +6,51 @@ import (
...
-			cqrs.WithEventBatchFunctionID(id.WorkflowID),
-			cqrs.WithEventBatchRunID(id.RunID),
-			cqrs.WithEventBatchEventIDs(id.EventIDs),
-			cqrs.WithEventBatchExecutedTime(executedTime),
-		)
-
-		if batch.IsMulti() {
-			_ = l.cqrs.InsertEventBatch(ctx, *batch)
+	/*
+		if id.BatchID != nil {
```
darwin67, line 37: I guess batchID is still needed for now for backward compatibility like you said.
This will likely break batching UI for dev server.

pkg/execution/driver/driver.go
```diff
@@ -29,39 +32,57 @@ type Driver interface {
...
 	attempt int,
 ) ([]byte, error) {
-	md := s.Metadata()
 
 	req := &SDKRequest{
-		Events:  s.Events(),
-		Event:   s.Event(),
-		Actions: s.Actions(),
+		Event:   map[string]any{},
+		Events:  []map[string]any{},
```
darwin67, line 45: should this be just `[]event.Event` instead?
don't know if we need to serialize it to JSON bytes if the HTTP requester is gonna do it for us anyways.

pkg/execution/driver/driver.go
```diff
@@ -29,39 +32,57 @@ type Driver interface {
...
+				return nil, fmt.Errorf("error unmarshalling event in driver marshaller: %w", err)
+			}
+		}
+		req.Event = evts[0]
+		req.Events = evts
+
+		// We do not need to unmarshal state, as it's already marshalled.
+		for k, v := range state.Steps {
+			req.Actions[k] = v
+		}
```
darwin67, line 85: You need to set `UserAPI = false` in here, or else this whole block is just wasted CPU cycles.

pkg/execution/execution.go
```diff
@@ -87,37 +88,24 @@ type Executor interface {
...
-		id state.Identifier,
-		item queue.Item,
-		edge inngest.Edge,
-		resp *state.DriverResponse,
-	) error
-
-	// HandleGeneratorResponse handles all generator responses.
-	HandleGeneratorResponse(ctx context.Context, resp *state.DriverResponse, item queue.Item) error
-	// HandleGenerator handles an individual generator response returned from the SDK.
-	HandleGenerator(ctx context.Context, gen state.GeneratorOpcode, item queue.Item) error
```
darwin67, line 102: are these no longer needed because they're not required anymore and will always be using the OSS logic?
if so, we could just make the methods private.

pkg/execution/executor/validate.go
```diff
@@ -78,26 +77,18 @@ func (r *runValidator) checkStepLimit(ctx context.Context) error {
...
-
-			// Create a new driver response to map as the function finished error.
-			resp := state.DriverResponse{}
-			resp.SetError(state.ErrFunctionOverflowed)
-			resp.SetFinal()
+	if limit > 0 && r.md.Metrics.StepCount >= limit {
+		// Create a new driver response to map as the function finished error.
+		resp := state.DriverResponse{}
+		resp.SetError(state.ErrFunctionOverflowed)
+		resp.SetFinal()
```
darwin67, line 84: should this also run `SaveStep`?
otherwise, how will the status for the function be set if it's going to return early?

pkg/execution/executor/validate.go
```diff
@@ -57,18 +63,11 @@ func (r *runValidator) validate(ctx context.Context) error {
 	return nil
 }
 
-func (r *runValidator) checkCancelled(ctx context.Context) error {
```
darwin67, line 60: why is this no longer needed?

pkg/execution/state/v2/interfaces.go
```diff
@@ -0,0 +1,61 @@
...
+	"io"
+
+	"github.com/google/uuid"
+)
+
+type CreateState struct {
+	Metadata Metadata
+	// Events contains a slice of JSON-encoded events.
+	Events []json.RawMessage
+	// XXX: You cannot start state with pre-existing steps yet.
```
darwin67, line 15: what does this mean?

darwin67 (changes requested):
The `extractTraceCtx` need to be added back or it'll break trace propagation.
Also added some questions.

pkg/execution/executor/executor.go
```diff
@@ -574,49 +624,49 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
-		// don't return an error.
-		// XXX: Handle retries with error types and return a non-retryable error here.
-		return nil, nil
+		// don't return an error - assume the function finishes and delete state.
+		return nil, e.smv2.Delete(ctx, md.ID)
 	}
 
-	// Store the metadata in context for future use and propagate trace
-	// context. This can be used to reduce reads in the future.
-	ctx = e.extractTraceCtx(WithContextMetadata(ctx, md), id, &item)
```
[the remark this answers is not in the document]
tonyhb: Can we pair on this later?  I've noticed that `WithContextMetadata` is not necessary and would like to run what it does by you, and why it's included.

pkg/execution/driver/driver.go
tonyhb: I'm leaving this as-is due to the gateway.  You're right that drivers could do with a major refactor.

pkg/execution/driver/driver.go
tonyhb: Haha, thanks, good catch.

pkg/execution/executor/executor.go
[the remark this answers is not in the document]
darwin67: Sure. I actually don't know what the `WithContextMetadata` is for, and we probably don't really rely on it.
We just need the trace context to be extracted from the `Metadata.Context` field, and have it injected into `ctx`, which is what `extractTraceCtx` is for.

It's also used before `onFunctionCancelled` hook is called.

pkg/execution/executor/validate.go
darwin67: Ah I see. I was asking saving for history, not state. if that's handled, I think it should be fine.

pkg/execution/executor/executor.go
```diff
@@ -724,36 +774,30 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
 			}
 		}
 	}
 
-	// Ensure that if users requeue steps we never re-execute.
-	incoming := edge.Incoming
-	if edge.IncomingGeneratorStep != "" {
-		incoming = edge.IncomingGeneratorStep
-	}
-	if resp, _ := s.ActionID(incoming); resp != nil {
```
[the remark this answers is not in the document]
tonyhb: Yep — thank you for checking!  This is a remnant.

pkg/execution/executor/executor.go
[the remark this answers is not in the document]
tonyhb: Re-added, thank you!

pkg/execution/driver/driver.go
```diff
@@ -29,39 +32,65 @@ type Driver interface {
...
+		UseAPI:  true,
 	}
 
-	// empty the attrs that consume the most
-	if req.IsBodySizeTooLarge() {
-		req.Events = []map[string]any{}
-		req.Actions = map[string]any{}
-		req.UseAPI = true
-		req.Context.UseAPI = true
+	if md.Metrics.StateSize <= consts.MaxBodySize {
```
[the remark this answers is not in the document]
tonyhb: Great catch, resolved.  Thank you!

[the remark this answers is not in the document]
tonyhb: Great catch, resolved.  Thank you!

jpwilliams (approved):

PR #1327, opened 2024-05-07
Add additional metrics for queue

darwin67:
## Description

Add additional metrics for instrumenting queue performance.
Also re-organize some of the existing counters into `telemetry`.

- [ ] Docs

## Checklist


pkg/telemetry/counter.go
```diff
@@ -0,0 +1,89 @@
...
+
+type CounterOpt struct {
+	PkgName string
+	Tags    map[string]any
+}
+
+func IncrQueuePeekedCounter(ctx context.Context, incr int64, opts CounterOpt) {
+	recordCounterMetric(ctx, incr, counterOpt{
+		Name:        opts.PkgName,
+		MetricName:  "queue_peeked_total",
```
darwin67, line 13: this is the new metric added.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -1133,8 +1137,8 @@ func (q *queue) capacity() int64 {
 // queue items.
 func (q *queue) peekSize() int64 {
 	f := q.capacity()
-	if f > QueuePeekMax {
-		return QueuePeekMax
+	if f > QueuePeekDefault {
+		return QueuePeekDefault
```
darwin67, line 1141: maybe this should be an environment variable setting instead of a const.
and just use the default value if it's not set.

tonyhb: Thinking that we we shouldn't downcast from max to default — if we do that, we should lower the max.  We could either lower the Default and the Max, or update the consts that we use when we initialize the queue in OSS and the cloud.

tonyhb (approved):

PR #1335, opened 2024-05-09
Add run details pop out page

goodoldneon:
> :warning: Depends on https://github.com/inngest/monorepo/pull/2614

## Description
Add a new `/env/production/runs/[runID]` route that shows the new run details. This will be the "pop out" page where users can look at run details outside of the run list.

## Testing
image

- [ ] Docs

## Checklist


ui/packages/components/src/TimelineV2/TraceHeading.tsx
```diff
@@ -92,12 +93,15 @@ function TimeWithText({ trace }: { trace: Props['trace'] }) {
...
-      text = 'Failed';
+
+    if (isFunctionRunStatus(trace.status)) {
+      if (trace.status === 'CANCELLED') {
+        text = 'Cancelled';
+      } else if (trace.status === 'COMPLETED') {
+        text = 'Completed';
+      } else if (trace.status === 'FAILED') {
+        text = 'Failed';
+      }
```
djfarrelly, line 104: **suggestion (non-blocking)**: Since we're importing a function to check the type then use a switch-like statement to return the text, we could also just put this entire thing into a util fn if we use it in multiple places like: `getStatusText(trace.status)`. Also, it would be ideal to check equality against constants and avoid string matching.

ui/packages/components/src/TimelineV2/Trace.tsx
```diff
@@ -38,13 +38,11 @@ export function Trace({ depth, getOutput, isExpandable = true, maxTime, minTime,
...
-    trace: {
-      endedAt: trace.endedAt ? new Date(trace.endedAt) : new Date(),
-      queuedAt: new Date(trace.queuedAt),
-      startedAt: trace.startedAt ? new Date(trace.startedAt) : null,
-    },
+    ended: toMaybeDate(trace.endedAt)?.getTime() ?? null,
+    max: maxTime.getTime(),
+    min: minTime.getTime(),
+    queued: new Date(trace.queuedAt).getTime(),
+    started: toMaybeDate(trace.startedAt)?.getTime() ?? null,
```
djfarrelly, line 45: Much cleaner!

djfarrelly (approved):

ui/packages/components/src/TimelineV2/TraceHeading.tsx
goodoldneon: Where's a good spot for that? `ui/packages/components/src/types/functionRun.ts` doesn't seem right since the user-facing status text is copy and not a type

PR #1344, opened 2024-05-10
Implement batch keys in OSS

BrunoScheufler:
## Description

This pull request introduces a new `key` option on the function batching configuration. When storing batches, the batch pointer now includes a hashed representation of the evaluated expression. Unlike the previous implementation where one function would have at most one associated active batch, batch keys create many batches per function.

## Motivation

Batch keys enable event aggregation or grouping by tenants: Events could be counted for user notifications, or processed together to enable other features.

- [ ] Docs

## Checklist
- [x] I've linked any associated issues to this PR. -> https://linear.app/inngest/issue/INN-3010/implement-batch-keys-in-oss-executor

### DIFF
--- a/pkg/execution/batch/redis.go
+++ b/pkg/execution/batch/redis.go
@@ -48,7 +48,7 @@ func (b redisBatchManager) batchKey(ctx context.Context, evt event.Event, fn inn
 			Str("expression", *fn.EventBatch.Key).
 			Interface("event", evt.Map()).
 			Msg("error evaluating batch key expression")
-		return "<invalid>", nil
+		return fn.ID.String(), fmt.Errorf("invalid expression: %w", err)
 	}
 	if str, ok := out.(string); ok {
 		return str, nil
--- a/pkg/execution/batch/redis.go
+++ b/pkg/execution/batch/redis.go
@@ -163,10 +163,44 @@ func (b redisBatchManager) RetrieveItems(ctx context.Context, batchID ulid.ULID)
 
 // StartExecution sets the status to `started`
 // If it has already started, don't do anything
-func (b redisBatchManager) StartExecution(ctx context.Context, batchID ulid.ULID, batchPointer string) (string, error) {
+func (b redisBatchManager) StartExecutionWithBatchPointer(ctx context.Context, batchID ulid.ULID, batchPointer string) (string, error) {
 	keys := []string{
 		b.k.BatchMetadata(ctx, batchID),
 		batchPointer,
+	}
+	args := []string{
+		enums.BatchStatusStarted.String(),
+		ulid.Make().String(),
+	}
+
+	status, err := scripts["start"].Exec(
+		ctx,
+		b.r,
+		keys,
+		args,
+	).AsInt64()
+	if err != nil {
+		return "", fmt.Errorf("failed to start batch execution: %w", err)
+	}
+
+	switch status {
+	case 0: // haven't started, so mark mark it started
+		return enums.BatchStatusReady.String(), nil
+
+	case 1: // Already started
+		return enums.BatchStatusStarted.String(), nil
+
+	default:
+		return "", fmt.Errorf("invalid status for start batch ops: %d", status)
+	}
+}
+
+// StartExecution sets the status to `started`
+// If it has already started, don't do anything
+func (b redisBatchManager) StartExecution(ctx context.Context, fnID uuid.UUID, batchID ulid.ULID) (string, error) {
+	keys := []string{
+		b.k.BatchMetadata(ctx, batchID),
+		b.k.BatchPointer(ctx, fnID),
 	}
 	args := []string{
 		enums.BatchStatusStarted.String(),
--- a/pkg/execution/batch/redis.go
+++ b/pkg/execution/batch/redis.go
@@ -39,7 +39,7 @@ type redisBatchManager struct {
 
 func (b redisBatchManager) batchKey(ctx context.Context, evt event.Event, fn inngest.Function) (string, error) {
 	if fn.EventBatch.Key == nil {
-		return fn.ID.String(), nil
+		return "default", nil
 	}
 
 	out, _, err := expressions.Evaluate(ctx, *fn.EventBatch.Key, map[string]any{"event": evt.Map()})
--- a/pkg/execution/batch/redis.go
+++ b/pkg/execution/batch/redis.go
@@ -38,14 +38,14 @@ type redisBatchManager struct {
 }
 
 func (b redisBatchManager) batchKey(ctx context.Context, evt event.Event, fn inngest.Function) (string, error) {
-	if fn.Debounce.Key == nil {
+	if fn.EventBatch.Key == nil {
 		return fn.ID.String(), nil
 	}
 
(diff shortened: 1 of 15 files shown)


vendor/github.com/inngest/inngestgo/handler.go
```diff
@@ -341,6 +341,7 @@ func (h *handler) register(w http.ResponseWriter, r *http.Request) error {
 			f.EventBatch = map[string]any{
 				"maxSize": c.BatchEvents.MaxSize,
 				"timeout": c.BatchEvents.Timeout,
+				"key": c.BatchEvents.Key,
```
BrunoScheufler, line 344: The only reason I manually updated this was to make the tests pass (otherwise the test function doesn't include the key in its register request to the devserver).

We can remove this or update the entire vendor once we introduce SDK support for batch keys!

pkg/execution/batch/redis.go
```diff
@@ -33,6 +37,43 @@ type redisBatchManager struct {
...
+		return fn.ID.String(), nil
+	}
+
+	out, _, err := expressions.Evaluate(ctx, *fn.EventBatch.Key, map[string]any{"event": evt.Map()})
+	if err != nil {
+		log.From(ctx).Error().Err(err).
+			Str("expression", *fn.EventBatch.Key).
+			Interface("event", evt.Map()).
+			Msg("error evaluating batch key expression")
+		return "<invalid>", nil
```
darwin67, line 51: if it's invalid, maybe just return it as the default `fn.ID.String()` instead?
hardcoding `"<invalid>"` here is no different from just using the function ID.

maybe also return an `error` just in case. I suspect it's not going to be used anytime soon, but hopefully we can surface this kind of invalid expression errors to users better.

vendor/github.com/inngest/inngestgo/handler.go
darwin67: You can make the change in `inngestgo` with a PR, and just vendor the hash for that commit.

BrunoScheufler: Great point, updated!

pkg/execution/batch/redis.go
BrunoScheufler: Update this to return an error and use the fn, just to clarify, should we pass this up already or silently ignore the error? I guess it does make sense to warn early and condition users to use valid expressions returning strings, what do you think?

darwin67: We don't have a way to pass this error back up to the user atm.
It's the same for all expression related evaluation, so we should bubble the evaluation error up somehow so users are aware.

But regardless we don't want to `fail` the execution because there's an invalid expression since there are no ways to recover from that and runs are essentially lost at that point.

pkg/execution/batch/redis.go
```diff
@@ -33,6 +37,43 @@ type redisBatchManager struct {
...
+		log.From(ctx).Error().Err(err).
+			Str("expression", *fn.EventBatch.Key).
+			Interface("event", evt.Map()).
+			Msg("error evaluating batch key expression")
+		return fn.ID.String(), fmt.Errorf("invalid expression: %w", err)
+	}
+	if str, ok := out.(string); ok {
+		return str, nil
+	}
+	return fmt.Sprintf("%v", out), nil
```
darwin67, line 56: ratelimit hashes the key before returning it. could be a something to copy if it makes sense, and can save you that type assertion above.
https://github.com/inngest/inngest/blob/16f8e28ddcb6eca3b28df3319b45f36a7ba363c6/pkg/execution/ratelimit/ratelimit.go#L27-L51

darwin67: I see you're already base64 encoding it, so it might not matter.

pkg/execution/batch/redis.go
```diff
@@ -33,6 +37,43 @@ type redisBatchManager struct {
 	q redis_state.QueueManager
 }
 
+func (b redisBatchManager) batchKey(ctx context.Context, evt event.Event, fn inngest.Function) (string, error) {
+	if fn.EventBatch.Key == nil {
+		return fn.ID.String(), nil
```
darwin67, line 42: sorry, I just realized that `BatchPointer` is already namespace with the functionID.
https://github.com/inngest/inngest/blob/16f8e28ddcb6eca3b28df3319b45f36a7ba363c6/pkg/execution/state/redis_state/key_generator.go#L324-L326

a fnID string could be redundant, so a `default` string might be enough in this case. not a big deal as long as the fallback string is consistent.

pkg/inngest/batch.go
```diff
@@ -47,6 +47,8 @@ func NewEventBatchConfig(conf map[string]any) (*EventBatchConfig, error) {
 // - The batch is full
 // - The time to wait is up
 type EventBatchConfig struct {
+	Key *string `json:"key,omitempty"`
```
darwin67, line 50: there's a `IsValid` function right below.
https://github.com/inngest/inngest/blob/16f8e28ddcb6eca3b28df3319b45f36a7ba363c6/pkg/inngest/batch.go#L64-L83

good to add a simple check to make sure the CEL expression itself is valid, so we can surface issues early when users try to sync the function.

tests/golang/batch_test.go
```diff
@@ -140,3 +141,74 @@ func TestBatchInvoke(t *testing.T) {
...
+
+		<-time.After(5 * time.Second)
+		require.EqualValues(t, 1, batchInvokedCounter["a"])
+		require.EqualValues(t, 3, batchEventsCounter["a"])
+		require.EqualValues(t, 1, batchInvokedCounter["b"])
+		require.EqualValues(t, 3, batchEventsCounter["b"])
+		require.EqualValues(t, 1, batchInvokedCounter["c"])
+		require.EqualValues(t, 2, batchEventsCounter["c"])
+
+		require.EqualValues(t, 8, atomic.LoadInt32(&totalEvents))
```
darwin67, line 212:
```suggestion
		require.EqualValues(t, len(sequence), atomic.LoadInt32(&totalEvents))
```
nit: changing test values don't need you to update the validation.

tests/golang/batch_test.go
```diff
@@ -140,3 +141,74 @@ func TestBatchInvoke(t *testing.T) {
...
+			batchEventsCounter[input.Events[0].Data.UserId] += len(input.Events)
+			mut.Unlock()
+			atomic.AddInt32(&totalEvents, int32(len(input.Events)))
+			return true, nil
+		},
+	)
+	h.Register(a)
+	registerFuncs()
+	// calls pkg/devserver/api.go:256
+	// wrong type in pkg/sdk/function.go:11
```
darwin67, line 181: what are these comments?

darwin67 (approved):
Looks great! Nice tests as well.
Have a couple of comments and suggestions.
Only need you to add the CEL expr check in `IsValid`, others are minor suggestions that are not blocking.

Approving the PR so you can merge it once you made the changes.

pkg/execution/batch/redis.go
BrunoScheufler: Good point, changed to `default`

pkg/inngest/batch.go
BrunoScheufler: Added, thanks for the pointer!

PR #1356, opened 2024-05-14
Complete pause support in batch handling

cdzombak:
## Description

This PR allows our batch scheduling logic to take the function's paused_at column into account at the point in time that the batch is being scheduled for imminent execution.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/batch/batch.go
+++ b/pkg/execution/batch/batch.go
@@ -78,11 +78,11 @@ type ScheduleBatchOpts struct {
 }
 
 type ScheduleBatchPayload struct {
-	BatchID          ulid.ULID  `json:"batchID"`
-	AccountID        uuid.UUID  `json:"acctID"`
-	WorkspaceID      uuid.UUID  `json:"wsID"`
-	AppID            uuid.UUID  `json:"appID"`
-	FunctionID       uuid.UUID  `json:"fnID"`
-	FunctionVersion  int        `json:"fnV"`
-	FunctionPausedAt *time.Time `json:"fpAt,omitempty"`
+	BatchID                    ulid.ULID  `json:"batchID"`
+	AccountID                  uuid.UUID  `json:"acctID"`
+	WorkspaceID                uuid.UUID  `json:"wsID"`
+	AppID                      uuid.UUID  `json:"appID"`
+	FunctionID                 uuid.UUID  `json:"fnID"`
+	FunctionVersion            int        `json:"fnV"`
+	DeprecatedFunctionPausedAt *time.Time `json:"fpAt,omitempty"` // deprecated
 }
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -2439,9 +2439,13 @@ func (e *executor) extractTraceCtx(ctx context.Context, id state.Identifier, ite
 	return ctx
 }
 
-// AppendAndScheduleBatch appends a new batch item. If a new batch is created, it will be scheduled to run
+func (e *executor) AppendAndScheduleBatch(ctx context.Context, fn inngest.Function, bi batch.BatchItem) error {
+	return e.AppendAndScheduleBatchWithOpts(ctx, fn, bi, execution.BatchExecOpts{})
+}
+
+// AppendAndScheduleBatchWithOpts appends a new batch item. If a new batch is created, it will be scheduled to run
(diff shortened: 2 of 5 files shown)


pkg/execution/executor/executor.go
```diff
@@ -2439,9 +2439,13 @@ func (e *executor) extractTraceCtx(ctx context.Context, id state.Identifier, ite
 	return ctx
 }
 
-// AppendAndScheduleBatch appends a new batch item. If a new batch is created, it will be scheduled to run
+func (e *executor) AppendAndScheduleBatch(ctx context.Context, fn inngest.Function, bi batch.BatchItem) error {
```
darwin67, line 2442: why not do variadic variables for opts instead? like
```
AppendAndScheduleBatch(ctx context.Context, fn inngest.Function, bi batch.BatchItem, opts ...BatchOpt)
```

This way, you can call the function with or without the 4th argument.

cdzombak: I started down that path, but:

If the options are a trivial struct (as I've written in this PR), we need to handle the case where someone passes multiple options. (In particular, what if someone passes conflicting options?) We could return an error if given >1 options structure, or we could say that the first or last one takes priority, but overall it makes for a confusing API.

If the options are functions (following e.g. [the functional options pattern](https://davidbacisin.com/writing/golang-options-pattern)), then we still need to add some sort of structure for those functions to modify. It's not clear to me what that structure should be, particularly since this option is used to build different types of requests.

Ultimately, I couldn't come up with a good, usable API using varargs, and I think it's clear how to use this API correctly. However, if you feel strongly that we don't want to add new methods here, I'm willing to switch back to varargs and make the API as ergonomic as I can.

darwin67 (approved):

pkg/execution/batch/batch.go
```diff
@@ -78,11 +78,11 @@ type ScheduleBatchOpts struct {
...
-	FunctionID       uuid.UUID  `json:"fnID"`
-	FunctionVersion  int        `json:"fnV"`
-	FunctionPausedAt *time.Time `json:"fpAt,omitempty"`
+	BatchID                    ulid.ULID  `json:"batchID"`
+	AccountID                  uuid.UUID  `json:"acctID"`
+	WorkspaceID                uuid.UUID  `json:"wsID"`
+	AppID                      uuid.UUID  `json:"appID"`
+	FunctionID                 uuid.UUID  `json:"fnID"`
+	FunctionVersion            int        `json:"fnV"`
+	DeprecatedFunctionPausedAt *time.Time `json:"fpAt,omitempty"` // deprecated
```
BrunoScheufler, line 87: How long do we need to keep this around? Just until the old consumers are completely phased out and we run an updated version of pkg/execution/executor/service.go:300 that passes FunctionPausedAt to RetrieveAndScheduleBatchWithOpts?

BrunoScheufler (commented):
Looking good! Should we also update [pkg/execution/executor/service.go:321](https://github.com/inngest/inngest/blob/cdz/batches-support-new-pause-feature/pkg/execution/executor/service.go#L321) to use the new `RetrieveAndScheduleBatchWithOpts` and [pkg/execution/runner/runner.go:619](https://github.com/inngest/inngest/blob/cdz/batches-support-new-pause-feature/pkg/execution/runner/runner.go#L619) to use the new `AppendAndScheduleBatchWithOpts` or do you want to do this separately?

cdzombak:
> Looking good! Should we also update [pkg/execution/executor/service.go:321](https://github.com/inngest/inngest/blob/cdz/batches-support-new-pause-feature/pkg/execution/executor/service.go#L321) to use the new `RetrieveAndScheduleBatchWithOpts` and [pkg/execution/runner/runner.go:619](https://github.com/inngest/inngest/blob/cdz/batches-support-new-pause-feature/pkg/execution/runner/runner.go#L619) to use the new `AppendAndScheduleBatchWithOpts` or do you want to do this separately?

Done in bf2cce039c9318015d37b0e49f8bda98954b53a7

pkg/execution/batch/batch.go
cdzombak: Theoretically, I believe we could remove it now, since it's `omitempty` and no code we've shipped in prod has ever set this field.

For safety, I think we should keep this around until all payloads created before this PR is merged have been processed.

PR #1357, opened 2024-05-14
INN-3033: make code editor support tailwind/system color schema and overrides

jacobheric:
## Description

Code editor was previously hard coded to use a dark theme. Now it will use the tailwind/system color scheme by default and also support the ability to override the color scheme per use. This is useful for things like the current function view in the dashboard that we still want dark despite the fact that the dashboard itself use a light color scheme.

Here the webhooks editor gets the default light:
Screenshot 2024-05-14 at 3 31 06 PM

And the function editor is overriden to dark
Screenshot 2024-05-14 at 12 47 51 PM

- [x ] Chore (refactors, upgrades, etc.)
- [ ] Docs

## Checklist
- [ x] I've linked any associated issues to this PR.

https://linear.app/inngest/issue/INN-3033/add-light-theme-to-code-block


jacobheric:
There are a couple visual quirks to the editor that light mode exposed (see lower border corners). I'll clean those up.

jacobheric:
Do we want to make the editor toolbar light as well? I've left it dark for now as it more or less matches the rest of the current design.

anafilipadealmeida:
> Do we want to make the editor toolbar light as well? I've left it dark for now as it more or less matches the rest of the current design.

It's a great question. Yes! In light mode, the border of the entire block should be `border-slate-200`, we should replace the backgrounds by `bg-white`, and text and description `text-white` by `text-slate-700`. We can ignore the button colors for now

jacobheric:
I believe I've addressed all the items we talked about in review with the exception of the transparent scrollbars. I couldn't find any easy way to do that via monaco options so I just let that be for now. 

Updates are: 
1. use figma colors for all light theme attributes
2. update header to respect light theme
3. use outline buttons that support dark/light in header
4. remove manual dark override and instead update helper dark detector to find any parent level tailwind dark class
5. fixe broken border corners on all editors after the first

They now look like:
Screenshot 2024-05-15 at 2 19 37 PM
Screenshot 2024-05-15 at 2 19 57 PM

ui/packages/components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -133,15 +196,6 @@ export function CodeBlock({ header, tabs, actions = [] }: CodeBlockProps) {
     setActiveTab(index);
   };
 
-  function handleEditorDidMount(editor: MonacoEditorType) {
```
jacobheric, line 136: This was the cause of the broken borders as it was not adding this class when there were more than one editor on the page. So instead I just removed it and applied this globally via css, see above.

ui/packages/components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -52,8 +149,10 @@ interface CodeBlockProps {
 }
 
 export function CodeBlock({ header, tabs, actions = [] }: CodeBlockProps) {
+  const [dark, setDark] = useState(isDark());
   const [activeTab, setActiveTab] = useState(0);
   const editorRef = useRef<MonacoEditorType>(null);
+  const wrapperRef = useRef<any>(null);
```
anafilipadealmeida, line 155:
```suggestion
  const wrapperRef = useRef<HTMLDivElement>(null);
```

ui/packages/components/src/utils/theme.ts
```diff
@@ -0,0 +1,12 @@
+//
+// This will futureproof us a bit if we switch to manual or system color scheme.
+// If an element ref is provided we look for a dark class in any parents.
+// This is useful for components that might have dark theme sections inside a general light theme
+export const colorScheme = (elementRef?: any): 'light' | 'dark' =>
```
anafilipadealmeida, line 5: I think both colorScheme and isDark can have a `HTMLElement` type instead of `any`

ui/packages/components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -66,62 +165,26 @@ export function CodeBlock({ header, tabs, actions = [] }: CodeBlockProps) {
   const language = tabs[activeTab]?.language ?? 'json';
   const handleChange = tabs[activeTab]?.handleChange ?? undefined;
 
+  useEffect(() => {
+    //
+    // We don't have a dom ref until we're renedered, so check for dark theme parent classes then
```
anafilipadealmeida, line 170: Nitpick: 
```suggestion
    // We don't have a DOM ref until we're rendered, so check for dark theme parent classes then
```

anafilipadealmeida (commented):
Looking good @jacobheric ! Left a few small comments 🙂

ui/packages/components/src/utils/theme.ts
jacobheric: Yup, will do!

jacobheric:
Thanks @anafilipadealmeida for the careful review, I think I've addressed all your comments! Let me know if you see anything else.  

Also, I see I'm getting an invalid dom nesting error now that I've switched to the new buttons. I'll give that a look and fix before we merge.

ui/packages/components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -309,9 +367,9 @@ export function CodeBlock({ header, tabs, actions = [] }: CodeBlockProps) {
                     className={classNames(
                       `px-5 py-2.5 text-xs`,
                       isSingleTab
-                        ? 'text-slate-400'
+                        ? 'text-slate-700 dark:text-slate-400'
                         : 'block border-b outline-none transition-all duration-150',
-                      isActive && 'border-indigo-400 text-white',
+                      isActive && 'border-indigo-400 text-indigo-500 dark:text-white',
                       !isActive && !isSingleTab && 'border-transparent text-slate-400'
```
anafilipadealmeida, line 373:
```suggestion
                      !isActive && !isSingleTab && 'border-transparent text-slate-700 dark:text-slate-400'
```

anafilipadealmeida (approved):
Thank you for all the hard work, looking great! Left a final small comment, but approving this already!

PR #1360, opened 2024-05-15
Add `runID` to `step.invoke()` when the run is scheduled

darwin67:
## Description

Annotate the `step.invoke` span with the target runID, so it can be used as reference when showing the trace.
This mainly helps with in-process runs where the invoke target might not have finished yet.

This PR also tweaks the wait virtual span to make sure it's reconstruction works as expected.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -5,12 +5,14 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/inngest/inngest/pkg/consts"
 	"github.com/oklog/ulid/v2"
+	"go.opentelemetry.io/otel/trace"
 )
 
 const (
@@ -141,6 +143,7 @@ type InngestMetadata struct {
 	SourceFnVersion     int    `json:"source_fn_v"`
 	InvokeFnID          string `json:"fn_id"`
 	InvokeCorrelationId string `json:"correlation_id,omitempty"`
+	InvokeParentSpanID  string `json:"psid,omitempty"`
 	InvokeSpanID        string `json:"sid,omitempty"`
 	InvokeTimestamp     int64  `json:"ts"`
 	InvokeExpiresAt     int64  `json:"expire"`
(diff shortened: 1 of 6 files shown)


pkg/execution/executor/executor.go
```diff
@@ -2168,24 +2218,35 @@ func (e *executor) handleGeneratorInvokeFunction(ctx context.Context, gen state.
 	opcode := gen.Op.String()
 	now := time.Now()
 
+	sid := telemetry.NewSpanID(ctx)
+	sidstr := sid.String()
+
 	// Always create an invocation event.
 	evt := event.NewInvocationEvent(event.NewInvocationEventOpts{
```
jpwilliams, line 2225: We may also need to change a caller of this when we invoke via the API.

pkg/event/event.go
```diff
@@ -222,14 +262,27 @@ func NewInvocationEvent(opts NewInvocationEventOpts) Event {
...
+		InvokeFnID:          opts.FnID,
+		InvokeCorrelationId: correlationID,
+		InvokeSpanID:        spanID,
+		InvokeTimestamp:     opts.SpanTimestamp,
+		InvokeExpiresAt:     opts.ExpiresAt,
+		InvokeGroupID:       opts.GroupID,
+		InvokeDisplayName:   opts.DisplayName,
+		SourceAppID:         opts.SourceAppID,
+		SourceFnID:          opts.SourceFnID,
+		SourceFnVersion:     opts.SourceFnVersion,
```
jpwilliams, line 285: Makes perfect sense, though it's a shame we have to push so much data into the event, as it's visible to users and has mostly a tonne of concepts they don't otherwise have access to.

jpwilliams (approved):

pkg/execution/executor/executor.go
darwin67: what do you mean?

jpwilliams: We also create invocation events in the [REST API](https://github.com/inngest/inngest/blob/1e9c9f2d0deaf2be338d21627add65f07d1502d0/pkg/api/api.go#L248-L252) and the [GraphQL API](https://github.com/inngest/inngest/blob/main/pkg/coreapi/graph/resolvers/app_mutations.go); we'd need toa dd these details to those calls too.

jpwilliams: Nevermind - all of these keys are under `_inngest` anyway, which is a solid marker for "don't use this please."

PR #1367, opened 2024-05-16
INN-3061: swap in new date picker for function replay range

jacobheric:
## Description

I swapped out the old time range pickers in favor of the new date range pickers. I didn't see a figma for the actual modal, so I made it look like the old layout except I moved the inputs to their own line since the new component is larger by default. I added support to customize the internal date button style so we can condense things back to a single row if we want.

There are a few rough edges like scrolling on the nested popover, but I wanted to get this pass up to make sure it's on the right track. I wasn't sure if we wanted to instead expand the datepicker itself to support range selection, currently it only supports a single selection.

Screenshot 2024-05-16 at 3 47 22 PM

Screenshot 2024-05-16 at 3 37 39 PM

Screenshot 2024-05-16 at 3 45 01 PM

## Motivation
Upgrade to new date input and picker components.

- [ x] Chore (refactors, upgrades, etc.)
- [ ] Docs

## Checklist
- [ x] I've linked any associated issues to this PR.


ui/packages/components/src/Modal/Modal.tsx
```diff
@@ -39,39 +39,40 @@ export function Modal({
...
-                  'dark:bg-slate-910 transform overflow-hidden rounded-lg bg-white shadow-xl transition-all'
-                )}
-              >
-                {(title || description) && <Header description={description}>{title}</Header>}
-                {children}
-                {footer && <Footer>{footer}</Footer>}
-              </Dialog.Content>
-            </motion.div>
-          </div>
+              {/* Full-screen container to center the panel */}
```
jacobheric, line 43: I moved the contents inside the overlay to get the overflow scrolling working on the y-axis. I couldn't get it working outside, not 100% sure why but it might be something with the framer motion div.

jacobheric:
I added some basic validation using the same toast convention as the other higher order components on this form.

Screenshot 2024-05-19 at 3 30 17 PM

jacobheric:
Also, I got scrolling working inside the modal and inside the date picker for smaller format screens. The modal scrolling fix is for all use cases of the modal component.

anafilipadealmeida (commented):
Just a few minor tweaks, this is looking really good! 🎉

anafilipadealmeida (approved):

PR #1382, opened 2024-05-22
Add step kind specific info to run details

goodoldneon:
## Description
Add step kind specific info to run details. In other words, the configuration and result for `step.invoke`, `step.sleep`, and `step.waitForEvent`.

For invoke:
image

- [ ] Docs

## Checklist


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/signing-key/page-old.tsx
```diff
@@ -4,12 +4,12 @@
  */
 
 import { Alert } from '@inngest/components/Alert';
+import { InlineCode } from '@inngest/components/InlineCode';
```
Comment (anafilipadealmeida) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/signing-key/page-old.tsx:7]: These are out of scope for this PR, but is the old page still used? Isn't the feature enabled for everybody already?

ui/apps/dashboard/src/components/RunDetails/RunDetails.tsx
```diff
@@ -21,6 +22,15 @@ export function RunDetails({ runID, standalone = true }: Props) {
   const cancelRun = useCancelRun({ envID: env.id, runID });
   const rerun = useRerun({ envID: env.id, envSlug: env.slug, runID });
 
+  const internalPathCreator = useMemo(() => {
```
anafilipadealmeida, line 25: Is there a way to do something similar to this instead of passing a pathCreator, as it would be more reusable for the Dev Server?
https://github.com/inngest/inngest/blob/main/ui/apps/dashboard/src/components/RunDetails/RunDetails.tsx#L61

anafilipadealmeida (commented):
Left a few comments. Something else: this Step information (Trace Info) component is not matching the designs. Should be using the same space between elements as the RunInfo one,  have a "Step information" header when it's a step?, and there's a CSS issue with the color applied on top of the card borders

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/signing-key/page-old.tsx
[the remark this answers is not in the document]
Comment (goodoldneon) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/signing-key/page-old.tsx:7]: The old page can go -- just haven't had time to think about it 😭

ui/apps/dashboard/src/components/RunDetails/RunDetails.tsx
goodoldneon: I don't think so, since we don't know the new run ID until we actually rerun it deep down in the view

ui/packages/components/src/TimelineV2/TraceInfo.tsx
```diff
@@ -1,34 +1,88 @@
...
   className?: string;
+  pathCreator: {
+    runPopout: (params: { runID: string }) => Route;
+  };
   trace: Trace;
 };
 
-export function TraceInfo({ className, trace }: Props) {
+export function TraceInfo({ className, pathCreator, trace }: Props) {
+  console.log(trace);
```
jacobheric, line 20: Should probable remove console.log

ui/packages/components/src/TimelineV2/types.ts
```diff
@@ -9,6 +9,51 @@ export type Trace = {
...
+
+type StepInfoWait = {
+  eventName: string;
+  expression: string | null;
+  timeout: string;
+  foundEventID: string | null;
+  timedOut: boolean;
+};
+
+export function isStepInfoInvoke(stepInfo: Trace['stepInfo']): stepInfo is StepInfoInvoke {
```
jacobheric, line 37: seems to me these three function could be consolidated to a single function if they took an additional param of 'triggeringEventID' | 'sleepUntil' | 'foundEventID'.

jacobheric (approved):
Looks ok to me. I added a couple non-blocking comments.

PR #1391, opened 2024-05-24
IINN-2932: rough out relative picker

jacobheric:
## Description

First pass on the Range Date picker is complete and it matches the UX pretty closely. I review it with John and he wants the absolute date time input cleaned up a bit. I'll do that before making this a proper PR.

Screenshot 2024-05-24 at 11 16 51 AM
Screenshot 2024-05-24 at 11 16 41 AM

- [ ] Docs
- [ x] New feature (non-breaking change that adds functionality)

## Checklist
- [x ] I've linked any associated issues to this PR.
- [x ] I've tested my own changes.


ui/packages/components/src/DatePicker/timeSchemas.ts
```diff
@@ -13,6 +15,20 @@ function getMinMaxHour(is24Format: boolean) {
   }
 }
 
+export const daysInMonth = (year: number, month: months) => {
```
anafilipadealmeida, line 18: I think you don't need this one, as it seems you are using the date-fns helper instead: https://date-fns.org/v3.6.0/docs/getDaysInMonth

ui/packages/components/src/DatePicker/timeSchemas.ts
```diff
@@ -47,3 +63,15 @@ export const periodSchema = z
...
+export const yearSchema = z.coerce
+  .number()
+  .int()
+  .min(1970, { message: 'Year must be a 4 digit number' })
+  .max(3000, { message: 'Year must be a 4 digit number' });
+
+export const monthSchema = z.coerce
+  .number()
+  .int()
+  .min(1, { message: 'Month must be 1 - 12' })
```
anafilipadealmeida, line 76:
```suggestion
  .min(1, { message: 'Month must be between 1 and 12' })
```
It's slightly more verbose but it's better for screen readers

ui/apps/dashboard/src/components/Forms/Input.tsx
```diff
@@ -52,7 +52,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>((props, ref) => {
         placeholder={placeholder}
         value={props.value}
         className={cn(
-          'border border-slate-300 text-sm leading-none placeholder-slate-500 shadow outline-2 outline-offset-2 outline-indigo-500 transition-all focus:outline',
+          'border border-slate-400 text-sm leading-none placeholder-slate-400 shadow outline-2 outline-offset-2 outline-indigo-500 transition-all focus:outline',
```
anafilipadealmeida, line 55: I'd not change these colors now, as this component is used next to other elements with similar border colors - and the DateInputButton is using similar border colors so they would better match. 
Screenshot 2024-05-31 at 16 15 36

ui/packages/components/src/DatePicker/Calendar.tsx
```diff
@@ -24,7 +29,7 @@ export function Calendar({ selected, onSelect }: CalendarProps) {
 
 const classNames: DayPickerDefaultProps['classNames'] = {
   caption: 'flex justify-center items-center h-6',
-  root: 'text-slate-900',
+  root: 'text-slate-900 dark:bg-white',
```
anafilipadealmeida, line 32: Did you mean `bg-white` ? I think the dark mode of this will be darker

ui/packages/components/src/DatePicker/DateTimeInput.tsx
```diff
@@ -0,0 +1,329 @@
...
+}: TimeInputProps) {
+  const daysRef = useRef<HTMLInputElement | null>(null);
+  const yearsRef = useRef<HTMLInputElement | null>(null);
+  const hoursRef = useRef<HTMLInputElement | null>(null);
+  const minutesRef = useRef<HTMLInputElement | null>(null);
+  const secondsRef = useRef<HTMLInputElement | null>(null);
+  const millisecondsRef = useRef<HTMLInputElement | null>(null);
+  const meridiemRef = useRef<HTMLInputElement | null>(null);
+
+  const [dayInput, setDayInput] = useState<string>('00');
```
anafilipadealmeida, line 53: Nit pick: If your default is a string already, you don't need to add the `<string>`

ui/packages/components/src/DatePicker/DateTimeInput.tsx
```diff
@@ -0,0 +1,329 @@
...
+  date ? formatDate(date, format) : defaultValue;
+
+export function DateTimeInput({
+  selectedDateTime,
+  onSelect,
+  is24HourFormat = false,
+  setValid,
+  valid,
+}: TimeInputProps) {
+  const daysRef = useRef<HTMLInputElement | null>(null);
```
anafilipadealmeida, line 45:
```suggestion
  const daysRef = useRef<HTMLInputElement>(null);
```
You don't need to specify the null

ui/packages/components/src/DatePicker/DateTimePIcker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [calendarDate, setCalendarDate] = useState<Date | undefined>(defaultValue);
+  const [inputDate, setInputDate] = useState<Date | undefined>(defaultValue);
+
+  useEffect(() => {
+    onChange(dateTime);
+  }, [dateTime]);
+
+  return (
+    <div>
+      <div className="mt-2 p-2 dark:bg-white">
```
anafilipadealmeida, line 32: I think you probably wanted just `bg-white`

ui/packages/components/src/DatePicker/RangePicker.tsx
```diff
@@ -0,0 +1,314 @@
...
+          </div>
+          {showAbsolute && (
+            <div className="flex w-[354px] flex-col">
+              <Tabs.Root className="flex flex-col" value={tab} onValueChange={setTab}>
+                <Tabs.List
+                  className="border-mauve6 flex shrink-0 border-b"
+                  aria-label="Manage your account"
+                >
+                  <Tabs.Trigger
+                    className="flex h-[52px] flex-1 cursor-pointer select-none items-center
```
anafilipadealmeida, line 219: Could we use h-11 instead? It's more standard, I know the design has more, but it's unnecessary vertical space that creates more need for a scroll. For some users it might be the difference between having to scroll or not, or even realizing that there's a button at the bottom of the picker

anafilipadealmeida:
A few notes on UX:
- There should be an error when I try to pick an end date that is inferior to the start date.
- I'm not sure about the behaviour of the relative time input. Perhaps something to confirm with John, because when I type something there, I don't realize I need to press enter to assume the value

ui/packages/components/src/utils/date.ts
```diff
@@ -87,3 +116,15 @@ export function toMaybeDate<T extends string | null | undefined>(value: T): Date
 
   return new Date(value);
 }
+
+export const parseDuration = (duration: string): Duration => {
+  if (!DURATION_STRING_REGEX.test(duration)) {
+    throw Error('invalid duration, must be in the form of 1s, 5m, 10h, 12d, 15w, 17M, 20y');
+  }
+  const n = duration.slice(0, duration.length - 1);
```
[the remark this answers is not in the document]
jacobheric: yup, will use more descriptive variables!

ui/packages/components/src/DatePicker/timeSchemas.ts
jacobheric: I wrote that one before I realized it existed in date-fns. I'll delete.

ui/apps/dashboard/src/components/Forms/Input.tsx
jacobheric: 👍

ui/packages/components/src/DatePicker/Calendar.tsx
jacobheric: Yeah, I was hard coding bg-white because the calendar picker would otherwise be illegible in dark mode. I'll switch to just `bg-white`.

ui/packages/components/src/DatePicker/DateTimeInput.tsx
jacobheric: 👍

ui/packages/components/src/DatePicker/RangePicker.tsx
jacobheric: 👍

jacobheric:
> A few notes on UX:
> 
> * There should be an error when I try to pick an end date that is inferior to the start date.
> * I'm not sure about the behaviour of the relative time input. Perhaps something to confirm with John, because when I type something there, I don't realize I need to press enter to assume the value

I've added start/end validation and now show an error like so if end is before start:
Screenshot 2024-06-03 at 10 06 22 AM

Also, I did have a convo with John last week about when to "commit" the relative time input. I suggested an apply button might be most explicit be he decided that onEnter was fine for now. 

I believe I've addressed all of your feedback here. Let me know if missed anything or you see anything else!

ui/packages/components/src/DatePicker/Calendar.tsx
jacobheric: done

ui/packages/components/src/DatePicker/DateTimePIcker.tsx
sylwiavargas, line 1: I was reading the code out of curiosity and just noticed that there's a capital "I" in the file name (`DateTimePIcker.tsx`). Thought I'd mention it in case it's not intentional 💜

jacobheric: 👍 good catch...thanks! fixed!

djfarrelly (approved):
Hey @jacobheric - great work! The component is fantastic! I did some QA on this, all changes are non-blocking merging this and some are more design/UX questions. It's your call how/when you want to address any of these or push the question back to design. These can all be fixed after this is merged to keep future changes/PRs small and get this out.

**issue (ux)**: The plan limits work for relative timestamps, but the absolute time range can be selected before the allowed time range. It's more complex, but we'll need to add a limit/validation for the time range to ensure it doesn't exceed the plan's limits, so a user couldn't work around their limit by using absolute time range. To replicate - use account on plan that limits to 30 days history then select a start date 60 days ago.

**issue (non-blocking)**: You can select a time in the future. For the Replay modal and most other cases, this shouldn't be possible. It's not harmful, but it's strange. It's not going to cause issues, but it feels strange being able to select the future. We will want support for future dates in the component though as our future "Bulk cancel" feature will need future timestamps.

**issue (minor, non-blocking)**: If you select a start date then select end and paste an ISO timestamp that is before the start date, it does not validate and you can apply a range with the end date _before_ the start date. If you just select the date via the day-time parts the component correctly validates.

**question (non-blocking)**: In replay modal, clicking outside of the date picker on the mask for the Replay modal, it closes the reply modal, not the date picker, my instinct felt like the date picker should close but not yet the modal. Maybe a transparent mask should be used with the date picker to have a similar UX to a modal/pop out that closes on external click?

**issue (non-blocking)**: days of week in month are not centered. Not sure if design calls for even padding on each side or numbers that are more spread out. 
Screen Shot 2024-06-05 at 8 42 42 AM

**nit-pick (non-blocking)**: I tried to insert my cursor between the A and M in AM and delete the A via backspace key to then type PM, but it doesn't work. It moves my cursor to the back of the input where I then can delete the M then the A then type PM. It's a small thing, but something I noticed and would be nice if a user could delete from between letters.

https://github.com/inngest/inngest/assets/1509457/30443591-5e07-4424-b5de-803483f56035



**question (non-blocking) - UX/design:** Scrolling the date picker feels a little weird as you lose the header. It's not blocking for this PR, but we may want to work with design to think about how the date picker works on smaller screens - does it move to left of the button instead of below it to be able to show the whole thing? might it have a more condensed view?
Screen Shot 2024-06-05 at 8 41 44 AM

**nit-pick (non-blocking)**: extra padding at bottom of right panel: 
image

**design feedback (non-blocking)**: Someone may have mentioned similar feedback in Slack, but selecting start and end feels a little disjointed with the range at the bottom left, it could be interesting to denote some sort of indicator that something is valid and has been selected for the start or end. This is a design question - again - not blocking. 
image

anafilipadealmeida (approved):

jacobheric:
Thanks for the thorough going over @djfarrelly! I've ticketed all the feedback [here](https://linear.app/inngest/issue/AC-16/relative-date-range-picker-ux-and-design-improvements). I'll merge this and then push a PR today with a bunch of the smaller ones and carve out some time for the bigger ones going forward.

PR #1402, opened 2024-05-29
Include error message in history item output

BrunoScheufler:
## Description

Currently, we do not properly enrich the history run item with errors included in the driver response. This causes step limit errors to vanish in the UI. This fix simply adds the error message to the history item output. Long-term, we should properly set and pass on the error code to be rendered nicely in the UI.

https://linear.app/inngest/issue/INN-3095/verify-that-errors-are-properly-set-when-interrupting-function-on

- [ ] Docs

## Checklist


pkg/execution/history/lifecycle.go
```diff
@@ -805,6 +805,12 @@ func applyResponse(
 		return nil
 	}
 
+	if resp.Error() != "" {
```
goodoldneon, line 808:
```suggestion
  // Only set the output to the response error string if there isn't already output. This prevents overriding errors in the user's function
	if resp.Output == nil && resp.Error() != "" {
```

tonyhb (approved):

goodoldneon (approved):

PR #1409, opened 2024-05-31
INN-3106: fix event stream pagination by using ulid as cursor

jacobheric:
## Description

Previously the dev server event stream gql fetch paginated by createdAt timestamp. This made any events beyond 40 (the ui page side) that shared the same timestamp impossible to view. 

Switched the fetch to use the ulid as a cursor instead so the user can properly page through all events.

## Motivation
User reported bug, they were not seeing events beyond 40. 

- [x ] Bug fix (non-breaking change that fixes an issue)
- [ ] Docs

## Checklist
- [x ] I've linked any associated issues to this PR.


goodoldneon:
Found a small bug where we keep fetching forever. As I scroll, it keeps loading a page of 1 with the same event in each. I realized that when I sent 20 events:
```
for i in {1..20}; do curl localhost:8288/e/foo -d '{"name": "foo"}'; done
```

And tried to scroll in the stream UI. The network tab repeatedly showed the same page size of 1 being returned

jacobheric:
> ```
> for i in {1..20}; do curl localhost:8288/e/foo -d '{"name": "foo"}'; done
> ```

Ok, I pushed the fix you found. Thanks for testing!

goodoldneon (approved):

PR #1418, opened 2024-06-04
Experiment with Turbo

djfarrelly:
## Description

Use Turbo pack and repo to run the Next.js dev server and build in production.

## Motivation
Pushing a spike from a couple weeks ago to improve the build times during dev and deployment. Also a secondary goal of reducing the memory footprint during local dev.

- [ ] Docs

## Checklist


djfarrelly:
Currently running turbo pack (`next dev --turbo`) fails as it can't seem to find the packages workspaces and components: 
image

ui/apps/dashboard/package.json
```diff
@@ -111,7 +111,7 @@
     "vercel": "32.4.1"
   },
   "engines": {
-    "node": "18.x",
+    "node": "20.x",
```
djfarrelly, line 114: Turbo requires Node 20.x which is already LTS, so I think it's good to use it.

ui/turbo.json
```diff
@@ -0,0 +1,16 @@
...
+    },
+    "type-check": {
+      "dependsOn": ["^type-check"]
+    },
+    "dev": {
+      "persistent": true,
+      "cache": false
+    }
+  }
+}
```
djfarrelly, line 16: This is added for the turbo repo work which can build and cache files more quickly across a monorepo. Mostly, I was interested in Turbo pack for local dev which seems to work alright without this file.

ui/packages/tsconfig/base.json
```diff
@@ -13,7 +13,10 @@
     "declarationMap": true,
     "esModuleInterop": true,
     "verbatimModuleSyntax": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "paths": {
+      "@inngest/components/*": ["../components/src/*"]
+    }
```
djfarrelly, line 19: This is needed for Turbo pack as per their docs. Before this change, the main `dashboard` app would be able to load the component (e.x. `Alert.tsx`) from the package (I believe due to the `package.json`'s `dependencies`), but it wouldn't be able to load the nested deps from within that component, e.g. `@inngest/components/utils/classNames` until this was added.

jacobheric (approved):

PR #1429, opened 2024-06-06
Fix parallel steps resulting in multiple finished events

goodoldneon:
## Description
Fix a bug where parallel steps caused the "function run finished" logic to be run multiple times for a single run. This happens because each parallel step has a "discovery" step, so the SDK can send multiple "the function run is done" responses back to the Executor.

This caused multiple `inngest/function.finished` (or `inngest/function.failed`) events to send. It also caused multiple runs in history

- [ ] Docs

## Checklist


pkg/execution/driver/driver.go
```diff
@@ -39,9 +39,22 @@ func MarshalV1(
...
+		return nil, fmt.Errorf("error loading state in driver marshaller: %w", err)
+	}
+
+	if len(state.Events) == 0 {
+		return nil, fmt.Errorf("no events found in state")
+	}
+	var evt map[string]any
+	if err := json.Unmarshal(state.Events[0], &evt); err != nil {
+		return nil, fmt.Errorf("error unmarshalling event in driver marshaller: %w", err)
+	}
```
jpwilliams, line 54: Can we remove this fix from the PR and use #1427 for that instead?

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -721,14 +725,20 @@ func (m mgr) Delete(ctx context.Context, i state.Identifier) error {
...
 	for _, k := range keys {
 		cmd := m.r.B().Del().Key(k).Build()
-		if err := m.r.Do(callCtx, cmd).Error(); err != nil {
-			return err
+		result := m.r.Do(callCtx, cmd)
+		if count, _ := result.ToInt64(); count > 0 {
+			performedDeletion = true
+		}
+		if err := result.Error(); err != nil {
+			return false, err
```
jpwilliams, line 737: Looks like `performedDeletion` is true if _any_ of the keys are removed. Seeing as we may have multiple executors hitting this path at the same time for the same run, each could return `1` for a subset and then still run finalization.

Could we change `performedDeletion` to be based on a single key? `m.kf.RunMetadata()` is probably a good one.

pkg/execution/state/state.go
```diff
@@ -305,7 +305,7 @@ type Mutater interface {
 	UpdateMetadata(ctx context.Context, runID ulid.ULID, md MetadataUpdate) error
 
 	// Delete removes state from the state store.
-	Delete(ctx context.Context, i Identifier) error
+	Delete(ctx context.Context, i Identifier) (bool, error)
```
jpwilliams, line 308: cc @BrunoScheufler - sharing context as it's a change to the state interfaces we'll need to account for when we upstream this.

pkg/execution/state/v2/interfaces.go
```diff
@@ -33,7 +33,7 @@ type RunService interface {
 	// Delete deletes state, metadata, and - when pauses are included - associated pauses
 	// for the run from the store.  Nothing referencing the run should exist in the state
 	// store after.
-	Delete(ctx context.Context, id ID) error
+	Delete(ctx context.Context, id ID) (bool, error)
```
jpwilliams, line 36: cc @BrunoScheufler - sharing context as it's a change to the state interfaces we'll need to account for when we upstream this.

pkg/execution/executor/validate.go
```diff
@@ -83,8 +83,10 @@ func (r *runValidator) checkStepLimit(ctx context.Context) error {
 		resp.SetError(state.ErrFunctionOverflowed)
 		resp.SetFinal()
 
-		if err := r.e.finalize(ctx, r.md, r.evts, r.f.GetSlug(), resp); err != nil {
+		if performedFinalization, err := r.e.finalize(ctx, r.md, r.evts, r.f.GetSlug(), resp); err != nil {
 			logger.From(ctx).Error().Err(err).Msg("error running finish handler")
+		} else if !performedFinalization {
+			return nil
```
jpwilliams, line 89: Did we check if we also need to set `r.stopWithoutRetry = true` here?

jpwilliams (changes requested):

pkg/execution/driver/driver.go
goodoldneon: Oops, forgot to remove that

jpwilliams (approved):

tonyhb (approved):

PR #1435, opened 2024-06-07
Surface expression compilation errors for waitForEvent expressions

BrunoScheufler:
## Description

This adds error handling and visibility improvements for expresssion parsing errors, specifically when interpolating waitForEvent `if`expressions.

### Before

image

### After

image

## Motivation

Currently, whenever a user configures an invalid expression for waitForEvent, a function run will simply log an error and get stuck, without surfacing any details as to what happened to the user.

Take the following Inngest function:

```
inngest.createFunction(
  {
    id: "invalid-expression",
  },
  {
    event: "test/invalid.expression",
  },
  async ({ step, event }) => {
    await step.run("do something", async () => {
      console.log("did something", event);
    });

    await step.waitForEvent("wait until something happens", {
      event: "test/continue",
      timeout: "6h",
      if: "event.data.userId === async.data.userId",
    });

    await step.run("finalize", async () => {
      console.log("wrapping up!");
    });
  },
);
```

Did you notice the third equal sign in the expression? Errors like this can happen quickly, and right now, there's no way for a developer to know that this breaks the function. They are led to believe that something in the Inngest cloud broke, with no way to fix the issue.

- [ ] Docs

## Checklist


pkg/expressions/expressions.go
```diff
@@ -334,3 +334,24 @@ func (e *expressionEvaluator) parseAttributes(ctx context.Context) error {
 	e.attrs = attrs
 	return nil
 }
+
+type CompileError struct {
+	Err error
+}
+
+func NewCompileError(err error) *CompileError {
```
BrunoScheufler, line 342: I've introduced a custom error to

- easily check for the occurrence in executor.go
- provide the underlying expression compilation error to the user without including the intermediate error messages

pkg/execution/executor/executor.go
```diff
@@ -2556,11 +2561,7 @@ func (e *executor) handleGeneratorWaitForEvent(ctx context.Context, i *runInstan
 			"event": evt.Map(),
 		})
 		if err != nil {
-			logger.StdlibLogger(ctx).Warn(
-				"error interpolating waitForEvent expression",
-				"error", err,
-				"expression", *opts.If,
-			)
+			return fmt.Errorf("error interpolating wait for event expression: %w", err)
```
BrunoScheufler, line 2564: This is where we currently end with a stuck function run. Now, we simply return the error and handle it a few layers above.

pkg/execution/executor/executor.go
```diff
@@ -1103,8 +1103,13 @@ func (e *executor) HandleResponse(ctx context.Context, i *runInstance, resp *sta
 	if len(resp.Generator) > 0 {
 		// Handle generator responses then return.
 		if serr := e.HandleGeneratorResponse(ctx, i, resp); serr != nil {
+			var compileError *expressions.CompileError
+
 			// If this is an error compiling async expressions, fail the function.
-			if strings.Contains(serr.Error(), "error compiling expression") {
+			if errors.As(serr, &compileError) {
+				resp.Output = nil
+				resp.Err = compileError.Message()
```
BrunoScheufler, line 1111: This might not be the smoothest way to surface the error. If you have any other ideas, please let me know!

pkg/execution/executor/executor.go
```diff
@@ -1103,8 +1103,13 @@ func (e *executor) HandleResponse(ctx context.Context, i *runInstance, resp *sta
...
+
```
tonyhb, line 1112: I suspect we may need to also add the onFunctionFailed lifecycle hook just below the finalize call, too, actually - unless this is already handled when returning an error from the handleResponse call?

Without this lifecycle hook we won't be properly storing logs/traces/etc.

It's actually a little annoying that we have to call the lifecycle hooks _and_ finalize separately;  maybe in the future we can refactor the executor code to combine these.

pkg/execution/executor/executor.go
tonyhb: combined with the custom error this seems entirely reasonable

pkg/execution/executor/executor.go
BrunoScheufler: We do use the `OnFunctionFinished` one, this might be the wrong hook then? It does already handle some errors (step limit exceeded), though.

darwin67 (approved):
Looks good as is.

While non blocking, we currently have some implicit expectation for step error data structure.
It'll be great to actually define that struct in this repo, and we can translate the `CompileError` into a `StepError`.
https://github.com/inngest/inngest/blob/main/docs/SDK_SPEC.md
image

pkg/execution/execution.go
```diff
@@ -264,3 +264,32 @@ func (h HandlePauseResult) Processed() int32 {
 func (h HandlePauseResult) Handled() int32 {
 	return h[1]
 }
+
+type GracefulError struct {
```
BrunoScheufler, line 268: I would have liked the `expressions.CompileError` to implement the GracefulError, however, this would require importing packages or code that isn't immediately related, so we have to convert the "internal" errors to graceful/public/user-facing ones one layer above.

Happy to accept suggestions on whether we should refactor this into a separate module and start implementing a common interface.

pkg/execution/executor/validate.go
```diff
@@ -80,7 +80,14 @@ func (r *runValidator) checkStepLimit(ctx context.Context) error {
 	if limit > 0 && r.md.Metrics.StepCount >= limit {
 		// Create a new driver response to map as the function finished error.
 		resp := state.DriverResponse{}
-		resp.SetError(state.ErrFunctionOverflowed)
+
+		gracefulErr := state.StandardError{
+			Error:   state.ErrFunctionOverflowed.Error(),
+			Name:    "ErrFunctionOverflowed",
```
jpwilliams, line 86: Can we make this name a `const`? atm I'm starting to:
- Add these errors over here: https://github.com/inngest/inngest/blob/b32061954f142aa79c247fc26ac3cb07b225ddff/pkg/execution/execution.go#L22
- Prefix them with `Inngest` so users can easily differentiate these from their own errors (and we can more easily provide UI help)

jpwilliams (changes requested):

jpwilliams (approved):

PR #1446, opened 2024-06-12
tailwind consolidation and design tokens

jacobheric:
## Description

This separates out the work to consolidate tailwind configs and implement design tokens so it can be merge sooner than the integrations work if necessary.

## Motivation
We want to de-duplicate our tailwind configs and establish out design token pattern so we can start using them.

- [x ] Chore (refactors, upgrades, etc.)
- [ ] Docs

## Checklist


ui/packages/components/tailwind.config.ts
```diff
@@ -15,10 +15,56 @@ export default {
...
+        tertiary: {
+          '2xSubtle': 'rgb(var(--color-tertiary-2xSubtle) / <alpha-value>)',
+          xSubtle: 'rgb(var(--color-tertiary-xSubtle) / <alpha-value>)',
+          subtle: 'rgb(var(--color-tertiary-subtle) / <alpha-value>)',
+          moderate: 'rgb(var(--color-tertiary-moderate) / <alpha-value>)',
+          intense: 'rgb(var(--color-tertiary-intense) / <alpha-value>)',
+          xIntense: 'rgb(var(--color-tertiary-xIntense) / <alpha-value>)',
+          '2xIntense': 'rgb(var(--color-tertiary-2xIntense) / <alpha-value>)',
+        },
+        foreground: {
```
anafilipadealmeida, line 36: I was thinking we could go with the foreground colors as textColor for now and change that that to colors if we really need it in the future. What do you think?

anafilipadealmeida (approved):
Approving this already, could we also extend the sharedConfig to the `dev-server-ui/tailwind.config.ts`?

ui/apps/dashboard/tailwind.config.ts
```diff
@@ -1,68 +1,9 @@
 import type { Config } from 'tailwindcss';
-import defaultTheme from 'tailwindcss/defaultTheme';
+
+import sharedConfig from '../../packages/components/tailwind.config';
```
anafilipadealmeida, line 3: Non blocking: 
```suggestion
import sharedConfig from '@inngest/components/tailwind.config';
```

ui/packages/components/tailwind.config.ts
jacobheric: I think textColor takes care of most cases, but I needed to use foreground tokens on icon fills...so I added them like this so I could do: `fill-foreground-onContrast` etc. I don't *think* textColor would work in the case...but maybe!

jacobheric:
> sharedConfig

Oh yeah I forgot to do that. I can test it now.

jacobheric:
> Approving this already, could we also extend the sharedConfig to the `dev-server-ui/tailwind.config.ts`?

Yup, done. 

Oh, I noticed that the change to @inngest import of the shared config in tailwind didn't actually work. I can look into that more when I get a chance. I switched it back to relative import for now.

ui/packages/components/tailwind.config.ts
anafilipadealmeida: Then we could probably do this with [fill](https://tailwindcss.com/docs/theme#configuration-reference), the same way we do it with textColor!

PR #1459, opened 2024-06-14
fix rate limit for small limits + add rate-limit test for limits over 1

### LINKED ISSUE #1460: Rate-Limit doesn't work for small rate-limits
**Describe the bug**
For rate-limits that are small, like 5req/s per 24h, the rate-limit doesn't work correct. It let's through only one request and then blocks for 24h.

**To Reproduce**
I created a PR here to test the behavior:
https://github.com/tclass/inngest-js/pull/1

to test this, have function with a low (smaller then 10) limit per 24h and try out the rate-limit, it will send the first call through, but will block then for 24h

**Expected behavior**
The rate-limit doesn't work correct, it should let through at least 5 req in 24h

**System info**
- inngest-cli version: 0.29.1
- OS: macOS

tclass:
Fixes https://github.com/inngest/inngest/issues/1460

## Description

If we have a small limit for the rate-limiter (something under 10), the maxBurst ends up being 0, which only let's through one request. The maxBurst should probably be at least 1, so for small limits we let through at least one request.

I also added another test that actually tests the rate-limit, the one that is there, doesn't hit the bug.

## Motivation
I had a limit of 5 per 24h and it didn't work, ended up digging through  the code and found that it does:
limit / 10, which in this case means 5 / 10 = 0,5 which in integer results in a 0.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/ratelimit/ratelimit.go
+++ b/pkg/execution/ratelimit/ratelimit.go
@@ -66,7 +66,7 @@ func rateLimit(ctx context.Context, store throttled.GCRAStoreCtx, key string, c 
 
 	quota := throttled.RateQuota{
 		MaxRate:  throttled.PerDuration(int(c.Limit), dur),
-		MaxBurst: int(c.Limit) / 10,
+		MaxBurst: max(1, int(c.Limit)/10),
 	}
 
 	limiter, err := throttled.NewGCRARateLimiterCtx(store, quota)
--- a/pkg/execution/ratelimit/ratelimit.go
+++ b/pkg/execution/ratelimit/ratelimit.go
@@ -54,9 +54,9 @@ func hash(res any, id uuid.UUID) string {
 
 // RateLimit checks the given key against the specified rate limit, returning true if limited.
 //
-// This allows bursts of up to 1/10th the given rate limit, by default.
+// This allows bursts of up to max(1,x/10) the given rate limit, by default.
 //
-// Tihs returns the duration until the next request will be permitted, or -1 if the rate limit
+// This returns the duration until the next request will be permitted, or -1 if the rate limit
 // has not been exceeded.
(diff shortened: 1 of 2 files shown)


goodoldneon:
It sounds like you'd like to directly configure the burst. Is that correct? Changing the default burst could be problematic since it's a breaking change to our rate limit behavior

tclass:
@goodoldneon I think the way it is implemented right now is broken for small limits. If we have a limit under 10, we will set a burst of 0, which won't let any other request through. We at least need a maxBurst of 1 to be able to handle a request. The problem is just not happening, with limits that are over 10 because then the maxBurst is big enough.

I also looked through the `throttled` library, which states it here: ```In fact a MaxBurst of zero will *never* permit a request with a quantity greater than one because it will immediately exceed the limit.```
https://github.com/tback/throttled/blob/9791826f034a62d4fdb9cdf44f62caaa7486c443/rate.go#L97

I also have this PR here, to test it out: https://github.com/tclass/inngest-js/pull/1

tclass:
@goodoldneon I would like to find a solution somehow for this. Just to get an understanding of how you understand it currently. When I have a rate limit of 5 for 1day, shouldn't it be possible to call the function 5x a day? Could you test that on your side? Do understand it wrong?

goodoldneon:
@tclass, sorry for taking a while to respond. This seems to be expected behavior since we enforce rate limiting with GCRA, which works by dividing the rate limit amount into buckets:
https://www.inngest.com/docs/guides/rate-limiting#how-rate-limiting-works

GCRA is popular rate limiting algorithm because it's much more performant than fixed window or sliding window. However, GCRA is admittedly less intuitive. Think of GCRA as a way to performantly enforce "you will not exceed N requests per T time period", but it _doesn't_ guarantee that it'll allow all N requests per T time period.

We'll work on exposing the max burst in our SDKs

PR #1472, opened 2024-06-18
Add GQL API for retrieving span output and trigger

darwin67:
## Description

Add APIs for 
1. retrieving the run span outputs, for both success and error.
2. retrieving the run's trigger

Add more integration tests to verify behavior.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/run/trace.go
+++ b/pkg/run/trace.go
@@ -235,7 +235,7 @@ func (tb *runTree) constructSpan(ctx context.Context, s *cqrs.Span) (*rpbv2.RunS
 		SpanId:       s.SpanID,
 		Name:         name,
 		Status:       rpbv2.SpanStatus_RUNNING,
-		QueuedAt:     timestamppb.New(ulid.Time(runID.Time())),
+		QueuedAt:     timestamppb.New(ulid.Time(runID.Time())), // TODO: need to be different values for steps
 		StartedAt:    timestamppb.New(s.Timestamp),
 		EndedAt:      timestamppb.New(endedAt),
 		DurationMs:   dur,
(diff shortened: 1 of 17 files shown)


pkg/run/trace.go
```diff
@@ -235,7 +241,7 @@ func (tb *runTree) constructSpan(ctx context.Context, s *cqrs.Span) (*rpbv2.RunS
 		SpanId:       s.SpanID,
 		Name:         name,
 		Status:       rpbv2.SpanStatus_RUNNING,
-		QueuedAt:     timestamppb.New(ulid.Time(runID.Time())),
+		QueuedAt:     timestamppb.New(ulid.Time(runID.Time())), // TODO: need to be different values for steps
```
jpwilliams, line 244: Do we have any other timestamp for this? A change in the future may be to allow run IDs to be created _before_ being scheduled, which might skew the `QueuedAt` time.

pkg/run/trace.go
```diff
@@ -489,27 +546,28 @@ func (tb *runTree) processInvoke(ctx context.Context, span *cqrs.Span, mod *rpbv
...
 	fnID, ok := invoke.SpanAttributes[consts.OtelSysStepInvokeTargetFnID]
 	if !ok {
-		return fmt.Errorf("missing invoke target function ID for invoke")
+		return fmt.Errorf("missing target function ID")
 	}
 
 	// run ID
-	if str, ok := invoke.SpanAttributes[consts.OtelSysStepInvokeRunID]; ok {
+	if str, ok := invoke.SpanAttributes[consts.OtelSysStepInvokeRunID]; ok && str != "" {
+		fmt.Println("RunID: ", str)
```
jpwilliams, line 560: Dangling log?

```suggestion
```

jpwilliams (approved):

pkg/run/trace.go
darwin67: good catch

pkg/run/trace.go
darwin67: This is fine for function spans but not the execution spans, which is what the comment is referring to.
I'm gonna annotate the queue item's timestamp for queued at since that's what it is for executions.

darwin67: seems like we don't have that information atm, so I'm just going to calculate the queue time from the delay.

PR #1477, opened 2024-06-18
Run tree tweaks

darwin67:
## Description

Update run tree logic with individual step processing and group processing based on context.

Group processing works well when a group needs to be created due to retries, and we need to make sure we construct the group accordingly.
Due to function level retries being grouped into the next step execution, all steps can potentially have multiple attempts.

Whereas there are times it's better to process the step itself.
These can be the happy path for most execution types, and also parallelism as well, where the discovery of steps could result in multiple attempts of execution.

Added more integration tests to verify scenarios.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -889,7 +889,7 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 
 	ctx, span := telemetry.NewSpan(ctx,
 		telemetry.WithScope(consts.OtelScopeExecution),
-		telemetry.WithName("execute"),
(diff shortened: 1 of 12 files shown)


pkg/execution/executor/executor.go
```diff
@@ -887,9 +887,9 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		))
 	}
 
-	ctx, span := telemetry.NewSpan(ctx,
+	_, span := telemetry.NewSpan(ctx,
```
darwin67, line 890: not passing down the execution span context anymore, so all spans will be a flat structure under the function span.

jpwilliams (approved):

PR #1511, opened 2024-06-26
Sharding

BrunoScheufler:
- [ ] Docs

## Checklist


pkg/devserver/devserver.go
```diff
@@ -93,33 +93,35 @@ func start(ctx context.Context, opts StartOpts) error {
...
+		redis_state.WithShardedClient(shardedClient),
+		redis_state.WithUnshardedClient(unshardedClient),
 	)
 	if err != nil {
 		return err
 	}
 	smv2 := redis_state.MustRunServiceV2(sm)
 
-	queueKG := &redis_state.DefaultQueueKeyGenerator{
-		Prefix: "{queue}",
```
BrunoScheufler, line 116: Instead of defining a global fallback and specific values for key prefixes, we reuse constants and require them at the time of client creation.

pkg/devserver/devserver.go
```diff
@@ -182,12 +184,12 @@ func start(ctx context.Context, opts StartOpts) error {
 			backoff.GetLinearBackoffFunc(time.Duration(opts.RetryInterval)*time.Second),
 		))
 	}
-	queue := redis_state.NewQueue(rc, queueOpts...)
+	queue := redis_state.NewQueue(unshardedClient.Queue(), queueOpts...)
```
BrunoScheufler, line 187: Each "feature" now receives its own client, including a Redis connection and key generator.

pkg/execution/state/redis_state/client.go
```diff
@@ -0,0 +1,193 @@
...
+
+import (
+	"github.com/oklog/ulid/v2"
+	"github.com/redis/rueidis"
+)
+
+const StateDefaultKey = "estate"
+const QueueDefaultKey = "queue"
+
+type FunctionRunStateClient struct {
```
BrunoScheufler, line 11: Clients and keys are colocated, so that users are forced into using only keys that work with a specific client.

pkg/execution/state/redis_state/client.go
```diff
@@ -0,0 +1,193 @@
...
+	kg            RunStateKeyGenerator
+	client        RetriableClient
+	unshardedConn RetriableClient
+}
+
+func (f *FunctionRunStateClient) KeyGenerator() RunStateKeyGenerator {
+	return f.kg
+}
+
+func (f *FunctionRunStateClient) Client(runID ulid.ULID) RetriableClient {
```
BrunoScheufler, line 21: Sharded clients dynamically decide which instance to use based on a switchover key. For function run state, we use the Run ID to place runs after a specific time and date on a new instance.

pkg/execution/state/redis_state/cluster.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"context"
+	"crypto/sha1"
+	"encoding/hex"
+	"github.com/redis/rueidis"
+	"runtime"
+	"time"
+)
+
+type RetriableClient interface {
```
BrunoScheufler, line 12: For Redis clustering, we need to add retry handling to our Redis client, as resizing and scaling operations may cause transient `CLUSTERDOWN` errors, which are not handled by Rueidis.

pkg/execution/state/redis_state/cluster.go
```diff
@@ -0,0 +1,107 @@
...
+func (r retryClusterDownClient) Do(ctx context.Context, cmd func(client rueidis.Client) rueidis.Completed) (resp rueidis.RedisResult) {
+	return r.do(ctx, cmd, 0)
+}
+
+func newRetryClusterDownClient(r rueidis.Client) RetriableClient {
+	return &retryClusterDownClient{r: r}
+}
+
+// NewClusterLuaScript creates a Lua instance whose Lua.Exec uses EVALSHA and EVAL.
+func NewClusterLuaScript(script string) *RetriableLua {
```
BrunoScheufler, line 69: We also need to introduce a forked version of Rueidis Lua scripts that works with our retriable client.

pkg/execution/state/redis_state/cluster.go
```diff
@@ -0,0 +1,107 @@
...
+type retryClusterDownClient struct {
+	r rueidis.Client
+}
+
+func (r retryClusterDownClient) B() rueidis.Builder {
+	return r.r.B()
+}
+
+func (r retryClusterDownClient) do(ctx context.Context, cmd func(client rueidis.Client) rueidis.Completed, attempts int) rueidis.RedisResult {
+	resp := r.r.Do(ctx, cmd(r.r))
```
BrunoScheufler, line 41: Unfortunately, Rueidis cannot reuse commands unless Pinned (in which case the commands are not recycled properly). For this reason, we needed to change the signature of cmd to expect a function which is executed for each attempt.

pkg/execution/state/redis_state/key_generator.go
```diff
@@ -3,24 +3,22 @@ package redis_state
...
 	osqueue "github.com/inngest/inngest/pkg/execution/queue"
 	"github.com/inngest/inngest/pkg/execution/state"
 	"github.com/oklog/ulid/v2"
 )
 
-var (
-	defaultQueueKey = DefaultQueueKeyGenerator{}
-)
+// TODO Set this properly, don't start sharding any time soon before we are sure everything works
+var switchover = time.Date(2100, 07, 10, 0, 0, 0, 0, time.UTC)
```
BrunoScheufler, line 15: This is set to a fixed future value for everyone right now. When moving this into production, we should determine the value for specific user groups.

pkg/execution/state/redis_state/key_generator.go
```diff
@@ -3,24 +3,22 @@ package redis_state
...
-// as the key for data stored in redis for workflows, events, actions, and
-// errors.
-type KeyGenerator interface {
-	// Workflow returns the key for the current workflow ID and version.
-	Workflow(ctx context.Context, workflowID uuid.UUID, version int) string
+func IsSharded(runID ulid.ULID) bool {
+	return ulid.Time(runID.Time()).After(switchover)
+}
 
+type RunStateKeyGenerator interface {
```
BrunoScheufler, line 21: Each feature now receives its own key generator.

pkg/execution/state/redis_state/key_generator.go
```diff
@@ -44,119 +42,105 @@ type KeyGenerator interface {
...
-type DefaultKeyFunc struct {
-	Prefix string
+type runStateKeyGenerator struct {
+	stateDefaultKey string
 }
 
-func (d DefaultKeyFunc) Idempotency(ctx context.Context, id state.Identifier) string {
-	return fmt.Sprintf("%s:key:%s", d.Prefix, id.IdempotencyKey())
+func (s runStateKeyGenerator) Prefix(defaultPrefix string, runID ulid.ULID) string {
+	if IsSharded(runID) {
```
BrunoScheufler, line 57: Some clients and their key generators support sharding, so the keys will include a dynamic hash tag as well. This way, we will assign keys to cluster slots uniformly.

pkg/execution/state/redis_state/lua/consumePause.lua
```diff
@@ -4,66 +4,26 @@ Consumes a pause.
...
-		-- Add an index of when the pause was added.
-		redis.call("ZREM", keyPauseAddIdx, pauseID)
-		-- Add an index of when the pause expires.  This lets us manually
-		-- garbage collect expired pauses from the HSET below.
-		redis.call("ZREM", keyPauseExpIdx, pauseID)
-	end
-	return 1
-end
-
-redis.call("DEL", pauseKey)
```
BrunoScheufler, line 43: Since pauses are unsharded, we cannot delete pauses while updating function run state in the same script, as we can only access one Redis cluster per script. We have moved the pause cleanup logic into a separate step. While we lose atomicity, deleting a pause is idempotent, so this does not introduce any new issues.

pkg/execution/state/redis_state/lua/consumePause.lua
```diff
@@ -4,66 +4,26 @@ Consumes a pause.
...
 if actionKey ~= nil and pauseDataKey ~= "" then
-	redis.call("RPUSH", stackKey, pauseDataKey)
-	redis.call("HSET", actionKey, pauseDataKey, pauseDataVal)
-	redis.call("HINCRBY", keyMetadata, "step_count", 1)
-	redis.call("HINCRBY", keyMetadata, "state_size", #pauseDataVal)
-end
-
-if invokeCorrelationId ~= false and invokeCorrelationId ~= "" and invokeCorrelationId ~= nil then
-	redis.call("HDEL", pauseInvokeKey, invokeCorrelationId)
+  -- idempotency check: only ever consume a pause once
```
BrunoScheufler, line 18: Updating function state is not idempotent, so we have to introduce a check to prevent running this logic twice for the same pauseDataKey.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -30,8 +30,9 @@ var embedded embed.FS
 
 var (
 	// scripts stores all embedded lua scripts on initialization
-	scripts = map[string]*rueidis.Lua{}
-	include = regexp.MustCompile(`-- \$include\(([\w.]+)\)`)
+	scripts          = map[string]*rueidis.Lua{}
+	retriableScripts = map[string]*RetriableLua{}
```
BrunoScheufler, line 34: In addition to our regular scripts, we also prepare a map of retriable scripts (i.e. scripts that can be called with a retriable client).

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -152,85 +163,62 @@ type Opt func(r *mgr)
 // By default, this connects to a local Redis server.  Use WithConnectOpts to
 // change how we connect to Redis.
 func New(ctx context.Context, opts ...Opt) (state.Manager, error) {
-	m := &mgr{
-		kf: DefaultKeyFunc{},
-	}
+	m := &mgr{}
```
BrunoScheufler, line 166: The top-level manager will delegate requests to a sharded or unsharded manager, which in turn supports different features (pauses, queue, etc.). By splitting the manager, we ensure that users cannot easily invoke cross-cluster operations or use keys on the wrong client.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -277,14 +267,14 @@ func (m mgr) New(ctx context.Context, input state.Input) (state.State, error) {
 		return nil, err
 	}
 
-	status, err := scripts["new"].Exec(
+	status, err := retriableScripts["new"].Exec(
```
BrunoScheufler, line 270: This function shows how we supply sharding-enabled keys to a retriable script on a cluster client. Since New() is placed on a sharded manager, we could not access unsharded keys or clients in here.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -339,34 +332,45 @@ func (m mgr) UpdateMetadata(ctx context.Context, runID ulid.ULID, md state.Metad
...
 	return !bytes.Equal(val, []byte("0")), nil
 }
 
-func (m mgr) Exists(ctx context.Context, runID ulid.ULID) (bool, error) {
-	cmd := m.r.B().Exists().Key(m.kf.RunMetadata(ctx, runID)).Build()
-	return m.r.Do(ctx, cmd).AsBool()
+func (m shardedMgr) Exists(ctx context.Context, runID ulid.ULID) (bool, error) {
+	fnRunState := m.s.FunctionRunState()
+	r := fnRunState.Client(runID)
+	return r.Do(ctx, func(client rueidis.Client) rueidis.Completed {
```
BrunoScheufler, line 350: Unfortunately, we needed to update all occurrences of Do() calls with the new retriable syntax. The actual commands did not change other than using the new key generators which may return sharded keys if the Run ID is newer than the switchover date.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -322,11 +312,14 @@ func (m mgr) UpdateMetadata(ctx context.Context, runID ulid.ULID, md state.Metad
...
 		input[2] = strconv.FormatInt(md.StartedAt.UnixMilli(), 10)
 	}
-	status, err := scripts["updateMetadata"].Exec(
+
+	fnRunState := m.s.FunctionRunState()
+
+	status, err := retriableScripts["updateMetadata"].Exec(
 		ctx,
-		m.r,
+		fnRunState.Client(runID),
```
BrunoScheufler, line 320: Not only do we construct old or new keys based on the Run ID, we also pick the sharded (feature-specific) or unsharded client to execute the operation on.

tonyhb (approved):

PR #1528, opened 2024-06-28
INN-3247 Add Select with search component

anafilipadealmeida:
## Description
- Add Select with search

Screenshot 2024-07-05 at 00 01 11
Screenshot 2024-07-05 at 00 00 40

## Motivation
We need the search inside a combobox to have apps and function filters in runs. This footer will include the ability to apply changes as well in the future

- [ ] Docs

## Checklist


ui/packages/components/src/Select/Select.tsx
```diff
@@ -44,12 +45,14 @@ export function Select({
     <Listbox value={defaultValue} onChange={onChange} multiple={multiple}>
```
goodoldneon, line 45: Is it correct to set `value` to `defaultValue`? When I'm using a component, I think the `defaultValue` prop will initially set the value and then from there the component is in charge of internally managing state. Should we rename our `defaultValue` prop to `value` to reflect that?

jacobheric (approved):
It would be nice to exp

PR #1536, opened 2024-07-02
Shard batch keys

BrunoScheufler:
## Description

Solves https://linear.app/inngest/issue/INN-3281

> [!NOTE]
> A follow-up PR for cleanup is open at https://github.com/inngest/inngest/pull/1541 but may be merged only after this PR is fully rolled out

Currently, batches are stored in the unsharded data store. With batch keys, we want to move batches to the sharded data store to offload pressure from our main instance.

With this PR

- All new Append() calls are routed to the sharded cluster, creating a new batch (even if an existing batch is present in the old cluster)
- Existing batches will simply start execution after the maximum timeout of 60s 
- After 60s, all old batches will have been processed and only batches on the sharded cluster will remain
- Once this mark is hit, we can refactor the batching code to strip all unsharded clients/keys.
     - [ ] Remove `redisBatchManager.isSharded` and always use sharded client + true for keys
     - [ ] Refactor `BatchKeyGenerator.Batch()` and `BatchKeyGenerator.BatchMetadata()` to default to using sharding `(ctx context.Context, functionId uuid.UUID, batchId ulid.ULID) string`

### Comparing to sharding function run state

While both run state and batch data end up in the same sharded cluster, the approach to sharding is different for batching in multiple ways:

- All batches in the system elapse after 60s, so we don't need to keep the `redisBatchManager.isSharded()` code around for longer than that period. This allows us to clean up swiftly after rolling out, so we streamline the interfaces.
  - For function run state, old runs can exist for a long time until they're run to completion. We will have to use the dynamic client/key switching for at least a few months before we can remove it. If we dropped this now, old runs would not be able to resume and complete.
- We determine whether a batch is sharded during creation and then pass the batch pointer key around. This serves as an identifier to determine whether we need to use the old client and Batch()/BatchMetadata() keys for the transition phase. We essentially default to using sharding for new batches in Append(), so there's no granular rollout here.
   - Function run state is access in many places, where we only get a run ID (embedding a certain timestamp) and the account ID. We have to determine whether the account at hand is enrolled to sharding _and_ whether the run was created past the specific switchover time, only then can we use the sharded client and keys. We do not have a stateful token like the batchPointerKey that tells us where we can find the data.
- Batches exist before a run is scheduled, so we cannot shard on the runId. That's why I'm proposing to shard batches by function ID to get a better distribution than simply sharding on the account or environment level.
   - Function run state is contained entirely within the scope of a runId, so we can naturally shard on that key.

### Considerations

- Batching was initially a loose part of the queue (at least logically). This will be the first feature to depart from the strict separation between unsharded queue and sharded run state, entering a hybrid model based on each feature. I believe that's the way to go, but I'm open to feedback.
- Do we want to store batches on the function run state instance? This is probably inconsistent in terms of naming, but otherwise fine

- [ ] Docs

## Checklist


pkg/execution/batch/redis.go
```diff
@@ -90,6 +91,8 @@ func (b redisBatchManager) Append(ctx context.Context, bi BatchItem, fn inngest.
 		return nil, fmt.Errorf("no batch config found for for function: %s", fn.Slug)
 	}
 
+	client := b.b.ShardedClient()
```
BrunoScheufler, line 94: Append is always sharded from now on. This means that new batches will be created on the sharded cluster. Previous batches will simply run into the configured or max timeout (60s) and execute. This way we don't need to specify a rollout policy, while still handling all batches in the system. After a minute past rolling out, all existing batches will have been processed and all new batches are sharded.

pkg/execution/batch/redis.go
```diff
@@ -133,14 +136,28 @@ func (b redisBatchManager) Append(ctx context.Context, bi BatchItem, fn inngest.
...
+	if isSharded {
+		client = b.b.ShardedClient()
+	}
+
+	return client, isSharded
+}
+
 // RetrieveItems retrieve the data associated with the specified batch.
-func (b redisBatchManager) RetrieveItems(ctx context.Context, batchID ulid.ULID) ([]BatchItem, error) {
+func (b redisBatchManager) RetrieveItems(ctx context.Context, functionId uuid.UUID, batchID ulid.ULID, batchPointer string) ([]BatchItem, error) {
```
BrunoScheufler, line 152: Retrieving can happen immediately after `Append` when the batch is full, or after the batch timeout has passed and execution is started. We need to check whether we're dealing with an old (unsharded) batch or a new (sharded) batch. To find out, we can simply compare the prefix with the sharded queue prefix.

pkg/execution/batch/batch.go
```diff
@@ -31,10 +31,10 @@ import (
 //	#4 needs to happen in one transaction in order to make sure there will not be any race conditions.
 type BatchManager interface {
 	Append(ctx context.Context, bi BatchItem, fn inngest.Function) (*BatchAppendResult, error)
-	RetrieveItems(ctx context.Context, batchID ulid.ULID) ([]BatchItem, error)
-	StartExecution(ctx context.Context, batchID ulid.ULID, batchPointer string) (string, error)
+	RetrieveItems(ctx context.Context, functionId uuid.UUID, batchID ulid.ULID, batchPointer string) ([]BatchItem, error)
```
BrunoScheufler, line 34: We need to include batch pointers for most operations as they will tell us whether the batch, at the time of appending, was sharded or now.

pkg/execution/batch/redis.go
```diff
@@ -133,14 +136,28 @@ func (b redisBatchManager) Append(ctx context.Context, bi BatchItem, fn inngest.
 	return result, nil
 }
 
+func (b redisBatchManager) isSharded(ctx context.Context, batchPointer string, functionId uuid.UUID) (redis_state.RetriableClient, bool) {
```
BrunoScheufler, line 139: We can remove this function 60s after rolling out, when all batches are sharded.

pkg/execution/state/redis_state/key_generator.go
```diff
@@ -288,47 +288,54 @@ func (u queueKeyGenerator) ConcurrencyFnEWMA(fnID uuid.UUID) string {
...
-	BatchPointer(context.Context, uuid.UUID) string
+	BatchPointer(ctx context.Context, functionId uuid.UUID) string
 	// BatchPointerWithKey returns the key used as the pointer reference to the
 	// actual batch for a given batchKey
-	BatchPointerWithKey(context.Context, uuid.UUID, string) string
+	BatchPointerWithKey(ctx context.Context, functionId uuid.UUID, key string) string
 	// Batch returns the key used to store the specific batch of
 	// events, that is used to trigger a function run
-	Batch(context.Context, ulid.ULID) string
+	Batch(ctx context.Context, isSharded bool, functionId uuid.UUID, batchId ulid.ULID) string
```
BrunoScheufler, line 303: Batch and BatchMetadata still require `isSharded` to be passed in to support older batches. This requirement drops after 60s of releasing this change, as all old batches will have been processed.

pkg/execution/execution.go
```diff
@@ -109,13 +109,9 @@ type Executor interface {
 	// InvokeFailHandler invokes the invoke fail handler.
 	InvokeFailHandler(context.Context, InvokeFailHandlerOpts) error
 
-	AppendAndScheduleBatchWithOpts(ctx context.Context, fn inngest.Function, bi batch.BatchItem, opts *BatchExecOpts) error
-	// deprecated; use AppendAndScheduleBatchWithOpts in new code
-	AppendAndScheduleBatch(ctx context.Context, fn inngest.Function, bi batch.BatchItem) error
+	AppendAndScheduleBatch(ctx context.Context, fn inngest.Function, bi batch.BatchItem, opts *BatchExecOpts) error
```
BrunoScheufler, line 112: This is just cleanup, as we don't use the old methods anymore.

jpwilliams (approved):

PR #1540, opened 2024-07-03
Fix internal metadata not disallowed for user events

goodoldneon:
## Description
Disallow internal metadata (`event.data._inngest`) for user events

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


tonyhb (approved):

pkg/api/api.go
```diff
@@ -163,6 +164,14 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 				return err
 			}
 
+			if _, ok := evt.Data["_inngest"]; ok {
+				// User event data must not have internal metadata since it can
+				// cause issues. For example, if an invoked function's event
+				// data is forwarded into a new event then it may accidentally
+				// fulfill the invocation
+				return errors.New("event data must not contain internal metadata (the _inngest key)")
+			}
```
jpwilliams, line 173: Should we just remove `_inngest` instead? Throwing an error could be frustrating if I just want to spread one event into another, e.g.

```ts
async ({ event, step }) => {
  await step.sendEvent("notify", {
    name: "app/some.event",
    data: {
      ...event.data,
      anotherProperty: "foo",
    },
  });
}
```

goodoldneon: The downside of that is users intentionally adding their own `_inngest` for whatever reason (e.g. tracking Inngest-related stuff for their own purposes) will silently lose that info. Dunno if that's too niche to worry about

jpwilliams: @goodoldneon Aye, fair shout. I think the likelihood of that is much lower than someone copying data from one event to another; I believe it's a fair trade-off to silently strip it.

jpwilliams (approved):

PR #1557, opened 2024-07-10
Fix missing transform when creating a webhook

goodoldneon:
## Description
Fix missing transform when creating a webhook. It's probably better to default the transform in the backend, but we need a fix now

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/CreateKeyButton.tsx
+++ b/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/CreateKeyButton.tsx
@@ -12,6 +12,7 @@ import { useMutation } from 'urql';
 import { useEnvironment } from '@/app/(organization-active)/(dashboard)/env/[environmentSlug]/environment-context';
 import Input from '@/components/Forms/Input';
 import { graphql } from '@/gql';
+import { defaultTransform } from './[keyID]/TransformEvent';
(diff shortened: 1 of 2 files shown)


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/CreateKeyButton.tsx
```diff
@@ -43,7 +44,11 @@ export default function CreateKeyButton() {
           name: inputValue,
           source: currentContent.type,
           metadata: {
-            transform: undefined,
+            // We must specify a transform, otherwise the webhook will be in a
+            // broken state due to a missing transform. It might be better to
+            // specify the default transform in the backend, but this is a quick
+            // fix
+            transform: defaultTransform,
```
Comment (djfarrelly) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/manage/[ingestKeys]/CreateKeyButton.tsx:51]: We should only set this when `currentContent.type` is `"webhook"`. We can't set this for event keys as within the Event API, the transform will execute even if it's an event key and not a webhook, but the user would never see this in the UI and it would lead to bugs: https://github.com/inngest/monorepo/blob/0fbc39c6e27d7fff41b7a8532df4495b2e7a0608/pkg/applogic/events/ingestapi/ingestapi.go#L181-L183

Additionally, to prevent a similar change like this from accidentally causing a but, we should probably also land a fix in monorepo to only execute a transform for webhook `source` keys only, not event keys.

djfarrelly (changes requested):
There is a required change before shipping as this will create issues due to the way the event api currently chooses to execute transforms.

djfarrelly (approved):

PR #1575, opened 2024-07-17
fix fonts in storybook

jacobheric:
## Description
Remove old inline style font import from storybook and add the new cdn font css import.

## Motivation
Make storybook work.

- [ x] Bug fix (non-breaking change that fixes an issue)
- [ ] Docs

## Checklist


ui/packages/components/src/AppRoot/AppRoot.tsx
```diff
@@ -15,7 +13,7 @@ export function AppRoot({
   return (
     <html
       lang="en"
-      className={`${devServer ? interTight.variable : ''} ${devServer ? robotoMono : ''} ${
+      className={`${devServer ? interTight.variable : ''} ${devServer ? robotoMono.variable : ''} ${
```
anafilipadealmeida, line 16: Why is the Dev Server not changing to the new font as well?

anafilipadealmeida (commented):
Can we also add the font to the Monaco editors, as it's not being applied to them?

ui/packages/components/src/AppRoot/AppRoot.tsx
jacobheric: Because there was push back that we shouldn't be doing a new font at all....because the change is disruptive....I put it in in the least disruptive was possible for me to build the new dashboard IA with it. 

As I said in the original PR, I'm happy to add it to the dev server as well if people are ok with the disruption. IMHO it looks fine there as well.

jacobheric:
I updated the code editor shared component.

ui/packages/components/src/AppRoot/AppRoot.tsx
anafilipadealmeida: The goal of the design system is to keep the consistency for users, so it builds trust in our product and the brand. By changing the font, we only achieve it if we do it in both products, otherwise we are introducing more inconsistency. Since there's no extra work in making it happen in the dev server, and we've seen it looks good, let's do it everywhere!

I'll approve this PR since there's great work here

anafilipadealmeida (approved):

ui/packages/components/src/AppRoot/AppRoot.tsx
jacobheric: Yup, will do.

PR #1590, opened 2024-07-22
Queue uses a mock-able replacement for time package

cdzombak:
## Description

This PR replaces usage of `time.Now` and related functions in `queue` and `queue_processor` by https://github.com/jonboulle/clockwork . It also replaces the (hacky and only partly-correct) `getNow` stub.

## Motivation

This ought to allow more stable and reliable tests of the queue.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/queue.go
+++ b/pkg/execution/state/redis_state/queue.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/inngest/inngest/pkg/telemetry/redis_telemetry"
 	"math"
 	"strconv"
 	"strings"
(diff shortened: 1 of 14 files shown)


cdzombak:
Interestingly, only a few tests in `queue_test` actually modified the `getNow` stub. Those have been fixed up to use the new mock; other references were simply replaced by calls to `time.Now` since that's all they were anyway.

pkg/execution/state/redis_state/queue.go
```diff
@@ -6,37 +6,38 @@ import (
...
-	"github.com/redis/rueidis"
-	"github.com/rs/zerolog"
-	"gonum.org/v1/gonum/stat/sampleuv"
-	"lukechampine.com/frand"
 )
 
-var (
-	PartitionSelectionMax int64 = 100
-	PartitionPeekMax      int64 = PartitionSelectionMax * 3
+const (
```
cdzombak, line 38: Unrelated change; these never seem to be modified (and it's unclear that they could be modified safely at runtime).

BrunoScheufler (approved):

cdzombak:
I would like very careful review from both @tonyhb and @BrunoScheufler on this PR; in particular, it's important that all usages of `time.Now()` and things that call it (like `time.Since`) get replaced by calls to `q.clock` instead.

(There is one usage of time.Now that I decided is okay for now, in the `duration` helper function. We can always figure out how to inject a clock in there later.)

cdzombak:
I also need to double check that no Lua scripts read the current time independent from Go code.

tonyhb (approved):

PR #1608, opened 2024-07-29
CEL filter for runs

darwin67:
Related: https://github.com/inngest/expr/pull/22

## Description

This change introduces using CEL syntax for filtering function runs.
The target objects are the triggers and the output of function runs.

_Example_
With function runs of output from 1-12.
Filtering by the timestamp and output > 10 should return 2 results.

image

- [ ] Docs

## Checklist


pkg/run/cel.go
```diff
@@ -0,0 +1,306 @@
+package run
```
goodoldneon, line 1: Could you write unit tests for this file? It'd be sweet to see a bunch of small "this CEL string becomes this SQL string" tests

darwin67: sure, it's probably not gonna be much since the SQL translation only really happens for some portion of the event queries, and there are non for output.

darwin67: CEL -> SQL unit tests added.
It's still pretty basic since it's rather limited in usage.

djfarrelly (commented):
Copying feedback from Slack convo - some of which may already be incorporated.

Ideally we can wrap the internal CEL errors and provide clearer or friendly errors, for example:

Or if we could somehow translate CEL errors into something less ugly, for example:
failed to compile the CEL expression: 
```
ERROR: <input>:10:5: Syntax error: mismatched input 'and' expecting <EOF>
 |     and (account.overdraftProtection
 | ....^
```
Could become:
```
Invalid query syntax:
    and (account.overdraftProtection
....^
```
Or even for more easily showing in our UI:
```
Invalid query syntax: "and (account.overdraftProtection"
```

Also, the Monaco editor in the UI probably won’t be able to do syntax highlighting well due to no CEL support, so we need to at least try to produce some good errors (probably from the API) to the user for common mistakes to help them self fix. One example might be if they use triple equals (b/c javascript) or maybe use “and” since they think it’s more SQL-like.

ui/apps/dev-server-ui/src/app/(dashboard)/runs/page.tsx
```diff
@@ -63,6 +64,7 @@ export default function Page() {
         endTime: endTime,
         status: filteredStatus,
         timeField,
+        celQuery,
```
darwin67, line 67: unclear how to update the `GetRunsDocument` to generate the correct params.
it needs to accept the `$query` param to make sure the CEL is passed over to the API.

cc @anafilipadealmeida

anafilipadealmeida: https://github.com/inngest/inngest/blob/main/ui/apps/dev-server-ui/src/coreapi.ts#L251
This one 😄

BrunoScheufler (approved):
🚢

PR #1617, opened 2024-07-31
INN-3434: Graceful pause deletion

BrunoScheufler:
## Description

This PR introduces a grace period of 10s before we delete an expired pause. This should help in reducing stuck function runs due to racing pause timeout handling. We also remove gc'ing logic in LoadEvaluablesSince as we already drop expired pauses in all consuming locations _after_ evaluation. 

https://linear.app/inngest/issue/INN-3434/fix-stuck-pauses

## Motivation

In some cases, function runs with waitForEvent will be stuck, with the only logs being "deleting expired pause". It looks like in some cases when using waitForEvent

- the related pause expires without receiving a matching event
- the timeout queue item/job has not run yet
- another matching event comes in after expiry and enters this branch, then drops the pause before the timeout job can run, leading to a stuck function run

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/redis_state.go
```diff
@@ -1238,11 +1230,6 @@ func (m unshardedMgr) LoadEvaluablesSince(ctx context.Context, workspaceID uuid.
 		}
 	}
 
-	// GC pauses on fetch.
-	for _, pause := range expired {
```
BrunoScheufler, line 1242: @tonyhb as far as I can see, we will delete expired pauses either

- when the timeout queue item is run
- when another event comes in after the grace period introduced above

I believe deleting pauses in the iterator for aggregate matching may introduce races where the timeout job hasn't run but the pause already expired.

We could also keep this if we introduced the same grace period.

pkg/execution/executor/executor.go
```diff
@@ -1451,9 +1452,15 @@ func (e *executor) handlePausesAllNaively(ctx context.Context, iter state.PauseI
 			// time to process.  We handle that here and assume that the event
 			// did not occur in time.
 			if pause.Expires.Time().Before(time.Now()) {
-				// Consume this pause to remove it entirely
-				l.Debug().Msg("deleting expired pause")
-				_ = e.pm.DeletePause(context.Background(), *pause)
+				l.Debug().Msg("encountered expired pause")
+
+				shouldDelete := pause.Expires.Time().Add(PauseExpiredDeletionGracePeriod).Before(time.Now())
```
BrunoScheufler, line 1457: This should never happen but it is possible that
- the pause expires without receiving a matching event
- the timeout job has not run yet
- another matching event comes in _after expiry_ and enters this branch, then drops the pause before the timeout job can run, leading to a stuck function run

pkg/execution/state/redis_state/redis_state.go
BrunoScheufler: We _should_ probably delete pauses from the index to prevent fetching expired pauses over and over. We should not drop the entire pause, though.

tonyhb (approved):

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -1238,11 +1230,6 @@ func (m unshardedMgr) LoadEvaluablesSince(ctx context.Context, workspaceID uuid.
...
-		_ = m.DeletePause(ctx, *pause)
```
BrunoScheufler, line 1243: TODO: Keep this in _but_ add the threshold!

PR #1618, opened 2024-07-31
Guaranteed Capacity

BrunoScheufler:
- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -680,12 +462,30 @@ func (q *queue) scan(ctx context.Context) error {
...
+		// Pick a random guaranteed capacity if we leased multiple
+		i := rand.Intn(len(existingLeases))
+		guaranteedCapacity = &existingLeases[i].GuaranteedCapacity
+
+		// Backwards-compatible metrics names
+		metricShardName = "<guaranteed-capacity>:" + guaranteedCapacity.Key()
+
+		// When account is leased, process it
+		partitionKey = q.u.kg.AccountPartitionIndex(guaranteedCapacity.AccountID)
+		return q.scanPartition(ctx, partitionKey, peekSize, peekUntil, guaranteedCapacity, metricShardName)
```
tonyhb, line 479: Here, are guaranteed capacity partitions single partitions of queue items, or are the pointer partitions which point to other partitions?  The reason I ask:

- It looks as though we call `scanPartition` on a guaranteed capacity partition.
- This implies that we have a single partition of queue items for each account.

If figured that guaranteed capacity would instead be a partition pointer, acting as an alternative to the global pointer.  In this situation, we'd peek other partitions from the guaranteed capacity partition, then peek those — allowing us to store all key queues, etc, as pointers in guaranteed capacity.

BrunoScheufler: Guaranteed Capacity is essentially a pointer to the global account partitions. This way, one or more executors instance can exclusively process an entire account's partitions configured with guaranteed capacity.

This slots neatly into our previous work on account queues, but we can also implement it in a different way, though I understood guaranteed capacity to be enabled for processing specific accounts, which we now do. Maybe that wasn't what you thought about?

PR #1625, opened 2024-08-01
Disable password manager by default in Input

goodoldneon:
## Description
In our `Input` component, disable password manager autocomplete by default. The autocomplete is really annoying when it isn't necessary.

Before:
Screenshot 2024-08-01 at 10 27 51

After:
Screenshot 2024-08-01 at 10 28 05

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/Forms/Input.tsx
+++ b/ui/packages/components/src/Forms/Input.tsx
@@ -19,7 +19,8 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
     { allowPasswordManager = false, className, type = 'text', inngestSize = 'base', ...props },
     ref
   ) => {
-    let passwordManagerProps: Record<string, boolean> = {
+    let passwordManagerProps: Record<string, unknown> = {
+      autocomplete: 'off',
       'data-1p-ignore': true,
       'data-bwignore': true,
       'data-lpignore': true,
--- a/ui/packages/components/src/Forms/Input.tsx
+++ b/ui/packages/components/src/Forms/Input.tsx
@@ -1,6 +1,7 @@
 import { forwardRef, type InputHTMLAttributes } from 'react';
 
 type InputProps = InputHTMLAttributes<HTMLInputElement> & {
+  allowPasswordManager?: boolean;
   label?: string;
   error?: string | undefined;
   inngestSize?: 'base' | 'lg';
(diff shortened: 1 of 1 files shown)


ui/packages/components/src/Forms/Input.tsx
```diff
@@ -14,7 +15,19 @@ const sizeStyles = {
...
-  ({ className, type = 'text', inngestSize = 'base', ...props }, ref) => {
+  (
+    { allowPasswordManager = false, className, type = 'text', inngestSize = 'base', ...props },
+    ref
+  ) => {
+    let passwordManagerProps: Record<string, boolean> = {
+      'data-1p-ignore': true,
+      'data-bwignore': true,
+      'data-lpignore': true,
+    };
```
djfarrelly, line 26: I think to also disable Chrome's autocomplete for password management we need to add `autocomplete="off"` as well. Can we add that?

djfarrelly (changes requested):

ui/packages/components/src/Forms/Input.tsx
```diff
@@ -14,7 +15,20 @@ const sizeStyles = {
...
-  ({ className, type = 'text', inngestSize = 'base', ...props }, ref) => {
+  (
+    { allowPasswordManager = false, className, type = 'text', inngestSize = 'base', ...props },
+    ref
+  ) => {
+    let passwordManagerProps: Record<string, unknown> = {
+      autocomplete: 'off',
+      'data-1p-ignore': true,
+      'data-bwignore': true,
+      'data-lpignore': true,
```
djfarrelly, line 26: I didn't realize, but since this is react, do these props need to be camelCase? e.g. autocomplete is really `autoComplete` in react: https://legacy.reactjs.org/docs/dom-elements.html#all-supported-html-attributes

goodoldneon: For whatever reason I usually see `data-` attributes written in kebab-case. Dunno why

djfarrelly: It seems from testing, React doesn't care when passing props like this, even for `autoComplete`. Not sure if there is a warning in the console, but I'm good with shipping this then.

djfarrelly (approved):

PR #1633, opened 2024-08-02
Fetch run trace and output in parallel

amh4r:
## Description
Fetch run trace and output in parallel.

This required fetching the run output from a resolver that results a string instead of parsed output. That's a change we needed anyway since the output parsing in the backend has some edge case bugs

- [ ] Docs

## Checklist


ui/packages/components/src/RunResult.tsx
```diff
@@ -1,46 +1,91 @@
...
       )}
 
-      {result.error && (
+      {!parsedResult.isSuccess && (
         <CodeBlock
           header={{
-            title:
-              (result.error.name || 'Error') +
-              (result.error.message ? ': ' + result.error.message : ''),
+            title: `${parsedResult.error.name}: ${parsedResult.error.message}`,
```
djfarrelly, line 34: **issue**: Only add the colon when there is a message like we did before
```suggestion
            title: `${parsedResult.error.name}${parsedResult.error.message?.length > ": ": ""}${parsedResult.error.message}`,
```

ui/packages/components/src/RunResult.tsx
```diff
@@ -1,46 +1,91 @@
...
   className?: string;
-  result: Result;
-  isSuccess?: boolean;
+  result: string;
+  isSuccess: boolean;
 };
 
 export function RunResult({ className, result, isSuccess }: Props) {
-  const prettyResult = result.data && usePrettyJson(result.data);
+  const parsedResult = useParsedResult({ isSuccess, result });
```
djfarrelly, line 15: **question**: Before this, we used `result.data` and now `useParsedResult` is using `result` - is this due to the way that the different API endpoint returns the response? Or is the string `result` include the `data` key? 

I ask this because I'm looking for it to be the same experience, and not have the successful output nested within `{ "data": { "message": "cool" }` - Confirming it should show to the user as `{ "message": "cool" }.

ui/apps/dashboard/src/components/RunDetails/useGetTraceResult.ts
```diff
@@ -1,35 +1,29 @@
...
+  query TraceResult($envID: ID!, $runID: String!) {
     workspace(id: $envID) {
-      runTraceSpanOutputByID(outputID: $traceID) {
-        data
-        error {
-          message
-          name
-          stack
-        }
+      run(runID: $runID) {
```
djfarrelly, line 10: **question**: does this endpoint also use Clickhouse as the source of truth or is this an old endpoint that uses Big Query? I want to make sure that it uses the same source of truth as the rest of the new UI

ui/packages/components/src/RunDetailsV2/RunDetails.tsx
```diff
@@ -57,13 +56,8 @@ export function RunDetails(props: Props) {
     enabled: Boolean(outputID),
```
djfarrelly, line 56: **issue**: It still looks like this query will not execute until the `outputID` is fetched so this won't run in parallel. I think this needs to change to `runID` or remove the enabled config option all together as everything else is passed by props.

djfarrelly (changes requested):
Hey @goodoldneon, I have some questions and have highlighted a couple of issues that I think might block this from actually running in parallel. Catch me on Slack if you want to discuss live!

ui/packages/components/src/RunResult.tsx
goodoldneon: We had to make this change because the normal run resolver returns unstructured output data. The old, non-parallelizable one returned structured data. However, there are edge cases the structured resolver didn't handle.

I don't see nested data when I test it:
image

ui/apps/dashboard/src/components/RunDetails/useGetTraceResult.ts
goodoldneon: It's ClickHouse. This is the same field that the run details queries:
https://github.com/inngest/inngest/blob/aaron/inn-3422-execute-run-trace-and-output-queries-in-parallel-not/ui/apps/dashboard/src/components/RunDetails/useGetRun.ts#L46

ui/packages/components/src/RunDetailsV2/RunDetails.tsx
goodoldneon: Oops, good call

ui/packages/components/src/RunResult.tsx
djfarrelly: Ok great! Just wanted to confirm the output wasn't nested as we've had regressions in the past.

PR #1667, opened 2024-08-09
NATS span exporter

darwin67:
## Description

Add a NATS span exporter that can be used to publish otel spans to NATS directly.

- Add protobuf schema for the message schema
- Move span.go into `run` pkg to resolve circular dependency (also makes more sense)

- [ ] Docs

## Checklist


pkg/pubsub/broker/nats.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func NewNATSConnector(ctx context.Context, opts NatsConnOpt) (*NatsConnector, error) {
+	host, err := os.Hostname()
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving hostname: %w", err)
+	}
+	name := fmt.Sprintf("%s-%s", opts.Name, host)
+
+	urls := "nats://127.0.0.1:4222"
```
djfarrelly, line 58: suggestion: could be good as a const

pkg/pubsub/broker/nats.go
```diff
@@ -0,0 +1,157 @@
...
+	if cons, ok := c.consumers[key]; ok {
+		return cons, nil
+	}
+
+	return nil, fmt.Errorf("no consumer named '%s' initialized for stream: '%s'", consumer, stream)
+}
+
+// Shutdown drains the connection to flush data in buffer before fully closing the
+// connection
+func (c *NatsConnector) Shutdown(ctx context.Context) error {
```
djfarrelly, line 145: **question**: Do we have to do anything to clean up the consumers? Like `DeleteConsumer`?

proto/run/v2/run.proto
```diff
@@ -49,6 +53,9 @@ enum SpanStepOp {
   INVOKE = 1;
   SLEEP = 2;
   WAIT_FOR_EVENT = 3;
+  STEP = 4;
```
djfarrelly, line 56: **question**: What is the difference between `STEP` and `RUN`?

cmd/commands/dev.go
```diff
@@ -88,10 +89,14 @@ func doDev(cmd *cobra.Command, args []string) {
...
-		ServiceName:   "devserver",
-		Type:          itrace.TracerTypeOTLPHTTP,
+		ServiceName: "devserver",
+		Type:        itrace.TracerTypeOTLPHTTP,
+		// Type:          itrace.TracerTypeNATS,
 		TraceEndpoint: "localhost:8288",
 		TraceURLPath:  "/dev/traces",
+		NATS: &exporters.NatsExporterOpts{
+			Subjects: []string{"inngest.run.spans"},
+		},
```
djfarrelly, line 99: **question**: when this is added, will it swap out to an in-memory pubsub via the gocloud package when running the dev server? Maybe I'm just not sure how this was working before with the previous tracer anyway.

djfarrelly (approved):
Code looks good, left some non-blocking questions

pkg/pubsub/broker/nats.go
darwin67: No, it shouldn't be clean up.
A consumer is a nats resource, created on the nats server side. we merely authenticate as a consumer to pull messages to work on.

proto/run/v2/run.proto
darwin67: `RUN` is the function run itself, `STEP` is the steps within the run.

darwin67: `RUN` is the function run itself, `STEP` is the steps within the run.

cmd/commands/dev.go
darwin67: nope it won't. to enable NATS, the `Type` a couple of lines above also need to be changed to `itrace.TracerTypeNATS`.
So just setting this configuration won't have any impact.

pkg/pubsub/broker/nats.go
darwin67: done

proto/run/v2/run.proto
djfarrelly: Thanks - that was misleading to me.

PR #1675, opened 2024-08-13
Continuations

tonyhb:
This PR introduces continuations as a queueing optimization for discussion.  To be looked at after the current queue changes.

- [ ] Docs

## Checklist


pkg/execution/executor/service.go
```diff
@@ -168,17 +168,21 @@ func (s *svc) getFinishHandler(ctx context.Context) (func(context.Context, sv2.I
 
 func (s *svc) Run(ctx context.Context) error {
 	logger.From(ctx).Info().Msg("subscribing to function queue")
-	return s.queue.Run(ctx, func(ctx context.Context, info queue.RunInfo, item queue.Item) error {
+	return s.queue.Run(ctx, func(ctx context.Context, info queue.RunInfo, item queue.Item) (bool, error) {
```
cdzombak, line 171: [minor] I would prefer to introduce some sort of `result` type; in a lot of queue code we return sets of bools, ints, errors, etc.. This (a) requires care on the caller's part and (b) allows room to return invalid states (what does it mean if we return (`true, error(…)` from this function?).

Introducing more structured results ([`Either3`](https://github.com/samber/mo/blob/master/either3.go#L44) for inspiration) would reduce the room for programming errors and eliminate the need to handle invalid states.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -970,6 +1030,12 @@ func (q *queue) process(ctx context.Context, p QueuePartition, qi QueueItem, s *
 			Tags:    map[string]any{"status": "completed"},
 		})
 
+		if isContinuation {
+			// Add the partition to be continued again.  Note that if we've already
+			// continued beyond the limit this is a noop.
```
cdzombak, line 1035: If this is a noop in some cases, we may consider whether to use a more complicated locking arrangement that allows calling `addContinue` without _always_ taking a lock that might be unnecessary.

BrunoScheufler (approved):

PR #1705, opened 2024-09-03
Add ability to provide a `--redis-url` for queue/state in `inngest lite`

jpwilliams:
## Description

Adds ability to provide a `--redis-url` when starting `inngest lite`, ensuring the best possible performance and safer snapshotting.

- [ ] Docs

## Checklist


cmd/commands/lite.go
```diff
@@ -24,6 +24,7 @@ func NewCmdLite() *cobra.Command {
 	cmd.Flags().String("config", "", "Path to the configuration file")
 	cmd.Flags().String("host", "", "host to run the API on")
 	cmd.Flags().StringP("port", "p", "8288", "port to run the API on")
+	cmd.Flags().String("redis-url", "", "URL to an external Redis instance to use")
```
djfarrelly, line 27: Redis' own docs refer to this as the server URI, not URL, we should align those. From the CLI's `--help`:

```
  -u <uri>           Server URI.
 ```
 
We could make this option `redis-uri` and add clearer context:

```suggestion
	cmd.Flags().String("redis-uri", "", "External Redis server URI (Optional). Uses self-contained in-memory Redis if not provided.")
```

pkg/lite/lite.go
```diff
@@ -284,12 +285,21 @@ func start(ctx context.Context, opts StartOpts) error {
 	)
 
 	// The devserver embeds the event API.
-	persistenceInterval := consts.LiteDefaultPersistenceInterval
+	pi := consts.LiteDefaultPersistenceInterval
+	persistenceInterval := &pi
+	if opts.RedisURL != "" {
```
djfarrelly, line 290: **thought**: Instead of checking options to detect if it should be backed up, I think a better pattern would be to check the actual configuration of what is running, like `shardedRc.isInMemory`. I know we're passing the raw client, so that may be non-trivial, but this is a suggestion to check for the current set up explicitly rather than infer the current state from a CLI option and the internal logic of another function (`createRedisConnection`).

djfarrelly (changes requested):

pkg/lite/lite.go
jpwilliams: Agree - I'll leave for when we change to `start` and other shapes, as lots can be more explicit then.

djfarrelly (approved):

PR #1719, opened 2024-09-09
INN-3565: Always enqueue to default function partition

BrunoScheufler:
## Description

image

This PR refactors `ItemPartitions` to always return the function partition when otherwise enqueuing to key queues exclusively. This ensures backwards compatibility: When rolling back the key queues change, the default function partitions will include all queue items.

This change also breaks strict FIFO:
- The same function can now be processed by multiple executor instances: Executor A can lease the default function partition while Executor B can lease the concurrency key queue
- When multiple jobs are available, both workers will peek and start processing queue items for the same function in no particular order, concurrently.
- But: Queue Items are still processed only once due to leasing

https://linear.app/inngest/issue/INN-3565/keep-enqueuing-to-workflow-partition-even-when-custom-concurrency-keys

- [ ] Docs

## Checklist


cdzombak:
I think this looks good, but @tonyhb should be the one to approve 😉

tonyhb (approved):

PR #1720, opened 2024-09-09
Fix missing UI output when step output is too large

amh4r:
## Description
Fix missing UI output when step output is too large. UI output was already correct when function output was too large.

## Testing
Before:
image

After:
image

Using this function:
```ts
inngest.createFunction(
  {
    id: "fn-1",
    retries: 0,
  },
  { event: "foo" },
  async ({ step }) => {
    await step.run("a", () => {
      return "A".repeat(5 * 1024 * 1024);
    });
  }
);
```

- [ ] Docs

## Checklist


pkg/execution/history/lifecycle.go
```diff
@@ -780,7 +780,7 @@ func applyResponse(
 		return nil
 	}
 
-	if len(resp.Generator) > 0 {
+	if len(resp.Generator) > 1 {
```
tonyhb, line 783: Why have we changed this, and what are the implications?  IIRC, each generator _also_ calls history lifecycle methods so now we'd be doubling up.  Mainly worried this will break BQ and or traces.

goodoldneon: I just checked and you're right: it makes the full generator stuff in the output, so this change isn't good.

I `spew.Dump`ed `resp.Generator` for the `output_too_large` case and it looks like a zero value struct:
```go
(*state.GeneratorOpcode)(0x14001346000)({
 Op: (enums.Opcode) None,
 ID: (string) "",
 Name: (string) "",
 Opts: (interface {}) <nil>,
 Data: (json.RawMessage) <nil>,
 Error: (*state.UserError)(<nil>),
 DisplayName: (*string)(<nil>)
})
```

Wonder why

PR #1723, opened 2024-09-10
Fix flaky trace run tests

BrunoScheufler:
This refactors some assertion logic in [`TestInvoke/trace run should have appropriate data`](https://github.com/inngest/inngest/blob/ee8e50d0af2daad954694155ec6e89c89973543a/tests/golang/invoke_test.go#L74).

Previously, we used `require.EventuallyWithT` together with fail-fast assertions such as `require.NotNil` to retry until trace data was available or up to 10 seconds. Unfortunately, [`t.FailNow()`](https://github.com/inngest/inngest/blob/0eafe3138809d3de47b0bf6bf49f1b0491200a58/vendor/github.com/stretchr/testify/assert/assertions.go#L1923) achieves fast failures by causes panics, which are not handled by `require.EventuallyWithT`.

This occasionally breaks tests, if trace data does not become available within the awaited 3-4 seconds (`<-time.After(3 * time.Second)`).

To fix the flakiness, I've replaced fail-fast requires with a single wait loop and added early returns on nil checks, so we don't run into panics. The assertion either fails after the timeout elapses or returns valid run traces. This way, we keep all existing assertions while avoiding panics either due to missing data or overly aggressive assertions.

### Related issues on the web
- https://github.com/stretchr/testify/issues/1457
- https://github.com/superfly/flyctl/pull/2760

- [ ] Docs

## Checklist

### DIFF
--- a/tests/client/function_run.go
+++ b/tests/client/function_run.go
@@ -228,6 +228,25 @@ func (c *Client) MustRunTraces(ctx context.Context, runID string) *RunV2 {
 	return data
 }
 
+func (c *Client) WaitForRunTraces(ctx context.Context, t *testing.T, runID *string, status models.FunctionStatus) *RunV2 {
+	var traces *RunV2
+	require.NotNil(t, runID)
+	require.Eventually(t, func() bool {
+		if *runID == "" {
+			return false
+		}
+
+		run, err := c.RunTraces(ctx, *runID)
+		traces = run
+
+		// NOTE: we force the function to return early to prevent panics in the next assertion.
+		// We cannot use require.NotNil as this will cause an uncaught panic (see https://github.com/stretchr/testify/issues/1457)
+		return err == nil && run != nil && run.Status == status.String()
+	}, 10*time.Second, 2*time.Second)
+
+	return traces
+}
+
 func (c *Client) RunTraces(ctx context.Context, runID string) (*RunV2, error) {
 	c.Helper()
 
--- a/tests/client/function_run.go
+++ b/tests/client/function_run.go
@@ -215,20 +215,7 @@ func (c *Client) WaitForRunStatus(
 	return run
 }
 
-// retrieve run with traces
-// TODO: add the traces once implemented
-func (c *Client) MustRunTraces(ctx context.Context, runID string) *RunV2 {
-	c.Helper()
-
-	data, err := c.RunTraces(ctx, runID)
-	if err != nil {
-		c.Fatalf(err.Error())
-	}
-
-	return data
-}
-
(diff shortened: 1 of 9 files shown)


tests/golang/invoke_test.go
```diff
@@ -69,29 +72,50 @@ func TestInvoke(t *testing.T) {
 	r.NoError(err)
 
 	t.Run("trace run should have appropriate data", func(t *testing.T) {
-		<-time.After(4 * time.Second)
+		// Wait for function run to complete
+		require.Eventually(t, func() bool {
+			return done
+		}, time.Second*4, time.Second)
```
goodoldneon, line 78: Could we use `WaitForRunStatus` instead?
https://github.com/inngest/inngest/blob/main/tests/golang/fn_output_test.go#L46

tests/client/function_run.go
```diff
@@ -215,13 +215,34 @@ func (c *Client) WaitForRunStatus(
...
-func (c *Client) RunTraces(ctx context.Context, runID string) *RunV2 {
+func (c *Client) WaitForRunTracesWithTimeout(ctx context.Context, t *testing.T, runID *string, status models.FunctionStatus, timeout time.Duration, interval time.Duration) *RunV2 {
+	var traces *RunV2
+	require.NotNil(t, runID)
+	require.Eventually(t, func() bool {
+		if *runID == "" {
+			return false
+		}
+
+		run, err := c.RunTraces(ctx, *runID)
```
BrunoScheufler, line 226: Here we simply keep retrying until we see no more errors, return data are present, and the status matches.

tests/golang/basic_step_test.go
```diff
@@ -153,117 +153,110 @@ func TestFunctionSteps(t *testing.T) {
...
-			require.Equal(t, models.FunctionStatusCompleted.String(), run.Status)
-			require.False(t, run.IsBatch)
-			require.Nil(t, run.BatchCreatedAt)
-
-			require.NotNil(t, run.Trace)
-			require.True(t, run.Trace.IsRoot)
-			require.Equal(t, 5, len(run.Trace.ChildSpans))
-			require.Equal(t, models.RunTraceSpanStatusCompleted.String(), run.Trace.Status)
-
+		run := c.WaitForRunTraces(ctx, t, &runID, models.FunctionStatusCompleted)
```
BrunoScheufler, line 156: All that's changed about all affected tests is that we move away from the previous flow of
- Waiting an arbitrary duration which causes flakiness (`<-time.After(3 * time.Second)`)
- Improperly using require which causes panics when flakiness is encountered (`require.NotNil(t, run)`)
- Switch away from `require.EventuallyWithT(t, func(ct *assert.CollectT) {` to using the helper to prevent running into accidental assertion panics

PR #1724, opened 2024-09-10
Reduce batch processor timeout for dev server

darwin67:
## Description

Batch processor has a default timeout of ~5s.
This is too slow for the dev server experience so lowering it to 100ms.

Also reducing the UI polling to 400ms.

## Type of change (choose one)
- [x] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [ ] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #1727, opened 2024-09-10
Fix button flicker when polling

djfarrelly:
## Description

Polling changes the state of the button too quickly and causes the UI to flicker unnecessarily. Disabling the button when polling doesn't matter too much as the view is always live and up to date. There is a potential better fix than this, but this is a simple improvement for situations where we're polling (the dev server).

## Motivation
With the faster polling in #1724, this problem got even worse.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/RunsPage/RunsPage.tsx
+++ b/ui/packages/components/src/RunsPage/RunsPage.tsx
@@ -243,6 +243,11 @@ export function RunsPage({
     return out;
   }, [columns]);
 
+  // Do not disable or show the button as loading if the poll interval is less than 1 second
+  // Changing state too quickly can cause the button to flicker
+  const disableRefreshButton =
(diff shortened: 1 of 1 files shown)


darwin67 (approved):

tonyhb (approved):

ui/packages/components/src/RunsPage/RunsPage.tsx
```diff
@@ -335,8 +335,18 @@ export function RunsPage({
...
+            // Changing state too quickly can cause the button to flicker
+            loading={
+              pollInterval && pollInterval < 1000
+                ? isLoadingInitial
+                : isLoadingMore || isLoadingInitial
+            }
+            disabled={
+              pollInterval && pollInterval < 1000
+                ? isLoadingInitial
+                : isLoadingMore || isLoadingInitial
```
darwin67, line 348: if it's the exact same logic, why not just put the result in a variable before passing it in as params?
this will result in execution twice iirc

PR #1729, opened 2024-09-11
Fix hanging runs when response body isn't JSON

goodoldneon:
## Description
Fix hanging runs when response body isn't JSON. For example, this happens on gateway timeouts

- [ ] Docs

## Checklist


pkg/execution/driver/httpdriver/parse.go
```diff
@@ -171,5 +171,5 @@ func parseResponse(byt []byte) any {
 		}
 	}
 
-	return json.RawMessage(byt)
+	return string(byt)
```
darwin67, line 174: `json.RawMessage` is just a type alias for `[]byte`, so I think we can just return `byt` as it instead of casting it to a string.

darwin67 (approved):

PR #1731, opened 2024-09-11
Replace go-batcher lib with in house implementation

darwin67:
## Description

For some reason, `go-batcher` is causing build issues so just implement the span batch processor in house.
Which should be more simple and don't need as much generics operations.

Build error:
https://github.com/inngest/inngest/actions/runs/10803761530/job/29969651382

- [ ] Docs

## Checklist


pkg/telemetry/exporters/batch_processor.go
```diff
@@ -2,123 +2,191 @@ package exporters
...
+	// pass span into the channel
+	b.in <- &s
+	metrics.IncrBatchProcessorEnqueuedCounter(context.TODO(), metrics.CounterOpt{PkgName: pkgName})
 }
 
 func (b *batchSpanProcessor) Shutdown(ctx context.Context) error {
-	b.batcher.Flush()
+	if err := b.flush(ctx); err != nil {
+		logger.StdlibLogger(ctx).Error("error flushing spans on shutdown", "error", err)
+	}
```
darwin67, line 81: I'm not closing the `in` channel here, since executors can have long grace periods before they shutdown, and keeping the channel alive until it completely finishes helps with spans getting processed even if a `ctx.Done()` signal is received.

Would this be a problem is a channel is not closed before shutting down?

BrunoScheufler: So you're saying we need to keep `b.in` open to ensure that we don't close the channel before all traces are sent, right?

I think that sounds reasonable, unless we have a signal that notifies the batcher when the last job is fully processed. Otherwise, we just don't know the right moment to close the channel.

pkg/telemetry/exporters/batch_processor.go
```diff
@@ -2,123 +2,191 @@ package exporters
...
+			if err := b.flush(ctx); err != nil {
+				logger.StdlibLogger(ctx).Error("error flushing spans on completion", "error", err)
+			}
+			return
 		}
-		spans = append(spans, span)
+	}
+}
+
+// append add the span into the buffer the pointer is currently pointing to
```
BrunoScheufler, line 104:
```suggestion
// append adds the span into the buffer the pointer is currently pointing to
```

BrunoScheufler (commented):
Looks great!

BrunoScheufler (approved):
Approving this for releasing the dev server, I'll reach out to @darwin67 with my remaining minor questions, which shouldn't block the dev server release, but could be answered before usptreaming!

pkg/telemetry/exporters/batch_processor.go
darwin67: yeah, services start shutting down after they get the termination signal in the container.
but it's unclear with the extended grace period, when exactly does `ctx.Done()` happens.

if it happens at the actual end when all the execution is done, then it won't be a problem. but if it happen any time before that, closing the channel prematurely will lead to lost spans.

PR #1741, opened 2024-09-12
Fix improperly handling JSON arrays

goodoldneon:
## Description
Fix improperly handling JSON arrays

- [ ] Docs

## Checklist


pkg/execution/driver/httpdriver/parse.go
```diff
@@ -141,13 +141,11 @@ func parseResponse(byt []byte) any {
...
-		// Is the response valid JSON?  If so, ensure that we don't re-marshal the
-		// JSON string.
-		respjson := map[string]interface{}{}
-		if err := json.Unmarshal(byt, &respjson); err == nil {
-			return respjson
-		}
+	// Is the response valid JSON?  If so, ensure that we don't re-marshal the
+	// JSON string.
+	var respjson json.RawMessage
+	if err := json.Unmarshal(byt, &respjson); err == nil {
```
tonyhb, line 147: Can we use https://pkg.go.dev/github.com/valyala/fastjson?utm_source=godoc#Validate here?

tonyhb, line 147: I think by adding this above line 158 we're bypassing that entire check for V1 of the SDK, which accidentally double encoded responses.  It will skip this check as those responses are valid JSON.

We should move this below the check.  Finally, is it an issue if we're no longer returning maps?  What code expects the `map[string]interface{}{}` response?

It might be safer if we add a fastjson.ValidateBytes response at the tail end of this function and return json.RawMessage if the validation passes: https://pkg.go.dev/github.com/valyala/fastjson?utm_source=godoc#ValidateBytes.

tonyhb (changes requested):

tonyhb (approved):

PR #1767, opened 2024-09-18
Fix redirect method

goodoldneon:
## Description
Fix redirect HTTP method. Before this change, we'd always use the original HTTP method even if the redirect was telling us to use a different one.

Also, this PR combines the headers in the original and redirect requests, rather than replacing the redirect request headers.

## Context
This bug became apparent for Modal users. Modal sends a `303: See Other` redirect after 150 seconds, and this redirect has a `GET` method even though the original method has a `POST`

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/driver/httpdriver/httpdriver_test.go
+++ b/pkg/execution/driver/httpdriver/httpdriver_test.go
@@ -34,7 +34,7 @@ func TestRedirect(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch count {
 		case 8:
-			require.Equal(t, http.MethodPost, r.Method)
+			require.Equal(t, http.MethodGet, r.Method)
 			byt, err := io.ReadAll(r.Body)
 			require.NoError(t, err)
 			require.Equal(t, input, byt)
--- a/pkg/execution/driver/httpdriver/util.go
+++ b/pkg/execution/driver/httpdriver/util.go
@@ -47,9 +47,6 @@ func CheckRedirect(req *http.Request, via []*http.Request) (err error) {
 		return fmt.Errorf("stopped after 10 redirects")
 	}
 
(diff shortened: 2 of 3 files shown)


tonyhb (approved):
Tests?

BrunoScheufler (approved):

pkg/execution/driver/httpdriver/httpdriver_test.go
```diff
@@ -34,7 +34,7 @@ func TestRedirect(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch count {
 		case 8:
-			require.Equal(t, http.MethodPost, r.Method)
+			require.Equal(t, http.MethodGet, r.Method)
```
tonyhb, line 37: This is scary.

A user might configure a gateway to use a 303.  This converts POSTs to GETs.  It's what this test covers.

We don't want that.  We actually ignore the HTTP spec here.  We do that because _we know what the intent of our request is_, and it *must* use the requested HTTP method after any redirects, instead of what the redirect asked for.

tonyhb (changes requested):
Have some questions.  303 tests were added specifically to ensure we keep the request method, but we're changing that and there's no reasoning as to why, or what's broken.

What is the intent here? 😅

pkg/execution/driver/httpdriver/httpdriver_test.go
goodoldneon: That isn't true. If a redirect tells us to use a `GET` then we _must_ use a `GET` after the redirect. Modal switches to a `GET` on the redirect:
1. We send `POST` to execute Inngest function hosted in Modal.
2. After 150 seconds, Modal sends a `303` redirect (which is always a `GET` by the HTTP spec).
3. If we send a `POST` after the redirect then Modal returns a `400` due an unsupported method. It only works if we send a `GET`.

tonyhb (approved):

PR #1769, opened 2024-09-19
INN-3761: Track pending event batches

BrunoScheufler:
## Description

This introduces a service similar to our queue instrumentation, which acquires an exclusive lease and periodically reports on  pending batches.

Pending batch counts are now tracked per account on batch creation and dropped on batch start, which helps us understand how many batches each account has in progress at any given time.

I had to move tracking out to dedicated Lua scripts, as batches are sharded by function ID. The newly-introduced hash storing account IDs <-> count of pending batches is global, so it cannot be modified in the same Lua scripts or else we would run into multi-slot errors.

I've abstracted this logic into our service interface to make it easy to plug in when upstreaming.

- [ ] Docs

## Checklist


pkg/execution/executor/executor.go
```diff
@@ -2407,6 +2407,10 @@ func (e *executor) AppendAndScheduleBatch(ctx context.Context, fn inngest.Functi
 				"account_id": bi.AccountID.String(),
 			},
 		})
+
+		if err := e.batcher.TrackBatchCreate(ctx, bi.AccountID); err != nil {
+			return fmt.Errorf("could not track batch create: %w", err)
```
cdzombak, line 2412: minor: should we return this error or just log it?

pkg/execution/executor/executor.go
```diff
@@ -2514,6 +2518,10 @@ func (e *executor) RetrieveAndScheduleBatch(ctx context.Context, fn inngest.Func
 		return err
 	}
 
+	if err := e.batcher.TrackBatchStart(ctx, payload.AccountID); err != nil {
+		return fmt.Errorf("could not track batch start: %w", err)
```
cdzombak, line 2522: minor: should we return this error or just log it?

BrunoScheufler:
We don't need this instrumentation anymore and the complexity of deploying a dedicated service is probably a bit much

PR #1805, opened 2024-09-24
Add enable-persistence option to start command

amh4r:
## Description
Add `enable-persistence` option to `start` command. This is useful for end user integration tests, since everything should be ephemeral.

- [ ] Docs

## Checklist

### DIFF
--- a/cmd/commands/start.go
+++ b/cmd/commands/start.go
@@ -42,6 +42,7 @@ func NewCmdStart(rootCmd *cobra.Command) *cobra.Command {
 	baseFlags.Int("tick", devserver.DefaultTick, "Interval, in milliseconds, of which to check for new work.")
 	baseFlags.String("signing-key", "", "Signing key used to sign and validate data between the server and apps.")
 	baseFlags.StringSlice("event-key", []string{}, "Event key(s) that will be used by apps to send events to the server.")
+	baseFlags.BoolP("enable-persistence", "", true, "Controls whether state is persisted. (default true)")
 	cmd.Flags().AddFlagSet(baseFlags)
 	groups = append(groups, FlagGroup{name: "Flags:", fs: baseFlags})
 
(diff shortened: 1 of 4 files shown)


djfarrelly (changes requested):
I'm up for adding this functionality behind a hidden flag (something the user cannot see or env var, ex. `INNGEST_EXPERIMENTAL_DISABLE_PERSISTENCE`), but combining `start` with disabling persistence for the end user is a strange UX that doesn't might not make sense outside of external testing.

As we're past 1.0, adding new flags mean that we'll need to support them long term. We need to discuss flags that affect the developer experience

cmd/commands/start.go
```diff
@@ -125,15 +125,16 @@ func doStart(cmd *cobra.Command, args []string) {
...
+		Config:            *conf,
+		PollInterval:      viper.GetInt("poll-interval"),
+		RedisURI:          viper.GetString("redis-uri"),
+		RetryInterval:     viper.GetInt("retry-interval"),
+		Tick:              time.Duration(tick) * time.Millisecond,
+		URLs:              viper.GetStringSlice("sdk-url"),
+		SQLiteDir:         viper.GetString("sqlite-dir"),
+		SigningKey:        viper.GetString("signing-key"),
+		EventKey:          viper.GetStringSlice("event-key"),
+		EnablePersistence: !config.IsEnvVarFalsy("INNGEST_EXPERIMENTAL_ENABLE_PERSISTENCE"),
```
djfarrelly, line 137: It kind of feels like a weird pattern to use an falsy value a truthy sounding env var that is already truthy by default. It makes the logic more complex. Like "Persistence is enabled by default. Do you want to disable persistence? Set the enable persistence var to false."

Since this is just on `inngest start` which enables persistence by default, I think we should just call this `INNGEST_EXPERIMENTAL_DISABLE_PERSISTENCE` - the logic becomes simpler.

jpwilliams: 👍 Agree - switching this to `INNGEST_EXPERIMENTAL_DISABLE_PERSISTENCE` feels good.

cc @amh4r - just this and LGTM

PR #1518, opened 2024-06-27
Key & Account Queues

tonyhb:
See the spec.  More details to follow.

_(nb. PR body & title rewritten by @cdzombak 2024-09-03; these are not @tonyhb's words)_

## Issues

- See: https://linear.app/inngest/project/key-and-account-level-queues-26f88c23df90/issues

## Type of change (choose one)
- [x] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

PR #1817, opened 2024-09-26
Always set inner queue name (i.Data.QueueName) during Enqueue()

BrunoScheufler:
## Description

> [!CAUTION]
> Roll this out before #1518 

This is required for ItemPartitions to work, as we previously used the inner queueName exclusively.

One potential improvement is to deduplicate queueNames altogether, there's no reason as to why we store the same attribute twice.

https://linear.app/inngest/issue/INN-3816

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/queue_processor.go
+++ b/pkg/execution/state/redis_state/queue_processor.go
@@ -105,6 +105,10 @@ func (q *queue) Enqueue(ctx context.Context, item osqueue.Item, at time.Time) er
 		item.QueueName = queueName
 	}
 
+	if queueName != nil && *queueName == "" {
+		return fmt.Errorf("encountered empty string for queue name")
+	}
+
 	metrics.IncrQueueItemStatusCounter(ctx, metrics.CounterOpt{
 		PkgName: pkgName,
 		Tags: map[string]any{
--- a/pkg/execution/state/redis_state/queue_processor.go
+++ b/pkg/execution/state/redis_state/queue_processor.go
@@ -100,6 +100,11 @@ func (q *queue) Enqueue(ctx context.Context, item osqueue.Item, at time.Time) er
 		queueName = item.QueueName
 	}
 
+	// ensure both QueueName attributes are always set (ItemPartitions depends on this)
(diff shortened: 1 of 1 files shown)


cdzombak:
@BrunoScheufler 

> One potential improvement is to deduplicate queueNames altogether, there's no reason as to why we store the same attribute twice.

This would also cut down on space usage for the queue, but is probably outside the scope of this PR and we already have a bunch of changes waiting to roll out. But I would like to do that sooner rather than later.

We have a Linear ticket tracking this, right?

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -100,6 +100,11 @@ func (q *queue) Enqueue(ctx context.Context, item osqueue.Item, at time.Time) er
 		queueName = item.QueueName
 	}
 
+	// ensure both QueueName attributes are always set (ItemPartitions depends on this)
+	if queueName != nil && item.QueueName == nil {
+		item.QueueName = queueName
+	}
+
```
cdzombak, line 107: Can we please return an error instead of enqueueing if the resulting queue name is a pointer to an empty string?

I'd like to detect unexpected inputs here, rather than having to discover them via debugging a problem deep in the queue.

cdzombak (changes requested):

pkg/execution/state/redis_state/queue_processor.go
BrunoScheufler: queueNames _can_ be nil as they're only used for overrides when we process system partitions, that's important. But they should never be an empty strings!

tonyhb (approved):

BrunoScheufler:
> @BrunoScheufler
> 
> > One potential improvement is to deduplicate queueNames altogether, there's no reason as to why we store the same attribute twice.
> 
> This would also cut down on space usage for the queue, but is probably outside the scope of this PR and we already have a bunch of changes waiting to roll out. But I would like to do that sooner rather than later.
> 
> We have a Linear ticket tracking this, right?

Yes: https://linear.app/inngest/issue/INN-3584

PR #1842, opened 2024-10-07
Upstream changes

BrunoScheufler:
- [ ] Docs

## Checklist


pkg/execution/state/redis_state/lua/includes/enqueue_to_partition.lua
```diff
@@ -8,7 +8,7 @@ local function enqueue_get_partition_item(partitionKey, id)
 end
 
 local function enqueue_to_partition(keyPartitionSet, partitionID, partitionItem, keyPartitionMap, keyGlobalPointer, keyGlobalAccountPointer, keyAccountPartitions, queueScore, queueID, partitionTime, nowMS)
-	if partitionID == "" then
+	if partitionID == "" or partitionID == nil or partitionID == false then
```
BrunoScheufler, line 11: partitionID will always be set to an (empty) string, so the nil or false case will not happen in the current implementation, but it's good to have a safeguard here.

pkg/execution/state/redis_state/lua/includes/enqueue_to_partition.lua
```diff
@@ -60,7 +60,8 @@ local function enqueue_to_partition(keyPartitionSet, partitionID, partitionItem,
 		--       we've forced a partition to have a delay.
 		--
 		--       Here, we do those checks.
-		if nowMS > existing.forceAtMS then
+
+		if nowMS == nil or nowMS == false or existing == false or existing == nil or existing.forceAtMS == nil or nowMS > tonumber(existing.forceAtMS) then
```
BrunoScheufler, line 64: We were hitting partitions without forceAtMS set. `tonumber` returns nil if the passed argument is nil or false, so we need to guard against that case.

image

pkg/execution/state/redis_state/lua/includes/enqueue_to_partition.lua
```diff
@@ -117,7 +118,8 @@ local function requeue_to_partition(keyPartitionSet, partitionID, partitionItem,
 		--       we've forced a partition to have a delay.
 		--
 		--       Here, we do those checks.
-		if existing ~= nil and nowMS > existing.forceAtMS then
+
+		if nowMS == nil or nowMS == false or existing == false or existing == nil or existing.forceAtMS == nil or nowMS > tonumber(existing.forceAtMS) then
```
BrunoScheufler, line 122: [Same as above](https://github.com/inngest/inngest/pull/1842/files#r1790539155)

pkg/execution/state/redis_state/lua/includes/update_pointer_score.lua
```diff
@@ -2,6 +2,12 @@
 -- score.  This score should almost always be the value from `get_fn_partition_score`.
 -- It's a separate function as > 1 pointer queue may be updated at a time.
 local function update_pointer_score_to(fnID, pointerQueueKey, updateTo)
+    -- if fnID contains
```
BrunoScheufler, line 5: To prevent enqueueing to key queues, we need to prevent existing runs from updating the account and global partition pointers for key queues. We detect this by filtering out all fnIDs starting with a curly brace, as this is guaranteed to be a fully-qualified Redis key, which is only used for key queues.

This fix should be refactored to a different level, with a proper feature toggle.

pkg/execution/state/redis_state/lua/queue/lease.lua
```diff
@@ -122,30 +122,9 @@ local function handleLease(keyPartition, keyConcurrency, partitionID)
...
-	-- NOTE: The global partition ID isn't the actual partition zset key for backwards compatibility.
-	-- Instead, they are the partition IDs, which is either the partition ZSET (for new concurrency
-	-- key partitions) OR the function ID (for default partitions).
-	-- 
-	-- The first version of the queue used function UUIDs as queue names, and the global pointer
-	-- expected just the function UUIDs instead of a fully defined redis key.
-	update_pointer_score_to(partitionID, keyGlobalPointer, earliestScore)
-
-	-- Update account partitions and account pointers with new score of next item
-	update_account_queues(keyGlobalAccountPointer, keyAccountPartitions, partitionID, accountId, earliestScore)
```
BrunoScheufler, line 143: We don't need to update pointers and account queues for each individual lease, as the partition itself is leased and thus prevents other executors from picking it up, so moving the score back and forth doesn't do anything.

pkg/execution/state/redis_state/lua/queue/partitionPeek.lua
```diff
@@ -22,21 +22,12 @@ if count > limit and sequential == 0 then
...
+if #partitionIds == 0 then
 	return {}
 end
 
-local potentiallyMissingPartitions = redis.call("HMGET", partitionKey, unpack(items))
-local missingPartitionIds = {}
-local validPartitions = {}
+local potentiallyMissingPartitions = redis.call("HMGET", partitionKey, unpack(partitionIds))
 
-for i, partition in ipairs(potentiallyMissingPartitions) do
```
BrunoScheufler, line 34: We moved from filtering out missing partitions in Lua to doing it in Go itself. This should speed up Lua scripts considerably.

pkg/execution/state/redis_state/lua/queue/partitionRequeue.lua
```diff
@@ -71,11 +71,15 @@ local items = redis.call("ZRANGE", keyPartitionZset, "-inf", "+inf", "BYSCORE",
...
     local encoded = redis.call("HMGET", queueKey, unpack(items))
     for k, v in pairs(encoded) do
-        local item = cjson.decode(v)
-        if (item.leaseID == nil or item.leaseID == cjson.null) and math.floor(item.at / 1000) < atS then
-            atS = math.floor(item.at / 1000)
-            break
-        end
+				-- when an old executor processes a default partition, it does not
+				-- remove pointers from key queues. we need to skip nil items in here
+				if v ~= nil and v ~= false then
```
BrunoScheufler, line 76: This handles missing queue items gracefully.

pkg/execution/state/redis_state/queue.go
```diff
@@ -984,7 +1000,7 @@ func (q *queue) ItemPartitions(ctx context.Context, i QueueItem) ([]QueuePartiti
 		fnPartition.ConcurrencyLimit = limits.AccountLimit
 	}
 
-	if len(ckeys) > 0 {
+	if enableKeyQueues && len(ckeys) > 0 {
```
BrunoScheufler, line 1003: We decided to roll back key queues. This line makes sure we only ever return the default partition, and no concurrency key queues.

pkg/execution/state/redis_state/queue.go
```diff
@@ -1382,29 +1398,48 @@ func (q *queue) Peek(ctx context.Context, partition *QueuePartition, until time.
...
-		missingQueueItems, ok = returnedSet[1].([]any)
+		allQueueItemIds, ok = returnedSet[1].([]any)
 		if !ok {
 			return nil, fmt.Errorf("unexpected first item in set returned from peek: %T", peekRet)
 		}
 	} else if len(returnedSet) != 0 {
 		return nil, fmt.Errorf("expected zero or two items in set returned by peek: %v", returnedSet)
 	}
 
+	items := make([]any, 0, len(allQueueItemIds))
```
BrunoScheufler, line 1424: This finds and gracefully handles missing items. We moved this out of Lua and into the executor to speed up script processing.

pkg/execution/state/redis_state/queue.go
```diff
@@ -1586,7 +1613,7 @@ func (q *queue) Lease(ctx context.Context, p QueuePartition, item QueueItem, dur
 	}
 
 	// Grab all partitions for the queue item
-	parts, acctLimit := q.ItemPartitions(ctx, item)
+	parts, acctLimit := q.ItemPartitions(ctx, item, true)
```
BrunoScheufler, line 1616: We do want to consider key queues on Lease, to add items to all concurrency queues. This is important because we wouldn't apply custom concurrency key limits otherwise.

pkg/execution/state/redis_state/queue.go
```diff
@@ -1237,7 +1253,7 @@ func (q *queue) EnqueueItem(ctx context.Context, i QueueItem, at time.Time) (Que
 		partitionTime = q.clock.Now()
 	}
 
-	parts, _ := q.ItemPartitions(ctx, i)
+	parts, _ := q.ItemPartitions(ctx, i, false)
```
BrunoScheufler, line 1256: We do not want to enqueue new items to key queue partitions. This is because we
- significantly grow the global and account partition pointers which may slow down peeks
- spend a lot of resources on peeking and leasing almost-empty partitions vs. processing partitions. QPS is limited and every operation we don't spend on processing actual work is wasted
-

pkg/execution/state/redis_state/queue.go
```diff
@@ -2208,27 +2235,48 @@ func (q *queue) partitionPeek(ctx context.Context, partitionKey string, sequenti
...
-		missingPartitions, ok = returnedSet[2].([]any)
+		allPartitionIds, ok = returnedSet[2].([]any)
 		if !ok {
 			return nil, fmt.Errorf("unexpected third item in set returned from partitionPeek: %T", peekRet)
 		}
 	} else if len(returnedSet) != 0 {
 		return nil, fmt.Errorf("expected zero or three items in set returned by partitionPeek: %v", returnedSet)
 	}
 
+	encoded := make([]any, 0)
```
BrunoScheufler, line 2259: This is the same improvement as [above](https://github.com/inngest/inngest/pull/1842/files#r1790611790), just for partition peeks.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -500,7 +491,7 @@ func (q *queue) scan(ctx context.Context) error {
 	}
 
 	processAccount := false
-	if q.runMode.Account && (!q.runMode.Partition || rand.Intn(2) == 1) {
+	if q.runMode.Account && (!q.runMode.Partition || rand.Intn(100) <= q.runMode.AccountWeight) {
```
BrunoScheufler, line 494: We made the ratio of scanning accounts vs partitions configurable. Setting an AccountWeight of 85 means we scan accounts 85% of each scan and partitions 15%.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -523,25 +516,32 @@ func (q *queue) scan(ctx context.Context) error {
...
 
-		eg := errgroup.Group{}
-
 		// Reduce number of peeked partitions as we're processing multiple accounts in parallel
 		// Note: This is not optimal as some accounts may have fewer partitions than others and
 		// we're leaving capacity on the table. We'll need to find a better way to determine the
 		// optimal peek size in this case.
-		peekSize := int64(math.Round(float64(PartitionPeekMax / int64(len(peekedAccounts)))))
+		// accountPartitionPeekMax := int64(math.Round(float64(PartitionPeekMax / int64(len(peekedAccounts)))))
+		accountPartitionPeekMax := int64(10)
```
BrunoScheufler, line 524: This is the result of manual tuning, balancing hitting QPS limits (running into context canceled and context deadline exceeded) and processing workloads.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -523,25 +516,32 @@ func (q *queue) scan(ctx context.Context) error {
...
-			eg.Go(func() error {
+
+			wg.Add(1)
+			go func(account uuid.UUID) {
+				defer wg.Done()
 				partitionKey := q.u.kg.AccountPartitionIndex(account)
 
-				return q.scanPartition(ctx, partitionKey, peekSize, peekUntil, nil, metricShardName, &account)
-			})
+				if err := q.scanPartition(ctx, partitionKey, accountPartitionPeekMax, peekUntil, nil, metricShardName, &account); err != nil {
```
BrunoScheufler, line 536: Accounts are processed in parallel to avoid stalling other accounts. For each account, we peek account partitions, which are again processed in parallel.

BrunoScheufler: We peek 30 accounts at once, and for each account peek 10 partitions.

When tuning, we prefer peeking more accounts than partitions per account to ensure fairness. For folks with high-throughput requirements, we set aside workers with guaranteed capacity.

pkg/execution/state/redis_state/queue.go
```diff
@@ -2368,6 +2408,9 @@ func (q *queue) partitionPeek(ctx context.Context, partitionKey string, sequenti
 		// If we have an allowlist, only accept this partition if its in the allowlist.
 		if len(q.allowQueues) > 0 && !checkList(item.Queue(), q.allowQueueMap, q.allowQueuePrefixes) {
 			// This is not in the allowlist specified, so do not allow this partition to be used.
+			if q.name == "pause-queue" {
+				q.logger.Debug().Interface("queue", item.Queue()).Msg("skipping queue via allowlist")
+			}
```
darwin67, line 2413: let's drop this.

darwin67 (approved):

pkg/execution/state/redis_state/lua/includes/update_pointer_score.lua
cdzombak: It looks like we have introduced some sort of flag `enableKeyQueues` below — can we use a similar one here?

cdzombak (approved):

tonyhb (approved):

pkg/execution/state/redis_state/lua/queue/dequeue.lua
```diff
@@ -89,6 +92,11 @@ local function handleDequeueConcurrency(keyConcurrency, keyPartitionSet, partiti
 		end
 	end
 
+	if partitionType ~= 0 then
+		-- If this is not a default partition, we don't need to update the global or account pointer.
```
cdzombak, line 95: Thanks for adding explanatory comments, @BrunoScheufler -- these will 💯% be helpful when we come back here in the future.

cdzombak:
@BrunoScheufler

This would require some Git wrangling, but: is it feasible split this into 2 PRs, one for general bugfixes and one that specifically disables key queues?

That would allow us to merge those in 2 squash commits, and when we circle back to key queues it would be simple to revert the "disable key queues" squash commit. Then we wouldn't need to worry about reconstructing tests, etc. — we could just revert it and know we're back where we started.

PR #1881, opened 2024-10-16
Add internal queue kind to handle migration between queues

darwin67:
## Description

Provide a new queue kind that only handles migrations of queue items between different queue clusters.
This is expected to be target agnostic eventually.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue.go
```diff
@@ -1546,50 +1686,34 @@ func (q *queue) Peek(ctx context.Context, partition *QueuePartition, until time.
...
-		return nil, nil
-	}
-	// The nested osqueue.Item never has an ID set;  always re-set it
-	qi.Data.JobID = &qi.ID
-	return qi, nil
+		qi := &osqueue.QueueItem{}
+		if err := json.Unmarshal(unsafe.Slice(unsafe.StringData(str), len(str)), qi); err != nil {
+			return nil, fmt.Errorf("error unmarshalling peeked queue item: %w", err)
+		}
+		if !opts.IgnoreLease && qi.IsLeased(now) {
```
BrunoScheufler, line 1709: Just making sure, we skip leased (in progress) queue items? Though, if they're in the partition queue, they wouldn't be leased in the first place.

darwin67: there's a window that they're leased but also execution is paused because of the `migrate` flag in metadata.
if the lease didn't expire by then, then it'll need to move as well.

BrunoScheufler (approved):

PR #1898, opened 2024-10-21
INN-3936 fetch keys in step 2

anafilipadealmeida:
## Description
- Wire signing key and event key
Note: the event key in the future should probably just come from the API directly, instead of doing the logic in the FE
Screenshot 2024-10-21 at 14 19 07

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/queries/useDefaultEventKey.ts
+++ b/ui/apps/dashboard/src/queries/useDefaultEventKey.ts
@@ -1,0 +1,51 @@
+import { graphql } from '@/gql';
+import { useGraphQLQuery } from '@/utils/useGraphQLQuery';
+
+const query = graphql(`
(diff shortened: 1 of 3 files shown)


ui/apps/dashboard/src/queries/useDefaultEventKey.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+`);
+
+export function useDefaultEventKey({ envID }: { envID: string }) {
+  const res = useGraphQLQuery({
+    query,
+    variables: { environmentID: envID },
+  });
+
+  if (res.data) {
```
jacobheric, line 22: Maybe we should at least log an error here so that we can distinguish error condition versus no data when troubleshooting.

jacobheric (approved):

PR #1908, opened 2024-10-23
Improve scavenger

BrunoScheufler:
## Description

scavenging is important to move abandoned in-progress items back to the "backlog" queue partition, so they get picked up by executors again. we observed some in-progress jobs not extending their lease for a prolonged duration, indicating that scavenging is not moving at the required pace, or at all.

- key queues should never be in there anymore: key queues are only used for tracking concurrency, NOT used as an execution entrypoint, so we don't care about key queues in the scavenger
- we ALWAYS want to add a pointer to the default fn partition to the scavenger queue, so that stalling jobs get picked up
- we should gracefully clean up leftover key queue data, and remove pointers to empty concurrency queues atomically. otherwise, we may accrue unnecessary pointers, which leaves the scavenger spinning without doing any work.

with these changes
- the scavenger queue (concurrency index) should gradually empty, only including pointers to function concurrency queues with actual in progress items
- key queues should be filtered out and never added again

context
- we have concurrency queues that contain stalled items. these should have been picked up by the scavenger
- the scavenger is logging key queues, which should no longer be in there
- the scavenger must clean up leftover items fast, so we cannot build up unnecessary cruft

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/queue.go
+++ b/pkg/execution/state/redis_state/queue.go
@@ -1442,9 +1442,9 @@ func (q *queue) SetFunctionPaused(ctx context.Context, accountId uuid.UUID, fnID
 	return nil
 }
 
-// SetFunctionPaused sets the "Paused" flag (represented in JSON as "off") for the given
-// function ID's queue partition.
-// If a function is unpaused, we requeue the partition with a score of "now" to ensure that it is processed.
+// dropPartitionPointerIfEmpty atomically drops a pointer queue member if the associated
+// ZSET is empty. This is used to ensure that we don't have pointers to empty ZSETs, in case
+// the cleanup process fails.
 func (q *queue) dropPartitionPointerIfEmpty(ctx context.Context, shard QueueShard, keyIndex, keyPartition, indexMember string) error {
 	ctx = redis_telemetry.WithScope(redis_telemetry.WithOpName(ctx, "SetFunctionPaused"), redis_telemetry.ScopeQueue)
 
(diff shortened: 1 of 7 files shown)


pkg/execution/state/redis_state/queue.go
```diff
@@ -1441,6 +1442,41 @@ func (q *queue) SetFunctionPaused(ctx context.Context, accountId uuid.UUID, fnID
 	return nil
 }
 
+// SetFunctionPaused sets the "Paused" flag (represented in JSON as "off") for the given
+// function ID's queue partition.
+// If a function is unpaused, we requeue the partition with a score of "now" to ensure that it is processed.
```
darwin67, line 1447: what does this comment have to do with the function?

darwin67 (approved):

pkg/execution/state/redis_state/queue.go
```diff
@@ -3114,28 +3166,48 @@ func (q *queue) Scavenge(ctx context.Context, limit int) (int, error) {
...
 			Key(queueKey).
 			Min("-inf").
 			Max(now).
 			Byscore().
-			Limit(0, 100).
+			Limit(0, ScavengeConcurrencyQueuePeekSize).
 			Build()
-		itemIDs, err := q.primaryQueueShard.RedisClient.unshardedRc.Do(ctx, cmd).AsStrSlice()
+		itemIDs, err := client.Do(ctx, cmd).AsStrSlice()
 		if err != nil && err != rueidis.Nil {
```
darwin67, line 3189: so what happens here if it's an error but it's `rueidis.Nil`?
does it just move forward?

pkg/execution/state/redis_state/queue.go
BrunoScheufler: nothing, thanks for catching 🙃

pkg/execution/state/redis_state/queue.go
BrunoScheufler: If it's rueidis nil. the ZSET is empty and will be handled by the next clause

PR #1928, opened 2024-10-31
WebSocket connections

BrunoScheufler:
## Description

This PR introduces the first iteration of WebSocket connections. This is not meant to be used, but it's in a mergeable state, so that we can split on future tasks.

- [ ] Docs

## Checklist


pkg/connect/pubsub/proxy.go
```diff
@@ -0,0 +1,295 @@
...
+)
+
+type ProxyResponse struct {
+	Status string
+
+	SdkResponse *connect_sdk.SdkResponse
+}
+
+type RequestForwarder interface {
+	Proxy(ctx context.Context, data connect_sdk.GatewayMessageTypeExecutorRequestData) (*connect_sdk.SdkResponse, error)
```
darwin67, line 24: the `SdkResponse` is better to be defined in this repo, and have the SDK reference it instead.
Otherwise, Go SDK becomes the source of truth, which will be weird.

pkg/connect/pubsub/proxy.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/oklog/ulid/v2"
+	"github.com/redis/rueidis"
+	"sync"
+	"time"
+)
+
+type ProxyResponse struct {
+	Status string
+
+	SdkResponse *connect_sdk.SdkResponse
```
darwin67, line 20: the `SdkResponse` is better to be defined in this repo, and have the SDK reference it instead.
Otherwise, Go SDK becomes the source of truth, which will be weird.

pkg/connect/pubsub/proxy.go
```diff
@@ -0,0 +1,295 @@
...
+		subscribersLock: sync.RWMutex{},
+	}
+}
+
+func (i *redisPubSubConnector) Proxy(ctx context.Context, data connect_sdk.GatewayMessageTypeExecutorRequestData) (*connect_sdk.SdkResponse, error) {
+	if data.RequestId == "" {
+		data.RequestId = ulid.MustNew(ulid.Now(), rand.Reader).String()
+	}
+
+	dataBytes, err := json.Marshal(data)
```
darwin67, line 59: Do we want to use a binary serializer instead of JSON?
WS allows binary transport so it can reduce the message sizes going back and forth.

pkg/connect/pubsub/proxy.go
```diff
@@ -0,0 +1,295 @@
...
+		go func() {
+			err = i.subscribe(withAckTimeout, i.channelAppRequestsAck(data.AppId, data.RequestId), func(msg string) {
+				acked = true
+			}, true)
+			ackErrChan <- err
+		}()
+	}
+
+	replyErrChan := make(chan error)
+	var reply ProxyResponse
```
darwin67, line 78: Same here, if this should be binary or JSON

darwin67 (approved):

pkg/connect/pubsub/proxy.go
BrunoScheufler: Good point, done

pkg/connect/pubsub/proxy.go
BrunoScheufler: We should, the question is which one. Protobuf?

darwin67: unless there's a strong argument for something else, we can default to protobuf for now.

PR #1939, opened 2024-11-05
INN-4004 Add onboarding click tracking

anafilipadealmeida:
## Description
- Trigger inngest tracking events for onboarding clicks

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Onboarding/useOnboardingTracking.ts
+++ b/ui/apps/dashboard/src/components/Onboarding/useOnboardingTracking.ts
@@ -58,10 +58,23 @@ export function useOnboardingTracking() {
     });
   };
 
+  const trackDeployAppAction = (type: string, hostingProvider: string) => {
+    trackEvent({
+      name: 'onboarding/deploy-app.action',
+      data: {
+        type: type,
+        hostingProvider: hostingProvider,
+      },
+      user: trackingUser,
+      v: '2024-11-04.1',
+    });
+  };
+
   return {
     trackOnboardingStepCompleted,
     trackWidgetDismissed,
     trackOnboardingOpened,
     trackCreateAppAction,
+    trackDeployAppAction,
   };
 }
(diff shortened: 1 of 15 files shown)


ui/apps/dashboard/src/components/Onboarding/useOnboardingTracking.ts
```diff
@@ -47,5 +47,60 @@ export function useOnboardingTracking() {
...
+    trackEvent({
+      name: 'onboarding/create-app.action',
+      data: {
+        type: type,
+      },
+      user: trackingUser,
+      v: '2024-11-04.1',
+    });
+  };
+
```
djfarrelly, line 60: The only question I have is whether we should change this to be more generic like:
```js
{
  name: "onboarding/step.completed",
  data: {
    step: "create-app",
    order: 0, // 1 or zero indexed step #
    isFinalStep: false, // boolean so if we change order or add future steps we can know the user completed the final step
    metadata: {
      hostingProvider: 'xyz'
    }
  },
  user: trackingUser,
}
```

From an analytics perspective, it may be easier over the long haul to have a single event name with many parameters so it can evolve over time.

Are there other ideas or thoughts on the multi-events? I'm open minded, just thinking about it now.

djfarrelly (approved):
The code looks good, but I have a key question on how we want to add these events. We can maybe chat over Slack - make a call then ship it!

ui/apps/dashboard/src/components/Onboarding/useOnboardingTracking.ts
anafilipadealmeida: 🤔 We are tracking 2 different things - steps being completed and clicks. Sometimes clicks lead to steps being completed, sometimes they don't. I'd keep them as two separated events, meaning that we have two event names (onboarding/step.completed and onboarding/click). But it's not a bad idea to keep 1) their data structure more similar, 2) make all the click events the same. Let me make some tweaks to this!

PR #1969, opened 2024-11-15
Handle disconnections from gateway side

darwin67:
## Description

Add implementation for connection deletion from state.
Also change move the run loop out of goroutine to make sure the function can exit or it'll keep spinning.

- [ ] Docs

## Checklist


proto/connect/v1/connect.proto
```diff
@@ -82,13 +82,12 @@ message SDKResponse {
...
-	string env_id = 4;
-	string app_id = 5;
-	SessionDetails session = 6;
-	string group_id = 7;
+	string id = 1;
+	string instance_id = 2;
+	string group_id = 3;
+	string language = 4;
+	string version = 5;
+	SystemAttributes attributes = 8;
```
darwin67, line 90: we can still shuffle this around since it's not used.

pkg/connect/gateway.go
```diff
@@ -0,0 +1,435 @@
+package connect
+
+import (
```
darwin67, line 3: this file is just splitting out the gateway related logic out of the generic service functions.

BrunoScheufler (approved):

pkg/connect/gateway.go
BrunoScheufler: any way to split this to preserve git history? np if not

PR #1983, opened 2024-11-20
Implement connection draining for connect gateway

BrunoScheufler:
## Description

This PR includes the following changes

- Gateways will store a minimal state in Redis on startup, periodically updated with a heartbeat goroutine
- The routing layer will ensure connections on inactive/unhealthy gateways are filtered out
- Two new API routes to drain and re-activate a gateway: This can be used for scenarios where we want to drain gateways without shutting them down
- A draining flow where connections will gracefully shut down with a proper WebSocket close frame, which is picked up by the client
- On shutdown, drain the gateway and wait until all connections have been closed before stopping the server

To drain a gateway, run

```
curl -X POST http://localhost:8289/v0/gateway/drain | jq
```

- [ ] Docs

## Checklist


pkg/connect/gateway.go
```diff
@@ -56,16 +75,28 @@ func (c *connectGatewaySvc) Handler() http.Handler {
...
 			svc:        c,
 			log:        c.logger,
 			ws:         ws,
 			updateLock: sync.Mutex{},
 		}
 
+		c.connectionSema.Add(1)
 		defer func() {
-			c.logger.Debug("Closing WebSocket connection")
+			defer c.connectionSema.Done()
```
darwin67, line 93: why `defer` within a `defer` function?

BrunoScheufler: good one, can just always call done
```suggestion
			c.connectionSema.Done()
```

BrunoScheufler: Left a comment to describe this behavior as expected:

We always want to update the semaphore and a couple lines down, we branch out with a condition. This can be done for two cases, but it's easy to forget to update the semaphore, thus preventing a graceful shutdown. If we update the semaphore too early, we will potentially shut down without a proper reason, which is also bad.

PR #1989, opened 2024-11-22
Add cancel event when function is cancelled

darwin67:
## Description

Provide a new cancel event to be sent when a function is cancelled.
This can be used for cancellation handling.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -1128,14 +1128,26 @@ func (e *executor) finalize(ctx context.Context, md sv2.Metadata, evts []json.Ra
 			Data:      data,
 		})
 
-		// Legacy - send inngest/function.failed, except for when the function has been cancelled.
-		if resp.Err != nil && !strings.Contains(*resp.Err, state.ErrFunctionCancelled.Error()) {
-			freshEvents = append(freshEvents, event.Event{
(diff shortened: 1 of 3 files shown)


pkg/execution/executor/executor.go
```diff
@@ -1128,14 +1128,26 @@ func (e *executor) finalize(ctx context.Context, md sv2.Metadata, evts []json.Ra
...
+				})
+			}
+
+			// Add function cancelled event
+			if *resp.Err == state.ErrFunctionCancelled.Error() {
+				freshEvents = append(freshEvents, event.Event{
+					ID:        ulid.MustNew(uint64(now.UnixMilli()), rand.Reader).String(),
+					Name:      event.FnCancelledName,
+					Timestamp: now.UnixMilli(),
+					Data:      data,
```
djfarrelly, line 1148: Only improvement would be to add a cancelled reason in the data. Since you can cancel via API or event, that might be interesting, maybe the event name or even the entire event payload. Maybe that doesn't need to be in for initial release though since it's additive.

djfarrelly (approved):

pkg/execution/executor/executor.go
darwin67: let's do this as a follow up. there are no reasoning being passed around, so that'll need more involvement.

PR #2012, opened 2024-12-09
INN-4110 Use history entitlement instead of features.log_retention

anafilipadealmeida:
## Description
- Replace deprecated API fields
  - features by entitlements
  - entitlementsUsage by entitlements

Note: I'm gonna change the billing requests again soon - https://linear.app/inngest/issue/INN-4158/make-billing-requests-on-client-side-instead-of-server-side

- [ ] Docs

## Checklist


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx
```diff
@@ -98,12 +106,7 @@ export default function DashboardTimeRangeFilter({
     query: GetBillingPlanDocument,
   });
 
-  // Since "features" is a map, we can't be 100% sure that there's a log
-  // retention value. So default to 7 days.
-  let logRetention = 7;
-  if (typeof data?.account.plan?.features.log_retention === 'number') {
-    logRetention = data.account.plan.features.log_retention;
-  }
+  const logRetention = data?.account.plan?.entitlements.history.limit || 7;
```
Comment (amh4r) [inline ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx:109]:
```suggestion
  const logRetention = data?.account.entitlements.history.limit || 7;
```
We want to use the `Account.entitlements` since `BillingPlan.entitlements` doesn't consider account overrides. Also, if we ever create a history addon then it'll reflect in `Account.entitlements` and not `BillingPlan.entitlements`

Comment (anafilipadealmeida) [reply ui/apps/dashboard/src/app/(organization-active)/(dashboard)/env/[environmentSlug]/functions/[slug]/(dashboard)/DashboardTimeRangeFilter.tsx:109]: The reason I was doing it is because I didn't wanna mess with those old time filters, that map the upgrades to specific plans. That's the reason for my comment in the PR about centralizing the logic, so we can get rid of this old code. To do what you are suggesting I need to rip that apart, which is fine but additional scope
Screenshot 2024-12-10 at 18 32 51

amh4r (approved):

PR #2015, opened 2024-12-10
fix: Update url for useFeatureFlags hook to support non default port

cmanou:
## Description

This pull request includes a small change to the `useFeatureFlags` function in the `useFeatureFlags.ts` file. The change modifies the URL used to fetch feature flags to use a relative path instead of an absolute URL.

* [`ui/apps/dev-server-ui/src/hooks/useFeatureFlags.ts`](diffhunk://#diff-c3a1048870870be173548b2354dc338f6cdf0092e7752da5485d519b00f2338cL15-R15): Changed the fetch URL from 'http://localhost:8288/dev' to '/dev' to use a relative path.

This follows the pattern of in the devApi of using a relative path https://github.com/inngest/inngest/blob/07e770d510baa55e4a7a20849193a0fb606af7b7/ui/apps/dev-server-ui/src/store/devApi.ts#L31

## Motivation
When running the dev server on a non default port, the feature flag request fails, meaning that users can not use function run search even with the flag set. Additionally this also fixes it for user who are self hosting. 

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dev-server-ui/src/hooks/useFeatureFlags.ts
+++ b/ui/apps/dev-server-ui/src/hooks/useFeatureFlags.ts
@@ -12,7 +12,7 @@ export function useFeatureFlags() {
   useEffect(() => {
     async function fetchFeatureFlags() {
       try {
-        const response = await fetch('/dev');
+        const response = await fetch(createDevServerURL('/dev'));
         if (!response.ok) {
           throw new Error('Failed to fetch feature flags');
         }
@@ -29,4 +29,16 @@ export function useFeatureFlags() {
   }, []);
 
   return { featureFlags, loading, error };
+}
+
+/**
+ * Creates a Dev Server URL from a path. If Dev Server host is unknown, it
+ * returns the path.
+ */
+function createDevServerURL(path: string) {
+  const host = process.env.NEXT_PUBLIC_API_BASE_URL;
+  if (!host) {
+    return path;
+  }
+  return new URL(path, host).toString();
 }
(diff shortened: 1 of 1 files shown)


amh4r:
Thank you for the contribution!

Your change works in our compiled binary, but we'll need to make some other changes to support our local dev flow. When developing the Dev Server UI, we run NextJS on port `3000`. This means that `/dev` will 404 since it'll try `localhost:3000/dev` instead of `localhost:8288/dev`.

We can solve this by creating a helper function that the `NEXT_PUBLIC_API_BASE_URL` env var (which we set [here](https://github.com/inngest/inngest/blob/main/ui/apps/dev-server-ui/.env.development#L1)):
```ts
/**
 * Creates a Dev Server URL from a path. If Dev Server host is unknown, it
 * returns the path.
 */
function createDevServerURL(path: string) {
  const host = process.env.NEXT_PUBLIC_API_BASE_URL;
  if (!host) {
    return path;
  }
  return new URL(path, host).toString();
}
```

(Feel free to put that function in `useFeatureFlags.ts`. We can reorganize it into a utility package after this PR merges.)

Then you can run `const response = await fetch(createDevServerURL("/dev"))` inside the `useFeatureFlags` hook

cmanou:
@amh4r  have added the fix like you said, apologies for not catching this when I looked into the other code from the devapi.ts. Will leave it to you to decide how you want to use the helper for the other files

amh4r (approved):

amh4r:
Looks good! Thanks for the fix

PR #2018, opened 2024-12-11
Fix invalid waitForEvent expression not failing run

amh4r:
## Description
Fix invalid `waitForEvent` expression not causing the run to fail. Instead, the `waitForEvent` would timeout.

The fix was a simple expression validation and returned error, but more refactoring was necessary to make history look correct. There's till an outstanding issue with the `step.waitForEvent` missing in the timeline, but that's a more significant change.

## Testing
Existing automated tests and manual testing.
image


pkg/execution/executor/executor.go
```diff
@@ -887,6 +887,48 @@ func (e *executor) HandleResponse(ctx context.Context, i *runInstance) error {
 		"workflow_id", i.md.ID.FunctionID.String(),
 	)
 
+	// This is a success, which means either a generator or a function result.
+	if i.resp.Err == nil && len(i.resp.Generator) > 0 {
```
jpwilliams, line 891: A method for these `len(i.resp.Generator) > 0` checks would be awesome.

I remember there's some quirk here with an empty generator response also being represented by a single element with an `OpcodeNone`. 🤔

jpwilliams (approved):

pkg/execution/executor/executor.go
amh4r: Like this?
```go
// IsFunctionResult returns true if the response is a function result. It will
// return false if it's a step result.
func (r *DriverResponse) IsFunctionResult() bool {
	for _, op := range r.Generator {
		if op.Op != enums.OpcodeNone {
			return true
		}
	}
	return false
}
```

jpwilliams:
LGTM 🥳

PR #2036, opened 2024-12-17
Fix `RunJobs` not selecting the correct queue shard

BrunoScheufler:
## Description

`RunJobs` was hardcoded to use the primary queue shard, which defaults to the default shard for services that didn't configure the primary queue shard environment variable. This means we never returned items from the proper queue shards for `finalize`/`Cancel` invocations.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/apiv1/apiv1.go
+++ b/pkg/api/apiv1/apiv1.go
@@ -13,6 +13,7 @@ import (
 	"github.com/inngest/inngest/pkg/cqrs"
 	"github.com/inngest/inngest/pkg/execution"
 	"github.com/inngest/inngest/pkg/execution/queue"
+	"github.com/inngest/inngest/pkg/execution/state/redis_state"
 	"github.com/inngest/inngest/pkg/headers"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/api/apiv1/apiv1.go
```diff
@@ -37,6 +38,8 @@ type Opts struct {
 	JobQueueReader queue.JobQueueReader
 	// CancellationReadWriter reads and writes cancellations to/from a backing store.
 	CancellationReadWriter cqrs.CancellationReadWriter
+	// QueueShardSelector determines the queue shard to use
+	QueueShardSelector redis_state.ShardSelector
```
djfarrelly, line 42: `QueueShardSelector` is not added added in the dev or start commands for the CLI. Should it be? It looks like `AddRoutes` only handles `AuthFinder` or `AuthMiddleware` as nil, so it might be good to implement it or use a default for the CLI if `GetFunctionRunJobs` is called, which it could be.
 https://github.com/inngest/inngest/blob/b84a0e5fe2437e722145ca0938ffedb6dec39701/pkg/devserver/devserver.go#L389-L396
https://github.com/inngest/inngest/blob/b84a0e5fe2437e722145ca0938ffedb6dec39701/pkg/devserver/devserver.go#L389-L396

djfarrelly (commented):
I think `QueueShardSelector` needs to be implemented by default in the CLI. I may be wrong, but we need to confirm prior to merging.

pkg/api/apiv1/apiv1.go
BrunoScheufler: Oh funny, somehow this didn't get suggested in my IDE -- will fix!

amh4r (approved):

PR #2049, opened 2024-12-23
Fix flaky tests

amh4r:


tests/client/function_run.go
```diff
@@ -426,16 +426,28 @@ func (c *Client) ExpectSpanOutput(t require.TestingT, expected string, output *m
...
-	require.Nil(t, output.Data)
-	require.NotNil(t, output.Error)
+func (c *Client) ExpectSpanErrorOutput(
+	t require.TestingT,
+	msg string,
+	stack string,
+	output *models.RunTraceSpanOutput,
+) {
+	a := assert.New(t)
+	if !a.NotNil(output) {
```
tonyhb, line 436: If output is nil, shouldn't we throw an error here?  We're expecting some sort of msg and trace within a nil span which to looks like it should be a failure

amh4r: I'm not sure what you mean. We're asserting that output is nil so we'll get an assertion error here and fail the test if it's nil

tonyhb: SG, thanks!

tonyhb (approved):

PR #2063, opened 2025-01-06
Introduce Connect Gateway Metrics

BrunoScheufler:
- [ ] Docs

## Checklist


djfarrelly (approved):
Tags look good. Not too much cardinality which is good. 

I know this follows existing conventions, but I'm partial to the Datadog style dot notation for let-to-right namespacing so it's easier to read and understand how things are grouped:
```bash
# Note - maybe total suffix isn't need either?
"connect_gateway.router.pubsub_messages.total"
"connect_gateway.worker_messages.total"

# Maybe these three could be grouped?
"connect_gateway.connections.attempts.total" 
"connect_gateway.connections.successes.total"
"connect_gateway.connections.active.total" # instead of _active_connections?

# instead of connect_active_gateway_total, etc. - 
# these also could just be tags for the same metric 🤔
"connect_gateway.gateways.active"
"connect_gateway.gateways.draining"
"connect_gateway.gateways" {tag: status: draining|active}

# Not really sure how to nicely group these, but perhaps there is a 
# good name spacing pattern beyond "connect_proxy"
"connect_proxy.executor_request_time_to_worker_ack.duration"
```

PR #2081, opened 2025-01-13
Fix history for v2 SDK parallel steps

amh4r:
## Description
Fix history incorrectly showing the run as failed for v2 SDK parallel steps.

## Testing
The following function with `v2.7.2` of the TS SDK:
```ts
inngest.createFunction(
  {
    id: "fn-1",
    name: "fn-1",
    retries: 1,
  },
  { event: "event-1" },
  async ({ step }) => {
    await Promise.all([
      step.run("b.1", async () => {}),
      step.run("b.2", async () => {}),
    ]);
  }
);
```

Results in the following history before this PR's change:
image

And the following history after this PR's change:
image

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -986,7 +986,9 @@ func (e *executor) HandleResponse(ctx context.Context, i *runInstance) error {
 		return i.resp
 	}
 
-	if i.resp.IsFunctionResult() {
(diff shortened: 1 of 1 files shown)


djfarrelly (approved):
Tested myself locally with a parallel 2.7.2 function.

Linked related PR which caused regression for posterity: https://github.com/inngest/inngest/pull/2018

PR #2088, opened 2025-01-17
Ensure `cause` if captured in `inngest/function.failed`

jpwilliams:
## Description

Fixes the `error` property of an `inngest/function.failed` event missing a `cause` if one had been specified.

> [!NOTE]
> We have two competing definitions here for this type of error: `StandardError` and `UserError`. We should unify these.

- [ ] Docs

## Checklist

## Related

- Fixes inngest/inngest-js#797


pkg/execution/state/driver_response.go
```diff
@@ -402,6 +405,12 @@ func (r *DriverResponse) StandardError() StandardError {
 				}
 			}
 		}
+
+		if cause, ok := processed["cause"].(map[string]any); ok {
+			if causeByt, err := json.Marshal(cause); err == nil {
+				_ = json.Unmarshal(causeByt, &ret.Cause)
```
tonyhb, line 411: Is it possible to do this without marshalling then unmarshalling into a map?  This seems like (at scale) a bunch of throwaway CPU and memory

tonyhb (approved):

pkg/execution/state/driver_response.go
jpwilliams: `cause` can now be any value, so we're no longer doing the dance here.

PR #2097, opened 2025-01-22
Add app_id to function system events

djfarrelly:
## Description

This change adds a new `app_id` field which is the user-facing App ID that they define when they create an Inngest client and see this in the UI. This will enable the user to filter system events by App ID or use the value of the app in external systems. 

Seeking feedback on this change if there would be any further ramifications or challenges to this as it's a change to a system event that requires proper review.

Screen Shot 2025-01-22 at 10 51 40 AM

## Motivation
[User request](https://discord.com/channels/842170679536517141/1329076807080738848/1331304612086415401) to be able to use this ID for event filtering.

- [ ] Docs

## Checklist


pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -93,6 +93,13 @@ func (w wrapper) LoadFunction(ctx context.Context, envID, fnID uuid.UUID) (*stat
 		return nil, err
 	}
 
+	app, err := w.GetAppByID(ctx, fn.AppID)
+	if err != nil {
+		return nil, err
+	}
+	// Append the app name as the user facing "app id"
+	def.AppID = app.Name
```
djfarrelly, line 101: Note - this approach would need to be changed in prod to use the cache for app lookups due to the throughput. Even with the cache, adding an additional lookup for every execution would add more load on the cache and potentially a perf hit in execution. May need to re-consider to perhaps only load this upon finalize.

PR #2109, opened 2025-01-27
Dev Server UI for Connect

BrunoScheufler:
## Description

This PR includes the initial iteration of the app details page wired up to the dev server GraphQL API. 

There are some limitations
- The `workerConnections` GraphQL query only fetches the first page

- [ ] Docs

## Checklist


pkg/coreapi/gql.schema.graphql
```diff
@@ -550,3 +551,12 @@ type ConnectV1WorkerConnectionEdge {
   node: ConnectV1WorkerConnection!
   cursor: String!
 }
+
+enum AppConnectionType {
+	SERVERLESS
+	CONNECT
```
anafilipadealmeida, line 557: According to the designs the types would be serverless and persistent. Did the term change?

ui/packages/components/src/types/workers.ts
```diff
@@ -13,7 +13,7 @@ export const convertWorkerStatus = (status: WorkerStatus): GroupedWorkerStatus |
     case 'READY':
       return 'ACTIVE';
     case 'DISCONNECTED':
-      return 'FAILED';
+      return 'DISCONNECTED';
```
anafilipadealmeida, line 16: Have we moved away from 
Screenshot 2025-01-28 at 12 34 20
and are we replacing failed by disconnected (with grey instead of red)? I couldn't find where that decision was made on Slack

ui/packages/components/src/types/workers.ts
```diff
@@ -39,7 +39,7 @@ export type Worker = {
   functionCount: number;
 };
 
-export const groupedWorkerStatuses = ['ACTIVE', 'INACTIVE', 'FAILED'] as const;
+export const groupedWorkerStatuses = ['ACTIVE', 'INACTIVE', 'FAILED', 'DISCONNECTED'] as const;
```
anafilipadealmeida, line 42: In your current implementation 'FAILED' doesn't exist anymore

ui/apps/dev-server-ui/src/components/App/AppActions.tsx
```diff
@@ -1,4 +1,5 @@
 import { useState } from 'react';
+import { useRouter } from 'next/navigation';
```
anafilipadealmeida, line 2: The additions in this file are not used. However, the "View Details" link in ui/apps/dev-server-ui/src/components/App/AppCardContent.tsx should actually be a button in this file instead

pkg/coreapi/gql.schema.graphql
BrunoScheufler: Going by Dan's comment here, distinguishing not on the app type but the connection type (serverless vs. worker/connect). Persistent feels somewhat confusing

ui/packages/components/src/types/workers.ts
BrunoScheufler: Disconnected and failed are two separate states, workers can disconnect gracefully and that's totally fine and probably shouldn't be marked as a failure. Connections can also fail, which we don't track separately right now but could do in the future.

ui/packages/components/src/types/workers.ts
BrunoScheufler:
```
enum ConnectionStatus {
	CONNECTED = 0;
	READY = 1;
	DRAINING = 2;
	DISCONNECTING = 3;
	DISCONNECTED = 4;
}
```

These are the connection statuses used at the moment. We don't have a separate failed status, so we could remove that from the UI for the moment.

BrunoScheufler: We do expose a `disconnectReason` which we could display, adding more information.

PR #2121, opened 2025-01-28
ux tweak when clicking on signing keys from a branch

jacobheric:
## Description

When clicking on signing keys in a branch env the user gets directed to an interstitial page which tells them then need to manage keys from the parent branch env. But if they click on the manage keys from there it takes them to event keys not signing keys so their original intent was lost. I just added another button to the interstitial so they don't get lost.

Screenshot 2025-01-28 at 1 12 14 PM

## Motivation
Slightly better ux

- [ ] Docs

## Checklist


jacobheric:
> I'm not a huge fan of two solid primary CTAs, if they have similar importance perhaps they would work better as outlined buttons instead? 🤔

FYI, I switched it to outlined buttons!

PR #2136, opened 2025-01-31
Idea: Use the HTTP hostname in the returned WebSocket URL.

djfarrelly:
## Description

When the dev server, or self-host is running on something other than 127.0.0.1, like on a public URL or in a Docker container using a custom network hostname, the dev server should attempt to automatically re-write the websocket URL on the same hostname that the HTTP request was received on. This saves the end user from having to intercept and re-write the gateway endpoint.

## Motivation
When running an app in Docker and the dev server on a host machine or another Docker container, connect does not work without writing a custom `rewriteGatewayEndpoint` function. We should aim to avoid as much config/customization out of the box during development. For example:

```
docker run --rm --name connect-app -e INNGEST_DEV=http://host.docker.internal:8288 connect-app
```

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/connect/rest/v0/v0.go
+++ b/pkg/connect/rest/v0/v0.go
@@ -2,6 +2,8 @@ package v0
 
 import (
 	"context"
+	"net/url"
+
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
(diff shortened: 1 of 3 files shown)


BrunoScheufler:
Implemented in #2142, closing this draft 👍

PR #2143, opened 2025-02-03
Add .tool-versions file, specifying node and pnpm versions

cdzombak:
This allows working with this repository seamlessly using the 'asdf' tool version manager: https://asdf-vm.com/manage/configuration.html


.tool-versions
```diff
@@ -0,0 +1,2 @@
+nodejs 20.18.2
```
djfarrelly, line 1: Is is possible to leave a comment in this file to add what tool this file is for?

djfarrelly (approved):

.tool-versions
cdzombak: Will do in a subsequent PR!

PR #2151, opened 2025-02-04
INN-4424: new run AI traces

jacobheric:
## Description

Here's the latest wip for the new runs traces. There is a ways to go yet. But we're at least at parity with the feature set from the old traces views. Since it's getting large, I'd like to get it merged before adding too much more. It's still behind a feature flag so no real danger here. 

Of note is some new shared context plumbing that begins to to pave the way out of props drilling hell. The existing rerun from step mutation and the new invoke mutation both use it. I'll comment more about it below.

https://github.com/user-attachments/assets/6211373c-d6fb-4b16-ac51-52647e6b57cc

## Motivation
More useful traces, especially for AI.

- [ ] Docs

## Checklist


ui/packages/components/src/RunDetailsV3/InlineSpans.tsx
```diff
@@ -0,0 +1,100 @@
...
+    queuedAt: string;
+    startedAt: string | null;
+    endedAt: string | null;
+  };
+}) {
+  const queuedAt = new Date(span.queuedAt);
+  const startedAt = toMaybeDate(span.startedAt);
+  const endedAt = toMaybeDate(span.endedAt);
+
+  const delay = (startedAt ?? new Date()).getTime() - queuedAt.getTime();
```
jacobheric, line 83: I've copied most of this logic over from the old run trace views but I moved it away from our old go-like let init then conditionally mutate pattern to const and ternaries so it reads more like idiomatic typescript. Also fun fact, doing the following hides dead code from the linter:

``` let delay = 0;
  if (spans.length === 1) {
    delay = something;
  }
```

ui/packages/components/src/RunDetailsV3/StepInfo.tsx
```diff
@@ -0,0 +1,222 @@
...
+  type StepInfoWait,
+} from './types';
+import { maybeBooleanToString, type PathCreator, type StepInfoType } from './utils';
+
+type StepKindInfoProps = {
+  stepInfo: StepInfoType['trace']['stepInfo'];
+  pathCreator: StepInfoType['pathCreator'];
+};
+
+const InvokeInfo = ({
```
jacobheric, line 38: Again copied over from the old components but broke out the component building stuff the from conditional logic just to make it a bit more readable.

ui/packages/components/src/Shared/SharedContext.tsx
```diff
@@ -0,0 +1,47 @@
+'use client';
+
+import React, { createContext, useContext } from 'react';
+
+import type { InvokeRunPayload, InvokeRunResult } from './useInvokeRun';
+import type { RerunFromStepPayload, RerunFromStepResult } from './useRerunFromStep';
+
+export type SharedDefinitions = {
```
jacobheric, line 8: Heres a little shared context glue that allows us to add operations to our shared component library that defer to context specific implementations without props drilling.

ui/packages/components/src/Shared/useInvokeRun.ts
```diff
@@ -0,0 +1,40 @@
+import { useState } from 'react';
+
+import { useShared, type SharedDefinitions } from './SharedContext';
+
+export interface InvokeRunPayload {
```
jacobheric, line 5: We still have to make up type signatures for our shared operations like we did before, I don't see away around that as long as we have to different sets of generated gql types.

djfarrelly (approved):

PR #2200, opened 2025-02-21
Parallelism optimization; pending step tracking

jpwilliams:
## Description

If an SDK opts in, parallel steps are tracked, reducing the number of discovery steps during step resolution and instead waiting for all steps to complete before coalescing.

This PR does not include tooling for `group.race()` or `group.parallel()` which can come later.

A broad overview is:
- SDK must opt in by declaring they wish to use Request Version V2
- When the Executor finds multiple opcodes in an SDK response, it adds each step as a pending item
- Step resolution (success, error, or pause consumption) removes the relevant pending item
- The next discovery step is only enqueued if there are no pending items remaining

- [ ] Docs

## Checklist


pkg/execution/executor/executor.go
```diff
@@ -1712,14 +1718,15 @@ func (e *executor) Resume(ctx context.Context, pause state.Pause, r execution.Re
...
-			err := e.pm.ConsumePause(ctx, pause.ID, nil)
+			_, _, err = e.pm.ConsumePause(ctx, pause.ID, nil)
 			if err == nil || err == state.ErrPauseNotFound {
 				return nil
 			}
 			return fmt.Errorf("error consuming pause via timeout: %w", err)
 		}
 
-		if err = e.pm.ConsumePause(ctx, pause.ID, r.With); err != nil {
+		didConsume, hasPendingSteps, err := e.pm.ConsumePause(ctx, pause.ID, r.With)
```
jpwilliams, line 1728: This is the complex new return from the state store. We could replace this with a single `canContinue` bool instead of these individual vars, but it feels strange for the state store to be making that decision.

pkg/execution/executor/executor.go
```diff
@@ -1856,8 +1867,15 @@ func (e *executor) HandleGeneratorResponse(ctx context.Context, i *runInstance,
...
 
-	groups := opGroups(resp.Generator).All()
-	for _, group := range groups {
+	groups := opGroups(resp.Generator)
+
+	if stepCount > 1 && i.md.ShouldCoalesceParallelism(resp) {
+		if err := e.smv2.SavePending(ctx, i.md.ID, groups.IDs()); err != nil {
+			return fmt.Errorf("error saving pending steps: %w", err)
+		}
+	}
```
jpwilliams, line 1876: Here we save pending steps when handling the response but before handling each opcode.

pkg/execution/executor/executor.go
```diff
@@ -1977,13 +1996,16 @@ func (e *executor) handleGeneratorStep(ctx context.Context, i *runInstance, gen
...
+
+	if !hasPendingSteps {
+		err = e.queue.Enqueue(ctx, nextItem, now, queue.EnqueueOpts{})
+		if err == redis_state.ErrQueueItemExists {
+			return nil
+		}
+		if err != nil {
+			logger.StdlibLogger(ctx).Error("error scheduling step queue item", "error", err)
+			return err
+		}
```
jpwilliams, line 2008: Most enqueue calls are now wrapped in this, checking if we should continue now or whether a later execution should have that privilege.

Lifecycles are untouched and still run.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -1013,44 +1049,47 @@ func (m unshardedMgr) DeletePause(ctx context.Context, p state.Pause) error {
...
 	}
 
 	// The pause was now consumed, so let's clean up
-	return m.unshardedMgr.DeletePause(ctx, *p)
+	err = m.unshardedMgr.DeletePause(ctx, *p)
+	return
 }
 
-func (m shardedMgr) consumePause(ctx context.Context, p *state.Pause, data any) error {
+func (m shardedMgr) consumePause(ctx context.Context, p *state.Pause, data any) (didConsume bool, hasPendingSteps bool, err error) {
```
jpwilliams, line 1068: Using named return variables like this can be confusing, but I'm specifying them here so that the caller has a better idea what they are; seeing `bool, bool, err` as the return doesn't give  much info.

We could refactor to a small struct if that feels more appropriate.

pkg/execution/state/redis_state/lua/saveResponse.lua
```diff
@@ -34,4 +36,5 @@ redis.call("HINCRBY", keyMetadata, "step_count", 1)
 redis.call("HSET", keyStep, stepID, outputData)
 redis.call("RPUSH", keyStack, stepID)
 
-return 0
+redis.call("SREM", keyStepsPending, stepID)
+return redis.call("SCARD", keyStepsPending) > 0 and 1 or 0
```
BrunoScheufler, line 40: Lol is that a ternary in Lua 😄

BrunoScheufler (commented):
Mostly looks good to me, does this affect existing runs in any way?

jpwilliams:
Addressed notes, @BrunoScheufler.

Nope - no existing runs will be affected and only those where the user has opted in will use these new branches.

BrunoScheufler (approved):

pkg/execution/executor/executor.go
jpwilliams: Changed to a struct.

pkg/execution/executor/executor.go
```diff
@@ -745,6 +745,12 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		if !errors.Is(err, state.ErrDuplicateResponse) && err != nil {
 			return nil, err
 		}
+
+		if hasPendingSteps {
+			// Other steps are pending before we re-enter the function, so
+			// we're now done with this execution.
+			return nil, nil
```
amh4r, line 752: Does this introduce `nil, nil` as a return tuple for the first time? Do we know that all callers are checking if both return values are `nil`, rather than just error?

PR #2212, opened 2025-02-27
Channel APIs

tonyhb:
This introduces authenticated APIs for arbitrarily publishing data to channels, including streaming data - which will be chunked and sent to the subscribers in a streaming fashion as soon as possible.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/realtime/broadcaster_redis.go
+++ b/pkg/execution/realtime/broadcaster_redis.go
@@ -59,18 +59,7 @@ func (b *redisBroadcaster) Publish(ctx context.Context, m Message) {
 
 	for _, t := range m.Topics() {
 		go func(t Topic) {
-			cmd := b.pubc.B().Publish().Channel(t.String()).Message(string(content)).Build()
-			for i := 0; i < redisPublishAttempts; i++ {
-				err := b.pubc.Do(ctx, cmd).Error()
-				if err == nil {
-					break
-				}
-				logger.StdlibLogger(ctx).Error(
-					"error publishing to realtime redis pubsub",
-					"error", err,
-				)
-				<-time.After(redisRetryInterval)
-			}
+			b.publish(ctx, t.String(), string(content))
 		}(t)
 	}
 }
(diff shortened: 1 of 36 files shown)


pkg/execution/realtime/realtime.go
```diff
@@ -149,18 +166,27 @@ type ReadWriteSubscription interface {
 // Each message is published to one or more topics.
 type Topic struct {
 	// Kind represents the topic kind, ie. whether this topic is for events or run data.
+	// This allows consumers to stream events and data from runs separately.
 	Kind TopicKind `json:"kind"`
-	// RunID represents the run that this topic represents, if this is a
-	// topic for a run.
-	RunID ulid.ULID `json:"run_id"`
```
jpwilliams, line 155: Can we keep `RunID` in the `Topic`?

`Channel` is the run ID atm but that could change, and keeping the `RunID` here means we also don't need to do `ulid.MustParse()` checks on `Channel` and it can remain _just a string_ which feels better.

pkg/execution/realtime/realtime.go
```diff
@@ -206,21 +233,39 @@ type Message struct {
...
 	Data json.RawMessage `json:"data"`
+	// CreatedAt is the time that this message was created.
+	CreatedAt time.Time `json:"created_at"`
 
-	// FnID is the function ID that this message is related to.
-	FnID uuid.UUID `json:"fn_id,omitempty,omitzero"`
-	// FnSlug is the function slug that this message is related to.
-	FnSlug string `json:"fn_slug,omitempty,omitzero"`
-	// RunID is the run ID that this message is related to.
-	RunID ulid.ULID `json:"run_id,omitempty,omitzero"`
```
jpwilliams, line 215: Can we keep `RunID` in the `Message`?

`Channel` is the run ID atm but that could change, and keeping the `RunID` here means we also don't need to do `ulid.MustParse()` checks on `Channel` and it can remain _just a string_ which feels better.

pkg/execution/realtime/broadcaster_redis.go
```diff
@@ -75,6 +75,27 @@ func (b *redisBroadcaster) Publish(ctx context.Context, m Message) {
 	}
 }
 
+func (b *redisBroadcaster) PublishStream(ctx context.Context, m Message, data string) {
+	for _, t := range m.Topics() {
+		go func(t Topic) {
+			cmd := b.pubc.B().Publish().Channel(t.String()).Message(
+				string(m.Data) + ":" + data,
+			).Build()
+			for i := 0; i < redisPublishAttempts; i++ {
```
jpwilliams, line 84: Would a log indicating that we ran out of attempts and gave up for this message be relevant if we exhaust all retires?

It'd be nice to distinguish between recoverable retries and full failures.

pkg/execution/realtime/broadcaster_redis.go
```diff
@@ -75,6 +75,27 @@ func (b *redisBroadcaster) Publish(ctx context.Context, m Message) {
...
+				err := b.pubc.Do(ctx, cmd).Error()
+				if err == nil {
+					break
+				}
+				logger.StdlibLogger(ctx).Error(
+					"error publishing to realtime redis pubsub",
+					"error", err,
+				)
+				<-time.After(redisRetryInterval)
```
jpwilliams, line 93: A nit - we'll still wait here on the last retry even though we won't do anything afterwards.

pkg/execution/realtime/broadcaster_redis.go
```diff
@@ -75,6 +75,27 @@ func (b *redisBroadcaster) Publish(ctx context.Context, m Message) {
 	}
 }
 
+func (b *redisBroadcaster) PublishStream(ctx context.Context, m Message, data string) {
+	for _, t := range m.Topics() {
+		go func(t Topic) {
```
jpwilliams, line 80: Seeing as this might run for a little while with the retries, shall we also check for context cancellation?

jpwilliams (commented):
Nits only.

Exciting!!!!

PR #2217, opened 2025-02-28
Migrate debounce to new system cluster

BrunoScheufler:
## Description

This PR introduces a new just-in-time migration mode for debounces and creates new debounces on the new system cluster, as soon as a migration feature flag is enabled.

I have added in-depth debounce tests to validate the current implementation, and a special test suite covering the migration logic.

### Goals

- Keep debounce creations and updates flowing (debounces must not be interrupted)
  - In particular: Always use *latest event*, respect timeout
- Do not touch atomicity
   - Debounce state and debounce system queue are still colocated on the same unsharded Redis instance
- Do not change existing debounce design
- Do not run expensive migration for all debounces at once, focus on individual debounces + enable migration path to backfill inactive debounces

### High-level ideas

- Create new debounces on new system cluster
- Migrate existing debounce when it's about to be updated
  - Retrieve existing debounce details: ID and timeout 
  - In the same atomic step, prevent processing this debounce on old cluster
  - Write debounce state to new system cluster, enqueue timeout job
    - Respect timeout: Adjust TTL if timeout is earlier than next TTL
    - Maintain old ID for transparency
  - Delete old debounce state and queue item

### Follow-up

- Run over `debounce` system queue and move existing debounces to new cluster, delete on old with same method (albeit without changing the TTL/timeout)
   - This is necessary to bring over existing debounces that haven't seen updates in a while and are still running (debounce period can be very large: weeks, months, potentially years)

- [ ] Docs

## Checklist


pkg/execution/executor/service.go
```diff
@@ -361,6 +361,15 @@ func (s *svc) handleDebounce(ctx context.Context, item queue.Item) error {
 		if f.ID == d.FunctionID {
 			di, err := s.debouncer.GetDebounceItem(ctx, d.DebounceID)
 			if err != nil {
+				if errors.Is(err, debounce.ErrDebounceNotFound) {
+					// This is expected after migrating items to a new primary cluster
+					logger.StdlibLogger(ctx).Info("debounce not found during timeout job, skipping",
+						"fn_id", d.FunctionID.String(),
+						"debounce_id", d.DebounceID.String(),
+					)
+					return nil
```
darwin67, line 370: probably good to just do `continue` instead of `return`?
so it can still process others.

darwin67 (approved):

PR #2233, opened 2025-03-07
Connect for inngest lite

BrunoScheufler:
- [ ] Docs

## Checklist

### DIFF
--- a/pkg/lite/lite.go
+++ b/pkg/lite/lite.go
@@ -491,7 +491,6 @@ func start(ctx context.Context, opts StartOpts) error {
 		RequireKeys:    true,
 	})
 
-	// waiterSvc := connectpubsub.NewConnectorWaiterSvc(gatewayProxy, connGateway, connRouter)
 	return service.StartAll(ctx, ds, runner, executorSvc, ds.Apiservice, connGateway, connRouter)
 }
 
--- a/pkg/lite/lite.go
+++ b/pkg/lite/lite.go
@@ -81,7 +81,8 @@ type StartOpts struct {
 	// given key.
 	EventKey []string `json:"event_key"`
 
-	ConnectGatewayPort int `json:"connect-gateway-port"`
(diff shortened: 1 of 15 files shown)


pkg/lite/lite.go
```diff
@@ -415,11 +452,46 @@ func start(ctx context.Context, opts StartOpts) error {
...
+		return fmt.Errorf("failed to create connect pubsub connector: %w", err)
+	}
+
+	connGateway := connect.NewConnectGatewayService(
+		connect.WithConnectionStateManager(connectionManager),
+		connect.WithRequestReceiver(gatewayRequestReceiver),
+		connect.WithGatewayAuthHandler(auth.NewJWTAuthHandler(consts.DevServerConnectJwtSecret)),
+		connect.WithDev(),
+		connect.WithGatewayPublicPort(opts.ConnectGatewayPort),
+		connect.WithApiBaseUrl(fmt.Sprintf("http://127.0.0.1:%d", opts.Config.EventAPI.Port)),
```
djfarrelly, line 481: Two notes:
* Should this hostname come from the config instead of hard coded? ex. `opts.Config.CoreAPI.Addr` Is the hostname. 
* Should the port be `opts.Config.CoreAPI.Port` instead of `EventAPI`? I know they're the same currently, but they could change in the future.

Looking at `start.go`, we also would need to set the CoreAPI port/host in additional to the eventapi port/host for this to work I think. Making host driven from config/ops could fix a couple weird things for self-hosting on different hostnames within a network.

djfarrelly: Can we also pass this API or hostname into the `RetrieveGateway` function in `service.go`? Right now it's hard coded to `127.0.0.1`, but pushing the `host` flag/config option would allow the dev server and start to work without necessarily needing `rewriteGatewayEndpoint` at all.

jpwilliams (approved):
🥳

PR #2250, opened 2025-03-15
Fix connect syncing issue

BrunoScheufler:
## Description

This PR fixes a bug with syncing for connect in the dev server.

With connect, we store metadata for connections and their related worker groups. Each group represents an app version at a given point in time. If the last connection for a worker group is disconnected, the worker group is dropped from the metadata store.

If another connection for the same app is established, the connect infrastructure will not find an existing worker group and attempt to send an out-of-band sync request. This should return the App ID and Sync ID. In the special case of re-connecting with the same app version, these should match the initial sync ID and app ID.

The function registration/syncing backend must thus support the IdempotencyKey sent in the connect sync request and return the same sync ID and app ID in case the current deployment matches the idempotency key. If this is not the case (for example because a worker with a new version is connecting), a new sync should be triggered.

In all cases, the sync API should return an app ID and sync ID. Otherwise, the connect backend is unable to update the internal state used for routing, causing "no healthy connections found" errors.

This PR updates the devserver sync API to always return the app ID, which fixes re-connecting with the same app version.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/connect/state/state.go
+++ b/pkg/connect/state/state.go
@@ -146,7 +146,13 @@ func (c *Connection) AppNames() []string {
 	return appNames
 }
 
-// Sync attempts to sync the worker group configuration
+// Sync attempts to sync the worker group configuration.
+//
+// - If a previous worker group with the same hash exists, the current worker group is updated, so App ID and Sync ID are provided.
+// - In case no previous worker group with the same hash exists (or all existing connections disconnected and triggered a cleanup),
+// an out-of-band Sync request is sent to the API. This is expected to handle idempotency, so subsequent calls return the same App ID and Sync ID
+// given the same idempotency key.
+// - To enable rollback functionality, the API should trigger a new sync if, and only if, the requested idempotency key does not match the current deploy.
 func (g *WorkerGroup) Sync(ctx context.Context, groupManager WorkerGroupManager, apiBaseUrl string, initialReq *connpb.WorkerConnectRequestData) error {
 	// The group is expected to exist in the state, as UpsertConnection also creates the group if it doesn't exist
 	existingGroup, err := groupManager.GetWorkerGroupByHash(ctx, g.EnvID, g.Hash)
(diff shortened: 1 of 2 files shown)


pkg/connect/state/state.go
```diff
@@ -268,15 +274,18 @@ func (g *WorkerGroup) Sync(ctx context.Context, groupManager WorkerGroupManager,
...
-	if syncReply.IsSuccess() {
-		g.AppID = syncReply.AppID
-		g.SyncID = syncReply.SyncID
-
-		// Update the worker group with the syncID so it's aware that it's already sync'd before
-		// Always update the worker group for consistency, even if the context is cancelled
-		if err := groupManager.UpdateWorkerGroup(context.Background(), g.EnvID, g); err != nil {
-			return fmt.Errorf("error updating worker group: %w", err)
-		}
+	if !syncReply.IsSuccess() {
```
BrunoScheufler, line 277: Previously, we ignored invalid sync responses. That's not good. We should always prevent connections from being established in case of a connection requiring a successful sync and not receiving a successful response.

pkg/connect/state/state.go
```diff
@@ -268,15 +274,18 @@ func (g *WorkerGroup) Sync(ctx context.Context, groupManager WorkerGroupManager,
...
+		// We always expect the App ID & Sync ID to be included in a sync result, representing either the idempotent reply or the new sync.
+		return fmt.Errorf("invalid sync result")
+	}
+
+	g.AppID = syncReply.AppID
```
BrunoScheufler, line 282: This is the crucial line for routing messages. If this is not set, the next UpsertConnection call will not create the proper indexes in the metadata store and messages will fail to route to the connection because it can't be identified as responsible for the app.

pkg/devserver/api.go
```diff
@@ -202,11 +202,18 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*cqrs.Sync
...
+	syncID := uuid.New()
+
 	if app, err := a.devserver.Data.GetAppByChecksum(ctx, consts.DevServerEnvID, sum); err == nil {
 		if !app.Error.Valid {
 			// Skip registration since the app was already successfully
 			// registered.
-			return &cqrs.SyncReply{OK: true}, nil
+			return &cqrs.SyncReply{
+				OK:     true,
+				AppID:  &app.ID,
```
BrunoScheufler, line 214: We always, always, always have to return the App ID in case of a connect sync. Otherwise, we can't associate connections with apps and routing will fail

tonyhb (approved):

PR #2262, opened 2025-03-21
Add dialer

tonyhb:
Performs DNS lookups and intercepts dialing.

- [x] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/driver/httpdriver/dialer.go
+++ b/pkg/execution/driver/httpdriver/dialer.go
@@ -1,0 +1,128 @@
+package httpdriver
+
+import (
+	"context"
+	"fmt"
+	"net"
+)
+
+var privateIPBlocks []*net.IPNet
+var nat64blocks []*net.IPNet
+
+func init() {
+	for _, cidr := range []string{
+		"127.0.0.0/8",    // IPv4 loopback
+		"10.0.0.0/8",     // RFC1918
+		"172.16.0.0/12",  // RFC1918
+		"192.168.0.0/16", // RFC1918
+		"169.254.0.0/16", // RFC3927 link-local
+		"::1/128",        // IPv6 loopback
+		"fe80::/10",      // IPv6 link-local
+		"fc00::/7",       // IPv6 unique local addr
+		"ff00::/8",       // multicast
+		"fec0::/10",      // deprecated
+	} {
+		_, block, err := net.ParseCIDR(cidr)
+		if err != nil {
+			panic(fmt.Errorf("parse error on %q: %v", cidr, err))
+		}
+		privateIPBlocks = append(privateIPBlocks, block)
+	}
+
+	for _, cidr := range []string{
+		"64:ff9b::/96",       // RFC 6052 suffix
+		"2001:db8::/32",      // RFC 6052 bits 32-63
+		"2001:db8:aaaa::/48", // RFC 6052 bits 48-87
+	} {
+		_, block, err := net.ParseCIDR(cidr)
+		if err != nil {
+			panic(fmt.Errorf("parse error on %q: %v", cidr, err))
+		}
+		nat64blocks = append(nat64blocks, block)
+	}
+
+}
+
+type SecureDialerOpts struct {
+	AllowHostDocker bool
+	AllowPrivate    bool
+	AllowNAT64      bool
+}
+
+type DialFunc = func(ctx context.Context, network, addr string) (net.Conn, error)
+
+func SecureDialer(o SecureDialerOpts) DialFunc {
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		// network will be one of the well defined networks as per
+		// https://pkg.go.dev/net#Dial, eg "tcp", "tcp4", "tcp6", etc.
+		//
+		// addr may be a domain or ip and port: "example.com:443", "192.0.2.1:http",
+		// "[fe80::1%lo0]:53".
+		//
+		// We always want to ensure we translate the domains to IP addresses.
+		host, _, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+
+		if !o.AllowHostDocker && isDockerHost(host) {
+			return nil, fmt.Errorf("Unable to make request to %s at IP %s: accessing docker host", addr, host)
+		}
+
+		// Ensure that the current hostname is not
+		addrs, err := net.DefaultResolver.LookupHost(ctx, host)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, a := range addrs {
+			if !o.AllowPrivate && isPrivateHost(a) {
+				return nil, fmt.Errorf("Unable to make request to %s at IP %s: private IP range", addr, a)
+			}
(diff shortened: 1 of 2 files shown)


BrunoScheufler (approved):

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+		return dialer.DialContext(ctx, network, addr)
+	}
+}
+
+func isDockerHost(host string) bool {
+	return host == "host.docker.internal"
+}
+
+func isPrivateHost(host string) bool {
+	if host == "localhost" || host == "0.0.0.0" {
```
cdzombak, line 97: This is missing `localhost.localdomain` which is commonly defined in `/etc/hosts`.

Also, `0.0.0.0` is not a good check here; it's missing the IPv6 equivalent `::/0`. And it fails to consider alternate representations (e.g. `0000:0000:0000:0000:0000:0000:0000:0000/0` etc). This should use a proper IP address API like the checks below rather than string comparisons.

Edited to add: or, better, use [IsUnspecified](https://pkg.go.dev/net#IP.IsUnspecified).

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+
+		// Return the default dialer in the http package
+		return dialer.DialContext(ctx, network, addr)
+	}
+}
+
+func isDockerHost(host string) bool {
+	return host == "host.docker.internal"
```
cdzombak, line 93: We may want to consider denying all `*.internal` addresses, not just this specific one. Also `.lan` and the following common private TLDs:

- `.intranet`
- `.internal`
- `.private`
- `.corp`
- `.home`

per https://www.rfc-editor.org/rfc/rfc6762#appendix-G .

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+		host, _, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+
+		if !o.AllowHostDocker && isDockerHost(host) {
+			return nil, fmt.Errorf("Unable to make request to %s at IP %s: accessing docker host", addr, host)
+		}
+
+		// Ensure that the current hostname is not
```
cdzombak, line 72: Half of this comment is missing.

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+			if !o.AllowPrivate && isPrivateHost(a) {
+				return nil, fmt.Errorf("Unable to make request to %s at IP %s: private IP range", addr, a)
+			}
+			if !o.AllowNAT64 && isNat64(a) {
+				return nil, fmt.Errorf("Unable to make request to %s at IP %s: NAT64 address", addr, a)
+			}
+		}
+
+		// Return the default dialer in the http package
+		return dialer.DialContext(ctx, network, addr)
```
cdzombak, line 88: Why not pass the resolved addr to go's net.Dialer?

This would:
(1) avoid a double DNS lookup
(2) prevent a theoretical vuln if `net.Dialer` sees a different DNS result than `SecureDialer`

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+func init() {
+	for _, cidr := range []string{
+		"127.0.0.0/8",    // IPv4 loopback
+		"10.0.0.0/8",     // RFC1918
+		"172.16.0.0/12",  // RFC1918
+		"192.168.0.0/16", // RFC1918
+		"169.254.0.0/16", // RFC3927 link-local
+		"::1/128",        // IPv6 loopback
+		"fe80::/10",      // IPv6 link-local
+		"fc00::/7",       // IPv6 unique local addr
```
cdzombak, line 21: Add RFC to comment: https://www.rfc-editor.org/rfc/rfc4193.html

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"net"
+)
+
+var privateIPBlocks []*net.IPNet
+var nat64blocks []*net.IPNet
+
+func init() {
+	for _, cidr := range []string{
+		"127.0.0.0/8",    // IPv4 loopback
```
cdzombak, line 14: Why not use [`IsLoopback`](https://pkg.go.dev/net#IP.IsLoopback) instead?

pkg/execution/driver/httpdriver/dialer.go
```diff
@@ -0,0 +1,128 @@
...
+		"10.0.0.0/8",     // RFC1918
+		"172.16.0.0/12",  // RFC1918
+		"192.168.0.0/16", // RFC1918
+		"169.254.0.0/16", // RFC3927 link-local
+		"::1/128",        // IPv6 loopback
+		"fe80::/10",      // IPv6 link-local
+		"fc00::/7",       // IPv6 unique local addr
+		"ff00::/8",       // multicast
```
cdzombak, line 22: Why not use [`IsMulticast`](https://pkg.go.dev/net#IP.IsMulticast)?

cdzombak (changes requested):

pkg/execution/driver/httpdriver/dialer.go
tonyhb: We can only pass one address and a domain name may resolve to many.  The net.Dialer then has its own connection logic for managing which IPs to hit..  We don't want to rebuild this.

pkg/execution/driver/httpdriver/dialer.go
tonyhb: we can, but this is now one loop for all banned IPs.

pkg/execution/driver/httpdriver/dialer.go
tonyhb: we can, but this is now one loop for all banned IPs.

pkg/execution/driver/httpdriver/dialer.go
tonyhb: This is a fast path.  We do check for ::/0 in the banned IPs, but will add.

pkg/execution/driver/httpdriver/dialer.go
tonyhb: We also do call that in `isPrivateIP` :)

cdzombak (approved):

pkg/execution/driver/httpdriver/dialer.go
tonyhb: Added that domain, thanks!

pkg/execution/driver/httpdriver/dialer.go
tonyhb: When those resolve to internal IPs we should be set.  Thank you for being thorough!

cdzombak (approved):

PR #2287, opened 2025-03-28
Add support for `x-inngest-event-seed-id` header

amh4r:
## Description
Add support for the `x-inngest-event-seed-id` header. Its value contains the data needed to deterministically generate a ULID (time and entropy). When it's set (and is valid), the Event API uses it to deterministically generate the internal event ID.

This is useful for idempotent retries when clients send events. It solves 2 issues:
- Our UI shows 2 IDs (idempotency ID and internal ID), which is confusing.
- There are some scenarios where `client.send` returns a different internal ID than the one that _actually_ triggered the run.

## Consequences
It's possible that there could be multiple DB rows with the same internal event ID. This is rare and we can't think of any significant issues caused by it, but we should still solve it in the future (e.g. replacing merge tree in ClickHouse)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -261,7 +261,7 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 			defer span.End()
 
 			seed := event.SeededIDFromString(
-				r.Header.Get("x-inngest-idempotency-key"),
+				r.Header.Get(headers.HeaderEventIDSeed),
 				index,
 			)
 			id, err := a.handler(ctx, &evt, seed)
@@ -333,7 +333,7 @@ func (a API) Invoke(w http.ResponseWriter, r *http.Request) {
 	evt := event.NewInvocationEvent(newInvOpts)
 
 	seed := event.SeededIDFromString(
-		r.Header.Get("x-inngest-idempotency-key"),
+		r.Header.Get(headers.HeaderEventIDSeed),
 		0,
 	)
 	evtID, err := a.handler(r.Context(), &evt, seed)
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -24,7 +24,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-type EventHandler func(context.Context, *event.Event) (string, error)
+type EventHandler func(context.Context, *event.Event, *event.SeededID) (string, error)
 
 type Options struct {
 	Config config.Config
@@ -220,7 +220,9 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 		// Close the idChan so that we stop appending to the ID slice.
 		defer close(idChan)
 
+		index := 0
 		for s := range stream {
+			index++
 			evt := event.Event{}
 			if err := json.Unmarshal(s.Item, &evt); err != nil {
 				return err
@@ -258,7 +260,11 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 				)
 			defer span.End()
 
-			id, err := a.handler(ctx, &evt)
+			seed := event.SeededIDFromString(
+				r.Header.Get("x-inngest-idempotency-key"),
+				index,
+			)
+			id, err := a.handler(ctx, &evt, seed)
 			if err != nil {
 				a.log.Error().Str("event", evt.Name).Err(err).Msg("error handling event")
 				return err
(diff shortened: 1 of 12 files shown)


pkg/event/event.go
```diff
@@ -205,9 +277,17 @@ func (e Event) InngestMetadata() (*InngestMetadata, error) {
...
+	if seed != nil {
+		newInternalID, err := seed.ToULID()
+		if err == nil {
+			internalID = newInternalID
+		}
+	}
+
 	if e.ID == "" {
 		e.ID = internalID.String()
 	}
```
jpwilliams, line 293: Shouldn't we use this to only overwrite the `ID` and not touch the `internalID`? `ossTrackedEvent.Id` should always be ours and protected.
```suggestion
	if seed != nil && e.ID == "" {
		newIdempotencyKey, err := seed.ToULID()
		if err == nil {
			e.ID = newIdempotencyKey.String()
		}
	}
```

jpwilliams (commented):
Does this also address the issue whereby a user will see an `id` in their payload that they did not generate?

It may be, for that, that we actually use a top-level `idempotencyKey` in `ossTrackedEvent`?

pkg/event/event.go
jpwilliams: No. Events are not filtered at the point of ingestion - see inngest/monorepo#2682.

tests/golang/send_idempotent_retry_test.go
```diff
@@ -0,0 +1,163 @@
...
+			{Name: eventName},
+			{Name: eventName},
+		},
+	)
+	r.NoError(err)
+
+	// Sleep long enough for the Dev Server to process the events.
+	time.Sleep(5 * time.Second)
+	r.Equal(int32(2), atomic.LoadInt32(&proxyCounter))
+	r.Equal(int32(2), atomic.LoadInt32(&fnCounter))
```
jpwilliams, line 104: (Discussed) Can we also add an assertion here that we expected to successfully have ingested 4 events?

Intended as a way to prove the behaviour that causes some of the decisions in this and other PRs.

tests/golang/send_idempotent_retry_test.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// sendEvents mimics the idempotent retry logic of the inngestgo client.
+//
+// TODO: Move this to the inngestgo package once we implement the idempotent
+// retry logic in the Dev Server and Cloud.
+func sendEvents(
```
jpwilliams, line 111: An AI bot that scans for `TODO`s after you release code in another repo and figures out what you can now update/change would be great. 😄

pkg/event/event.go
```diff
@@ -205,9 +277,17 @@ func (e Event) InngestMetadata() (*InngestMetadata, error) {
...
+func NewOSSTrackedEvent(e Event, seed *SeededID) TrackedEvent {
 	// Never use e.ID as the internal ID, since it's specified by the sender
 	internalID := ulid.MustNew(ulid.Now(), rand.Reader)
+
+	if seed != nil {
+		newInternalID, err := seed.ToULID()
+		if err == nil {
+			internalID = newInternalID
+		}
+	}
```
jpwilliams, line 289: Can we add a comment here to note the potential downstream effect of having duplicate IDs in the store?

As you say, it shouldn't actually affect downstream queries, but will be useful if someone finds two events and is confused.

pkg/api/api.go
```diff
@@ -258,7 +260,11 @@ func (a API) ReceiveEvent(w http.ResponseWriter, r *http.Request) {
 				)
 			defer span.End()
 
-			id, err := a.handler(ctx, &evt)
+			seed := event.SeededIDFromString(
+				r.Header.Get("x-inngest-idempotency-key"),
```
jpwilliams, line 264: (Discussed - using as a reminder)
```suggestion
				r.Header.Get("x-inngest-event-id-seed"),
```
Nit: Could this also be a const somewhere?

pkg/api/api.go
```diff
@@ -326,7 +332,11 @@ func (a API) Invoke(w http.ResponseWriter, r *http.Request) {
 	}
 	evt := event.NewInvocationEvent(newInvOpts)
 
-	evtID, err := a.handler(r.Context(), &evt)
+	seed := event.SeededIDFromString(
+		r.Header.Get("x-inngest-idempotency-key"),
```
jpwilliams, line 336: (Discussed - using as a reminder)
```suggestion
		r.Header.Get("x-inngest-event-id-seed"),
```
Nit: Could this also be a const somewhere?

jpwilliams (commented):
> [!NOTE]
> Events are not filtered at the point of ingestion, and the idempotency key (regardless of the field used) is only utilized to filter _runs_.
>
> This is done for performance as well as allowing users to see duplicate events and understand why runs were not triggered.
>
> See inngest/monorepo#2682.

PR #2298, opened 2025-04-02
Migrate debounces by account

BrunoScheufler:
## Description

This PR adds account IDs to the debounce migration feature flag. This allow us to toggle the migration (create new debounces on new system queue + migrate existing debounces JIT) for individual accounts. When the flag evaluates to false, the system will work exactly as before (creating/updating/deleting on the old cluster)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/debounce/debounce.go
+++ b/pkg/execution/debounce/debounce.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/inngest/inngest/pkg/enums"
+	"github.com/inngest/inngest/pkg/telemetry/metrics"
 	"github.com/jonboulle/clockwork"
 	"io/fs"
 	"regexp"
@@ -26,6 +27,10 @@ import (
 	"github.com/oklog/ulid/v2"
 	"github.com/redis/rueidis"
 	"github.com/xhit/go-str2duration/v2"
+)
+
+const (
+	pkgName = "execution.debounce"
 )
 
 //go:embed lua/*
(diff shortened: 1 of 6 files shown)


pkg/execution/debounce/debounce.go
```diff
@@ -238,12 +248,34 @@ func (d debouncer) queueManager(shouldMigrate bool) redis_state.QueueManager {
...
+	if queueShard.Name == "" {
+		return fmt.Errorf("queueShard did not return QueueShard")
+	}
+
+	err := d.deleteDebounceItem(ctx, debounceID, client)
+	if err != nil {
+		return fmt.Errorf("could not delete debounce item: %w", err)
+	}
+
+	metrics.IncrQueueDebounceOperationCounter(ctx, metrics.CounterOpt{
```
darwin67, line 270: I would prefer this metrics to be called in a `defer` function.
So we can get metrics regardless of success of failure.
Obviously you'll need to change the tag if it fails so we can differentiate the metric.

BrunoScheufler: Sure can do

pkg/execution/debounce/debounce.go
```diff
@@ -307,6 +355,14 @@ func (d debouncer) StartExecution(ctx context.Context, di DebounceItem, fn innge
 
 	switch res {
 	case 0, 1:
+		metrics.IncrQueueDebounceOperationCounter(ctx, metrics.CounterOpt{
```
darwin67, line 358: probably good to add this metric on `default` case as well to provide better visibility with a different `op` value.

pkg/execution/debounce/debounce.go
```diff
@@ -238,12 +248,34 @@ func (d debouncer) queueManager(shouldMigrate bool) redis_state.QueueManager {
 }
 
 // DeleteDebounceItem removes a debounce from the map.
-func (d debouncer) DeleteDebounceItem(ctx context.Context, debounceID ulid.ULID) error {
-	shouldMigrate := d.shouldMigrate(ctx)
+func (d debouncer) DeleteDebounceItem(ctx context.Context, debounceID ulid.ULID, accountID uuid.UUID) error {
+	// Determine the flag value once and pass down to prevent inconsistent values mid-deletion
+	shouldMigrate := d.shouldMigrate(ctx, accountID)
```
darwin67, line 253: nit:

if `d.shouldMigrate` is not an expensive call, it seem pretty redundant to pass the value around for other internal function calls.
seems okay to just embed into `d.client` and `d.queueShard` since they both can just call `d.shouldMigrate` themselves.

darwin67 (approved):

PR #2318, opened 2025-04-07
feat: Add indexes to increase performance of self hosting when using postgres

cmanou:
## Description

Add strategic database indexes to improve query performance for common queries in the PostgreSQL database. This includes indexes on function_runs (for run_id, event_id, and 
  time-based function lookups), function_finishes (for run status lookups), events (for ID and time-range queries), history (for run-related queries), and traces tables.

  Note: I initially planned to use CREATE INDEX CONCURRENTLY to avoid table locks, but since the migration system runs migrations within transactions and CONCURRENTLY cannot be used
  inside transactions, I've used regular CREATE INDEX statements. Alternatively we could add look at turning on multi statement execution (https://github.com/golang-migrate/migrate/tree/master/database/postgres) which I think would work. 

## Motivation
After a month of using self hosting with postgres the dashboard becomes unusable nothing loads and it basically tanks the DB any time you try load a trace

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/cqrs/base_cqrs/migrations/postgres/000012_add_pg_indexes.down.sql
+++ b/pkg/cqrs/base_cqrs/migrations/postgres/000012_add_pg_indexes.down.sql
@@ -1,0 +1,19 @@
+-- Drop indexes from Function Runs Table
+DROP INDEX IF EXISTS idx_function_runs_run_id;
+DROP INDEX IF EXISTS idx_function_runs_event_id;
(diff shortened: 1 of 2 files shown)


cmanou:
@jpwilliams and @djfarrelly  as discussed a while back in discord have tested some indexes which greatly improve perf for self hosted postgres

tonyhb (approved):

PR #2325, opened 2025-04-09
Block storage for pauses

tonyhb:
- [ ] Docs

## Checklist


pkg/execution/pauses/serializer.go
```diff
@@ -0,0 +1,49 @@
+package pauses
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const (
+	encodingJSON = 0x00
```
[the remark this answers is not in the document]
darwin67: also probably good to have some unit test for serializer/deserializer.

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	// Unmarshal the block, using the first byte to figure out encoding.
+	return Deserialize(byt)
+}
+
+// Delete deletes a pause from the backing blob.  Note that blobs are immutable;  we cannot
+// yeet the pause out of a blob as-is.  Instead, we track which blocks have deleted pauses
+// via indexes, and eventually compact blocks.
+func (b blockstore) Delete(ctx context.Context, index Index, pause state.Pause) error {
```
darwin67, line 290: should this operation be transactional?
what happens when there are network issues, io timeouts, context cancel, etc?
won't there be a ton of left over garbage?

also won't these much operations contribute to higher qps?

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+	if err := b.rc.Do(ctx, cmd).Error(); err != nil {
+		return err
+	}
+
+	return b.rc.Do(
+		ctx,
+		b.rc.B().
+			Hset().
+			Key(b.blockMetadataKey(idx)).
+			FieldValue().
```
darwin67, line 415: should these be transactions?
what do we do when we have failures?

pkg/execution/pauses/block_leaser.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Revoke drops a lease, allowing any other worker to flush an index.
+func (r redisBlockLeaser) Revoke(ctx context.Context, index Index, leaseID ulid.ULID) (err error) {
+	return r.rc.Do(ctx, r.rc.B().Del().Key(r.indexKey(index)).Build()).Error()
+}
+
+var (
+	lease = rueidis.NewLuaScript(`
+--[[
```
darwin67, line 91: let's just put this into an actual lua script file. still hard to read without syntax highlighting even if I'm used to it now.

pkg/execution/pauses/iterator.go
```diff
@@ -0,0 +1,244 @@
...
+	// Move the blocks to in-flight.
+	d.inflightBlocks = append(d.inflightBlocks, blockIDs...)
+
+	// And remove from unfetched
+	d.unfetchedBlocks = d.unfetchedBlocks[maxFetch:]
+
+	for _, bid := range blockIDs {
+		// Fetch each block, without capturing the ID.
+		id := bid
+		go d.fetchBlock(context.Background(), id)
```
darwin67, line 213: should this be done in a pool?

pkg/execution/pauses/iterator.go
```diff
@@ -0,0 +1,244 @@
...
+	inflightBlocks []ulid.ULID
+
+	// pauses represents the current pauses that have been fetched from downloaded
+	// blocks.
+	pauses []*state.Pause
+
+	err error
+
+	// l represents a lock held when mutating block slices or pauses.
+	l *sync.Mutex
```
darwin67, line 74: can this be a rw lock or does it need to be a mutex?

I see a couple of cases where it's more reading than writing to it.

pkg/execution/pauses/iterator.go
```diff
@@ -0,0 +1,244 @@
...
+	d.l.Lock()
+	defer d.l.Unlock()
+	if len(d.pauses) > 0 {
+		// Simple case:  pauses are downloaded, so yeah we have a next value.
+		return true
+	}
+
+	// If we're here and we've already processed all in-flight blocks, attempt
+	// to redownload blocks.
+	return d.Next(ctx)
```
darwin67, line 152: won't there be risk of deadlocks if this is a recursive call because of the `defer` a couple lines above?

pkg/execution/pauses/manager.go
```diff
@@ -0,0 +1,140 @@
...
+	// each pause.
+	//
+	// However, in order to eventually compact we need to handle the “pause not found”
+	// case when consuming, and always re-delete the pause.  that’s no big deal, but
+	// not the best.
+	//
+	// In the future, we could add two block indexes:  pending, and stored.  this is a
+	// pain, though, because we may die when uploading pending blocks, and that requires
+	// a bit of thought to work around, so we’ll just go with double deletes for now,
+	// assuming this won’t happen a ton.  this can eb optimized later
```
darwin67, line 63: got it, could we add metrics to see how often this happens?
and also warning logs.

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+	DefaultCompactionLimit = (DefaultPausesPerBlock / 3)
+
+	// DefaultCompactionSample gives us a 10% chance of running compactions after
+	// a delete.
+	DefaultCompactionSample = 0.1
+)
+
+// Block represents a block of pauses.
+type Block struct {
+	// ID is the block ID.
```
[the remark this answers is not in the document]
tonyhb: It should match the timestamp of the latest pause in the block.  I'll note that, thanks.

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+		func(ctx context.Context) error {
+			// Call this function and block, renewing leases in the background
+			// until this function is done.
+			return b.flushIndexBlock(ctx, index)
+		},
+		10*time.Second,
+	)
+}
+
+func (b blockstore) flushIndexBlock(ctx context.Context, index Index) error {
```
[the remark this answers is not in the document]
tonyhb: this will always be the # of items per block, so im not sure if that's worthwhile?

pkg/execution/pauses/block.go
tonyhb: On garbage: what garbage would you expect?  We're doing two reads, and inserting a pause into a single delete key - that single write is all we need to mark a deletion.

On transactions:  because this is a single write to delete, its not necessary.

pkg/execution/pauses/block.go
tonyhb: This is idempotent.  Writing the same block twice will cause this to rerun.

pkg/execution/pauses/iterator.go
tonyhb: AFAIK, there aren't concurrent reads so it doesn't matter.

pkg/execution/pauses/iterator.go
tonyhb: We only fetch up to 20 at a time so I don't think it's worth it.

pkg/execution/pauses/manager.go
```diff
@@ -0,0 +1,140 @@
...
+// This returns the total number of pauses in the buffer.
+func (m manager) Write(ctx context.Context, index Index, pauses ...*state.Pause) (int, error) {
+	n, err := m.buf.Write(ctx, index, pauses...)
+	if err != nil {
+		return n, err
+	}
+
+	// If this is larger than the max buffer len, schedule a new block write.
+	if m.bs != nil && n >= m.bs.BlockSize() {
+		go func() {
```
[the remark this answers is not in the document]
tonyhb: Not necessary.  Makes the code more complex, and this service scales so we have to use locks anyway.

pkg/execution/pauses/serializer.go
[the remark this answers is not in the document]
tonyhb: There are tests in iterator_test.go.

pkg/execution/state/pause.go
```diff
@@ -50,16 +57,11 @@ type PauseGetter interface {
...
+
 	// EventHasPauses returns whether the event has pauses stored.
 	EventHasPauses(ctx context.Context, workspaceID uuid.UUID, eventName string) (bool, error)
 
-	// PauseByStep returns a specific pause for a given workflow run, from a given step.
-	//
-	// This is required when continuing a step function from an async step, ie. one that
-	// has deferred results which must be continued by resuming the specific pause set
-	// up for the given step ID.
-	PauseByStep(ctx context.Context, i Identifier, actionID string) (*Pause, error)
```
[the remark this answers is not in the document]
tonyhb: Ya

pkg/execution/state/pause.go
```diff
@@ -156,16 +182,10 @@ type Pause struct {
...
+	Event *string `json:"event,omitempty"`
 	// Expression is an optional expression that must match for the pause
 	// to be resumed.
-	Expression *string `json:"expression"`
-	// ExpressionData _optionally_ stores only the data that we need to evaluate
-	// the expression from the event.  This allows us to load pauses from the
-	// state store without round trips to fetch the entire function state.  If
-	// this is empty and the pause contains an expression, function state will
-	// be loaded from the store.
-	ExpressionData map[string]any `json:"data"`
```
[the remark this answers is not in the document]
tonyhb: We dont need ExpressionData any more, as it's folded into the expression directly.

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -308,6 +312,23 @@ func (q *queue) claimSequentialLease(ctx context.Context) {
...
+	// - lock
+	// - set lease ID
+	// - unlock
+	//
+	// setup tick to attempt to configure lease
+	// or update lease if there's already one
+	//
+	// flush pauses to blob store on condition
+
+	panic("not implemented")
```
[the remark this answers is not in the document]
tonyhb: I didn't see this code.  This attributes to you ;)

pkg/util/lease.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func Lease(
+	ctx context.Context,
+	name string,
+	lease func(ctx context.Context) (ulid.ULID, error),
+	renew func(ctx context.Context, leaseID ulid.ULID) (ulid.ULID, error),
+	revoke func(ctx context.Context, leaseID ulid.ULID) error,
+	do func(ctx context.Context) error,
+	interval time.Duration,
```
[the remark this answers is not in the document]
tonyhb: it could be, but that's an extra alloc and I didn't feel it was necessary.

pkg/execution/pauses/block.go
tonyhb: We also _need_ this to be idempotent because we have already written the entire block to storage, and if this tx fails to commit we're in the same place.  So, it's optional to put this in a lua script and won't impact us.

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+
+const (
+	// DefaultPausesPerBlock is the number of pauses to store in a single block.
+	// A pause equates to roughly ~0.75-1KB of data, so this is a good default
+	// of roughly 25mb blocks.
+	DefaultPausesPerBlock = 25_000
+
+	// DefaultCompactionSample is the minimum number of deleted pauses within an
+	// index that should trigger a compaction.
+	DefaultCompactionLimit = (DefaultPausesPerBlock / 3)
```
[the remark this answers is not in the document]
tonyhb: I dont think it's worthwhile increasing the DefaultPausesPerBlock so that this number is round.  The comments in the DefaultPausesPerBlock indicate why it's 25K items.  If anything, we can change the compaction limit factor.

pkg/execution/pauses/block.go
```diff
@@ -0,0 +1,464 @@
...
+	Bufferer Bufferer
+	// Bucket is the backing blobstore for reading and writing blocks.
+	Bucket *blob.Bucket
+	// Leaser manages leases for a given index.
+	Leaser BlockLeaser
+	// BlockSize is the number of pauses to store in a single block.
+	BlockSize int
+	// CompactionLimit is the total number of pauses that should trigger a compaction.
+	// Note that this doesnt always trigger a compaction;
+	CompactionLimit int
```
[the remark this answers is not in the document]
tonyhb: Thanks, updated.

pkg/execution/pauses/iterator.go
tonyhb: aha, fixed - thank you!

pkg/execution/pauses/iterator.go
darwin67: not necessary if there are no benefits to it

pkg/execution/pauses/block.go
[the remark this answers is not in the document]
darwin67: more so for seeing this over time and frequency of flushing

pkg/execution/pauses/block.go
darwin67: more so of a reaction when I see multiple interactions with redis in one function.
if this is fine, then we're good.

pkg/util/lease.go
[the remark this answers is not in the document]
darwin67: don't see why an extra alloc make much difference if it increases readability though.

pkg/execution/state/redis_state/queue_processor.go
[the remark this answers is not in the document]
darwin67: sounds good

pkg/run/trace_lifecycle.go
```diff
@@ -1044,11 +1044,11 @@ func (l traceLifecycle) OnInvokeFunctionResumed(
...
-					attribute.String(consts.OtelSysAccountID, pause.Identifier.AccountID.String()),
-					attribute.String(consts.OtelSysWorkspaceID, pause.Identifier.WorkspaceID.String()),
-					attribute.String(consts.OtelSysAppID, pause.Identifier.AppID.String()),
-					attribute.String(consts.OtelSysFunctionID, pause.Identifier.WorkflowID.String()),
-					attribute.Int(consts.OtelSysFunctionVersion, pause.Identifier.WorkflowVersion),
+					attribute.String(consts.OtelSysAccountID, md.ID.Tenant.AccountID.String()),
+					attribute.String(consts.OtelSysWorkspaceID, md.ID.Tenant.EnvID.String()),
+					attribute.String(consts.OtelSysAppID, md.ID.Tenant.AppID.String()),
+					attribute.String(consts.OtelSysFunctionID, md.ID.FunctionID.String()),
+					attribute.Int(consts.OtelSysFunctionVersion, md.Config.FunctionVersion),
```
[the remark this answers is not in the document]
tonyhb: it's safer that this comes from the function config instead of the pause, actually

pkg/execution/pauses/block.go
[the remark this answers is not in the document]
tonyhb: ack, lets add metrics in a followup PR prior to release?

darwin67 (approved):

pkg/execution/state/redis_state/queue.go
```diff
@@ -714,6 +714,10 @@ type QueueRunMode struct {
 
 	// Continuations enables continuations
 	Continuations bool
+
+	// PauseFluster determines whether pauses are flushed from hot storage
```
BrunoScheufler, line 718:
```suggestion
	// PauseFlusher determines whether pauses are flushed from hot storage
```

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -308,6 +312,23 @@ func (q *queue) claimSequentialLease(ctx context.Context) {
 	}
 }
 
+func (q *queue) runPauseFlusher(ctx context.Context) {
```
BrunoScheufler, line 315: We don't have to do this now but it's probably good to consider moving this to a separate entrypoint. I think adding more responsibilities to the executor (even if you can enable/disable them for most executors in the case of flushing + deploy a separate instance just for this), it adds complexity and blurs the scope of our services.

pkg/execution/pauses/buffer.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Delete deletes a pause from the buffer, or returns ErrNotInBuffer if the pause is not in
+// the buffer.
+func (r redisAdapter) Delete(ctx context.Context, index Index, pause state.Pause) error {
+	// Check if pause is in buffer;  if not, return ErrNotInBuffer so that we can default
+	// to deleting the pause from the backing store.
+	if r.rsm.PauseExists(ctx, pause.ID) == state.ErrPauseNotFound {
+		return ErrNotInBuffer
+	}
+	return r.rsm.DeletePause(ctx, pause)
```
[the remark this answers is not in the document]
tonyhb: It largely doesn't matter actually.  The manager hits delete on both the buffer and the underlying block store in every case.

pkg/execution/state/redis_state/queue_processor.go
tonyhb: ya completely right — this will be handled in a queue.  we shouldn't run this in the executor / new runs / whichever service manages pauses due to the added memory pressure incurred.

PR #2370, opened 2025-04-24
Make sure to assign run_id in a couple other areas for root spans

darwin67:
## Description

Add run_id to root spans whenever possible, and provide additional fallbacks to kafka keys when exporting spans.

Also skip sending to Kafka if the spans is marked for deletion.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -2945,16 +2945,19 @@ func (e *executor) RetrieveAndScheduleBatch(ctx context.Context, fn inngest.Func
 
 	// Don't bother if it's already there
 	if err == redis_state.ErrQueueItemExists {
+		span.SetAttributes(attribute.Bool(consts.OtelSysStepDelete, true))
 		return nil
 	}
 
 	// If function is paused, we do not schedule runs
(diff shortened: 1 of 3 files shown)


pkg/execution/executor/executor.go
```diff
@@ -2969,8 +2972,6 @@ func (e *executor) RetrieveAndScheduleBatch(ctx context.Context, fn inngest.Func
 
 	if md != nil {
 		span.SetAttributes(attribute.String(consts.OtelAttrSDKRunID, md.ID.RunID.String()))
-	} else {
-		span.SetAttributes(attribute.Bool(consts.OtelSysStepDelete, true))
```
djfarrelly, line 2973: question, non-blocking: Do we still want to handle this or leave this blank and then if we don't see the delete/unset status we can investigate separately as another issue?

djfarrelly (approved):

pkg/execution/executor/executor.go
darwin67: it's been moved upwards to mark as delete on errors.
so if this still happens, we should definitely investigate.

PR #2387, opened 2025-04-29
Implement scanner with blanket refill

BrunoScheufler:
- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -498,30 +509,124 @@ func (q *queue) worker(ctx context.Context, f osqueue.RunFunc) {
...
+			}
+		}
+	}()
+
+	// Check if shadow partition cannot be processed (paused/refill disabled, etc.)
+	if shadowPart.PauseRefill {
+		q.removeShadowContinue(ctx, shadowPart, false)
+
+		forceRequeueAt := q.clock.Now().Add(ShadowPartitionRefillPausedRequeueExtension)
+		err = q.ShadowPartitionRequeue(ctx, shadowPart, *leaseID, &forceRequeueAt)
```
darwin67, line 574: do we actually need to requeue?
seems like a redundant work that only consume additional qps for no benefit.

besides when unpaused, it'll just start processing from the oldest stuff it sees right away.

pkg/execution/state/redis_state/queue.go
```diff
@@ -2738,6 +2782,357 @@ func (q *queue) Requeue(ctx context.Context, queueShard QueueShard, i osqueue.Qu
...
+
+	if err != nil {
+		return nil, fmt.Errorf("error peeking shadow partition backlogs: %w", err)
+	}
+	returnedSet, ok := peekRet.([]any)
+	if !ok {
+		return nil, fmt.Errorf("unknown return type from shadowPartitionPeek: %T", peekRet)
+	}
+
+	var potentiallyMissingBacklogs, allBacklogIDs []any
```
darwin67, line 2938: do we still need all these for shadow partition processing?
what does it mean to have potentially missing backlogs?

pkg/execution/state/redis_state/queue.go
```diff
@@ -39,9 +40,10 @@ import (
 )
 
 const (
-	PartitionSelectionMax = int64(100)
-	PartitionPeekMax      = PartitionSelectionMax * 3
-	AccountPeekMax        = int64(30)
+	PartitionSelectionMax          = int64(100)
+	PartitionPeekMax               = PartitionSelectionMax * 3
+	AccountPeekMax                 = int64(30)
+	ShadowPartitionPeekMaxBacklogs = int64(100)
```
darwin67, line 46: this should be overridable, like `peekMax` in the queue struct

darwin67: not sure if there are benefits of doing EWMA atm, so we can defer the decision for now

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -498,30 +509,124 @@ func (q *queue) worker(ctx context.Context, f osqueue.RunFunc) {
...
+	if err != nil {
+		return fmt.Errorf("could not peek backlogs for shadow partition: %w", err)
+	}
+
+	// Sequentially refill backlogs
+	for _, backlog := range backlogs {
+		err := q.BacklogRefill(ctx, backlog, shadowPart)
+		if err != nil {
+			// TODO Handle capacity reached and exit
+			isCapacityReached := false
```
darwin67, line 594: I assume there will be a follow up to convert this variable to true?
I think you can return it as part of `BacklogRefill` too so it's not a separate call to the storage system

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -498,30 +509,124 @@ func (q *queue) worker(ctx context.Context, f osqueue.RunFunc) {
...
 
-			// TODO: probably good to wrap this in a closure with the shadow partition logic
-			// so it can just do a defer
-			metrics.ActiveShadowScanerCount(ctx, -1, metrics.CounterOpt{PkgName: pkgName})
+	hasMoreBacklogs := false
+	if !hasMoreBacklogs {
+		// No more backlogs right now, we can continue the scan loop until new items are added
+		q.removeShadowContinue(ctx, shadowPart, false)
+
+		err = q.ShadowPartitionRequeue(ctx, shadowPart, *leaseID, nil)
```
darwin67, line 614: I wonder if the shadow partition should just be deleted if there are no more backlogs in it instead of requeueing

pkg/execution/state/redis_state/queue.go
BrunoScheufler: This is to clean up dangling backlog IDs in shadow partitions. I doubt this should happen but it’s probably still good to just clean them up, same for dangling queue items in backlogs themselves

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -498,30 +509,124 @@ func (q *queue) worker(ctx context.Context, f osqueue.RunFunc) {
...
+
+				return nil
+			}
+		}
+	}
 
-			// TODO: probably good to wrap this in a closure with the shadow partition logic
-			// so it can just do a defer
-			metrics.ActiveShadowScanerCount(ctx, -1, metrics.CounterOpt{PkgName: pkgName})
+	hasMoreBacklogs := false
```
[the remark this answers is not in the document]
BrunoScheufler: Hmm we’d need something like the total amount of backlogs

pkg/execution/state/redis_state/lua/queue/backlogRefill.lua
```diff
@@ -0,0 +1,78 @@
...
+-- TODO Enforce constraints and adjust capacity
+
+-- TODO Move item(s) out of backlog and into partition
+
+local items = redis.call("ZRANGE", keyBacklogSet, "-inf", "+inf", "BYSCORE", "LIMIT", 0, capacity, "WITHSCORES")
+
+-- Reverse the items to be added to the ready set
+local args = {}
+local remArgs = {}
+for i = 1, #items, 2 do
```
darwin67, line 27: why increment by 2 here?

pkg/execution/state/redis_state/queue_processor.go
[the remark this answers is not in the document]
BrunoScheufler: Because we need to know the current total number of backlogs for all keyd

pkg/execution/state/redis_state/lua/queue/backlogRefill.lua
BrunoScheufler: Every second item is the score

BrunoScheufler: So we need to skip by two every time

pkg/execution/state/redis_state/queue_processor.go
darwin67: right, but why the need to do so?

pkg/execution/state/redis_state/queue_processor.go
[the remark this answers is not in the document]
darwin67: well, if you're getting the full `limit` you asked for, that's a good assumption to believe there are more backlogs.
it's basically pagination right? if you ask for 5 items and you get 5, then there's likely more then 5, but if it's 4 then you know there's no more.

pkg/execution/state/redis_state/lua/queue/backlogRefill.lua
darwin67: oh I see. better to put that into comments.

pkg/execution/state/redis_state/queue.go
darwin67: yeah, agree it's a good thing to have. kinda curious what will cause it.

darwin67 (approved):
Non of the comments are blocking so feel free to merge whenever.

pkg/execution/state/redis_state/queue_processor.go
BrunoScheufler: Because we don't want to shadow scan paused functions?

pkg/execution/state/redis_state/queue.go
BrunoScheufler: Honestly this should rarely, if ever, happen. For queue items in backlogs it's much more likely since the same item can be in more than a single backlog

PR #2399, opened 2025-05-07
Pause block nits

tonyhb:
Misc exports and some minor cleanup and comments as we continue the pause block work.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/pauses/block_leaser.go
+++ b/pkg/execution/pauses/block_leaser.go
@@ -11,9 +11,14 @@ import (
 	"github.com/redis/rueidis"
 )
 
-type redisBlockLeaser struct {
-	rc rueidis.Client
+// NewRedisBlockLeaser creates a new lease manager across Redis, using the given
+// prefix for any keys stored.
+func NewRedisBlockLeaser(rc rueidis.Client, prefix string, duration time.Duration) BlockLeaser {
+	return &redisBlockLeaser{rc: rc, prefix: prefix, duration: duration}
+}
 
+type redisBlockLeaser struct {
+	rc       rueidis.Client
 	prefix   string
 	duration time.Duration
 }
--- a/pkg/execution/pauses/buffer.go
+++ b/pkg/execution/pauses/buffer.go
@@ -7,11 +7,16 @@ import (
 	"github.com/inngest/inngest/pkg/execution/state"
 )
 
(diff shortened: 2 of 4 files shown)


ajdinngest (approved):

pkg/execution/executor/executor.go
```diff
@@ -1846,16 +1857,6 @@ func (e *executor) Resume(ctx context.Context, pause state.Pause, r execution.Re
...
 	err = util.Crit(ctx, "consume pause", func(ctx context.Context) error {
-		// Lease this pause so that only this thread can schedule the execution.
-		//
-		// If we don't do this, there's a chance that two concurrent runners
-		// attempt to enqueue the next step of the workflow.
-		err = e.pm.LeasePause(ctx, pause.ID)
-		if err == state.ErrPauseLeased || err == state.ErrPauseNotFound {
-			// Ignore;  this is being handled by another runner.
-			return nil
-		}
```
darwin67, line 1857: no longer leasing?

pkg/execution/executor/service.go
```diff
@@ -212,7 +212,7 @@ func (s *svc) Run(ctx context.Context) error {
 		}
 
 		if err != nil && err.Error() != "NonRetriableError" {
-			s.log.Error("error handling queue item", "error", err, "item", item)
```
darwin67, line 215: why remove the item from the log?
it helps with debugging purposes and I don't think it contains PII.

darwin67 (commented):
Is pauses

pkg/execution/executor/service.go
tonyhb: it's extremely noisy in the dev server CLI, and contains lots of details that our users do not need to see.  given this runs in the dev server and will primarily be seen by our users... if we want the run ID, we should log the run ID.

pkg/execution/executor/executor.go
tonyhb: correct.  the entire code is idempotent:

- consuming a pause and saving step details is transactional and idempotent, and doesn't conflict if two concurrent threads attempt to save the same step
- and given that, only one thread can schedule the next step.

leasing is no longer necessary.

tonyhb:
> Is pause consumption still gonna be idempotent with block changes?

Yes :)

darwin67 (approved):

PR #2424, opened 2025-05-13
Function run scheduling updates

darwin67:
## Description

Due to the lack of transactions between state creation and enqueue, there's always room for failure to enqueue function runs if `Schedule` fails abruptly due to OOMs or panics.

This attempts to make both state creation and enqueue idempotent, so everything can be reran on failures.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -599,19 +599,16 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 
 	st, err := e.smv2.Create(ctx, newState)
 	switch err {
-	case nil, state.ErrIdentifierExists:
-		// no-op: this is safe, continue
+	case nil: // no-op
+	case state.ErrIdentifierExists:
+		// override metadata from the existing state
+		id := sv2.IDFromV1(st.Identifier())
+		metadata, err = e.smv2.LoadMetadata(ctx, id)
+		if err != nil {
+			return nil, err
+		}
 	default:
 		return nil, fmt.Errorf("error creating run state: %w", err)
-	}
-	// override the runID from the state if it doesn't match.
-	// this can happen on scheduling retries due to errors like networking, or abrupt failures
-	//
-	// TODO: update the code from here on to reference the `state` returned instead of using
-	// previous values to avoid weird mismatch issues.
-	// right now, the only mismatch should just be the runID
-	if metadata.ID.RunID != st.RunID() {
-		metadata.ID.RunID = st.RunID()
 	}
 
 	//
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -600,7 +600,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 	st, err := e.smv2.Create(ctx, newState)
 	switch err {
 	case nil, state.ErrIdentifierExists:
-		// no-op, continue
+		// no-op: this is safe, continue
 	default:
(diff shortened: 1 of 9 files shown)


pkg/execution/executor/executor.go
```diff
@@ -597,19 +597,29 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
...
+	// previous values to avoid weird mismatch issues.
+	// right now, the only mismatch should just be the runID
+	if metadata.ID.RunID != st.RunID() {
+		metadata.ID.RunID = st.RunID()
+	}
 
 	//
 	// Create cancellation pauses immediately, only if this is a non-batch event.
 	//
+	// NOTE: on scheduling retries, this will result in duplicated pauses for cancellation.
```
BrunoScheufler, line 620: Or should we create a deterministic pause ID to only create a single pause?

BrunoScheufler (approved):

pkg/execution/executor/executor.go
BrunoScheufler: https://github.com/inngest/inngest/blob/8eef881818dc6e9fd6c63b83146779c1800ed1a4/pkg/execution/executor/executor.go#L2525

BrunoScheufler: https://github.com/inngest/inngest/blob/8eef881818dc6e9fd6c63b83146779c1800ed1a4/pkg/execution/executor/executor.go#L2525

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -243,14 +243,26 @@ func (m shardedMgr) New(ctx context.Context, input state.Input) (state.State, er
...
+		// key to ULID values.
+		// So if this error is returned, we should just continue with creating a new state, since
+		// it could mean that the state is not actually created.
+		case state.ErrInvalidIdentifier: // no-op
+		default:
+			return nil, err
+		}
+
+		// If a state already exists with the idempotency key, override the input's runID and continue
+		if runID != nil && input.Identifier.RunID != *runID {
```
tonyhb, line 263: If state already exists, there's a chance that a step may have already ran. Shall we return state from the state store via a lookup here?

pkg/execution/executor/executor.go
```diff
@@ -597,22 +597,28 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
...
 	default:
 		return nil, fmt.Errorf("error creating run state: %w", err)
 	}
+	// override the runID from the state if it doesn't match.
+	// this can happen on scheduling retries due to errors like networking, or abrupt failures
+	//
+	// TODO: update the code from here on to reference the `state` returned instead of using
+	// previous values to avoid weird mismatch issues.
+	// right now, the only mismatch should just be the runID
+	if metadata.ID.RunID != st.RunID() {
```
tonyhb, line 613: We should be able to replace the metadata wholesale, in every case, because e.smv2.Create either returns `metadata` or the new details, right?

tonyhb (approved):

pkg/execution/state/redis_state/redis_state.go
darwin67: I've already tried that, but it causes a race condition which the idempotency testharness caught.

pkg/execution/executor/executor.go
darwin67: we can, it's just the metadata structure is different where the one here that's created and returned is the v2 shape, while the state holds it in the v1 shape.
there are currently no way to convert v1 to v2 shape afaik.

PR #2439, opened 2025-05-16
Idempotent pause resume

darwin67:
## Description

Make sure pauses are resumed in an idempotent way so we do not get functions stucked when there's a failure during resuming the pause.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/pauses/manager.go
+++ b/pkg/execution/pauses/manager.go
@@ -40,12 +40,12 @@ type manager struct {
 	flushDelay time.Duration
 }
 
-func (m manager) ConsumePause(ctx context.Context, pause state.Pause, opts state.ConsumePauseOpts) (state.ConsumePauseResult, error) {
+func (m manager) ConsumePause(ctx context.Context, pause state.Pause, opts state.ConsumePauseOpts) (state.ConsumePauseResult, func() error, error) {
 	if pause.Event == nil {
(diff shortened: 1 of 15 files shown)


BrunoScheufler (approved):

pkg/execution/pauses/manager.go
```diff
@@ -72,29 +72,36 @@ func (m manager) ConsumePause(ctx context.Context, pause state.Pause, data any)
...
-		logger.StdlibLogger(ctx).Error(
-			"error deleting pause once consumed",
-			"error", err,
-			"pause", pause,
-			"index", idx,
-		)
+
+	// override the cleanup with idx deletion
+	cleanup = func() error {
+		err := m.Delete(ctx, idx, pause)
```
tonyhb, line 90: this is a note for me:  we probably want to call the OG cleanup, and I'll handle that in the continued pause blob work to be shipped after this merge.

tonyhb (approved):

pkg/execution/pauses/manager.go
darwin67: honestly overriding like this seems fine, I don't have strong preferences because
the core delete doesn't consider index and kinda seems weird if it needs to.

the point of the closure is to encapsulate the cleanup logic so nothing else really needs to be aware of how it's done, and this looks okay to me.

PR #2441, opened 2025-05-16
Fix runs page crashing on invalid CEL expression

albertchae:
## Description

Fix fatal error for invalid CEL expression in runs search

- When we get a gql error from the backend, instead of throwing to the error boundary, parse out the combined error to find the one for `expression_invalid` and display that message
image

Also show error for frontend validation of CEL expression

- We don't send these queries to the backend if we detect it's invalid on the frontend when user clicks Search (but we do if user hits cmd + enter)
- Replace boolean check for error with presence of error itself

image

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Runs/Runs.tsx
+++ b/ui/apps/dashboard/src/components/Runs/Runs.tsx
@@ -8,7 +8,7 @@ import {
   useSearchParam,
   useStringArraySearchParam,
 } from '@inngest/components/hooks/useSearchParam';
-import { useQuery } from 'urql';
+import { CombinedError, useQuery } from 'urql';
 
 import { useEnvironment } from '@/components/Environments/environment-context';
 import { useGetRun } from '@/components/RunDetails/useGetRun';
@@ -35,6 +35,12 @@ type EnvProps = {
 };
 
 type Props = FnProps | EnvProps;
+
+const parseCelSearchError = (combinedError: CombinedError | undefined) => {
+  return combinedError?.graphQLErrors.find(
+    (error) => error.extensions.code == 'expression_invalid'
+  );
+};
 
 export const Runs = forwardRef<RefreshRunsRef, Props>(function Runs(
   { functionSlug, scope }: Props,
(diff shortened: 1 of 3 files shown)


ui/packages/components/src/CodeSearch/CodeSearch.tsx
```diff
@@ -370,8 +373,15 @@ export default function CodeSearch({
     }, VALIDATION_DELAY);
   };
 
+  const expressionError = searchError || editorValidationError;
```
albertchae, line 376: showing `editorValidationError` in the `Alert` makes the UI slightly more consistent, but the error messages from the frontend and backend validation are currently inconsistent

ui/packages/components/src/CodeSearch/CodeSearch.tsx
albertchae, line 1: John and I also discussed disabling the Search button if we get a frontend validation error. I will look into that in a follow up PR

ui/apps/dashboard/src/components/Runs/Runs.tsx
```diff
@@ -36,6 +36,12 @@ type EnvProps = {
 
 type Props = (FnProps | EnvProps) & { traceAIEnabled?: boolean };
 
+const parseCelSearchError = (combinedError: CombinedError | undefined) => {
+  return combinedError?.graphQLErrors.find(
+    (error) => error.extensions.code == 'expression_invalid'
```
albertchae, line 41: I didn't find an enum of error codes in the frontend anywhere, did I miss it?

Also I considered reviving https://github.com/inngest/inngest/pull/1220 for this but it seems like we might want these invalid expression errors specifically here rather than the first error with a valid `extensions.code`, wasn't sure 🤔

albertchae (commented):
It might be overkill for now but I can also look into extending the frontend validation to catch more types of invalid expressions and/or having the backend validation add errors that can be used in the monaco frontend.

ui/apps/dashboard/src/components/Runs/Runs.tsx
jacobheric: I don't see one either. You can add an enum to the frontend to mirror the backend equivalent but you'll need to take some care to handle when those things get out of sync. Here I think that is super low risk...as opposed to something like an entity type.

jacobheric:
> It might be overkill for now but I can also look into extending the frontend validation to catch more types of invalid expressions and/or having the backend validation add errors that can be used in the monaco frontend.

Yeah I'm all for handling all errors here as gracefully as possible! Even if we can't parse it we can still say sorry something went wrong and give them some recourse.

jacobheric (approved):

ui/packages/components/src/CodeSearch/CodeSearch.tsx
albertchae: I decided to flip the order here since currently the frontend validation error is more informative than the backend one

PR #2488, opened 2025-05-30
Add FunctionConfiguration to coreapi

albertchae:
## Description

This is the backend necessary so we can show function configuration details in the dev server frontend.

I copied over the FunctionConfiguration gql schema from monorepo as well as the serialization code, but refactored it slightly
so we should be able to easily share it between monorepo and inngest/inngest going forward

This can currently be tested by going to http://localhost:8288/v0 and using the following query
```
query GetFunction($functionSlug: String!) {
  functionBySlug(query: { functionSlug: $functionSlug }) {
    name
    id
    concurrency
    config
    configuration {
      cancellations {
        event
        timeout
        condition
      }
      retries {
        value
        isDefault
      }
      priority
      eventsBatch {
        maxSize
        timeout
        key
      }
      concurrency {
        scope
        limit {
          value
          isPlanLimit
        }
        key
      }
      rateLimit {
        limit
        period
        key
      }
      debounce {
        period
        key
      }
      throttle {
        burst
        key
        limit
        period
      }
    }
    slug
    url
  }
}
```

## Motivation

We want to show function configuration in the dev server

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/coreapi/gql.schema.graphql
+++ b/pkg/coreapi/gql.schema.graphql
@@ -130,7 +130,7 @@ type Function {
   name: String!
   slug: String!
   config: String!
-  configuration: FunctionConfiguration
+  configuration: FunctionConfiguration!
   concurrency: Int!
   triggers: [FunctionTrigger!]
   url: String!
@@ -172,7 +172,7 @@ type FunctionEvent {
 }
 
 type FunctionConfiguration {
-  cancellations: [CancellationConfiguration!]
+  cancellations: [CancellationConfiguration!]!
   retries: RetryConfiguration!
   priority: String
   eventsBatch: EventsBatchConfiguration
--- a/pkg/coreapi/gql.schema.graphql
+++ b/pkg/coreapi/gql.schema.graphql
@@ -130,6 +130,7 @@ type Function {
   name: String!
   slug: String!
   config: String!
+  configuration: FunctionConfiguration
   concurrency: Int!
   triggers: [FunctionTrigger!]
   url: String!
@@ -168,6 +169,73 @@ type FunctionEvent {
   type: FunctionEventType
   output: String
   createdAt: Time
+}
+
+type FunctionConfiguration {
+  cancellations: [CancellationConfiguration!]
+  retries: RetryConfiguration!
+  priority: String
+  eventsBatch: EventsBatchConfiguration
+  concurrency: [ConcurrencyConfiguration!]!
+  rateLimit: RateLimitConfiguration
+  debounce: DebounceConfiguration
+  throttle: ThrottleConfiguration
+}
+
+type CancellationConfiguration {
+  event: String!
+  timeout: String
+  condition: String
+}
+
+type RetryConfiguration {
+  value: Int!
+  isDefault: Boolean
+}
+
+type EventsBatchConfiguration {
+  "The maximum number of events a batch can have."
+  maxSize: Int!
+
+  "How long to wait before running the function with the batch."
+  timeout: String!
+
+  key: String
+}
+
+type ConcurrencyConfiguration {
+  scope: ConcurrencyScope!
+  limit: ConcurrencyLimitConfiguration!
+  key: String
+}
+
+enum ConcurrencyScope {
+  ACCOUNT
+  ENVIRONMENT
+  FUNCTION
+}
+
+type ConcurrencyLimitConfiguration {
+  value: Int!
+  isPlanLimit: Boolean
+}
+
+type RateLimitConfiguration {
+  limit: Int!
+  period: String!
+  key: String
+}
+
+type DebounceConfiguration {
+  period: String!
+  key: String
+}
+
+type ThrottleConfiguration {
+  burst: Int!
+  key: String
+  limit: Int!
+  period: String!
 }
 
 enum StepEventType {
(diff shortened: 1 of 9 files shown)


pkg/coreapi/gql.schema.graphql
```diff
@@ -170,6 +171,73 @@ type FunctionEvent {
   createdAt: Time
 }
 
+type FunctionConfiguration {
+  cancellations: [CancellationConfiguration!]
```
amh4r, line 175: Nit:

```suggestion
  cancellations: [CancellationConfiguration!]!
```

That'd simplify the types ever so slightly. So no cancellations would be `[]` instead of `null`

pkg/coreapi/gql.schema.graphql
```diff
@@ -170,6 +171,73 @@ type FunctionEvent {
...
+  retries: RetryConfiguration!
+  priority: String
```
amh4r, line 177: TS SDK shows this type as an object:
https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/InngestFunction.ts#L522-L534

pkg/coreapi/gql.schema.graphql
```diff
@@ -130,6 +130,7 @@ type Function {
   name: String!
   slug: String!
   config: String!
+  configuration: FunctionConfiguration
```
amh4r, line 133:
```suggestion
  configuration: FunctionConfiguration!
```

When would a function not have a config?

pkg/coreapi/graph/resolvers/function.go
```diff
@@ -12,3 +12,13 @@ func (r *functionResolver) App(ctx context.Context, obj *models.Function) (*cqrs
 	appID := uuid.MustParse(obj.AppID)
 	return r.Data.GetAppByID(ctx, appID)
 }
+
+func (qr *queryResolver) FunctionBySlug(ctx context.Context, query models.FunctionQuery) (*models.Function, error) {
+	dummyLocalWorkspaceID := uuid.MustParse("6c6f6361-6c00-0000-0000-000000000000")
```
amh4r, line 17: Might be good to use this:
https://github.com/inngest/inngest/blob/main/pkg/consts/devserver.go#L33

Sometimes things get weird when you don't use it. Also, it'll help discovery when we eventually add multi-environment support to the Dev Server

pkg/coreapi/gql.schema.graphql
albertchae: I noticed that as well but the cloud API is currently using a string https://github.com/inngest/monorepo/blob/dcaef52cdb93e28d0683af80e4a4d59c96f60147/api/schema.graphql#L891

Figured I'd leave it for now until we add more fields to priority unless you think it's worth changing now

pkg/coreapi/gql.schema.graphql
albertchae: makes sense, I'll backport to https://github.com/inngest/monorepo/blob/dcaef52cdb93e28d0683af80e4a4d59c96f60147/api/schema.graphql#L838 as well

pkg/coreapi/graph/resolvers/function.go
albertchae: cool didn't know about that, my joke of encoding "local" in hex and having it at the front of the UUID is toast 😆

amh4r: Ha, I didn't notice that!

amh4r (approved):

pkg/coreapi/graph/models/function_configuration_serializer_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.fn.Validate(context.Background())
+			if err != nil {
+				t.Errorf("Invalid function configuration for test %v: %v", tt.name, err)
+			}
+			//if got := ToFunctionConfiguration(tt.args.fn, tt.args.planConcurrencyLimit); !reflect.DeepEqual(got, tt.expected) {
+			//	gotJson, _ := json.MarshalIndent(got, "", "  ")
+			//	wantJson, _ := json.MarshalIndent(tt.expected, "", "  ")
+			//	t.Errorf("ToFunctionConfiguration() = \n%s, expected \n%s", gotJson, wantJson)
+			//}
```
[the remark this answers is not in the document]
KiKoS0: I think it's because cancellations isn't returning `nil` but an empty array when there are no cancels.

```diff
diff --git a/pkg/coreapi/graph/models/function_configuration_serializer.go b/pkg/coreapi/graph/models/function_configuration_serializer.go
index c8f809c70..31da077cb 100644
--- a/pkg/coreapi/graph/models/function_configuration_serializer.go
+++ b/pkg/coreapi/graph/models/function_configuration_serializer.go
@@ -76,6 +76,9 @@ func ToFunctionConfiguration(fn *inngest.Function, planConcurrencyLimit int) *Fu
 }

 func mapCancellations(cancels []inngest.Cancel) []*CancellationConfiguration {
+       if len(cancels) == 0 {
+               return nil
+       }
        cancellations := make([]*CancellationConfiguration, len(cancels))
        for i, cancel := range cancels {
                cancellations[i] = &CancellationConfiguration{
```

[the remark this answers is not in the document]
albertchae: oh that makes sense, actually we just changed that to not nullable in this PR so it works now after I run codegen

PR #2530, opened 2025-06-10
Add singletons cancel mode

KiKoS0:
## Description

WIP cancel mode for singletons

## Motivation
[INN-4981](https://linear.app/inngest/issue/INN-4981)

- [ ] Docs

## Checklist


pkg/execution/executor/executor.go
```diff
@@ -533,23 +533,50 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
...
+						FunctionID: req.Function.ID,
+						Tenant: sv2.Tenant{
+							AccountID: req.AccountID,
+							EnvID:     req.WorkspaceID,
+						},
+					}
+
+					_ = e.Cancel(ctx, runID, execution.CancelRequest{
+						EventID: &eventID,
+					})
```
KiKoS0: Should we only expect a nil error or a subset of safe to ignore errors here?

darwin67: should still log the error so we can see what's going on, but no action are needed for now.

pkg/execution/executor/executor.go
```diff
@@ -534,23 +534,50 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
...
 			}
 
-			if alreadyRunning {
-				// TODO: Handle cancellation mode
+			eventID := req.Events[0].GetInternalID()
 
-				// Immediately end before creating state
-				return nil, ErrFunctionSkipped
+			if singletonRunID != nil {
+				if req.Function.Singleton.Mode == enums.SingletonModeCancel {
```
darwin67: better to do a `switch` here

pkg/execution/singleton/singleton.go
```diff
@@ -17,7 +16,7 @@ var (
 )
 
 type Singleton interface {
-	Singleton(ctx context.Context, key string, c inngest.Singleton) (bool, error)
+	HandleSingleton(ctx context.Context, key string, c inngest.Singleton) (*string, error)
```
darwin67: if you're always expecting a runID here, why not return a `*ulid.ULID` instead of `*string`?

darwin67 (approved):
looks good overall, just some questions here and there

PR #2553, opened 2025-06-13
Use BillingPlan.isLegacy

amh4r:
> ⚠️ Do not merge this PR until https://github.com/inngest/monorepo/pull/4629 is released to prod

Use `BillingPlan.isLegacy` instead of inferring from the plan name. Currently, the UI infers it from the plan name but that won't work anymore now that we have 2 plans named "Pro": legacy and new


albertchae (approved):

albertchae:
@amh4r I realized I hadn't rerun codegen after some FunctionConfiguration backend changes, so I made https://github.com/inngest/inngest/pull/2560 to fix that, then I stacked your PR on top of that to fix the vercel preview deployment

PR #2613, opened 2025-06-23
Ignore singleton key evaluation error instead of retrying

KiKoS0:
## Description

Ignore singleton key evaluation error instead of retrying

## Motivation
Errors in logs and constantly pubsub retrying events.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -603,6 +603,9 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 				}
 			}
 			singletonConfig = &queue.Singleton{Key: singletonKey}
+		case errors.Is(err, singleton.ErrEvaluatingSingletonExpression):
+			// Ignore singleton expressions if we cannot evaluate them
+			logger.StdlibLogger(ctx).Warn("error evaluating singleton expression", "error", err)
 		case errors.Is(err, singleton.ErrNotASingleton):
 			// We no-op, and we run the function normally not as a singleton
 		default:


BrunoScheufler (approved):

pkg/execution/executor/executor.go
```diff
@@ -603,6 +603,9 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 				}
 			}
 			singletonConfig = &queue.Singleton{Key: singletonKey}
+		case errors.Is(err, singleton.ErrEvaluatingSingletonExpression):
+			// Ignore singleton expressions if we cannot evaluate them
+			logger.StdlibLogger(ctx).Warn("error evaluating singleton expression", "error", err)
```
tonyhb, line 608: Should we treat invalid expressions as a specific key so that singletons still work?  I'm wondering what's preferable:

- Having many more functions run than expected
- Or having far fewer runs than expected.

I also suspect we should shoot over an Inngest event here / an audit trail event so that we can alert and email on this in the future or something.  We need to define how to show alerts / warnings like this in our UI at some point... badly!

KiKoS0: Hmm we could treat it as a truthy value, but since this isn’t a user-facing error, I’m not sure if it would be slightly more confusing for users to run it as a singleton under a different key.

KiKoS0: Or a different specific key and document that

darwin67 (approved):

pkg/execution/executor/executor.go
KiKoS0: Merging to get rid of the error, we can follow it up if we want the other behavior of running as under a specific key.

PR #2643, opened 2025-06-29
Allow Vercel Marketplace integration project management

amh4r:
Allow Vercel Marketplace integration project management

### DIFF
--- a/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/integrations.tsx
+++ b/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/integrations.tsx
@@ -31,21 +31,11 @@ const INTEGRATIONS: Integration[] = [
     title: 'Vercel',
     slug: 'vercel',
     Icon: <IconVercel className="text-onContrast h-6 w-6" />,
-    actionButton: ({ enabled, hasError, isMarketplace }) => {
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/integrations.tsx
```diff
@@ -31,21 +31,11 @@ const INTEGRATIONS: Integration[] = [
...
       let target = undefined;
       let url: string;
-      if (isMarketplace) {
-        // Marketplace integration installed.
-
-        label = 'Manage';
-        target = '_blank';
-
-        // Managed in Vercel's dashboard.
-        url = 'https://vercel.com/integrations/inngest';
```
albertchae, line 45: does this mean we can't manage it from vercel's dashboard? or we can do it from either?

albertchae (approved):

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/integrations.tsx
amh4r: The same stuff can still be managed from Vercel's dashboard. We only linked to it before because we didn't know what other link to use 😆

PR #2646, opened 2025-07-01
Forward connect executor -> gateway through gRPC

KiKoS0:
## Description

- Create a new gRPC server on the connect gateway that listens to incoming requests from executors.
- Add a callback to control if we want to forward requests through gRPC or the regular pubsub method.
- Store outbound gateways IPs in the connect store metadata, so that executors are able to resolve which gRPC client to use for each request.

Still missing changes are:

- Return gateway forward errors to executors.
- Garbage collect gRPC clients of dead/gone gateways.

- [ ] Docs

## Checklist


pkg/connect/pubsub/grpc.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	logger.StdlibLogger(ctx).Debug("got connect gateways to connect to", "len", len(gateways))
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	for _, g := range gateways {
+		url := fmt.Sprintf("%s:%d", g.IPAddress, connectConfig.Gateway(ctx).GRPCPort)
+		conn, err := grpc.NewClient(url, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
BrunoScheufler: I wonder if we need to configure some settings on the client here

- Timeouts: Requests may run for up to 2h if we wait for the response to come back
- Idle/Keepalive: In the case of waiting for a long time, we have to ensure the connection isn't closed prematurely

KiKoS0: Hmm we aren't blocking till a worker response like sdk gateways though right? I thought requests from executor to gateways are just forwards to the websocket connection which shouldn't take that long ideally.

pkg/connect/service.go
```diff
@@ -59,11 +63,16 @@ type ConnectEntitlementRetriever interface {
...
 type connectGatewaySvc struct {
+	pb.ConnectGatewayServer
+
 	gatewayPublicPort int
 
 	gatewayRoutes  chi.Router
 	maintenanceApi chi.Router
 
+	grpcServer *grpc.Server
+	wsConnections sync.Map
```
KiKoS0, line 74: >The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

I’ve used a `sync.Map` here because I believe both conditions apply to our use case. Let me know if you think otherwise.

pkg/connect/state/state.go
```diff
@@ -156,6 +160,8 @@ type Gateway struct {
 	LastHeartbeatAtMS int64         `json:"last_heartbeat"`
 
 	Hostname string `json:"hostname"`
+
+	IPAddress net.IP `json:"ip"`
```
BrunoScheufler, line 164: Does marshaling/unmarshaling work fine with the built-in type? Do we want to store this as a string representation just in case? Had some trouble with time.Time previously which is why we're mostly using int64/string types

KiKoS0: Looks like `net.IP` does encode/decode fine: https://github.com/golang/go/blob/6c3b5a2798c83d583cb37dba9f39c47300d19f1f/src/net/ip.go#L360-L386
```
1751897150.508521 [0 10.30.0.30:55794] "HSET" "{connect}:gateways" "01JZJANEEXZHN333V000R1PDVF" "{\"id\":\"01JZJANEEXZHN333V000R1PDVF\",\"status\":\"active\",\"last_heartbeat\":1751897150508,\"hostname\":\"e418a375e3ba\",\"ip\":\"10.30.0.30\"}"
```

pkg/telemetry/metrics/counter.go
```diff
@@ -238,6 +238,42 @@ func IncrConnectGatewayReceivedRouterPubSubMessageCounter(ctx context.Context, v
...
+		MetricName:  "connect_gateway.grpc.forward_total",
+		Description: "Total number of messages forwarded via gRPC to connect gateways",
+		Tags:        opts.Tags,
+	})
+}
+
+func IncrConnectGatewayGRPCClientFailureCounter(ctx context.Context, value int64, opts CounterOpt) {
+	RecordCounterMetric(ctx, value, CounterOpt{
+		PkgName:     opts.PkgName,
+		MetricName:  "connect_gateway.grpc.client_failure_total",
```
darwin67, line 271: not a blocker, but this style of metrics likely won't work well when in-housing metrics later on.

don't need to change it for now, but just wanted to raise awareness.

pkg/connect/service.go
```diff
@@ -83,7 +92,8 @@ type connectGatewaySvc struct {
 	workerRequestExtendLeaseInterval                      time.Duration
 	workerRequestLeaseDuration                            time.Duration
 
-	hostname string
+	hostname  string
+	ipAddress net.IP
```
darwin67, line 96: this is IPv4 right?
Prod is IPv6 only so it might not work.

pkg/connect/gateway.go
```diff
@@ -386,11 +386,14 @@ func (c *connectGatewaySvc) Handler() http.Handler {
 		}
 
 		{
-			onSubscribed := make(chan struct{})
+			onSubscribedPubsub := make(chan struct{})
+			onSubscribedGRPC := make(chan struct{})
 			// Wait for relevant messages and forward them over the WebSocket connection
-			go ch.receiveRouterMessages(ctx, onSubscribed)
+			go ch.receiveRouterMessagesFromPubsub(ctx, onSubscribedPubsub)
+			<-onSubscribedPubsub
```
darwin67, line 393: won't this block the lines below for the `ch.receiveRouterMessagesFromGRPC` from running?

KiKoS0: oh that channel just waits for the subscription to be created, it closes the channel just before blocking
https://github.com/inngest/inngest/blob/234315b06da46518d7f1ad26b22683087ad29036/pkg/connect/pubsub/proxy.go#L609-L613

pkg/connect/service.go
KiKoS0: net.IP is a byte slice so it should cover both i think. If the returned `POD_IP` from Kubernetes is IPv6, ParseIP should cover that.
https://github.com/golang/go/blob/54c9d776302d53ab1907645cb67fa4a948e1500c/src/net/ip.go#L521-L527
```go
// ParseIP parses s as an IP address, returning the result.
// The string s can be in IPv4 dotted decimal ("192.0.2.1"), IPv6
// ("2001:db8::68"), or IPv4-mapped IPv6 ("::ffff:192.0.2.1") form.
// If s is not a valid textual representation of an IP address,
// ParseIP returns nil. The returned address is always 16 bytes,
// IPv4 addresses are returned in IPv4-mapped IPv6 form.
func ParseIP(s string) IP {
```

pkg/telemetry/metrics/counter.go
KiKoS0: do you mean the periods in the metric name or something else?

darwin67: yeah, the periods

pkg/connect/service.go
darwin67: cool 👍

pkg/connect/gateway.go
darwin67: cool, thanks for clarifying

PR #2652, opened 2025-07-03
Improve latency

BrunoScheufler:
## Description

This PR includes important latency improvements

- In-memory hints: During `EnqueueItem`, we add a shadow continue to hint to the executor that a shadow partition can be scanned. During backlog refills, we add a normal continue to hint to the executor that the ready partition can be scanned.
- GCRA retry after: When GCRA throttles run starts, we now return the timestamp at which the next item will be allowed. The retry after is capped to 5 minutes to avoid breaking user expectations when the throttle period is updated.

These improvements should reduce unnecessary QPS and lead to faster refills and processing.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/backlog.go
+++ b/pkg/execution/state/redis_state/backlog.go
@@ -664,7 +664,7 @@ type BacklogRefillResult struct {
 	Capacity          int
 	Refill            int
 	RefilledItems     []string
-	RetryAfter        time.Time
+	RetryAt           time.Time
 }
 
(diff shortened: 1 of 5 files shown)


pkg/execution/state/redis_state/backlog.go
```diff
@@ -687,6 +664,7 @@ type BacklogRefillResult struct {
 	Capacity          int
 	Refill            int
 	RefilledItems     []string
+	RetryAfter        time.Time
```
darwin67: nit: `RetryAt` is better. `RetryAfter` to me reads as a duration.

darwin67 (approved):

PR #2656, opened 2025-07-07
First pass at EventsV2 dev server resolver

albertchae:
## Description

This is the first PR implementing the EventsV2 resolver for the dev server

The EventsV2 gql schema is copied directly from monorepo.

Currently unimplemented functionality:
- function runs
- source
- filtering on internal events
- cel query

This PR introduces changes to the sqlc queries to support filtering on multiple event names and new queries for getting total count.
It's possible we will have to switch these to goqu to support additional filtering in the follow up work.

I am also punting on parity in Postgres to a follow up.

## Motivation

new events page work for dev server

- [ ] Docs

## Checklist


pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+
+		raw, err := marshalRaw(e)
+		if err != nil {
+			return nil, err
+		}
+
+		eventV2 := models.EventV2{
+			EnvID:          e.WorkspaceID,
+			ID:             e.InternalID(), // TODO this or event id?
+			IdempotencyKey: &e.EventID,
```
albertchae: am I interpreting these columns correctly?

pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+			Runs:           []*models.FunctionRunV2{}, // TODO
+			Version:        &e.EventVersion,
+		}
+
+		if e.SourceID != nil {
+			eventV2.Source = &models.EventSource{
+				ID:         e.SourceID.String(),
+				Name:       &e.Source,
+				SourceKind: "TODO",
+			}
```
albertchae, line 69: this will come in a follow up but we may end up hardcoding something here anyway, since we aren't using event keys in dev server

pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+		eventEdges = append(eventEdges, &models.EventsEdge{
+			Node:   &eventV2,
+			Cursor: base64.StdEncoding.EncodeToString(cursorByt),
+		})
+	}
+
+	totalCount, err := qr.Data.GetEventsCount(ctx, consts.DevServerAccountID, consts.DevServerEnvID, opts)
+	if err != nil {
+		return nil, err
+	}
```
albertchae: might move this to a field resolver later to follow the pattern of other `TotalCount` fields

pkg/cqrs/base_cqrs/sqlc/sqlite/queries.sql
```diff
@@ -170,11 +170,18 @@ WHERE
...
 
 -- name: WorkspaceNamedEvents :many
-SELECT * FROM events WHERE internal_id < @cursor AND received_at <= @before AND received_at >= @after AND event_name = @name ORDER BY internal_id DESC LIMIT ?;
+SELECT * FROM events WHERE internal_id < @cursor AND received_at <= @before AND received_at >= @after AND event_name in (sqlc.slice('@names')) ORDER BY internal_id DESC LIMIT ?;
+
+-- name: WorkspaceCountEvents :one
+SELECT count(*) FROM events WHERE received_at <= @before AND received_at >= @after;
+
+-- name: WorkspaceCountNamedEvents :one
+SELECT count(*) FROM events WHERE received_at <= @before AND received_at >= @after AND event_name in (sqlc.slice('@names'));
```
albertchae, line 184: will likely end up replacing all of these with goqu instead since it'll be easier to handle all the filtering we want

tests/golang/event_list_test.go
albertchae, line 1: mostly a copy of [test/integration/app/event_list_test.go](https://github.com/inngest/monorepo/blob/develop/test/integration/app/event_list_test.go)

pkg/api/apiv1/events.go
```diff
@@ -81,7 +81,7 @@ func (a router) getEvents(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if name := r.FormValue("name"); name != "" {
-		opts.Name = &name
+		opts.Names = []string{name}
```
amh4r, line 84: Is this a bug fix?

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: Yea, you are. I took the new schema as an opportunity to disambiguate the 2 event IDs. The non-internal event ID (the one the user sets when sending the event) is really an idempotency key

pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+
+		eventV2 := models.EventV2{
+			EnvID:          e.WorkspaceID,
+			ID:             e.InternalID(), // TODO this or event id?
+			IdempotencyKey: &e.EventID,
+			Name:           e.EventName,
+			OccurredAt:     time.UnixMilli(e.EventTS),
+			Raw:            raw,
+			ReceivedAt:     e.ReceivedAt,
+			Runs:           []*models.FunctionRunV2{}, // TODO
```
amh4r: `Raw` and `Runs` should both be resolvers. We want to lazy-load them to avoid over-fetching in the run list

pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+			Version:        &e.EventVersion,
+		}
+
+		if e.SourceID != nil {
+			eventV2.Source = &models.EventSource{
```
amh4r, line 65: Let's also make `Source` a resolver for the same reason as `Raw` and `Runs`

pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return &models.EventsConnection{
+		Edges: eventEdges,
+		PageInfo: &models.PageInfo{
+			HasNextPage:     hasNextPage,
+			HasPreviousPage: after != nil, // accurate as long as clients always send null cursor for first page
+			StartCursor:     startCursor,
+			EndCursor:       endCursor,
+		},
+		TotalCount: int(totalCount),
```
amh4r: `TotalCount` should be a resolver, since that lets the UI decouple the event list query from the total count query

jacobheric (approved):

tests/golang/event_list_test.go
```diff
@@ -0,0 +1,361 @@
...
+			//}
+
+			for _, edge := range res.Edges {
+				if edge.Node.ID == ulid.MustParse(eventID) {
+					runs = edge.Node.Runs
+				}
+			}
+			// TODO: punting on FunctionRunsV2 for now
+			if !a.Len(runs, 0) {
+				return
```
amh4r: The latest `testify` release lets us use `require` in `EventuallyWithT`! It obviates the need for the annoying early return stuff that `assert` necessitates

```suggestion
			r := require.New(t)
			res, err := c.GetEvents(ctx, client.GetEventsOpts{
				PageSize: 10,
				Filter: models.EventsFilter{
					EventNames: []string{eventName, "inngest/function.finished"},
					From:       time.Now().Add(-time.Minute),
				},
			})
			r.NoError(err)

			// Includes "inngest/function.finished".
			//if !a.Len(res.Edges, 2) {
			//	return
			//}

			for _, edge := range res.Edges {
				if edge.Node.ID == ulid.MustParse(eventID) {
					runs = edge.Node.Runs
				}
			}
			// TODO: punting on FunctionRunsV2 for now
			r.Len(runs, 0)
```

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: `Raw` - this is a resolver in monorepo because it is actually doing a separate fetch to the metrics service, but here we are just marshalling data that's already on the row we pulled in (see `marshalRaw` in this file), so there's no overfetching. IMO it's not worth making this a separate resolver

`Runs` - yes planning on making this a resolver with a data loader

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: agree although it mind end up being hardcoded on dev server so might not be worth the ceremony to make it a separate resolver. Might need to for self hosted though

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: yes I commented on that here https://github.com/inngest/inngest/pull/2656/files#r2195032643

from the principle of making fields separate resolvers when they issue separate DB queries I agree (and also maintaining consistency with the other `TotalCount` fields in our API), but I was wondering when will the frontend ever decouple querying `EventsConnection` and only want the total count and not the contents of a page? IMO it might not be worth the weird hack of adding the filter on the connection in augmented.go and then having a separate resolver 🤷

pkg/api/apiv1/events.go
albertchae: I changed the shape of the struct here https://github.com/inngest/inngest/pull/2656/files#diff-c04250f7155d49a574b9ee5626119dab207a5a11080b67fa83d3f928c3b9f7d8

This allows us to use the same SQL queries + cqrs code to support requesting for a single event name in the REST API vs multiple event names in the gql API

tests/golang/event_list_test.go
albertchae: great! I remember your slack post about that. will update here and the monorepo tests too

tests/golang/event_list_test.go
```diff
@@ -0,0 +1,361 @@
...
+
+		// Send 1 event and wait for it show in the API.
+		event1ID, err := ic.Send(ctx, inngestgo.Event{Name: event1Name})
+		r.NoError(err)
+		r.EventuallyWithT(func(t *assert.CollectT) {
+			a := assert.New(t)
+			res, err := c.GetEvents(ctx, client.GetEventsOpts{
+				PageSize: 10,
+				Filter: models.EventsFilter{
+					EventNames: []string{event1Name, event2Name},
```
albertchae, line 115: forgot to mention but without filtering for EventNames explicitly here (and other subsequent tests), we have test pollution from events sent in other cases. This also prevents us from parallelizing the tests.

@amh4r any ideas on solving this? the monorepo tests didn't rely on EventNames so not sure if they were isolating tests in some other way

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: If `Raw` isn't a resolver then memory is gonna spike if loading a bunch of large events. We'll never need the raw data for the event list -- only when loading an individual event

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: We may need it for self-hosted eventually (since it can have multiple event keys), but hardcoding for now sounds good

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: If we don't separate them then we'll recalculate the total count on every page fetch. Like if the user starts doing infinite scroll then we don't want to keep recalculating the total count

albertchae: oh so the frontend doesn't request total count for each page? it only requests for the first page?

tests/golang/event_list_test.go
amh4r: I like to use a `randomSuffix` helper to ensure the event names are unique
https://github.com/inngest/inngest/blob/main/tests/golang/parallel_test.go#L137

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: Yea, it's separate in Cloud because the total count is irrespective of pagination

tests/golang/event_list_test.go
albertchae: You can see that I'm already using `randomSuffix` in all the tests here. I wasn't asking about that, the issue is if we leave out the `EventNames` filter entirely, events from other test cases show up in the same time range that we're querying for here. However the `EventNames` filter wasn't needed in https://github.com/inngest/monorepo/blob/develop/test/integration/app/event_list_test.go for some reason, so not sure how those tests are being isolated.

amh4r: Ah sorry, misunderstood you. Maybe the only way is to constrain the time range to the test?

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: OK it makes sense to only request `totalCount` on the first page load (and thus have a separate resolver), but I don't think this is how our frontend is currently implemented.

We use this query for each page https://github.com/inngest/inngest/blob/c95a4d898c3fbc35e85107aba59024dc08796845/ui/apps/dashboard/src/components/Events/useEvents.ts#L46 so we end up computing the total count for each page. You can verify this in the network tab on our events page too

Also somewhat orthogonal but I think we end up treating infinite scroll with a moving time window which could be problematic, since we take [`Until` as `time.Now()` by default](https://github.com/inngest/monorepo/blob/698b688a8c362dca6e191080def6c374ef137c59/api/gql/resolver/workspace.go#L1966)

tests/golang/event_list_test.go
albertchae: hm I tried that and it worked for some tests but not all. I might end up making the tests looser (i.e. don't check for exact counts, just check that the events I'm expecting are returned in the response), but I'll evaluate that after adding the other cases. The ideal thing could be if each test ran as a separate inngest environment which I don't think the dev server supports yet

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: fair enough, I didn't see it in the event list query either. I guess the rule of thumb on using separate resolvers is whether we need to make an additional data fetch or would waste significant memory allocating structs that ultimately don't get sent over the network?

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: That's an oversight. Can you create a ticket for it? We separate total count in the run list:
https://github.com/inngest/inngest/blob/main/ui/apps/dashboard/src/components/Runs/Runs.tsx#L123-L128

Separating the total count is primarily a performance gain. Including the total count in the same query effectively negates the chunking optimization in the backend

pkg/coreapi/graph/resolvers/events_v2.go
amh4r: Yea, those are good rules of thumb. Using a resolver is good when there's an expense to fetching it (either requiring a separate query in the backend or using an appreciable amount of resources)
https://github.com/inngest/monorepo/blob/develop/api/gqlgen.yml#L375-L383

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: https://linear.app/inngest/issue/CON-25/fix-pagination-behavior-with-totalcount-on-events-page

albertchae: Moved `TotalCount` into its own resolver and was able to avoid adding input fields into the Connection struct in augmented.go by using `GetFieldContext` from gqlgen

- https://stackoverflow.com/a/73900329

These mention a similar `GetResolverContext` but it's deprecated now in favor of `GetFieldContext`
- https://github.com/99designs/gqlgen/issues/813#issuecomment-518990626
- https://github.com/99designs/gqlgen/issues/796#issuecomment-533866720
- https://github.com/99designs/gqlgen/issues/681#issuecomment-484264913

pkg/coreapi/graph/resolvers/events_v2.go
albertchae: Moved `Raw` into a resolver, at the moment it'll technically trigger an N+1 if we request `Raw` off an event list which we won't be from the frontend as discussed earlier in the thread. This will be mitigated anyway with a data loader in a pending branch I'm working on for the singular EventV2 resolver

PR #2658, opened 2025-07-08
Garbage-collect empty outdated backlogs, always clean up backlog/partition meta

BrunoScheufler:
## Description

This PR adds automatic garbage collection to empty, outdated backlogs.

During backlogNormalizePrep, we check if the backlog is empty. If it is, there's no reason to normalize it in the first place, as there are no items to move. In this case, we simply remove the dangling pointers and also drop the backlog meta (as there are no more items to be refilled).

This should prevent us from spinning on outdated backlogs. We still have to understand _why_ pointers to empty/non-existing backlog exist in shadow partitions. To this end, I've made sure to log whenever we garbage-collect backlogs. 

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/lua/includes/update_backlog_pointer.lua
+++ b/pkg/execution/state/redis_state/lua/includes/update_backlog_pointer.lua
@@ -1,13 +1,19 @@
-local function updateBacklogPointer(keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
+local function updateBacklogPointer(keyShadowPartitionMeta, keyBacklogMeta, keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
   -- Retrieve the earliest item score in the backlog in milliseconds
   local earliestBacklogScore = get_earliest_score(keyBacklogSet)
 
   -- If backlog is empty, update dangling pointers in shadow partition
   if earliestBacklogScore == 0 then
+    -- Remove meta
+    redis.call("HDEL", keyBacklogMeta, backlogID)
+
     redis.call("ZREM", keyShadowPartitionSet, backlogID)
 
     -- If shadow partition has no more backlogs, update global/account pointers
     if tonumber(redis.call("ZCARD", keyShadowPartitionSet)) == 0 then
+      -- Remove meta
+      redis.call("HDEL", keyShadowPartitionMeta, partitionID)
+
       redis.call("ZREM", keyGlobalShadowPartitionSet, partitionID)
       redis.call("ZREM", keyAccountShadowPartitionSet, partitionID)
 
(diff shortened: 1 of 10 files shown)


pkg/execution/state/redis_state/lua/queue/backlogPrepareNormalize.lua
```diff
@@ -9,29 +9,53 @@
...
+if backlogCount == nil or backlogCount == false or backlogCount == 0 then
+  -- Remove pointer from shadow partition
+  redis.call("ZREM", keyShadowPartitionSet, backlogID)
+
+  -- Remove meta
+  redis.call("HDEL", keyBacklogMeta, backlogID)
+
+  -- Update pointers
+  updateBacklogPointer(keyShadowPartitionMeta, keyBacklogMeta, keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
+
```
BrunoScheufler: This is the primary change: When we're preparing an outdated backlog for normalization, we now check if it is empty. In that case, we drop the dangling pointer and metadata, as the backlog will not be processed again.

We also update all pointers to reflect the changes and prevent spinning on an empty partition, etc.

pkg/execution/state/redis_state/lua/includes/update_backlog_pointer.lua
```diff
@@ -1,13 +1,19 @@
-local function updateBacklogPointer(keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
+local function updateBacklogPointer(keyShadowPartitionMeta, keyBacklogMeta, keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
   -- Retrieve the earliest item score in the backlog in milliseconds
   local earliestBacklogScore = get_earliest_score(keyBacklogSet)
 
   -- If backlog is empty, update dangling pointers in shadow partition
   if earliestBacklogScore == 0 then
+    -- Remove meta
+    redis.call("HDEL", keyBacklogMeta, backlogID)
```
BrunoScheufler, line 8: This script is used during backlog refills, dequeues, and the updated logic in this PR: https://github.com/inngest/inngest/pull/2658/files#r2192564610

When encountering an empty backlog or shadow partition, not only do we drop the pointer, we also clean up the metadata. This is important to prevent memory leaks.

pkg/execution/state/redis_state/lua/queue/backlogRequeue.lua
```diff
@@ -36,17 +37,8 @@ end
...
-  if tonumber(redis.call("ZCARD", keyShadowPartitionSet)) == 0 then
-    redis.call("ZREM", keyGlobalShadowPartitionSet, partitionID)
-    redis.call("ZREM", keyAccountShadowPartitionSet, partitionID)
-
-    if tonumber(redis.call("ZCARD", keyAccountShadowPartitionSet)) == 0 then
-      redis.call("ZREM", keyGlobalAccountShadowPartitionSet, accountID)
-    end
-  end
+  -- Update pointers
+  updateBacklogPointer(keyShadowPartitionMeta, keyBacklogMeta, keyGlobalShadowPartitionSet, keyGlobalAccountShadowPartitionSet, keyAccountShadowPartitionSet, keyShadowPartitionSet, keyBacklogSet, accountID, partitionID, backlogID)
```
BrunoScheufler: This change is equivalent, except that the reusable function also clears out the metadata now.

pkg/execution/state/redis_state/lua/includes/update_backlog_pointer.lua
darwin67: When I comment out the changes in this file, the new test still pass.
That feels like the cleanåup is already done properly elsewhere?
Or that this code has no effect somehow?

pkg/execution/state/redis_state/lua/queue/backlogPrepareNormalize.lua
```diff
@@ -9,29 +9,53 @@
...
+local backlogCount = redis.call("ZCARD", keyBacklogSet)
+
+-- If backlog is empty, garbage-collect it from shadow partition
+if backlogCount == nil or backlogCount == false or backlogCount == 0 then
+  -- Remove pointer from shadow partition
+  redis.call("ZREM", keyShadowPartitionSet, backlogID)
+
+  -- Remove meta
+  redis.call("HDEL", keyBacklogMeta, backlogID)
```
darwin67: looks like these lines were called before the `updateBacklogPointer` fn, so the cleanup already happened here.

darwin67: Considering `updateBacklogPointer` is used in a lot more areas, this section might be redundant since it essentially do the same thing right after.

Unless there's a case where this is needed but `updateBacklogPointer` doesn't cover?

PR #2662, opened 2025-07-09
Clean up source backlog pointers while normalizing

BrunoScheufler:
## Description

This PR improves normalization as follows

- Remove immediate normalization. Having an extra entrypoint to normalization caused a lot of edge cases with contention, potentially stalling shadow partition processing, and more
- Only marking backlogs as outdated if the constraints are at least the same version or newer. In case the function config is not yet updated but new items are enqueued, backlogs should not be normalized back to an old version, ever.
- Update backlog pointers during normalization in enqueue, after removing item from old backlog: When normalizing, items are gradually moved from the old source backlog to the desired target backlog. Previously, we left pointers to now empty source backlogs in shadow partitions. This is now fixed by always updating backlog pointers.

This should lead to 1) fewer false positive normalizations 2) significantly reduced contention 3) no impact on shadow processing for non-outdated backlogs

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/backlog.go
+++ b/pkg/execution/state/redis_state/backlog.go
@@ -946,13 +946,13 @@ func (q *queue) BacklogRequeue(ctx context.Context, backlog *QueueBacklog, sp *Q
 	}
 }
 
-func (q *queue) BacklogPrepareNormalize(ctx context.Context, b *QueueBacklog, sp *QueueShadowPartition, normalizeAsyncMinimum int) (int, bool, error) {
+func (q *queue) BacklogPrepareNormalize(ctx context.Context, b *QueueBacklog, sp *QueueShadowPartition) error {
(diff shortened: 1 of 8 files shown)


pkg/execution/state/redis_state/backlog.go
```diff
@@ -273,8 +275,9 @@ type BacklogThrottle struct {
 }
 
 type QueueBacklog struct {
-	BacklogID         string `json:"id,omitempty"`
-	ShadowPartitionID string `json:"sid,omitempty"`
+	BacklogID               string `json:"id,omitempty"`
+	ShadowPartitionID       string `json:"sid,omitempty"`
+	EarliestFunctionVersion int    `json:"fv,omitempty"`
```
darwin67, line 280: what's the reason for calling it `Earliest`?
it gives an impression that the version number should be low.

BrunoScheufler: Because the backlog metadata is stored after the first item is enqueued. We don't update it every time a new item is added, and the backlog may store items for multiple versions

darwin67 (approved):
have a non blocking comment, looks good in general.

PR #2670, opened 2025-07-10
Implement single EventV2 resolver with EventV2 data loader

albertchae:
## Description

Implement resolver for singular EventV2. This will get used by the dev server in a query very similar to https://github.com/inngest/inngest/blob/62a3360c72229ca39bae35c769b00c03f1f05c17/ui/apps/dashboard/src/components/Events/useEvents.ts#L132-L148

Use data loader for single `EventV2` as well as `Raw` so we don't fetch the same event twice from the database in each resolver. Query logs show a single batched call to `GetEventsByInternalId` instead of 2 to `GetEventByInternalId` per event.

Add event dataloader Prime from EventsV2 query. Because EventsV2 calls `GetEvents` which doesn't query the database by id, we need to explicitly prime with the results after. Query logs show a single call to `GetEvents` instead of one to both `GetEvents` and to `GetEventsByInternalId` when running `event_list_test.go`

## Motivation

Another resolver needed for new events page on dev server

- [ ] Docs

## Checklist


pkg/coreapi/graph/resolvers/events_v2.go
```diff
@@ -112,3 +101,26 @@ func (c *EventsV2ConnectionCursor) Decode(val string) error {
 type EventsV2ConnectionCursor struct {
 	ID string
 }
+
+func cqrsEventToGqlEvent(e *cqrs.Event) *models.EventV2 {
```
amh4r:
```suggestion
func cqrsEventToGQLEvent(e *cqrs.Event) *models.EventV2 {
```

Nit: Go's convention is to keep acronyms capitalized

albertchae: changed

PR #2681, opened 2025-07-13
Support multipart/form-data in ParseStream

amh4r:
## Description
Add support for processing events in `multipart/form-data` requests.

## Motivation
Support `multipart/form-data` webhooks.

- [ ] Docs

### DIFF
--- a/pkg/eventstream/eventstream.go
+++ b/pkg/eventstream/eventstream.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
+	"mime/multipart"
 
 	"github.com/inngest/inngest/pkg/consts"
 )
@@ -23,12 +25,13 @@ type StreamItem struct {
 // ensuring that no individual event is too large.
 //
 // This closes the given channel once the JSON stream in the given reader has been parsed.
+// Supports both JSON and multipart/form-data content types.
 //
 // Usage:
 //
 //			var err error
 //			go func() {
-//			        err = ParseStream(ctx, r, stream)
+//			        err = ParseStream(ctx, r, stream, contentType)
 //			()
 //
 //			for bytes := range stream {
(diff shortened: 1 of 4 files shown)


pkg/eventstream/eventstream.go
```diff
@@ -110,3 +127,65 @@ func ParseStream(ctx context.Context, r io.Reader, stream chan StreamItem, maxSi
...
+		}
+
+		// Skip empty parts
+		if len(data) == 0 {
+			continue
+		}
+
+		fieldName := part.FormName()
+		if fieldName != "" {
+			formData[fieldName] = string(data)
```
albertchae, line 168: part data can be arbitrary binary right? but it looks like go strings can store arbitrary binary too, so this should be OK

pkg/eventstream/eventstream.go
```diff
@@ -110,3 +127,65 @@ func ParseStream(ctx context.Context, r io.Reader, stream chan StreamItem, maxSi
...
+		}
+	}
+
+	if len(formData) > 0 {
+		byt, err := json.Marshal(formData)
```
albertchae, line 173: won't this possibly error if the part data was binary, since we can't directly represent arbitary binary in JSON?

pkg/eventstream/eventstream.go
```diff
@@ -110,3 +127,65 @@ func ParseStream(ctx context.Context, r io.Reader, stream chan StreamItem, maxSi
...
+		if err != nil {
+			return err
+		}
+
+		if len(byt) > maxSize {
+			return fmt.Errorf("%w: Max %d bytes / Size %d bytes", ErrEventTooLarge, maxSize, len(byt))
+		}
+
+		select {
+		case stream <- StreamItem{N: 0, Item: json.RawMessage(byt)}:
```
albertchae, line 183: what is N supposed to be? an index for the items streamed so far?

albertchae (approved):
looks fine for `multipart/form-data` requests that only contain data representable in utf-8, so don't want to block if that's all we need to support immediately, but I don't think this will work for arbitrary binary (that can't be represented by utf-8)

pkg/eventstream/eventstream.go
amh4r: I think it can be arbitrary bytes, but at this point I don't think we should support that. If we intentionally don't support it, do you see any issues with the code? For example, will arbitrary bytes just end up as gobbledygook when passed to `string`?

pkg/eventstream/eventstream.go
amh4r: It's the index for the event in the request body (like if multiple events are sent in the same webhook). AFAICT, there isn't a way to support multiple events in a single `multipart/form-data` request, so I hardcoded to 0

pkg/eventstream/eventstream.go
amh4r: It won't because we've stringified all the map values

pkg/eventstream/eventstream.go
albertchae: yeah if we're not intentionally supporting it I think this PR is good. Thanks for adding the test

PR #2640, opened 2025-06-27
Add tenant instrumentor

darwin67:
## Description

Provide a passthrough function that can be invoked for instrumentation.
Also an iterator for backlogs so we can run operations on each of them.

## Type of change (choose one)
- [x] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [ ] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #2683, opened 2025-07-14
Enroll future queue items to backlogs

BrunoScheufler:
## Description

This PR should be merged on top of https://github.com/inngest/inngest/pull/2640 and adds the following changes

- A new partition iterator that periodically (every 10s by default) iterates the global partition pointer (ZSET) and spins up a goroutine per partition, which then
  - runs the tenant instrumentation logic
  - reports the partition size gauge OTel metrics
  - waits until capacity is available (semaphore) and enrolls future queue items in the partition to their respective backlog

- Backlog enrollment simply uses the existing partition iterator for items enqueued at least 5s into the future. This prevents any soon-to-run items from being moved to the backlog. We also exclude items refilled from a backlog (as they're already accounted for), items currently leased (should never happen, but better to be safe). In case key queues are disabled, we exit the check early.

This consolidates existing methods to iterate over partitions to reduce the load on the queue and reuse existing work. It also allows us to trivially extend observability and background processing efforts on partitions.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue.go
```diff
@@ -3541,6 +3558,132 @@ func (q *queue) InProgress(ctx context.Context, prefix string, concurrencyKey st
...
+			Min("-inf").
+			Max("+inf").
+			Byscore().
+			Limit(offset, chunkSize).
+			Build()
+
+		pkeys, err := r.Do(ctx, cmd).AsStrSlice()
+		if err != nil {
+			return fmt.Errorf("error retrieving partitions for instrumentation: %w", err)
+		}
```
darwin67, line 3589: might be cool to make this a `PartitionIterator` with an option to ignore leases or something.

it'll be nice to reduce these sort of direct redis command usage over the codebase, so it's more concentrated and make the wrapper more flexible.

don't have to do this in a PR, so would be good to at least document it as a follow up.

darwin67 (approved):
cool, looks good

PR #2685, opened 2025-07-14
Implement EventV2.Runs to return FunctionRunV2

albertchae:
## Description

Add associated runs with events in EventsV2

Copied tests from monorepo and added several cases which I'll backport
- multiple event types 1-1 with multiple functions
- one event fan out to multiple functions
- same event type with multiple runs to the same function
- multiple events batched to the same function

Unfortunately the current implementation has an N+1. Ideally we would use a data loader to batch these into a single db query where given a list of eventIDs/triggerIDs we fetch all the associated function runs, but we have a few things working against us here.

- Because multiple events can trigger one function (batched events), `trace_runs.trigger_ids` is a comma separated string of ULID strings (stored as a BLOB). Because it's not a JSON array we can't use `json_each` and instead have to use substring matching with INSTR/POSITION
- `trace_runs` does have a foreign key to `event_batches`, but `event_batches.event_ids` is also a comma separated string so there's no JOIN that gets around the substring matching.
- This is still fine (modulo some performance issues) as long as we could provide a list of event_ids and join against that in a query, but sqlc has an open issue where we can't use `value` from `json_each` to unnest the list into a CTE https://github.com/sqlc-dev/sqlc/issues/1830. This prevents us from using this query
  ```sql
  WITH event_ids AS (
      SELECT value AS id
      FROM json_each(sqlc.slice('trigger_ids'))
  )
  SELECT tr.*
  FROM trace_runs tr
  CROSS JOIN event_ids s
  WHERE INSTR(CAST(tr.trigger_ids AS TEXT), s.id) > 0
  GROUP BY tr.run_id;
  ```

## Motivation

parity with cloud API

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/coreapi/graph/resolvers/event_v2.go
+++ b/pkg/coreapi/graph/resolvers/event_v2.go
@@ -36,8 +36,8 @@ func (er eventV2Resolver) Runs(ctx context.Context, obj *models.EventV2) ([]*mod
 		return nil, err
 	}
 
-	functionRuns := make([]*models.FunctionRunV2, len(traceRuns))
-	for i, r := range traceRuns {
+	functionRuns := make([]*models.FunctionRunV2, 0, len(traceRuns))
+	for _, r := range traceRuns {
 		// TODO dedupe cqrs.TraceRun to models.FunctionRunV2 transformation
 		var (
 			started   *time.Time
@@ -66,7 +66,7 @@ func (er eventV2Resolver) Runs(ctx context.Context, obj *models.EventV2) ([]*mod
 			continue
 		}
 
-		functionRuns[i] = &models.FunctionRunV2{
+		functionRuns = append(functionRuns, &models.FunctionRunV2{
 			ID:             runID,
 			AppID:          r.AppID,
 			FunctionID:     r.FunctionID,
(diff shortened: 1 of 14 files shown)


pkg/coreapi/graph/resolvers/event_v2.go
```diff
@@ -25,7 +26,67 @@ func (qr *queryResolver) EventV2(ctx context.Context, id ulid.ULID) (*models.Eve
...
+			sourceID = &r.SourceID
+		}
+		if len(r.Output) > 0 {
+			s := string(r.Output)
+			output = &s
+		}
+		runID := ulid.MustParse(r.RunID)
+		status, err := models.ToFunctionRunStatus(r.Status)
+		if err != nil {
+			continue
```
amh4r, line 66: This can lead to a sparse slice. Is that a problem?

albertchae: hm true, we'd return the zero struct of `FunctionRunV2` here but I think it'd be better to change it to not return a sparse slice

albertchae: ok changed it to initialize with 0 len and the same capacity and append instead

amh4r (approved):

PR #2690, opened 2025-07-14
Add step-level parallelism optimization disabling

amh4r:
## Description
Handle the new "parallel mode" option in SDK step responses. This only applies to v2 execution.

There are 2 modes:
- Wait: The existing behavior (only send 1 discovery request after all parallel steps end)
- Race: After the parallel step ends, send a discovery request (doesn't matter if other parallel steps haven't ended)

The new "race" mode is opt-out, so "wait" is the default.

## Motivation
Enables some desired logic in SDKs. Specifically, allow sequential steps in parallel groups to run independently

## Context
Spec: https://www.notion.so/inngest/Parallel-mode-race-22db64753bbd80749be8e247f8e3ca70

- [ ] Docs

### DIFF
--- a/cmd/commands/start.go
+++ b/cmd/commands/start.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/inngest/inngest/cmd/commands/internal/localconfig"
+	"github.com/inngest/inngest/pkg/authn"
 	"github.com/inngest/inngest/pkg/config"
 	"github.com/inngest/inngest/pkg/devserver"
 	"github.com/inngest/inngest/pkg/headers"
@@ -150,6 +151,11 @@ func doStart(cmd *cobra.Command, args []string) {
 	signingKey := viper.GetString("signing-key")
 	if signingKey == "" {
 		fmt.Println("Error: signing-key is required")
+		os.Exit(1)
+	}
+	_, err = authn.HashedSigningKey(signingKey)
+	if err != nil {
+		fmt.Printf("Error: signing-key must be a valid hexadecimal string\n")
 		os.Exit(1)
 	}
 
--- a/pkg/authn/authn.go
+++ b/pkg/authn/authn.go
@@ -1,0 +1,35 @@
+package authn
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+// AuthContext represents authentication context information
+type AuthContext struct {
+	// Add fields as needed for your authentication context
+	isAuthenticated bool
+	// Add other relevant fields
+}
+
+var (
+	// ErrIncorrectStrategy represents an attempt authenticating via an incorrect strategy,
+	// eg. via an API key when no API key is present in the authorization header.
+	ErrIncorrectStrategy = errors.New("incorrect auth strategy")
+
+	// ErrNoAuthentication is returned when no authorization strategy was successful and the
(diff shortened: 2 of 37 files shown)


pkg/execution/queue/item.go
```diff
@@ -503,6 +503,20 @@ func GetEdge(i Item) (*PayloadEdge, error) {
 	}
 }
 
+// OptimizedParallelism returns true if the item has optimized parallelism
+// enabled.
+func (i Item) OptimizedParallelism() bool {
+	optPar, ok := i.Metadata["opt_par"].(bool)
```
jpwilliams: Can we move `"opt_par"` to a const somewhere?

pkg/execution/queue/item.go
```diff
@@ -503,6 +503,20 @@ func GetEdge(i Item) (*PayloadEdge, error) {
 	}
 }
 
+// OptimizedParallelism returns true if the item has optimized parallelism
+// enabled.
```
jpwilliams: A note of the intention to default to `true` would be good here (like the `opcode.go` version).

pkg/execution/executor/executor.go
```diff
@@ -2455,8 +2455,9 @@ func (e *executor) handleGeneratorStep(ctx context.Context, i *runInstance, gen
 		Payload:               queue.PayloadEdge{Edge: nextEdge},
 		Metadata:              make(map[string]any),
 	}
+	nextItem.SetOptimizedParallelism(gen.OptimizedParallelism())
 
-	if !hasPendingSteps {
+	if !hasPendingSteps || !i.item.OptimizedParallelism() {
```
jpwilliams: `if !hasPendingSteps` felt bad to have everywhere, especially seeing as it's such a common check.

Would it be reasonable to have a helper like:
```go
func shouldEnqueueDiscovery(hasPendingSteps bool, qi queue.Item) bool {
  return !hasPendingSteps || !nextItem.item.OptimizedParallelism()
}

// ...

if shouldEnqueueDiscovery(hasPendingSteps, qi) {
  // ...
}
```

It's a total nit and feels a bit over the top, but at least then if we require more things to check etc we're not having to trawl through the code to find where we `Enqueue()`.

Maybe there are a few spots where it just doesn't make sense too. 'tis a nit - feel free to ignore.

jpwilliams (approved):

jpwilliams (changes requested):
I think we may also need to consider a couple more locations for the check?

- Discovery step following a `step.waitForEvent()`, `step.waitForSignal()`, or `step.invoke()` timeout https://github.com/inngest/inngest/blob/9ebbde880ffd3a8d3997ddbb83de4c1bb923b3cc/pkg/execution/executor/executor.go#L2018
- Discovery step following a pause, which is likely a `step.waitForEvent()`, `step.waitForSignal()`, or `step.invoke()` completion https://github.com/inngest/inngest/blob/9ebbde880ffd3a8d3997ddbb83de4c1bb923b3cc/pkg/execution/executor/executor.go#L2146
- Discovery step following a `step.sleep()` completion https://github.com/inngest/inngest/blob/9ebbde880ffd3a8d3997ddbb83de4c1bb923b3cc/pkg/execution/executor/executor.go#L861

PR #2707, opened 2025-07-16
Create ParallelMode enum; update OptKey enum

amh4r:
Update enums to reflect new "parallel mode" approach


pkg/enums/parallelmode.go
```diff
@@ -0,0 +1,25 @@
+//go:generate go run github.com/dmarkham/enumer -trimprefix=ParallelMode -type=ParallelMode -json -text
+
+package enums
+
+type ParallelMode int
+
+const (
+	// ParallelModeNone defaults to ParallelModeWait
+	ParallelModeNone ParallelMode = iota
```
jpwilliams, line 9: Interestingly this also depends on the execution version 🤔

PR #2710, opened 2025-07-17
Add event_cel package

amh4r:
## Description
Add `event_cel` package for event CEL search.

It's tested in a monorepo feature branch. We aren't using it in OSS yet so we'll wait to add tests in OSS


albertchae:
reviewing post merge, looks good, looked like the main diff was removing output based expressions and more error handling with default cases

any reason for putting this in `event_cel `and not `event/cel.go` like `run/cel.go` currently is?

Would be nice to share more code between the event/run versions but we can think more about that later

PR #2718, opened 2025-07-17
Add error reporting interface to logger

darwin67:
## Description

Provide error reporting as part of the logger interface.

This helps encapsulate how errors are being reported without littering the codebase with a bunch of sentry deps.

Log attributes are also inherited into the error tags as much as possible as long as it can be rendered into a string, so it should simplify replacing error logs with error reports.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -839,6 +839,15 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		return nil, fmt.Errorf("no function loader specified running step")
 	}
 
+	l := e.log.With(
+		"account_id", item.Identifier.AccountID,
+		"env_id", item.WorkspaceID,
+		"app_id", item.Identifier.AppID,
+		"fn_id", item.Identifier.WorkflowID,
+		"run_id", id.RunID,
+	)
+	ctx = logger.WithStdlib(ctx, l)
+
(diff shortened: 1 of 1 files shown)


pkg/execution/state/redis_state/backlog_normalization.go
```diff
@@ -340,8 +340,18 @@ func (q *queue) normalizeBacklog(ctx context.Context, backlog *QueueBacklog, sp
 			item := item // capture range variable
 			wg.Go(func() {
 				_, err := q.normalizeItem(logger.WithStdlib(ctx, l), shard, sp, latestConstraints, backlog, *item)
-				if err != nil {
-					l.Error("could not normalize item", "err", err)
+				if err != nil && !errors.Is(err, context.Canceled) {
+					l.ReportError(err, "could not normalize item",
```
KiKoS0, line 344: **non blocker:** I think this pattern will repeat a lot where we skip reporting on certain errors, I wonder if we should integrate it as an option to `ReportError`, even more so if we still want to log the error but not report it.

KiKoS0 (approved):

pkg/execution/state/redis_state/backlog_normalization.go
darwin67: Good point, let's see how it goes and we can make tweaks to it.

PR #2723, opened 2025-07-18
share feature flags between cloud and dev server

jacobheric:
## Description

Fleshes out our shared feature flag provider implementation for dev server such that you can use the same feature flag names from launchdarkly for the cloud in the dev server using env vars. 

If you have a launch darkly feature flag called `step-over-debugger`, you can do this in a shared component:

```
  const { value, isReady } = booleanFlag(
    'step-over-debugger',
    false
  );
```
And start the dev server like so locally:

`INNGEST_FEATURE_FLAGS="step-over-debugger=true,some-other-flag=false" go run ./cmd/main.go dev`

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/devserver/api.go
+++ b/pkg/devserver/api.go
@@ -152,6 +152,23 @@ func (a devapi) Info(w http.ResponseWriter, r *http.Request) {
 		features[flag] = enabled
 	}
 
+	//
+	// inngest feature flags are a map of arbitrary key value
+	// pairs such that we can use the same feature flag names as cloud
+	if ffEnv := os.Getenv("INNGEST_FEATURE_FLAGS"); ffEnv != "" {
+		pairs := strings.Split(ffEnv, ",")
+		for _, pair := range pairs {
+			kv := strings.SplitN(strings.TrimSpace(pair), "=", 2)
+			if len(kv) == 2 {
+				key := strings.TrimSpace(kv[0])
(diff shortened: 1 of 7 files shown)


pkg/devserver/api.go
```diff
@@ -152,6 +152,23 @@ func (a devapi) Info(w http.ResponseWriter, r *http.Request) {
...
+		for _, pair := range pairs {
+			kv := strings.SplitN(strings.TrimSpace(pair), "=", 2)
+			if len(kv) == 2 {
+				key := strings.TrimSpace(kv[0])
+				valStr := strings.TrimSpace(kv[1])
+				if enabled, err := strconv.ParseBool(valStr); err == nil {
+					features[key] = enabled
+				}
+			}
+		}
```
albertchae, line 170: might be nice to print something out just so we can confirm a feature flag was read and set properly, although it looks like they do get reported in http://localhost:8288/dev (which you are using in useBooleanFlag.ts) after so maybe checking there is sufficient

albertchae (approved):

pkg/devserver/api.go
jacobheric: I info log out the feature flag key (not value) when it's set now.

jacobheric:
> It would be awesome to add a brief explanation about this to our readme

done!

PR #2731, opened 2025-07-21
Environments: Update actions UX

doverdbng:
## Description

This adjusts the environments page actions according to updated Figma designs.

Screenshot 2025-07-22 at 11 25 38 AM

**Note:** A (near) future PR will make some finer adjustments to better match the designs in Figma.

**Note x2:** An additional Linear ticket was created to add explainers about shared branch keys to those key management pages; I don't think this changeset makes that peculiarity noticeably less clear in the meantime

- [ ] Docs

## Checklist


ui/apps/dashboard/src/components/Environments/CustomEnvironmentListTable.tsx
```diff
@@ -111,28 +99,12 @@ function TableRow(props: { env: Environment }) {
...
-              env={props.env}
-              onClose={() => setOpenDropdown(false)}
-            />
-          </DropdownMenuContent>
-        </DropdownMenu>
+      <td>
+        <div className="flex items-center gap-2 px-4">
+          <EnvViewButton env={props.env} />
+          <EnvKeysDropdownButton env={props.env} />
+          <EnvArchiveButton env={props.env} onClose={() => {}} />
```
doverdbng: **Note:** Need to analyze the `env`, `isArchived` caching that happens in here, as mounting is no longer hidden behind a dropdown.

doverdbng: **Update:** I haven't seen any actual issues with this. The state caching is definitely a bit unfortunate. However, now that (un)archiving an environment throws it in another list (because of the forced status filter), these sorts of staleness concerns should be less relevant.

ui/apps/dashboard/src/components/Environments/BranchEnvironmentActions.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <SplitButton
+      left={
+        <Button
+          appearance="solid"
+          className="rounded-r-none border-r-0 text-sm"
+          kind="primary"
+          label="Sync new app"
+          href={`/env/${branchParent.slug || 'branch'}/apps`}
```
doverdbng, line 30: I assume this shouldn't need the "branch" fallback based on usage, but it being here historically makes me wonder.

ui/apps/dashboard/src/components/Environments/BranchEnvironmentActions.tsx
```diff
@@ -0,0 +1,69 @@
...
+              Manage signing key
+            </DropdownMenuItem>
+          </DropdownMenuContent>
+        </DropdownMenu>
+      }
+    />
+  );
+}
+
+function slugOrBranchDefault(env: Environment) {
```
doverdbng: I assume this "" -> "branch" fallback shouldn't be necessary based on usage, but it being here historically makes me wonder. Maybe some early branch parents didn't have "branch" as their slug?

ui/apps/dashboard/src/components/Environments/CustomEnvironmentListTable.tsx
doverdbng: One thing that's slightly awkward about this approach is that it's inconsistent. For example, I mocked the refreshed environments list fetch to fail (awkwardly, with a useEffect) and the archive button is gray here because the caching in that component detected the update, but it's still in the "active" list and shows as such.

Screenshot 2025-07-21 at 4 52 26 PM

I don't think this is a new general type of client/server inconsistency and this **_probably_** isn't worth worrying about right now? But, if we do have a track record of being thorough with these sorts of client-side cache manipulations, I'm happy to chase it. I'm interested to hear thoughts here from someone with more codebase knowledge.

ui/apps/dashboard/src/components/Environments/BranchEnvironmentActions.tsx
doverdbng: Based on a direct 1:1 comparison with the existing code, `slugOrBranchDefault` would be used only for "Sync new app" and not the key route(s)...but that doesn't make a lot of obvious sense, so that's probably an oversight.

anafilipadealmeida: I ran a query in our db and I don't think there are any BranchChilds that have no parentID, or with a parentID that is not a BranchParent

doverdbng: In this case I believe the necessary check is whether there are ever Branch Parents that have a default value of `""` as their `slug` instead of `"branch"`. If so, without this check, we'll visit incorrect routes like `/env//manage/signing-key`.

My guess is that this probably isn't a real concern, but it's _possible_ that early Branch Parents didn't set "branch" as their slug. It's also possible that the previous check was really for the case where the branch parent doesn't exist, but this nested component requires it to exist, so that concern is void.

The only way this fallback is possibly _harmful_ is if Branch Parents can have a slug other than `"branch"`, but I haven't seen any evidence that that is the case.

doverdbng: Thanks @amh4r for pointing me toward the `getEnvironmentSlug` converter. It ensures that every branch parent will have "branch" as a vanity slug, so this fallback should not be necessary. I'll remove it.

PR #2732, opened 2025-07-21
Andy/inn 5302 create backlog sizequeue depth api endpoint

AndyInternet:
## Description

Add prometheus formatted metrics endpoint with queue_depth metric.

## Motivation
Allow self hosting users to view queue depth through an endpoint. Prometheus format used to support usage with KEDA and Graphana.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/queue.go
```diff
@@ -2829,6 +2832,70 @@ func (q *queue) partitionSize(ctx context.Context, partitionKey string, until ti
...
 }
 
+// TotalSystemQueueDepth returns the total number of items waiting to be processed
+// across all partitions in the system. This provides a global view of queue depth
+// by aggregating items from all partition queues (ready + backlog items).
+//
+// The returned count represents all items that are queued but not yet being
+// processed across the entire system. Items currently being executed are not
+// included in this count.
+func (q *queue) TotalSystemQueueDepth(ctx context.Context) (int64, error) {
```
darwin67, line 2839: this already exist here
https://github.com/inngest/inngest/blob/2e40704aa5f862300f15f43e367025ed3447f805/pkg/execution/state/redis_state/queue.go#L3425-L3513

darwin67: it could be split out into an iterator to make it easier for testing and other usage like this.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,339 @@
...
+				lines = append(lines, fmt.Sprintf("%s_count %d", metric.Name, count))
+			}
+			if sum, ok := histData["sum"].(float64); ok {
+				lines = append(lines, fmt.Sprintf("%s_sum %.2f", metric.Name, sum))
+			}
+		}
+	}
+
+	return strings.Join(lines, "\n")
+}
```
darwin67: prometheus has a library you can use so you don't have to handle all this yourself.

pkg/metrics/metrics_api_test.go
```diff
@@ -0,0 +1,237 @@
...
+	}{
+		{
+			name:           "authenticated request should return metrics",
+			authHeader:     "Bearer valid-test-token",
+			expectedStatus: http.StatusOK,
+			checkBody: func(t *testing.T, body string) {
+				// Should contain Prometheus format with queue_depth metric
+				require.Contains(t, body, "# HELP queue_depth Total system queue depth")
+				require.Contains(t, body, "# TYPE queue_depth gauge")
+				require.Contains(t, body, "queue_depth")
```
darwin67: convention wise, it should prefix with the service name so this should be
`inngest_queue_depth` instead

pkg/execution/state/redis_state/queue.go
AndyInternet: I took your suggestion, removed my implementation and updated the Instrument function to return the total (int64, error). This function was only being called in one place so i updated the call to discard the return value. Please let me know if this is what you were thinking.

pkg/metrics/metrics.go
AndyInternet: Updated to use prometheus libs

pkg/execution/state/redis_state/queue.go
darwin67: Sorry, I meant more that the partition iteration logic is already implemented so you didn't have to do the `ZRANGE` stuff you were previously doing.

The accounting of the queue depth will still need to happen, because the `total` value is just the total number of partitions that were iterated over, not the actual depth of it.

AndyInternet: I refactored the Instrument function and extracted the iterator logic into a new function called QueueIterator that returns two int64 values, the partition count and the queue item count. It also accepts two optional callback functions:
- OnPartitionProcessed, which runs inside the loop 
- OnIterationComplete, which runs after the iterator is complete
I reimplemented the Instrument function with these callbacks. It should function just like it did before, but please take a look and validate.

darwin67 (approved):

PR #2745, opened 2025-07-23
Incoming tracing - `has_output` attribute

jpwilliams:
## Description

Add an attribute to state that we have output in this span.

Output is set as an attribute, and then extracted to be placed elsewhere (mostly, at least in another column in the DB).

With some query patterns, we can pick the span that has output defined and pull out that span ID in the roll-up, but it's not available everywhere. This PR gives us an easy way of telling even if we can't do it with the DB we're using.

Also includes a `panic()` if we have an attribute serializer that has been defined but no value set. This is an operator error and would be caught in a huge amount of testing paths (every function execution for a start); it's there to make this very obvious in dev and CI so it can't get through to prod.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/tracing/meta/serializers.go
+++ b/pkg/tracing/meta/serializers.go
@@ -3,6 +3,7 @@
 package meta
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -401,7 +402,7 @@ func HttpHeaderAttr(key string) attr[*http.Header] {
 // ExtractTypedValues uses reflection to extract typed pointer values from the
 // Attrs struct given a map of attribute key-value pairs. It returns a properly
 // typed ExtractedValues struct with IDE support and compile-time type checking.
-func ExtractTypedValues(attrs map[string]any) *ExtractedValues {
+func ExtractTypedValues(ctx context.Context, attrs map[string]any) (*ExtractedValues, error) {
 	result := &ExtractedValues{}
 	resultValue := reflect.ValueOf(result).Elem()
 	attrsValue := reflect.ValueOf(Attrs)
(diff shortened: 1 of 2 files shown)


pkg/tracing/meta/serializers.go
```diff
@@ -422,6 +422,14 @@ func ExtractTypedValues(attrs map[string]any) *ExtractedValues {
 		attrValue := field.Interface()
 		if serializer, ok := attrValue.(Serializer); ok {
 			key := serializer.Key()
+
+			if key == "" {
+				// The serializer exists in the struct definition but hasn't
+				// actually been set. This is an operator error and needs to
+				// be addressed!
+				panic(fmt.Sprintf("Span attribute serializer for '%s' is empty - a span attribute has been defined but no (de)serializer has been set; this attribute will never be persisted or deserialized.", fieldName))
```
amh4r: Why panic instead of returning an error?

pkg/tracing/meta/attributes.go
```diff
@@ -50,6 +50,9 @@ var Attrs = struct {
 	// separately.
 	StepOutput    attr[*string]
 	StepOutputRef attr[*string]
+	// StepHasOutput is used to mark that a specific span has an output in the
+	// attributes, in place of the output itself.
+	StepHasOutput attr[*bool]
```
darwin67, line 55: what does it mean to not have an output?

jpwilliams: This one could zero-value; I just haven't made that pass yet! Gotta adjust some code generation for that 😄

pkg/tracing/meta/serializers.go
jpwilliams: > Also includes a panic() if we have an attribute serializer that has been defined but no value set. This is an operator error and would be caught in a huge amount of testing paths (every function execution for a start); it's there to make this very obvious in dev and CI so it can't get through to prod.

Just really don't want it to go unnoticed in dev/CI, and this would be caught in soo many paths.

amh4r: My worry about panics is how they can crash the whole server if not recovered. Even though we should discover this panic in dev/CI, there's always new opportunities to miss that when we make code changes. Are you open to using an error instead of a panic?

tonyhb (approved):

pkg/tracing/meta/serializers.go
tonyhb: Maybe see @darwin67's new error logging in logs?

tonyhb (approved):

pkg/tracing/meta/serializers.go
jpwilliams: Made it an error. Will alter the downstream PR when this goes in

PR #2769, opened 2025-07-29
Sync checkpointing APIs

tonyhb:
This introduces base sync checkpoint APIs necessary for zero.


pkg/cqrs/apps.go
```diff
@@ -69,7 +75,7 @@ type UpsertAppParams struct {
 	Error       sql.NullString
 	Checksum    string
 	Url         string
-	Method      string
+	Method      string // enums.AppMetho
```
jpwilliams: TODO?

pkg/devserver/devserver.go
```diff
@@ -501,6 +494,12 @@ func start(ctx context.Context, opts StartOpts) error {
 			Broadcaster:        broadcaster,
 			RealtimeJWTSecret:  consts.DevServerRealtimeJWTSecret,
 			TraceReader:        ds.Data,
+
+			AppCreator:      dbcqrs,
+			FunctionCreator: dbcqrs,
+			EventPublisher:  runner,
+			TracerProvider:  tracing.NewSqlcTracerProvider(base_cqrs.NewQueries(db, dbDriver)),
```
jpwilliams: Can we reuse the `TracerProvider` created further up the file?

pkg/execution/execution.go
```diff
@@ -170,6 +170,10 @@ type ScheduleRequest struct {
 	WorkspaceID uuid.UUID
 	// AppID is the app that this request belongs to.
 	AppID uuid.UUID
+	// RunID allows specifying a run ID for the scheduled run.  This is entirely
+	// optional, and allows clients to choose a run ID when scheduling.  We need this
+	// for run IDs with API-based checkpointing.
+	RunID *ulid.ULID
```
jpwilliams:
```suggestion
	// RunID allows specifying a run ID for the scheduled run.  This is entirely
	// optional, and allows clients to choose a run ID when scheduling.  We need this
	// for run IDs with API-based checkpointing.
	//
	// Note that this should never be provided by the user, as that could welcome
	// conflicts.
	RunID *ulid.ULID
```

pkg/execution/executor/executor.go
```diff
@@ -543,8 +550,9 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 	runSpanRef, err := e.tracerProvider.CreateSpan(
 		meta.SpanNameRun,
 		&tracing.CreateSpanOptions{
-			Debug:    &tracing.SpanDebugData{Location: "executor.Schedule"},
-			Metadata: &metadata,
+			Debug:     &tracing.SpanDebugData{Location: "executor.Schedule"},
+			Metadata:  &metadata,
+			StartTime: runID.Timestamp(),
```
jpwilliams: Hmm this field currently directly represents the _start_ of a span for users, as opposed to the _queued_ time, which we need here.

May need to add that as a top-level option instead.

pkg/execution/executor/executor.go
```diff
@@ -1461,6 +1480,8 @@ func (e *executor) run(ctx context.Context, i *runInstance) (*state.DriverRespon
 
 	// Execute the actual step.
 	response, err := e.executeDriverForStep(ctx, i)
+	spew.Dump(err, response)
+
```
jpwilliams:
```suggestion
```

pkg/execution/executor/executor.go
```diff
@@ -13,6 +13,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
```
jpwilliams:
```suggestion
```

pkg/api/apiv1/checkpoint.go
```diff
@@ -0,0 +1,435 @@
...
+		// 		// TODO: err
+		// 		return
+		// 	}
+		// 	if err := a.finalize(ctx, md, result); err != nil {
+		// 		// TODO: err
+		// 		return
+		// 	}
+		default:
+			// This is now async.  For now, do NOT allow this.
+			panic("TODO")
```
jpwilliams, line 250: Okay to leave this in to go out?

PR #2770, opened 2025-07-29
Add application/x-www-form-urlencoded support to Event API

amh4r:
- Add `application/x-www-form-urlencoded` support to Event API. This is only for webhooks in Cloud and won't actually work with the Dev Server Event API.
- Make `multipart/form-data` return a map of slices.


pkg/eventstream/eventstream.go
```diff
@@ -52,10 +53,14 @@ func ParseStream(
...
+	// happens
+	mediaType, params, _ := mime.ParseMediaType(contentType)
+	switch mediaType {
+	case "multipart/form-data":
 		return parseMultipartStream(ctx, r, stream, maxSize, params["boundary"])
+	case "application/x-www-form-urlencoded":
+		return parseFormUrlencodedStream(ctx, r, stream, maxSize)
 	}
 
 	// Default to JSON parsing
```
albertchae, line 66: minor but now that we have 3 cases it might be better structurally to move the rest of this JSON parsing code into `parseJSONStream` and call that as the default of the switch

pkg/eventstream/eventstream.go
```diff
@@ -158,33 +163,65 @@ func parseMultipartStream(
...
+	maxSize int,
+) error {
+	// Read all data from the reader
+	data, err := io.ReadAll(r)
+	if err != nil {
+		return err
+	}
+
+	// Parse the form data
+	values, err := url.ParseQuery(string(data))
```
albertchae, line 206: cool, looks like this groups repeated fields into an array for us for free https://pkg.go.dev/net/url#ParseQuery

and I see your tests cover that too

pkg/eventstream/eventstream_test.go
```diff
@@ -314,13 +331,164 @@ func TestParseStream_Multipart(t *testing.T) {
...
+
+		stream := make(chan StreamItem)
+		eg := errgroup.Group{}
+		eg.Go(func() error {
+			return ParseStream(ctx, body, stream, 512*1024, contentType)
+		})
+
+		n := 0
+		for item := range stream {
+			// The multipart parser creates a JSON object with form fields as keys
```
albertchae: copy pasta?

pkg/eventstream/eventstream.go
albertchae: I don't think we should do anything to handle these but FYI some libraries have a pattern of using `field[]` or `field[0]`/`field[1]` for sending arrays in form-urlencoded

https://www.baeldung.com/java-postman-send-array
https://brandur.org/fragments/application-x-wwww-form-urlencoded

This isn't part of the spec for `application/x-www-form-urlencoded` though, so I'd leave it as is until we see some webhook in the wild that adheres to this.

albertchae (approved):

pkg/eventstream/eventstream.go
amh4r: I think I'll implement the strategy pattern when we add a 4th content type

PR #2779, opened 2025-07-31
Synchronously commit spans in the Executor

jpwilliams:
## Description

Enables synchronous committing of spans for the incoming tracing work.

It still runs through the regular otel pipeline, so we use channels here to wait for each span to be sent individually.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/tracing/meta/notifications.go
+++ b/pkg/tracing/meta/notifications.go
@@ -1,0 +1,46 @@
+package meta
+
+import (
+	"sync"
+)
+
+var (
+	spanNotifications = make(map[string]chan error)
+	notificationMu    sync.RWMutex
+)
+
+// RegisterSpanNotification creates and registers a notification channel for the
+// given span ID. The caller should wait on the returned channel to be notified
+// when a span has been exported.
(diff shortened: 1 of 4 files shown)


pkg/tracing/meta/notifications.go
```diff
@@ -0,0 +1,46 @@
+package meta
+
+import (
+	"sync"
+)
+
+var (
+	spanNotifications = make(map[string]chan error)
```
tonyhb, line 8: i have some questions on how well this will scale as one executor many thousands of executions and spans at once, especially with eg. parallel steps that may each present 1K steps for spans also.  this may quickly balloon.  do we need this?  is it okay to say that as soon as we publish, it gets written to a buffer and sent to eg. kafka?

jpwilliams: Mm this is to ensure that it's part of the critical flow -  we need to know that it's been written to Kafka (and not just the buffer) before we're happy to continue.

Without this, there's always the risk that we write to the buffer and then fail to publish, resulting in lost data and a trace that appears as if something was missed.

Intentionally left this as a separate PR as I figured it would create discussion! 😄

PR #2782, opened 2025-07-31
Set database max_connection limits for postgres

AndyInternet:
## Description

Set max idle and max open connections for postgres driver
- set sane defaults
  - 100 for max
  - 10 for idle
- add advanced flag to set max open connections (--postgres_max_connections)
- add unpublished flag to set idle (--postgres_idle_connections)

## Motivation
Self-hosting currently can open as many connections as it requires and it throws errors in postgres when it runs out of available connections.

- [ ] Docs

## Checklist


pkg/devserver/devserver.go
```diff
@@ -172,8 +177,18 @@ func start(ctx context.Context, opts StartOpts) error {
...
+		MaxOpenConns:    opts.PostgresMaxOpenConns,
+		ConnMaxIdle:     opts.PostgresConnMaxIdleTime,
+		ConnMaxLifetime: opts.PostgresConnMaxLifetime,
+	})
+	hd := base_cqrs.NewHistoryDriver(db, dbDriver, sqlc_postgres.NewNormalizedOpts{
+		MaxIdleConns:    opts.PostgresMaxIdleConns,
+		MaxOpenConns:    opts.PostgresMaxOpenConns,
+		ConnMaxIdle:     opts.PostgresConnMaxIdleTime,
+		ConnMaxLifetime: opts.PostgresConnMaxLifetime,
+	})
```
albertchae, line 191: nit: looks like you could extract the normalized opts struct to a variable and reuse it in both constructors

cmd/start/start_test.go
```diff
@@ -0,0 +1,111 @@
...
+			postgresMaxIdleConns: 0,
+			postgresMaxOpenConns: 10,
+			expectError:          false,
+		},
+		{
+			name:                 "default values should be valid",
+			postgresMaxIdleConns: 10,
+			postgresMaxOpenConns: 100,
+			expectError:          false,
+		},
```
albertchae, line 94: nit: could replace the first test setup "valid connection pool settings" since the values are the same and it's almost the same intent?

albertchae (approved):

PR #2803, opened 2025-08-06
Ana/change fn list

anafilipadealmeida:
- [ ] Docs

## Checklist


doverdbng:
I'm noticing a significant number of calls to `GetFunctionUsage` when I load the Functions list, I believe one for each visible list entry. The fetching arguments also seem to have changed from `page` and `pageSize` based to `startTime` and `endTime` based. Are these changes expected?

ui/apps/dashboard/src/components/Functions/useFunctions.ts
```diff
@@ -0,0 +1,123 @@
...
+  const client = useClient();
+  return useCallback(
+    async ({ cursor, archived, nameSearch }: QueryVariables) => {
+      const result = await client
+        .query(
+          GetFunctionsDocument,
+          {
+            environmentID: envID,
+            page: cursor ?? 1,
+            pageSize: 30,
```
doverdbng, line 25: I _think_ this was previously 50 by comparison with prod? Is this change intended?

ui/packages/components/src/Functions/FunctionsTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+    },
+    initialPageParam: 1,
+  });
+
+  const mergedData = useMemo(() => {
+    return (
+      functionsData?.pages.flatMap((page) =>
+        page.functions.map((e) => ({
+          ...e,
+          usage: undefined,
```
doverdbng, line 116: Why is this set directly to `undefined`?

doverdbng:
Environment switches don't seem to be taken into account when I'm on the Functions page. Whatever set of functions I see for the original environment stick around through environment changes. It seems like there might be something going on with the cache key.

anafilipadealmeida:
> GetFunctionUsage

@doverdbng , great questions!
It's intended, those are slightly different queries, this already had the fetching arguments I'm using:
https://github.com/inngest/inngest/blob/main/ui/apps/dashboard/src/queries/functions.ts#L242
The difference is that the previous approach was querying the usage for all functions in a page, the current approach queries the usage for a single function. It's the same lazy loading approach we take in event types, events, runs.

ui/apps/dashboard/src/components/Functions/useFunctions.ts
anafilipadealmeida: Yep, to match the page size we use in places like events, but I can increase both to 50 in case someone is using really large screens

ui/packages/components/src/Functions/FunctionsTable.tsx
anafilipadealmeida: For a few reasons, mostly related to the tanstack implementation of columns. It avoids column rendering errors and makes the data shape consistent, since we fetch the usage later in each row. For instance if someone changes the shared functions query and for some reason adds the usage, this prevents it from breaking.

ui/packages/components/src/Functions/columns.tsx
```diff
@@ -156,11 +121,12 @@ function createColumns(environmentSlug: string) {
     }),
     columnHelper.accessor('failureRate', {
       cell: (info) => {
-        const value = info.getValue();
-        if (value === undefined) {
+        const functionID = info.row.original.id;
+        const { data, isLoading } = useFunctionVolume(functionID, getFunctionVolume);
```
doverdbng: It looks like this pattern is also used for `EventTypes`, but I'm a little surprised React doesn't complain because of the rule about only calling hooks at the top-level of components.

ui/packages/components/src/Functions/columns.tsx
```diff
@@ -170,33 +136,38 @@ function createColumns(environmentSlug: string) {
 
         return (
           <TextCell>
-            <span className="text-tertiary-intense">{value}%</span>
+            <span className="text-tertiary-intense">{data?.failureRate}%</span>
```
doverdbng: Will this render `undefined%` in an error case? Line 21 is a strict test against `0`.

ui/packages/components/src/Functions/columns.tsx
```diff
@@ -131,7 +97,7 @@ function createColumns(environmentSlug: string) {
       },
       header: 'Triggers',
     }),
-    columnHelper.accessor((row) => row.app.name, {
+    columnHelper.accessor((row) => row.app?.name, {
```
doverdbng: Out of curiosity, what happens here if the `?` becomes relevant?

djfarrelly (approved):

PR #2806, opened 2025-08-07
SYS-253: Source paused status from new getter, migration from lock key

BrunoScheufler:
## Description

- This PR removes existing usage of fnMeta to source state from the database for pausing and temporary lock keys for queue migrations
- Queue migrations: When peeking a partition, we will load lock keys from Redis in a single MGET. The value is expected to be a ULID, the timestamp of which determines how long the lock is valid for. When locked, the partition is requeued for the lock duration.
- Pausing: This PR introduces a new partition paused getter, which is implemented to return data from an in-memory map in dev and using the database in cloud. This may use caching and signal potential staleness using the `Stale` value. If fresh data is returned _and_ the function is paused, partition peeks will push back the paused partition by 24h.
  - The staleness flag prevents us from accidentally pushing back the partition if the function was unpaused recently but the in-memory cache hasn't expired
- SetFunctionPaused was completely removed in favor of the getter + requeue during partition peeks _and_ a new method `UnpauseFunction` which requeues the given function to its earliest item.
- Key Queues also retrieve the paused status and requeue the shadow partitions for 24h if the function is currently paused (and not stale), and UnpauseFunction requeues the shadow partition to its earliest item (if key queues enabled).

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/queue.go
+++ b/pkg/execution/state/redis_state/queue.go
@@ -1492,41 +1492,6 @@ func (q *queue) dropPartitionPointerIfEmpty(ctx context.Context, shard QueueShar
 	}
 }
 
-func (q *queue) UnpauseFunction(ctx context.Context, shardName string, acctID, fnID uuid.UUID) error {
-	shard, ok := q.queueShardClients[shardName]
-	if !ok {
-		return fmt.Errorf("invalid shard %q", shardName)
-	}
-
-	part := &QueuePartition{
-		ID:         fnID.String(),
-		FunctionID: &fnID,
-		AccountID:  acctID,
-	}
-
-	err := q.PartitionRequeue(ctx, shard, part, q.clock.Now(), false)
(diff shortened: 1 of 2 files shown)


pkg/execution/state/redis_state/queue.go
```diff
@@ -1573,14 +1492,45 @@ func (q *queue) dropPartitionPointerIfEmpty(ctx context.Context, shard QueueShar
 	}
 }
 
-func (q *queue) SetFunctionMigrate(ctx context.Context, sourceShard string, fnID uuid.UUID, migrate bool) error {
-	ctx = redis_telemetry.WithScope(redis_telemetry.WithOpName(ctx, "SetFunctionMigrate"), redis_telemetry.ScopeQueue)
+func (q *queue) UnpauseFunction(ctx context.Context, shardName string, acctID, fnID uuid.UUID) error {
```
darwin67: Move this function to a different file?
We should be removing things from this file when there's an opportunity to do so.

PR #2834, opened 2025-08-18
Interface & pb tweaks for state store proxy

KiKoS0:
## Description
- Generate a go-grpc version of the state proxy protobuf service.
- ~~Return a V2 state from the state store `Create` method as it will be much easier to deal with on the state proxy gRPC implementation in Cloud.~~ will be in a follow up, splitting the change.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/v2_adapter.go
```diff
@@ -69,15 +69,79 @@ func (v v2) Create(ctx context.Context, s state.CreateState) (statev1.State, err
...
+	}
+
+	metrics.IncrStateWrittenCounter(ctx, len(batchData), metrics.CounterOpt{
+		PkgName: "redis_state",
+		Tags: map[string]any{
+			"account_id": s.Metadata.ID.Tenant.AccountID,
+		},
+	})
+
+	// XXX: We do the exact same size calculations done in `mgr.New` to return a v2 state without changing the v1 interface.
```
KiKoS0: I’m not a fan of duplicating this, so I’m open to ideas. I just don’t want to rely on `LoadState` and add three extra Redis calls.

pkg/execution/state/redis_state/redis_state.go
```diff
@@ -346,11 +346,16 @@ func (m shardedMgr) New(ctx context.Context, input state.Input) (state.State, er
 			),
```
KiKoS0: `make([]string, 0)` The stack always ends up empty when we create a new state. Is that intentional? It feels like a bug, but since it’s unused we don’t see it. If we try to cache it in the proxy, though, it could cause trouble.

darwin67: yeah, because it's an entire new state, so there are no values in the stack.

proto/state/v2/buf.gen.yaml
```diff
@@ -0,0 +1,5 @@
+version: v1
+plugins:
+  - plugin: go-grpc
+    out: proto/gen
+    opt: paths=source_relative
```
darwin67, line 5: does it not generate the source properly if you don't have this file?

pkg/execution/state/redis_state/redis_state.go
KiKoS0: hmm i thought so too but state can be created with steps so the stack wouldn't really be empty
https://github.com/inngest/inngest/blob/a7620f3959d7295f34a044f0c6ffa1339d4d64dd/pkg/execution/executor/reconstruct.go#L125-L128

proto/state/v2/buf.gen.yaml
KiKoS0: nope I think this is the last service not using go-grpc though so we can have a followup PR that makes `go-grpc` in the root `buf.gen.yaml` config.

pkg/execution/state/redis_state/redis_state.go
darwin67: that one is new to me so wasn't aware of this.

PR #2847, opened 2025-08-19
Redo cron using queues

darwin67:
## Description

The fundamental issue with the existing cron implementation is it holds all schedule in memory, and there are no external state.
Meaning when the cron crashes, and if that crash happens after the top of minute of hour, it can result in missed schedules.

This change moves cron implementation using the queue so it automatically gets the reliability from the queue, and issues like timing of the cron schedule or crashes no longer result in missed scheduling.

- [ ] Docs

## Checklist


pkg/execution/state/redis_state/iterator.go
```diff
@@ -4,7 +4,9 @@ import "time"
 
 type QueueIterOpt func(o *queueIterOpt)
 
+// queueIterOpt provides options to be used for queue iteration operations
 type queueIterOpt struct {
+	// TODO figure how to embed queueOpOpt into this struct and share the setting overrides
```
BrunoScheufler: Does this still need to be done?

pkg/inngest/function.go
```diff
@@ -302,6 +303,20 @@ func (f Function) IsScheduled() bool {
 	return false
 }
 
+// ScheduleExpression returns the cron expression string for the function
+//
+// NOTE
+// the code technically allow multiple triggers, should this also return a
```
BrunoScheufler: Should we allow multiple schedules to be returned?

pkg/execution/cron/redis.go
```diff
@@ -0,0 +1,463 @@
...
+	rangeNs := int64(max - min + time.Nanosecond)
+	randomBytes := make([]byte, 8)
+
+	_, err := rand.Read(randomBytes)
+	if err != nil {
+		return 0
+	}
+
+	// Convert bytes to uint64 and get value in range
+	randomValue := uint64(randomBytes[0])<<56 | uint64(randomBytes[1])<<48 | uint64(randomBytes[2])<<40 | uint64(randomBytes[3])<<32 |
```
BrunoScheufler, line 210: This is a little intense, is there no current code that does something similar that we can reuse?

pkg/execution/cron/redis.go
```diff
@@ -0,0 +1,463 @@
...
+		return fmt.Errorf("error marshalling cron item: %w", err)
+	}
+
+	rc := c.c.Client()
+	kg := c.c.KeyGenerator()
+
+	cmd := rc.B().
+		Hset().
+		Key(kg.Schedule()).
+		FieldValue().
```
BrunoScheufler: Is this an extra field that should be removed?
```suggestion

```

pkg/execution/cron/redis.go
```diff
@@ -0,0 +1,463 @@
...
+func WithScheduleForwardDuration(dur time.Duration) RedisCronManagerOpt {
+	return func(c *redisCronManagerOpt) {
+		if dur > 0 {
+			c.scheduleForwardDur = dur
+		}
+	}
+}
+
+func NewRedisCronManager(
+	c *redis_state.CronClient,
```
BrunoScheufler, line 42: Just to confirm, we're storing the following new data

- map of [functionID:schedule] key-value pairs
- new system queue for cron syncs
- new system queue for cron queue

And these all reside on the same queue shard which will be the system queue?

pkg/execution/cron/redis.go
```diff
@@ -0,0 +1,463 @@
...
+	//
+	// we do this so we can make the actual run start as close as possible to the actual cron schedule.
+	jitterMin time.Duration
+	jitterMax time.Duration
+
+	// we need to move the time upwards on finding the next time due to how we use jitter
+	// to push the time back a little to coordinate the run start timing.
+	//
+	// considering cron's minimum granularity is a minute, the seconds range should work fine. defaults to 10s
+	scheduleForwardDur time.Duration
```
BrunoScheufler: Is this the lead time for scheduling runs a little before they need to start processing to have ample time to handle cron processing and scheduling the run? If so, should we rename it to lead time to ensure it's the same naming convention we applied previously?

pkg/execution/cron/redis.go
```diff
@@ -0,0 +1,463 @@
...
+
+	var res CronItem
+	if err := json.Unmarshal(byt, &res); err != nil {
+		return nil, fmt.Errorf("error unmarshalling cron item for function: %w", err)
+	}
+
+	return &res, nil
+}
+
+func (c *redisCronManager) setFunctionScheduleMap(ctx context.Context, ci CronItem) error {
```
BrunoScheufler: Just to confirm my mental model here

- We store the next schedule in the system queue with the score of the next expression minus some lead time to start in time
- There's a separate system queue for any sync operations that need to happen
- The mapping stores the next expected schedule and may change during a sync. This is used to ensure that any old schedules still in the queue don't run if they shouldn't?

Does this mapping essentially act like a layer of indirection storing state that's inferred from the current function configuration in the database?

And just for understanding (not that I want to change this), but could an alternative implementation simply use the cached current workflow to read from JIT to see if the schedule/function is still valid?

BrunoScheufler (commented):
Looks great, added some comments mostly around understanding and sharing knowledge

pkg/execution/cron/redis.go
BrunoScheufler: May want to add the new mapping key to redis-stats for proper accounting

pkg/execution/cron/redis.go
lkasinathan: @darwin67  @BrunoScheufler and I discussed this in person today.
This function mapping exists for two reasons
- In the dev server we don't maintain function versions, so we need this mapping to look for existing items in the queue to delete
- This would also be helpful in the debug API to pick up the next scheduled item for a particular partition. 


For the first, we could maybe update the function version field in the sql table `functions` and rely on checking function version on current config vs function version on the enqueued item as part of a JIT check to drop cron items enqueued for an older function config.

For the second, if we had some deterministic ID for the cron queue item, in the debug server we could reconstruct the expected queue item ID for the next schedule and use that to look into the details of the next item in the queue for a particular function.

With these two changes, we could get rid of this additional function mapping.

pkg/execution/state/redis_state/iterator.go
lkasinathan: doesn't seem to be related, at least to this PR. Reverted.

pkg/inngest/function.go
lkasinathan: YES!! I've made this change.

pkg/execution/cron/redis.go
lkasinathan: This bookkeeping was removed altogether as discussed in person, resolving.

pkg/execution/cron/redis.go
lkasinathan: This bookkeeping was removed altogether as discussed in person, resolving.

pkg/execution/cron/redis.go
lkasinathan: I don't think we need this `scheduleForwardDur` anymore, I've removed it.

Before:
- cron job's ID was set to {next_schedule-jitter}. Jitter is the lead time you're referring to here.
- For ex, next_schedule=`10:00:00`, jitter=`2s`, ID is set to `09:59:58` and the cron is enqueued at `09:59:58`
- to schedule the next cron job, we added this `scheduleForwardDur` to ensure that we don't schedule another job for `10:00:00` and instead move to `10:01:00`

Now:
- we maintain the same logic for jitter aka lead time and use it for `enqueueAt`, but we don't encode the jitter or actual enqueue time in the ID.
- ID just contains the actual schedule time and we can do away with `scheduleForwardDur`

pkg/execution/cron/redis.go
lkasinathan: >  For the first, we could maybe update the function version field in the sql table `functions`

https://github.com/inngest/inngest/pull/3114

> For the second, if we had some deterministic ID for the cron queue item, in the debug server we could reconstruct the expected queue item ID

https://github.com/inngest/inngest/pull/3132

pkg/execution/cron/redis.go
lkasinathan: punting on this for now, but we could potentially unify it with the Cloud implementation which is a bit simpler.

darwin67: no strong preferences. it's gonna be hard to beat bit manipulation in terms of spead 😏

lkasinathan (approved):

PR #2854, opened 2025-08-20
[Insights]: Integrate query fetching with BE

doverdbng:
## Description

TBD

- [ ] Docs

## Checklist


ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
```diff
@@ -0,0 +1,111 @@
...
+
+function parseValueByType(
+  value: string,
+  columnType: InsightsColumnType
+): string | number | Date | null {
+  switch (columnType) {
+    case InsightsColumnType.Number:
+      return parseFloat(value);
+    case InsightsColumnType.Date:
+      return new Date(value);
```
albertchae, line 76: the backend is returning time in RFC3339Nano format (although we won't be more precise than milliseconds). Should we more explicitly parse for that format here?

ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
```diff
@@ -0,0 +1,111 @@
...
+    case InsightsColumnType.String:
+    case InsightsColumnType.Unknown:
```
albertchae, line 78: we will probably need to track how many times we run into `Unknown` and add more types. I'll probably do that somewhere in the backend

ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
```diff
@@ -0,0 +1,111 @@
...
+  return columns.reduce((acc, column, index) => {
+    const value = values[index];
+    if (value === undefined) {
+      acc[column.name] = null;
+      return acc;
+    }
+
+    acc[column.name] = parseValueByType(value, column.columnType);
+    return acc;
+  }, {} as Record<string, string | number | Date | null>);
```
albertchae, line 97: this is essentially doing https://lodash.com/docs/4.17.15#keyBy right? but looks like we don't pull in anything from lodash besides debounce, so using standard lib makes sense

AndyInternet (approved):

ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
doverdbng: Yeah, it doesn't seem like `lodash` or similar packages are widely in use.

ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
doverdbng: I'll jot this down for later investigation, but I'm not immediately aware of a need to make a change.

PR #2867, opened 2025-08-22
[Insights]: Add keyword-based autocomplete

doverdbng:
## Description
This adds basic keyword-based (`SELECT`, `FROM`, ...) autocomplete for the query editor.

~**TODO:** Either fix issue with suggestion window clipping or create follow-up ticket to do so.~

## Motivation
This will make the process of writing SQL queries more efficient.

- [ ] Docs

## Checklist


ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+  'TRUE',
+  'WHERE',
+] as const;
+
+const FUNCTIONS = [
+  { name: 'AVG', signature: 'AVG(${1:column})' },
+  { name: 'COUNT', signature: 'COUNT(${1:column})' },
+  { name: 'MAX', signature: 'MAX(${1:column})' },
+  { name: 'MIN', signature: 'MIN(${1:column})' },
+  { name: 'SUM', signature: 'SUM(${1:column})' },
```
albertchae, line 31: what does `${1:column}` mean? does it mean the editor assumes one argument?

in the future we can try to generate this list from here https://github.com/inngest/monorepo/blob/develop/pkg/insights/allowed_clickhouse_functions.go

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+  'LIMIT',
+  'NOT',
+  'NULL',
+  'OFFSET',
+  'OR',
+  'ORDER BY',
+  'SELECT',
+  'TRUE',
+  'WHERE',
+] as const;
```
albertchae, line 24: looks good. compared it to a list based on https://github.com/inngest/monorepo/blob/e8c8c83c2073cdf690c57eb0422623320cb82462/pkg/insights/parser.go#L172

from claude
```javascript
function extractSQLKeywords(regexPattern) {
    // Remove the regex flags and boundaries
    const cleaned = regexPattern
        .replace(/^\(\?\w*\)/, '') // Remove (?i) flag
        .replace(/\\b/g, '')       // Remove word boundaries
        .replace(/^\(|\)$/g, '');  // Remove outer parentheses
    
    // Split by pipe character and sort alphabetically
    return cleaned.split('|').sort();
}

// Usage
const pattern = `(?i)\\b(SELECT|FROM|DISTINCT|WHERE|GROUP|BY|ORDER|LIMIT|OFFSET|TRUE|FALSE|NULL|IS|NOT|BETWEEN|AND|OR|LIKE|AS|IN|DESC|ASC)\\b`;
const keywords = extractSQLKeywords(pattern);
```

```
[
  'AND',
  'AS',
  'ASC',
  'BETWEEN',
  'BY',
  'DESC',
  'DISTINCT',
  'FALSE',
  'FROM',
  'GROUP',
  'IN',
  'IS',
  'LIKE',
  'LIMIT',
  'NOT',
  'NULL',
  'OFFSET',
  'OR',
  'ORDER',
  'SELECT',
  'TRUE',
  'WHERE'
]
```

only discrepancies I saw are you include `BY` with `ORDER BY ` and `GROUP BY` which is preferable anyway, and `IN` is currently missing but it's not supported outside of the lexer yet anyway. https://linear.app/inngest/issue/CON-291/syntax-support-in

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/constants.ts
albertchae: oh I see, it puts `column` as a placeholder to the function?

ui/packages/components/src/SQLEditor/hooks/useSQLCompletions.ts
```diff
@@ -0,0 +1,87 @@
...
+        columns.forEach((column) => {
+          if (labelMatchesPrefix(column, word.word)) {
+            suggestions.push({
+              kind: monaco.languages.CompletionItemKind.Field,
+              insertText: column,
+              label: column,
+              range,
+            });
+          }
+        });
```
albertchae, line 38: nit: this loop + prefix check + adding it to suggestions is repeated 4 times, could we factor that out into a common function?

also for the future – the approach of iterating over the array of columns/functions/keywords/tables and checking `startsWith` is fine for now since we only have a few entries but I wonder if it'll be accidentally quadratic and noticeable if we end up adding all the ~500 CH functions we're allowing to autocomplete. It might be worth building up a lookup table with all the prefixes or using a https://en.wikipedia.org/wiki/Trie

albertchae (approved):

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/constants.ts
doverdbng: That's right, and it highlights it so you can immediately modify it to whatever you want without having to move your cursor.

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/constants.ts
doverdbng: Cool. I let Claude work this out as well from the grammar.

ui/packages/components/src/SQLEditor/hooks/useSQLCompletions.ts
doverdbng: On the nit, the shape of the suggestion is a little different depending on the type. I'm sure we could factor it, and I'd like to, but I'm not sure it would be much cleaner?

doverdbng: The future point makes sense. A trie could be useful in the future, but not investing in that until/unless it becomes necessary seems pragmatic.

albertchae: > On the nit, the shape of the suggestion is a little different depending on the type. I'm sure we could factor it, and I'd like to, but I'm not sure it would be much cleaner?

yeah I noticed that too, so it's your call. I treat nits as generally fine to ignore so it's ok to leave as is

PR #2887, opened 2025-08-26
REST API v2 - Fetch Accounts

AndyInternet:
## Description

Implement `GET /v2/account`

Default Response:
```json
{
  "data": {
    "id": "00000000-0000-4000-a000-000000000000",
    "email": "dev@inngest.local",
    "name": "Dev Server",
    "createdAt": "2021-05-13T16:30:04Z",
    "updatedAt": "2021-05-13T16:30:04Z"
  },
  "metadata": {
    "fetchedAt": "2025-08-27T16:17:39.850969Z",
    "cachedUntil": null
  }
}
```

image

## Motivation
REST API v2

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/v2/service.go
+++ b/pkg/api/v2/service.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"github.com/inngest/inngest/pkg/consts"
 	apiv2 "github.com/inngest/inngest/proto/gen/api/v2"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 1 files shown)


pkg/api/v2/service.go
```diff
@@ -247,7 +258,27 @@ func (s *Service) FetchPartnerAccounts(ctx context.Context, req *apiv2.FetchAcco
 }
 
 func (s *Service) FetchAccount(ctx context.Context, req *apiv2.FetchAccountRequest) (*apiv2.FetchAccountResponse, error) {
-	return nil, NewError(http.StatusNotImplemented, ErrorNotImplemented, "Current account not implemented in OSS")
+	now := time.Now()
+
+	// First commit date: 2021-05-13 09:30:04 -0700
+	firstCommitTime := time.Date(2021, 5, 13, 16, 30, 4, 0, time.UTC)
```
darwin67, line 264: this might be easier to read.

```go
//                         layout                actual time
t, err := time.Parse("2006-01-02 15:04:05", "2021-05-13 09:30:04")
if err != nil {
  return nil, err // NewError something something
}
```

this should default to UTC

darwin67 (approved):

pkg/api/v2/service.go
```diff
@@ -257,7 +258,27 @@ func (s *Service) FetchPartnerAccounts(ctx context.Context, req *apiv2.FetchAcco
...
+
+	// Return the default dev server account
+	account := &apiv2.Account{
+		Id:        consts.DevServerAccountID.String(),
+		Email:     "dev@inngest.local",
```
djfarrelly, line 269: **question** - In cloud, I would assume this maps to the billing email, which by default is the email of the user that created the account. It may change though. I wonder if there is any benefit to making this `billingEmail` instead of `email` - this may have already been discussed last week, so I don't want to block if it's already been sorted. 

I also don't think it's that big of a deal if it's just "email" either, just asking for posterity as well!

AndyInternet: You are correct that this maps to billing email in cloud. Since it's already in the wild with a beta partner, I think the best approach is to update the docs to call it out that this is the billingEmail in cloud.
image

PR #2891, opened 2025-08-26
Implement OSS Endpoint fetch signing key

AndyInternet:
## Description

Implement `GET /v2/keys/signing`

## Motivation
REST API v2

- [ ] Docs

## Checklist


pkg/api/v2/signing_keys_provider.go
```diff
@@ -0,0 +1,32 @@
...
+		return nil
+	}
+	return signingKeysProvider(*signingKey)
+}
+
+func (key signingKeysProvider) GetSigningKeys(ctx context.Context) ([]*apiv2.SigningKey, error) {
+	return []*apiv2.SigningKey{{
+		Id:          "",
+		Name:        "",
+		Environment: "dev",
```
djfarrelly, line 28: **question** (non-blocking) (for cloud): This may depend on how API scopes will work, but will these endpoints return signing keys across multiple environments if it's an account-scoped api key? And if it's a signing key (which is associated with a single env), this would return just keys for the env?

If so, maybe this should be `EnvironmentID` - otherwise this may be an ambiguous - is it a name and id, a slug, etc.

pkg/api/v2/signing_keys_provider.go
```diff
@@ -0,0 +1,32 @@
...
+		Key:         string(key),
+		CreatedAt:   timestamppb.New(time.Now()),
+	}}, nil
```
djfarrelly, line 31: **thought**(non-blocking) (for cloud): For Cloud, we will likely want to add another field in this `SigningKey` struct/protobuf that returns some sort of field for if the key is "active" or "current" or similar to support multiple keys for signing key rotation. This is future and additive - not blocking at all, but will be nice to support later somehow where a key could be created via POST and maybe a key could be deleted to complete rotation. (TBD on ergonomics)

PR #2907, opened 2025-08-28
fix: only trigger 1 onFailure call on parallel step failure

Linell:
## Description

Use the existing deduplication based on the idempotency key to prevent duplicate onFailure calls when parallel steps
fail.

Note that this only prevents the duplicated onFailure event from actually scheduling -- the event itself is still sent and
recorded, which may be confusing to users. This could potentially be resolved by [this](https://linear.app/inngest/project/first-class-event-id-idempotency-84d7a96e57cd/overview) Linear project.

You can see an example of two runs of the example function here, with the earlier run executing the onFailure twice while the changes in this PR only execute the function once.
Screenshot 2025-08-29 at 12 15 30 PM

## Motivation
[EXE-194](https://linear.app/inngest/issue/EXE-194/duplicate-onfailure-calls-when-multiple-parallel-steps-fail)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/finalize.go
+++ b/pkg/execution/executor/finalize.go
@@ -223,7 +223,7 @@ func (e *executor) finalizeEvents(ctx context.Context, opts execution.FinalizeOp
 			// Legacy - send inngest/function.failed, except for when the function has been cancelled.
 			if !strings.Contains(*opts.Response.Err, state.ErrFunctionCancelled.Error()) {
 				freshEvents = append(freshEvents, event.Event{
-					ID:        ulid.MustNew(uint64(now.UnixMilli()), rand.Reader).String(),
+					ID:        opts.Metadata.ID.RunID.String(), // using the RunID as the ID prevents duped runs for parallel steps
 					Name:      event.FnFailedName,
 					Timestamp: now.UnixMilli(),
 					Data:      data,
@@ -233,7 +233,7 @@ func (e *executor) finalizeEvents(ctx context.Context, opts execution.FinalizeOp
 			// Add function cancelled event
 			if *opts.Response.Err == state.ErrFunctionCancelled.Error() {
 				freshEvents = append(freshEvents, event.Event{
-					ID:        ulid.MustNew(uint64(now.UnixMilli()), rand.Reader).String(),
+					ID:        opts.Metadata.ID.RunID.String(), // using the RunID as the ID prevents duped runs for parallel steps
 					Name:      event.FnCancelledName,
 					Timestamp: now.UnixMilli(),
 					Data:      data,
--- a/tests/golang/parallel_test.go
+++ b/tests/golang/parallel_test.go
@@ -616,3 +616,96 @@ func TestParallelStepsDuplicatePlan(t *testing.T) {
 
 	r.Equal(1, int(atomic.LoadInt32(&counter)))
 }
+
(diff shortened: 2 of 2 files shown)


pkg/execution/executor/finalize.go
```diff
@@ -223,7 +223,7 @@ func (e *executor) finalizeEvents(ctx context.Context, opts execution.FinalizeOp
 			// Legacy - send inngest/function.failed, except for when the function has been cancelled.
 			if !strings.Contains(*opts.Response.Err, state.ErrFunctionCancelled.Error()) {
 				freshEvents = append(freshEvents, event.Event{
-					ID:        ulid.MustNew(uint64(now.UnixMilli()), rand.Reader).String(),
+					ID:        opts.Metadata.ID.RunID.String(),
```
amh4r: Could you add a comment explaining why we want the idempotency key to be the failing run ID?

tests/golang/parallel_test.go
```diff
@@ -616,3 +616,93 @@ func TestParallelStepsDuplicatePlan(t *testing.T) {
...
+                Name: "test/parallel-fail",
+                Data: map[string]any{"test": true},
+        })
+        require.NoError(t, err)
+
+        <-time.After(5 * time.Second)
+
+        require.Eventually(t, func() bool {
+                return atomic.LoadInt32(&failureCount) == 1
+        }, 15*time.Second, 500*time.Millisecond)
```
amh4r:
```suggestion
        require.EventuallyWithT(t, func(t *assert.CollectT) {
                require.Equal(t, 1, int(atomic.LoadInt32(&failureCount)))
        }, 15*time.Second, 500*time.Millisecond)
```

Nit: if we use an assertion then we'll also see what `failureCount` is if it isn't 1

tests/golang/parallel_test.go
```diff
@@ -616,3 +616,93 @@ func TestParallelStepsDuplicatePlan(t *testing.T) {
...
+
+        registerFuncs()
+
+        _, err = inngestClient.Send(ctx, inngestgo.Event{
+                Name: "test/parallel-fail",
+                Data: map[string]any{"test": true},
+        })
+        require.NoError(t, err)
+
+        <-time.After(5 * time.Second)
```
amh4r: Is this sleep necessary? The poll in the the subsequent `Eventually` should take care of it

tests/golang/parallel_test.go
```diff
@@ -616,3 +616,93 @@ func TestParallelStepsDuplicatePlan(t *testing.T) {
...
+                inngestgo.EventTrigger(
+                        "inngest/function.failed",
+                        inngestgo.StrPtr("event.data.function_id == 'parallel-fail-parallel-fail-test'"),
+                ),
+                func(ctx context.Context, input inngestgo.Input[map[string]any]) (any, error) {
+                        atomic.AddInt32(&failureCount, 1)
+
+                        // Assert failure data is present
+                        require.NotEmpty(t, input.Event.Data["run_id"])
+                        require.Equal(t, "parallel-fail-parallel-fail-test", input.Event.Data["function_id"])
```
amh4r: I don't think `require` assertions work in Inngest functions. Under the hood, they work by panicking but we have our own panic recover in the Go SDK

krsjenswbp (approved):

PR #2908, opened 2025-08-28
debugger resolvers, loaders and fetch/map implementation

jacobheric:
## Description

Exact shape of debug run and session data may change but good enough to keep the UI parts moving for now.

## Motivation
Support debug run and session views in UI.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/telemetry/exporters/kafka.go
+++ b/pkg/telemetry/exporters/kafka.go
@@ -212,7 +212,12 @@ func (e *kafkaSpanExporter) ExportSpans(ctx context.Context, spans []trace.ReadO
 				status = "error"
 
 				if strings.Contains(err.Error(), consts.KafkaMsgTooLargeError) {
-					l.Error("error", err, "span proto size", proto.Size(span), "marhsalled span proto size", len(byt), "span.output size", len(span.Output))
+					batchSize := len(span.Events)
+					evtPayloadSizes := make([]int, len(span.Events))
+					for i, evt := range span.Events {
+						evtPayloadSizes[i] = len(evt.Name)
+					}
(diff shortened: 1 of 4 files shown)


pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -93,41 +93,125 @@ func (w wrapper) GetSpansByRunID(ctx context.Context, runID ulid.ULID) (*cqrs.Ot
...
 	for _, span := range spans {
-		st := strings.Split(span.StartTime.(string), " m=")[0]
-		startTime, err := time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", st)
+		// Type assert to get the fields directly
+		var traceID, runIDStr string
+		var dynamicSpanID, parentSpanID sql.NullString
+		var startTime, endTime interface{}
+		var spanFragments interface{}
+
+		switch s := any(span).(type) {
```
jacobheric: ~~This has a smell about it. I've abstracted the shared parts of the mapping for run, debug and session trace mappings but those row types are autogenerated so I any typed this and do some type assertions here. Lemme know if there is a better way to do this in go.~~

Switched to an interface approach.

ui/packages/components/src/Debugger/DebugRun.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const maxTime = toMaybeDate(debugRun.endedAt) ?? new Date();
+
+  return (
+    <div className={`w-full pb-4 pr-8`}>
+      <Trace
+        depth={0}
+        maxTime={maxTime}
+        minTime={minTime}
+        runID={debugRun.runID}
+        trace={{ ...(debugRun as any), name: 'Debug Run' }}
```
jacobheric, line 20: Will implement a DebugTrace component on follow up, for now I'm just using the old trace here and ignoring the errors.

jpwilliams (approved):

PR #2930, opened 2025-09-02
[Insights]: Copy template name to tab name

doverdbng:
## Description

Previously, opening a new tab by clicking on a query resulted in the tab being titled "Untitled query." After this change, the tab will take on the name of the template.

## Motivation
UX clarity

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsTabManager.tsx
+++ b/ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsTabManager.tsx
@@ -72,7 +72,12 @@ export function useInsightsTabManager(
         createTabBase(makeEmptyUnsavedQuery());
       },
       createTabFromQuery: (query: Query | QuerySnapshot | QueryTemplate) => {
-        if (isQueryTemplate(query) || isQuerySnapshot(query)) {
+        if (isQueryTemplate(query)) {
+          createTabBase({ ...makeEmptyUnsavedQuery(), query: query.query, name: query.name });
+          return;
+        }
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsTabManager.tsx
```diff
@@ -72,7 +72,12 @@ export function useInsightsTabManager(
         createTabBase(makeEmptyUnsavedQuery());
       },
       createTabFromQuery: (query: Query | QuerySnapshot | QueryTemplate) => {
-        if (isQueryTemplate(query) || isQuerySnapshot(query)) {
+        if (isQueryTemplate(query)) {
+          createTabBase({ ...makeEmptyUnsavedQuery(), query: query.query, name: query.name });
```
albertchae, line 76: maybe we want to add "From Template:" or something like that, but we can see what user feedback is first.

Also from playing around on preview, looks like it's fine if we have multiple tabs with the same name 👍

albertchae (approved):

ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsTabManager.tsx
doverdbng: Yeah, I also considered something like that but it's probably good to take the simplest approach first.

That's right, duplicate names shouldn't be an issue. Every query (~tab) has a generated `id` to differentiate it.

PR #2946, opened 2025-09-08
Add support for multiple prod envs in UI

amh4r:
> :warning: Do not merge until https://github.com/inngest/monorepo/pull/5113 is released

## Description

Add support for multiple production environments. This only affects a single customer, since they're the only one with multi prod env support. All other accounts are unaffected by this change.

Note that there were 2 shortcuts in this PR:
1. Prod envs don't appear in the top-left env dropdown
2. The "all environments" page looks weird at the top

Resolves https://linear.app/inngest/issue/CON-271

## Testing

image

### DIFF
--- a/ui/apps/dashboard/src/utils/environments.ts
+++ b/ui/apps/dashboard/src/utils/environments.ts
@@ -35,8 +35,8 @@ export function getDefaultEnvironment(
 ): Environment | null {
   let isMultiProd = false;
   let prodCount = 0;
-  for (const workspace of environments) {
-    if (workspace.type === EnvironmentType.Production) {
+  for (const env of environments) {
+    if (env.type === EnvironmentType.Production) {
       prodCount++;
     }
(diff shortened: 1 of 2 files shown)


ui/apps/dashboard/src/utils/environments.ts
```diff
@@ -33,7 +33,30 @@ export function getActiveEnvironment(
 export function getDefaultEnvironment(
   environments: NonEmptyArray<Environment>
 ): Environment | null {
-  return environments.find((e) => e.type === EnvironmentType.Production) || null;
+  let isMultiProd = false;
+  let prodCount = 0;
+  for (const workspace of environments) {
```
AndyInternet: nit - strange that it's env and environments everywhere else, but workspace here. Could be environment of environments.

amh4r: AI 😡

PR #2951, opened 2025-09-09
fix: problem with multiple event trigger expressions

Linell:
## Description

Fix a problem where event trigger expressions were being applied to all event triggers, not just the specific event for which it was defined.

## Motivation
[EXE-235](https://linear.app/inngest/issue/EXE-235/event-trigger-expression-problems-when-multiple-triggers)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/runner/runner.go
+++ b/pkg/execution/runner/runner.go
@@ -502,6 +502,8 @@ func (s *svc) functions(ctx context.Context, tracked event.TrackedEvent) error {
 	// Do this once instead of many times when evaluating expressions.
 	evtMap := evt.Map()
 
+	matchingPatterns := inngest.GenerateMatchingPatterns(evt.Name)
+
 	for _, fn := range fns {
 		// We want to initialize each function concurrently;  some of these
 		// may have expressions that take ~tens of milliseconds to run, and
@@ -523,6 +525,11 @@ func (s *svc) functions(ctx context.Context, tracked event.TrackedEvent) error {
 
 			for _, t := range copied.Triggers {
 				if t.EventTrigger == nil {
+					continue
+				}
+
+				// Only process triggers that match the current event
+				if !t.EventTrigger.MatchesAnyPattern(matchingPatterns) {
 					continue
 				}
 
--- a/pkg/execution/runner/runner.go
+++ b/pkg/execution/runner/runner.go
@@ -15,6 +15,7 @@ import (
 	"github.com/inngest/inngest/pkg/consts"
 	"github.com/inngest/inngest/pkg/cqrs"
 	"github.com/inngest/inngest/pkg/event"
+	"github.com/inngest/inngest/pkg/event_trigger_patterns"
 	"github.com/inngest/inngest/pkg/execution"
 	"github.com/inngest/inngest/pkg/execution/batch"
 	"github.com/inngest/inngest/pkg/execution/executor"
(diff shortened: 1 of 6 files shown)


tests/golang/multi_trigger_expression_test.go
```diff
@@ -0,0 +1,149 @@
...
+		
+		_, err = inngestClient.Send(ctx, inngestgo.Event{
+			Name: "user.created",
+			Data: map[string]any{"type": "premium"},
+		})
+		require.NoError(t, err)
+		time.Sleep(2 * time.Second)
+
+		assert.Equal(t, int32(1), atomic.LoadInt32(&executionCount))
+		assert.Equal(t, "user.created", lastExecutedEvent)
```
amh4r: Sleeps in tests increase the risk of flakiness. Could you use `require.EventuallyWithT` instead?

tests/golang/multi_trigger_expression_test.go
```diff
@@ -0,0 +1,149 @@
...
+		atomic.StoreInt32(&executionCount, 0)
+		
+		_, err = inngestClient.Send(ctx, inngestgo.Event{
+			Name: "user.created", 
+			Data: map[string]any{"type": "standard"}, // Doesn't match premium requirement
+		})
+		require.NoError(t, err)
+		time.Sleep(2 * time.Second)
+
+		assert.Equal(t, int32(0), atomic.LoadInt32(&executionCount))
```
amh4r: We do need to wait some time here because we're ensuring that something didn't happen. For this, I like to use `require.Never` since IMO it better communicates the purpose of the wait.

Also, I would wait longer than 2 seconds. Maybe 5 instead? We might have a other high load tests running concurrently, which could delay event processing longer than 2 seconds. 5 seconds is still arbitrary, but seems a little safer 😄

tests/golang/multi_trigger_expression_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+		assert.Equal(t, int32(1), atomic.LoadInt32(&executionCount))
+		assert.Equal(t, "user.updated", lastExecutedEvent)
+	})
+
+	t.Run("trigger without expression matches any data", func(t *testing.T) {
+		// Create a separate function for this test case
+		var noExprCount int32
+		
+		noExprFn, err := inngestgo.CreateFunction(
```
amh4r:
```suggestion
		_, err := inngestgo.CreateFunction(
```

This will prevent the unused variable error you're handling later

tests/golang/multi_trigger_expression_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// TestMultiTriggerExpressions tests that functions with multiple event triggers
+// only evaluate expressions for triggers that match the incoming event name.
+// This prevents incorrect cross-trigger expression evaluation.
+func TestMultiTriggerExpressions(t *testing.T) {
+	ctx := context.Background()
+
+	_ = client.New(t)
```
amh4r: What is this doing?

Linell: Absolutely nothing -- vestigial from a LLM helping me write the tests. Sorry about that!

pkg/execution/runner/runner.go
```diff
@@ -526,6 +526,11 @@ func (s *svc) functions(ctx context.Context, tracked event.TrackedEvent) error {
 					continue
 				}
 
+				// Only process triggers that match the current event
+				if t.EventTrigger.Event != evt.Name {
```
amh4r: This will prevent wildcards. For example:
```ts
inngest.createFunction(
  { id: "fn-1" },
  { event: "foo/*" },
  async () => {}
);
```

Given that this change didn't cause any test failures, I suspect we need to add a wildcard test 😄

amh4r (approved):

PR #2972, opened 2025-09-11
Hide Concurrency CTA for Marketplace Accounts

jakobevangelista:
## Description

Hides the concurrency CTA for marketplace accounts

**Note**: The build will fail in staging/preview environments but works locally. This is expected because the build script only runs `pnpm graphql-codegen` when `VERCEL_ENV=production`, but skips it for staging/preview environments. The generated GraphQL files work fine locally since `pnpm graphql-codegen` runs during development.

- [ ] Docs

## Checklist


ui/apps/dashboard/src/components/Metrics/AccountConcurrency.tsx
```diff
@@ -14,9 +14,10 @@ import { getLineChartOptions, getXAxis, lineColors, seriesOptions } from './util
 type Props = {
   data: MetricsResponse | undefined;
   limit?: number;
+  isMarketplace?: boolean;
```
amh4r:
```suggestion
  isMarketplace: boolean;
```

Make it required? Just to ensure we always pipe it thru. Same with the other spot in another file

PR #2976, opened 2025-09-12
Deprecate old events key from state store

KiKoS0:
## Description

Deprecate old `events:{fnID}:{runID}` state store key that held only one event before supporting batching. Also removes old history and errors keys that shouldn't exist in state store anymore as well.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/redis_state.go
+++ b/pkg/execution/state/redis_state/redis_state.go
@@ -1,7 +1,6 @@
 package redis_state
 
 import (
-	"bytes"
 	"context"
 	"embed"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


pkg/execution/state/redis_state/redis_state.go
```diff
@@ -1075,11 +1010,6 @@ func (m shardedMgr) delete(ctx context.Context, callCtx context.Context, i state
 		fnRunState.kg.RunMetadata(ctx, isSharded, i.RunID),
 		fnRunState.kg.Actions(ctx, isSharded, i.WorkflowID, i.RunID),
 		fnRunState.kg.Stack(ctx, isSharded, i.RunID),
-
-		// XXX: remove these in a state store refactor.
-		fnRunState.kg.Event(ctx, isSharded, i.WorkflowID, i.RunID),
```
lkasinathan, line 1080: I assume this is probably fine since batching was introduced a while ago, but just wanna clarify (more so for my understanding) if all the older event keys have already disappeared from the state store.

> When doing so, make sure to keep the TTL setting or deletion of the key around for some time before completely removing that so we make sure we don’t have things lingering around.

KiKoS0: Change has been made since 2023-05: https://github.com/inngest/inngest/pull/450/files#diff-58744e111813b3627e52c6d1b0280fd75fd8d60d6971023309c782851ba64dfaL299-L300 So you are right all runs with the old event key should have been cleared by now or is no longer a valid run. Bruno is also running a scan on the pattern to make sure we remove any that are leftover.

BrunoScheufler (approved):

PR #2996, opened 2025-09-16
[Insights]: Render prettified JSON data in results table

doverdbng:
## Description

This change allows the results table to detect when strings are actually JSON and then to render them in a prettified format. The JSON cells themselves are focusable and allow for vertical scrolling (when focused) subject to a max height restriction.

**NOTE:** This should eventually move to the BE via an actual `json` column type so that the FE doesn't have to parse. This will be better from a performance perspective.

**NOTE x2:** Going through a `JSON.parse` cycle could cause really large numbers to lose precision.

---

Screenshot 2025-09-16 at 3 59 49 PM

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ResultsState/JSONAwareTextCell.tsx
+++ b/ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ResultsState/JSONAwareTextCell.tsx
@@ -15,7 +15,7 @@ export function JSONAwareTextCell({ children }: JSONAwareTextCellProps) {
   return (
     <div className="text-basis text-sm font-medium">
       <pre
-        className="m-0 max-h-[150px] w-full max-w-none overflow-hidden whitespace-pre break-all outline-none [scrollbar-width:thin] group-focus-within:overflow-auto [&::-webkit-scrollbar]:w-1"
+        className="max-h-[160px] max-w-none overflow-hidden outline-none [scrollbar-width:thin] group-focus-within:overflow-auto [&::-webkit-scrollbar]:w-1"
         tabIndex={0}
       >
         {formattedJSON}


doverdbng:
This should be good to go. I just need to test on browsers other than Chrome since table style treatments can vary to a surprising degree.

ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ResultsState/json.ts
```diff
@@ -0,0 +1,17 @@
+export function getFormattedJSONObjectOrArrayString(value: string): null | string {
+  if (!mayBeJSONArray(value) && !mayBeJSONObject(value)) return null;
```
AndyInternet, line 2: nit - incorrect camelCasing on the word maybe. Should be `maybeJSONArray`

ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ResultsState/json.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  try {
+    return JSON.stringify(JSON.parse(value), null, 2);
+  } catch {
+    return null;
+  }
+}
+
+function mayBeJSONArray(str: string): boolean {
```
AndyInternet, line 11: same

ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ResultsState/json.ts
```diff
@@ -0,0 +1,17 @@
...
+  return str.startsWith('[') && str.endsWith(']');
+}
+
+function mayBeJSONObject(str: string): boolean {
```
AndyInternet, line 15: same

doverdbng:
> looks good, definitely an improvement. The scroll/focus behavior is a bit weird but I know it's what we decided on with @sanjanaladdha
> 
> re: JSON parsing and backend, are you suggesting the backend would return a formatted JSON for json types? because otherwise you would still have to `JSON.parse` in the frontend, but the savings would be in not having to use the `[]` or `{}` heuristic and also not calling `JSON.parse` if we have strings that are surrounded by `[]` or `{}` but aren't actually valid JSON.

Fair question. It would probably be an overall perf bump if the BE could do that, but maybe it's not necessary. So long as we keep the row limit, hopefully none of this matters too much, especially the very fast heuristics.

doverdbng:
> looks good, definitely an improvement. The scroll/focus behavior is a bit weird but I know it's what we decided on with @sanjanaladdha
> 
> re: JSON parsing and backend, are you suggesting the backend would return a formatted JSON for json types? because otherwise you would still have to `JSON.parse` in the frontend, but the savings would be in not having to use the `[]` or `{}` heuristic and also not calling `JSON.parse` if we have strings that are surrounded by `[]` or `{}` but aren't actually valid JSON.

Agreed that viewing JSON in the table is a bit strange, but I don't know that there's really an optimal approach where something doesn't feel weird...leaning into text wrapping felt very, very strange to me. I'm kind of curious how often people will actually even fetch JSON instead of extracting specific properties. It might not be that big of a deal either way?

PR #3000, opened 2025-09-17
[Insights]: Move query history to ephemeral react state

doverdbng:
## Description

This fulfills a request to move query history from LocalStorage to ephemeral react state.

**Aside**: It also removes trivial mocks for saved queries and query history, using the empty state instead.

---

Screenshot 2025-09-17 at 11 36 25 AM

- [ ] Docs

## Checklist


ui/apps/dashboard/src/components/Insights/QueryHelperPanel/StoredQueriesContext.tsx
```diff
@@ -77,11 +73,11 @@ export function StoredQueriesProvider({ children, tabManagerActions }: StoredQue
 
   const saveQuerySnapshot = useCallback(
     (snapshot: QuerySnapshot) => {
-      setQuerySnapshots(
-        withId(removeQuerySnapshotIfOverLimit(querySnapshots, 10), snapshot.id, snapshot)
+      setQuerySnapshots((current) =>
```
doverdbng, line 76: I noticed without using the "functional form" of the react updater that if two queries were run at the same time (unlikely), the snapshot history could get overwritten. This is because React can batch state updates, so there was a stale data issue.

I don't think the functional form was working with `useLocalStorage` previously (it should) and I had to use the non-functional form, but I can fix it now.

PR #3001, opened 2025-09-17
Scaffold event type APIs

tonyhb:
- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/v2/endpoints.go
+++ b/pkg/api/v2/endpoints.go
@@ -273,3 +273,40 @@ func (s *Service) ListWebhooks(ctx context.Context, req *apiv2.ListWebhooksReque
 	// 3. Return the list with proper pagination metadata
 	return nil, s.base.NewError(http.StatusNotImplemented, apiv2base.ErrorNotImplemented, "Webhooks not implemented in OSS")
 }
+
+func (s *Service) FetchEventTypes(ctx context.Context, req *apiv2.FetchEventTypesRequest) (*apiv2.FetchEventTypesResponse, error) {
+	// Validate pagination parameters
+	if req.Limit != nil {
+		if *req.Limit < 1 {
+			return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Limit must be at least 1")
+		}
+		if *req.Limit > 250 {
+			return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Limit cannot exceed 250")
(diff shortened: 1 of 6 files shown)


pkg/api/v2/endpoints.go
```diff
@@ -273,3 +273,40 @@ func (s *Service) ListWebhooks(ctx context.Context, req *apiv2.ListWebhooksReque
...
+		}
+		if *req.Limit > 250 {
+			return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Limit cannot exceed 250")
+		}
+	}
+
+	// For now, return empty list
+	// In a full implementation, this would:
+	// 1. Extract account/workspace from auth context
+	// 2. Query the database for unique event names with MIN(event_ts) as createdAt
```
albertchae, line 291: this is what we would do on dev server/self hosted right? since cloud has an `event_types` table? Any reason not to eventually make an actual table here too?

proto/api/v2/service.proto
```diff
@@ -1088,3 +1241,83 @@ message Webhook {
...
+  Page page = 3;
+}
+
+message EventType {
+  string id = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "The event type UUID identifier"
+      example: "\"550e8400-e29b-41d4-a716-446655440000\""
+    }
+  ];
```
albertchae, line 1271: per my other comment, if we don't have a physical `event_types` table, what would this ID be?

albertchae: also do we need an ID if event name will be unique for a type? or is that an incorrect assumption

djfarrelly: It's ok if this isn't a breaking change, but we discussed using user-defined params as much as possible in the Rest API v2 - so meaning that we'd use the event name as the param in the get one endpoint.
https://www.notion.so/inngest/REST-API-v2-Resource-IDs-269b64753bbd80298356d1af9544049f?source=copy_link

```
/v2/event-types/<url-encoded-event-name>
ex. 
/v2/event-types/user.created
/v2/event-types/app%2Fuser.created
```

Otherwise, to do much with this data from a developer standpoint, to get an event type id (as uuid), they'd have to fetch and filter `/v2/event-types` which isn't super ergonomic at all.

Also, then if we return the response shape of `/event-types` as `id` as a uuid it's something that isn't known by the user. We could support querying the endpoint by uuid OR event name (url encoded) as well.

PR #3002, opened 2025-09-17
[Insights]: Move query management from LS to BE

doverdbng:
## Description

This PR moves management of saved queries from LocalStorage to the BE.

Additionally, it:
- removes locally defined types like `Query` in favor of GQL types
- disambiguates `Tab` as its own type instead of being a duplicate of "Query"
- ensures tabs have their own unique ID instead of reusing that of the query associated with them
- adds toasts for successful and failed delete/save/update operations
- detects BE's duplicate name restriction and warns users

**NOTE:** Since we're still in beta, there's nothing in here to move existing queries from LS to the BE.

**NOTE x2:** This depends on an open BE PR.

**Review's Note:** One of the biggest changes in this PR is the disambiguation of `Query` (now using the GQL type) and the new `Tab` type. They are similar in many ways, and the code before used them interchangeably. Now, `Tab` is a slightly less involved type and essentially has a foreign key relationship to queries via its `savedQueryId` key. 

## Motivation

This allows user to persist their queries across machines.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/QueryHelperPanel/StoredQueriesContext.tsx
+++ b/ui/apps/dashboard/src/components/Insights/QueryHelperPanel/StoredQueriesContext.tsx
@@ -82,7 +82,7 @@ export function StoredQueriesProvider({ children, tabManagerActions }: StoredQue
         }
       }
     },
-    [beSaveQuery, beUpdateQuery, tabManagerActions]
+    [beSaveQuery, beUpdateQuery, refetchSavedQueries, tabManagerActions]
   );
 
   const deleteQuery = useCallback(


ui/apps/dashboard/src/components/Insights/QueryHelperPanel/useModifySavedQueries.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const ids = result.data.removeInsightsQuery?.ids;
+      if (ids === undefined) return { ok: false, error: 'other' };
+
+      return { ok: true, data: ids };
+    },
+    [executeMutation, runRemove]
+  );
+
+  const saveQuery = useCallback<(args: SaveQueryArgs) => Promise<MutationResult<InsightsQuery>>>(
```
doverdbng: There's a slight gap in that save/query/delete rely on the list query being re-fetched for the queries to be updated on the client. Naturally, that will work most of the time, and without good existing patterns for manual cache manipulations, I think the correct amount of engineering is probably to live with the edge case for now.

One alternate solution is to fetch the latest individual query on hover/click, but I'd hate to slow down opening a query to accommodate a rare edge case.

doverdbng: Any follow up save / update / delete is likely to fix the problem. And the user would probably try it in most cases (the orange dot will suggest the query isn't sync'd). We can keep an eye on it, but I don't want to over-engineer something for this edge case. It's a shame that the default cache we use for urql doesn't make it easy (possible?) to manually modify the cache.

albertchae:
Saving the query doesn't seem to update the saved queries list without a hard refresh. Not sure if this is due to the gql type change that led to some codegen and query changes 🤔

https://github.com/user-attachments/assets/93971c2f-c1cd-4b7d-a52c-addae43c67ea

albertchae:
> Saving the query doesn't seem to update the saved queries list without a hard refresh. Not sure if this is due to the gql type change that led to some codegen and query changes 🤔
> 

OK looks like this must have been due to the type issues. Latest preview works!

ui/apps/dashboard/src/components/Insights/InsightsStateMachineContext/useFetchInsights.ts
```diff
@@ -16,8 +16,8 @@ export interface FetchInsightsParams {
 
 type FetchInsightsCallback = (query: string, name: undefined | string) => void;
 
-const insightsQuery = graphql(`
-  query Insights($query: String!, $workspaceID: ID!) {
```
albertchae, line 20: I think this was causing the conflict with https://github.com/inngest/monorepo/pull/5257, but it's best to avoid by not having types end in `Query` anyway

albertchae (approved):
LGTM

doverdbng:
cc: @albertchae on this [pragmatic extra commit](https://github.com/inngest/inngest/pull/3002/commits/fe46d5b9a3ff4db65bb9eb3df13fcff298e9d9c9) to fix a bug you saw at one point. I'm certainly open to better fixes.

PR #3030, opened 2025-09-22
Add run claim JWTs for sync APIs

tonyhb:
This allows us to create single-scoped JWTs that allow access to a specific run ID's outputs.  Critically, we create these JWTs whenever a sync fn checkpoints and turns async.  The sync fns can then take these JWTs and give them to the original caller of the API, allowing them to fetch the output of the async fn in the future.

## Checklist

### DIFF
--- a/pkg/api/apiv1/checkpoint.go
+++ b/pkg/api/apiv1/checkpoint.go
@@ -54,13 +54,17 @@ type checkpointAPI struct {
 	// upserted tracks fn IDs and their associated config in memory once upserted, allowing
 	// us to prevent DB queries from hitting the DB each time a sync fn begins.
 	upserted *ccache.Cache
+
+	// runClaimsSecret is the secret for creating run claims JWTs
+	runClaimsSecret []byte
 }
 
-func NewCheckpointAPI(o Opts) CheckpointAPI {
+func NewCheckpointAPI(o Opts, runClaimsSecret []byte) CheckpointAPI {
 	api := checkpointAPI{
-		Router:   chi.NewRouter(),
-		Opts:     o,
-		upserted: ccache.New(ccache.Configure().MaxSize(10_000)),
+		Router:          chi.NewRouter(),
+		Opts:            o,
+		upserted:        ccache.New(ccache.Configure().MaxSize(10_000)),
+		runClaimsSecret: runClaimsSecret,
 	}
 
 	api.Post("/", api.CheckpointNewRun)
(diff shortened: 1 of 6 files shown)


pkg/api/apiv1/checkpoint.go
```diff
@@ -430,6 +492,40 @@ func (a checkpointAPI) CheckpointResponse(w http.ResponseWriter, r *http.Request
...
+		return
+	}
+
+	until := time.Now().Add(CheckpointOutputWaitMax)
+
+	for time.Now().Before(until) {
+		output, err := a.outputReader.RunOutput(r.Context(), claims.Env, claims.RunID)
+
+		if err == nil {
+			// XXX: (tonyhb) add status code handling here.
```
BrunoScheufler, line 518: Should this be handled in a follow-up or in here?

tonyhb: In a followup

pkg/devserver/devutil/devutil.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, err
+	}
+	if run == nil {
+		return nil, fmt.Errorf("run not found: %s", runID)
+	}
+
+	switch run.Status {
+	case enums.RunStatusCompleted, enums.RunStatusFailed:
+		return run.Output, nil
+	}
```
jpwilliams, line 65: Can we use `RunStatusEnded()` here?

Some of those other statuses may not have outputs yet, but it'd be fair for them to later, for example to say why they were skipped, cancel reasons, etc.

https://github.com/inngest/inngest/blob/a552ce57c0923e04ad8bdc13225efd80c2721f47/pkg/enums/run_status.go#L67-L74

pkg/api/apiv1/checkpoint.go
```diff
@@ -430,6 +492,40 @@ func (a checkpointAPI) CheckpointResponse(w http.ResponseWriter, r *http.Request
...
+		w.WriteHeader(401)
+		_ = publicerr.WriteHTTP(w, publicerr.Wrap(err, 401, "Unable to find run with auth token"))
+		return
+	}
+
+	if a.outputReader == nil {
+		w.Header().Set("content-type", "application/json")
+		_, _ = w.Write([]byte(`{"status":"unknown","message":"unable to fetch run output"}`))
+		return
+	}
```
jpwilliams, line 510: Can we log something internally here (or `reportError()`) as this indicates bad setup somewhere?

jpwilliams (approved):

PR #3052, opened 2025-09-23
Implement idempotence for appending events to a batch

lkasinathan:
## Description
Do not append a duplicate event into a batch if the event was already appended to the same batch earlier.

We add a idempotence set for each batch ID to store all the event IDs and use this set to check if an event has already been appended to a batch. If it was, the append script early returns with a `status=BatchItemExists`.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/batch/batch_test.go
+++ b/pkg/execution/batch/batch_test.go
@@ -143,10 +143,11 @@ func TestBatchAppendIdempotenceDifferentBatches(t *testing.T) {
 		Version: 0,
 	}
 
-	// var lastEventID ulid.ULID
 	var lastBatchID string
 	for i := range 10 {
+		// append a new event to the batch
 		bi.EventID = ulid.MustNew(ulid.Now(), rand.Reader)
+
 		res, err := bm.Append(context.Background(), bi, function)
 		require.NoError(t, err)
 		require.NotEmpty(t, res.BatchID)
@@ -159,7 +160,6 @@ func TestBatchAppendIdempotenceDifferentBatches(t *testing.T) {
 		default:
 			require.Equal(t, enums.BatchAppend, res.Status)
 		}
-		// lastEventID = bi.EventID
 		lastBatchID = res.BatchID
 	}
 


darwin67 (approved):

BrunoScheufler (approved):

pkg/execution/batch/lua/append.lua
```diff
@@ -5,13 +5,14 @@
 local batchPointerKey = KEYS[1]      -- key to the batch pointer
 
 local batchLimit = tonumber(ARGV[1]) -- max size configured for this batch
-local event = ARGV[2]                -- event to be appended to the batch
-local newULID = ARGV[3]              -- ULID to update the pointer with, either if the batch is full or doesn't exist
-local prefix = ARGV[4]               -- the prefix used for redis
+local eventID = ARGV[2]
+local event = ARGV[3]                -- event to be appended to the batch
```
amh4r, line 9: It looks like we're reordering positional args passed to `append.lua`. How does that coordinate with Go code changes? In other words, will we have a period of time where some Go is passing `event` as `ARGV[2]` and some is passing as `ARGV[3]`?

amh4r (changes requested):
I left a comment about the rollout coordination between Go and Lua. Just being extra paranoid 😬

pkg/execution/batch/lua/append.lua
BrunoScheufler: The Lua scripts are embedded in Go, there is decoupling so the old services use the old scripts while the new services use the new scripts

amh4r: Ok cool! Thanks

pkg/execution/batch/batch_test.go
```diff
@@ -59,6 +59,54 @@ func TestBatchSizeLimit(t *testing.T) {
 	require.Equal(t, enums.BatchMaxSize, res.Status)
 }
 
+func TestBatchAppendIdempotence(t *testing.T) {
```
amh4r, line 62: If you're feeling spicy, maybe another test that ensures the same event can be appended to 2 different batches. Reading the `append.lua`, there shouldn't be a problem with that. But a test would be awesome documentation of that!

lkasinathan: :fire::fire::fire: Excellent callout! 
Duplicate events can definitely get added to different batches - if the first event was the last one in a batch and it comes in again, it is possible that we could schedule it for execution again as part of a different batch. 

This is slightly more tricky to handle unless we keep recently scheduled eventIDs around. I've added a test to document the current behaviour.

PR #3054, opened 2025-09-24
Allow marketplace accounts to see usage page

amh4r:
Allow marketplace accounts to see usage page. They cannot see any other billing pages. This will affect DigitalOcean Marketplace, Vercel Marketplace, and Replit users

image

### DIFF
--- a/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/billing/MarketplaceAccessControl.tsx
+++ b/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/billing/MarketplaceAccessControl.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { usePathname } from 'next/navigation';
+import { redirect, usePathname } from 'next/navigation';
 
 import { pathCreator } from '@/utils/urls';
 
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/app/(organization-active)/(dashboard)/billing/MarketplaceAccessControl.tsx
```diff
@@ -0,0 +1,31 @@
...
+  isMarketplace,
+}: React.PropsWithChildren<Props>) {
+  const pathname = usePathname();
+
+  if (isMarketplace) {
+    const isAllowed = marketplaceAllowedPaths.some((allowedPath) => pathname.endsWith(allowedPath));
+
+    if (!isAllowed) {
+      // Redirect to usage page if trying to access non-whitelisted page
+      window.location.href = pathCreator.billing({ tab: 'usage' });
```
jakobevangelista: should we use nextjs's `useRouter` here instead of `window.location.href` `useRouter` would avoid a full page reload and be more consistent with nextjs' patterns

ui/apps/dashboard/src/app/(organization-active)/(dashboard)/billing/layout.tsx
jakobevangelista, line 1: should we put the logic for redirecting either directly in this server component or in middleware so the redirect can't get overridden client-side?

amh4r: I think it's risky to put in a layout since they aren't rerendered when the path changes (e.g. navigating between `/billing` sub pages).

For middleware, I dunno if we want that because it'd be a fetch on every request

PR #3055, opened 2025-09-25
fix hanging invoke runs when retrying from i/o timeouts

KiKoS0:
## Description

Fixes hanging invokes that get retried because of MDB i/o timeouts. This was a state proxy issue where ErrIdentifierExists was returned wrapped to the Schedule function while we are doing exact error matching instead of using errors.Is. I fixed both, made it use errors.Is just in case we encounter this issue again and also correctly unwrapped the error from the state proxy client.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -713,9 +713,10 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 	}
 
 	st, err := e.smv2.Create(ctx, newState)
-	switch err {
-	case nil, state.ErrIdentifierExists: // no-op
-	case state.ErrIdentifierTomestone:
+	switch {
+	case err == nil: // no-op
+	case errors.Is(err, state.ErrIdentifierExists): // no-op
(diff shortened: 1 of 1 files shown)


pkg/execution/executor/executor.go
```diff
@@ -713,9 +713,10 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 	}
 
 	st, err := e.smv2.Create(ctx, newState)
-	switch err {
-	case nil, state.ErrIdentifierExists: // no-op
-	case state.ErrIdentifierTomestone:
+	switch {
+	case err == nil: // no-op
+	case errors.Is(err, state.ErrIdentifierExists): // no-op
```
jpwilliams, line 718: These always look like fallthrough cases to me 😆

BrunoScheufler (approved):

pkg/execution/executor/executor.go
KiKoS0: same and i'm not getting used to them any soon lol also just knew about the `fallthrough` keyword

PR #3056, opened 2025-09-25
retry steps on system errors after receiving a successful SDK response

KiKoS0:
## Description
same fix as https://github.com/inngest/monorepo/pull/5297 but for dev server

The idea is if we've hit an internal system error after receiving an SDK response, we should allow it to retry instead of always checking if the response is retryable. The assumption is that handling the response should be idempotent and continue exactly where it failed. The main example that requires this fix are invokes:
- SDK tries to invoke a step
- we hit an i/o timeout when trying to save the timeout's pause for the current run inside [handleGeneratorInvokeFunction](https://github.com/inngest/inngest/blob/03581414262ccc5d215546884253ebdc4acebd73/pkg/execution/executor/executor.go#L3300).
- the error is returned all the way up to [handleExecutorJob](https://github.com/inngest/monorepo/blob/1c2d928013fbec0d23cb172ffcca8667750b4c43/pkg/execution/impls/internalredis/executor_queue.go#L22) which has now access to the response and can now enter this to check if the error is retryable or not: https://github.com/inngest/monorepo/blob/1c2d928013fbec0d23cb172ffcca8667750b4c43/pkg/execution/impls/internalredis/executor_queue.go#L102-L104
- and according to this clause in resp.Retryable we return that it''s not retryable because there was no error in the response (it was an internal system error i/o timeout) https://github.com/inngest/inngest/blob/03581414262ccc5d215546884253ebdc4acebd73/pkg/execution/state/driver_response.go#L164-L167

Allowing it to retry means that the SDK will send another invoke step opcode, it'll either create the pause that failed or just get it from the store (sometimes i/o timeouts will persist the data but still returns an error) and it will continue where it left of enqueuing the job that will timeout the pause and sending the invoke event.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/service.go
+++ b/pkg/execution/executor/service.go
@@ -313,7 +313,11 @@ func (s *svc) handleQueueItem(ctx context.Context, item queue.Item) (bool, error
 	if err != nil || (resp != nil && resp.Err != nil) {
 		// Accordingly, we check if the driver's response is retryable here;
 		// this will let us know whether we can re-enqueue.
-		if resp != nil && !resp.Retryable() {
+		//
+		// If the error did not come from the response (which is likely the case here)
+		// and is likely a system error we should skip checking if the response is
+		// retryable and always retry.
+		if resp != nil && resp.Err != nil && !resp.Retryable() {
 			return false, nil
 		}
 


jpwilliams (approved):
From internal comms:

---

Errors are messy here rn.
It looks to me simply like `DriverResponse#Retryable()` was written at a different time and quite sensibly returns `false` if there's no error - why retry if there's no reason to?

```go
func (r DriverResponse) Retryable() bool {
	if r.Err == nil {
		// There's no error, so no need to retry
		return false
	}

```

But that doesn't fit in with the context in which it's being used in `executor_queue.go` there.
As it looks like you discussed, you could totally do `runCtx.SetError(err)` and that's exactly what some of the gateways do:

```go
		// And, finally, if this is retryable return an error which will be retried.
		// Otherwise, we enqueue the next step directly so that the SDK can throw
		// an error on output.
		if runCtx.ShouldRetry() {
			// Set the response error, ensuring the response is retryable in the queue.
			runCtx.SetError(err)
```

But that's really just some ugly code to make it work.
I'm okay with your proposed change as it makes sense in that context.

Incidentally, all other uses of `DriverResponse#Retryable()` only call it `if resp.Err != nil`❗

Kind of a code smell that it has that check there at all.

pkg/execution/executor/service.go
```diff
@@ -313,7 +313,11 @@ func (s *svc) handleQueueItem(ctx context.Context, item queue.Item) (bool, error
 	if err != nil || (resp != nil && resp.Err != nil) {
```
jpwilliams, line 313: Incidentally, this case was already protected in OSS. 😄

PR #3059, opened 2025-09-25
Implement eager cancellation for start and finish timeouts.

lkasinathan:
## Description

TODO

## Motivation
TODO

- [ ] Docs

## Checklist


pkg/execution/executor/service.go
```diff
@@ -503,6 +507,159 @@ func (s *svc) handleCancel(ctx context.Context, item queue.Item) error {
...
+		})
+	}
+	// timeout was extended, requeue eager cancellation.
+	qm, ok := s.queue.(redis_state.QueueManager)
+	if !ok {
+		l.Error("queue does not conform to queue manager")
+		return nil
+	}
+	requeueAt := jobEnqueuedAt.Add(*timeout)
+	err = qm.Enqueue(ctx, item, requeueAt, queue.EnqueueOpts{})
```
BrunoScheufler, line 677: I'd have to check but this probably doesn't push back the queue item -- you may want to use .Requeue for this or alternatively enqueue a new item, after which the current item will be dequeued (essentially cloning with a new time to run and having the old one dropped)

lkasinathan: > or alternatively enqueue a new item, after which the current item will be dequeued (essentially cloning with a new time to run and having the old one dropped)

interesting, this is what I _thought_ I was doing here, is it not?

It worked when I tested this manually (I created a fn with timeout x, sent an event and increased the timeout to y) -- I noticed this system job get dequeued at x, realize that the timeout is actually not met and gets dequeued again at y as a result of this Enqueue agiain here.

lkasinathan: > I'd have to check but this probably doesn't push back the queue item

ftr, this was because bruno suspected idempotency (on `Item.JobID`) would stop us from re-queueing the same Item again.

What's actually happening:
- On the originial system job enqueue[[1]](https://github.com/inngest/inngest/blob/lakshmi/sys-363-eager-cancellation-for-timeouts/pkg/execution/executor/executor.go#L554), we set a job ID `Item.JobID=XYZ`.
    - Just before we actually write to the queue, `Item.JobID` is re-written to be `hash(XYZ)` [[2]](https://github.com/inngest/inngest/blob/main/pkg/execution/state/redis_state/queue.go#L1327-L1330) [[3]](https://github.com/inngest/inngest/blob/main/pkg/execution/state/redis_state/queue.go#L1350-L1352)
- When we first process this system job, `Item.JobID` is now `hash(XYZ)`
  -  During processing, when we realize we need to push this job back because timeout was extended on the recent config, we call Enqueue again on `Item`
  - Note that we do NOT set `EnqueueOpts.PassthroughJobId` when we Enqueue again [[4]](https://github.com/inngest/inngest/blob/lakshmi/sys-363-eager-cancellation-for-timeouts/pkg/execution/executor/service.go#L651-L658)
  - And this time, the job `item` gets enqueued with a idempotency key of `hash(hash(XYZ))` [[2]](https://github.com/inngest/inngest/blob/main/pkg/execution/state/redis_state/queue.go#L1327-L1330) -- which is why the enqueue doesn't fail here and a new queue item gets enqueued for later. Essentially, we're enqueueing a new system job with this Enqueue call, so basically this:

 > alternatively enqueue a new item, after which the current item will be dequeued (essentially cloning with a new time to run and having the old one dropped)

I'll add some comments to clarify that Enqueue here creates a _new_ queue item.

BrunoScheufler: Thanks for checking this! You're right, the ID will be different as it's hashed once again. I am a little concerned that we rely too much on this implicit implementation detail, though. If we ever changed Enqueue behavior to accept the existing Job ID by default (PassthroughJobID = true), which may not be totally unreasonable for _existing_ queue items, this may lead to issues.

I wonder if we should still override the Job ID on the outside so it's something completely unrelated for the new item, simply copying the item contents instead. That way, there's no question of whether the queue item will ever be skipped due to idempotency, or have any other side effects with global hashes, etc.

That said, this is a personal preference, not a law of physics, so I'm fine just doing it the way you're proposing while being clear about the assumptions being made. ~We could add a comment above the Enqueue usage to state that we are re-enqueueing the existing queue item, which will receive a new ID. We can even link to this discussion, if we want to. This way, we can save time in the future!~

Edit: The comment is perfect, fine with me!

lkasinathan: > I wonder if we should still override the Job ID on the outside so it's something completely unrelated for the new item, simply copying the item contents instead. That way, there's no question of whether the queue item will ever be skipped due to idempotency, or have any other side effects with global hashes, etc.

totally fair, done!

PR #3083, opened 2025-10-01
Check Errors For Singletons Before Logging

mar-inngest:
## Motivation

Bug Fix: We currently log all singleton cancel events as errors

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -739,7 +739,9 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 					err = e.Cancel(ctx, runID, execution.CancelRequest{
 						EventID: &eventID,
 					})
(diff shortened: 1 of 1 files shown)


KiKoS0 (approved):
looks good! We can also gracefully ignore this error in case the run to be cancelled has already finished: https://github.com/inngest/inngest/blob/edddb2939e1f2346cedbc355cf46041e161b871c/pkg/execution/state/state.go#L49

mar-inngest:
Yeah, this issue is relatively infrequent, otherwise, something like this would have been better (unless the state store went away). 
```go
 errCancelStr := "error canceling singleton run"
if errors.Is(err, state.ErrEventNotFound) {
	l.Warn(errCancelStr, "error", err)
} else if err != nil {
	l.ReportError(err, errCancelStr)
}
```

PR #3097, opened 2025-10-06
Accept trace_preview query param on event runs v1 API to use trace preview run status

albertchae:
## Description

There is a discrepancy sometimes when the traces-preview feature flag is on where the UI shows completed but the event runs V1 API is still showing running.

To work around this until we have a proper v2 api, this change allows
users to submit `?trace_preview=true` to their event runs call to get
the run status from the same source as the UI.

- [Frontend looking at trace.status](https://github.com/inngest/inngest/blob/4dcaa131741a7ffcb9ffccd58c21a3bca1b09aea/ui/packages/components/src/RunDetailsV3/RunDetailsV3.tsx#L203)
- [Trace resolver uses `RunTraceLoader` when preview is on](https://github.com/inngest/monorepo/blob/00e4f1d01b3651af13b70b943a673679d77854d2/api/gql/resolver/function_run_v2_resolver.go#L79)
- [`RunTraceLoader` goes to `GetRunTrace`](https://github.com/inngest/monorepo/blob/00e4f1d01b3651af13b70b943a673679d77854d2/api/gql/loader/loader.go#L175)
- [`GetRunTrace` calls metrics RPC `GetSpansByRunID`](https://github.com/inngest/monorepo/blob/00e4f1d01b3651af13b70b943a673679d77854d2/api/gql/loader/loader.go#L279-L281)
- [`TraceReader.GetSpansByRunID` also calls metrics RPC `GetSpansByRunID`](https://github.com/inngest/monorepo/blob/00e4f1d01b3651af13b70b943a673679d77854d2/pkg/traces/trace_reader.go#L40-L41)

So this change reads from the same data source.

## Motivation
Bug report from https://inngest.slack.com/archives/C06K5S9S3K6/p1759499961469429

- [ ] Docs

## Checklist


pkg/api/apiv1/events.go
```diff
@@ -154,9 +155,23 @@ func (a router) getEventRuns(w http.ResponseWriter, r *http.Request) {
...
 	}
 	runs, err := a.GetEventRuns(ctx, parsed)
+
 	if err != nil {
 		_ = publicerr.WriteHTTP(w, err)
 		return
 	}
+
+	// allow user to pass a flag and get the same run status as trace previews
+	if r.URL.Query().Get("trace_preview") == "true" {
```
albertchae, line 164: alternatively we could read directly from LD here but this seemed more flexible. Any thoughts?

pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -2753,20 +2753,7 @@ func (w wrapper) GetSpanRuns(ctx context.Context, opt cqrs.GetTraceRunOpt) ([]*c
...
-					case enums.StepStatusRunning:
-						status = enums.RunStatusRunning
-					case enums.StepStatusCancelled:
-						status = enums.RunStatusCancelled
-					case enums.StepStatusTimedOut:
-						status = enums.RunStatusCancelled // Map timeout to cancelled
-					case enums.StepStatusScheduled, enums.StepStatusWaiting, enums.StepStatusSleeping, enums.StepStatusInvoking:
-						status = enums.RunStatusRunning // These are all "in progress" states
-					}
+					status = StepStatusToRunStatus(stepStatus)
```
albertchae, line 2756: small refactor so we can reuse this enum conversion. It looks like the switch technically wasn't exhaustive, so I added a default case to return `enums.RunStatusRunning` which is also the zero value of RunStatus

KiKoS0 (approved):

pkg/cqrs/base_cqrs/cqrs.go
jpwilliams: For posterity, we really should get an "Unknown" status as the zero value. It's not supported across the UI yet, but I'd rather have it marked as "Unknown" rather than looking like it's hung in the UI later.

albertchae: it looks like we already have Unknown although unfortunately it's not the zero value and the comments above this enum definition seem to indicate we probably can't change it.
 https://github.com/inngest/inngest/blob/ebe71354322d1c9e30ff83e7308114c9e8ed3967/pkg/enums/run_status.go#L36-L39
 
 updated the default case to return Unknown

pkg/api/apiv1/events.go
albertchae: for posterity, we don't have LD set up in this codebase yet so kept the query param approach

PR #3120, opened 2025-10-14
fix: Unwrap extra StandardError wrapping

rhino1998:
## Description

This removes an extra layer of `{"error": ...}` wrapping from StandardError serialization in driver responses.

Before
image

After
image

## Motivation

StandardErrors were not appearing correctly in traces and were showing as raw json.

https://linear.app/inngest/issue/EXE-123/check-that-permanent-failures-are-displayed-with-new-tracing

Also fixes 
EXE-196

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/driver_response.go
+++ b/pkg/execution/state/driver_response.go
@@ -412,6 +412,10 @@ func (r *DriverResponse) GetFunctionOutput() (*string, error) {
 }
 
 func isWrappedError(maybeErr []byte) bool {
+	if len(maybeErr) == 0 || maybeErr[0] != '{' {
+		return false
+	}
+
 	// Unmarshal into a struct to check if it's already wrapped.
 	// We don't care about the full structure, just whether it has
 	// the right fields.


pkg/execution/state/driver_response.go
```diff
@@ -386,26 +386,49 @@ func (r *DriverResponse) GetFunctionOutput() (*string, error) {
...
+		keyedOutput = &s
 	}
-	s := string(keyedByt)
-	keyedOutput = &s
 
 	return keyedOutput, nil
 }
 
+func isWrappedError(maybeErr []byte) bool {
+	// Unmarshal into a struct to check if it's already wrapped.
```
tonyhb, line 419: Could we also do a fast check here to see:

```
if len(maybeErr) == 0 || maybeErr[1] != '{' {
  return false
}
```

This will always short-circuit if the response isn't valid JSON - which is minor, but saves some CPU cycles when we do this very, very often.

pkg/execution/state/driver_response.go
```diff
@@ -386,26 +386,49 @@ func (r *DriverResponse) GetFunctionOutput() (*string, error) {
...
-	// Now we have the output, we make sure it's keyed the same as regular step
-	// outputs are, either under `data` or `error`.
 	var keyedOutput *string
-	key := "data"
-	if r.Error() != "" {
-		key = "error"
-	}
+	if isWrappedError([]byte(*output)) {
+		// Error is already wrapped, return as-is.
+		keyedOutput = output
```
tonyhb: I know this is a nit, but over time this stacks up and makes a difference....

Let's just return &output here directly so that we don't have logic in an else.  This reduces nesting and makes code complexity simpler!

rhino1998: [9a7115a](https://github.com/inngest/inngest/pull/3120/commits/9a7115a9c089c8032e2eb3b1e51cd0dbfefd98e7)

pkg/execution/state/driver_response.go
rhino1998: [66d1b44](https://github.com/inngest/inngest/pull/3120/commits/66d1b4451ba459b5984712e77c6dd8935f63f67b)

jpwilliams (approved):
Great! 🙌 

Noting that the detection we do here always feels scary to me; what if the user has returned a successful data object of `{"error":...}`?

We're mostly protected by the code paths that get us to these checks, but I'd like to see this kind of code where we're forcibly wrapping/unwrapping/detecting things in the data to disappear in the future!

PR #3122, opened 2025-10-14
[Security]: Bump pnpm 8 to pnpm 10 and add minimumReleaseAge setting

doverdbng:
## Description

The primary purpose of this PR is to introduce pnpm's new [`minimumReleaseAge`](https://pnpm.io/settings#minimumreleaseage) setting to counteract supply chain attack risk.

Here is an example of an error generating from attempting to install a package that violates the `minimumReleaseAge` setting:

```
ERR_PNPM_NO_MATCHING_VERSION 

No matching version found for react@19.2.0 published by Wed Sep 24 2025 22:11:07 GMT-0500 (Central Daylight Time) while fetching it from https://registry.npmjs.org/.

Version 19.2.0 satisfies the specs but was released at Wed Oct 01 2025 16:38:32 GMT-0500 (Central Daylight Time)
```

In order to have access to this new setting, `pnpm` had to be upgraded to v10. Specifically, this PR bumps from `pnpm@8.15.8` to `pnpm@10.18.2`. An analysis of the changelog entries between these two versions is available in the [ticket](https://linear.app/inngest/issue/EXE-585/investigate-usingstandardizing-on-minimumreleaseage-dependency).

---

**NOTE:** This PR sets `minimumReleaseAge` to `1440` minutes, i.e. 1 day. This is the example value in the docs, but a reasonable argument could be made to increase or decrease it. It's a tradeoff between making it a high enough value that a malicious version would likely have been caught and removed from the registry versus making sure it's not overly frustrating to attempt to update packages.

**NOTE x2:** Reference [this comment](https://github.com/inngest/inngest/pull/3122#discussion_r2430570759) for a discussion about the new approval model for dependency lifecycle scripts and the opinions of this PR.

**NOTE x3:** Reference [this comment](https://github.com/inngest/inngest/pull/3122#discussion_r2430563382) for a discussion about the new `enable-pre-post-scripts` and the specific choice in this PR that _**diverges from default behavior**_.

**NOTE x4:** After this bump, pnpm no longer respects `NODE_ENV=production` as an indication to not install dev dependencies. I tried adding `--prod` to most of our installs in CI and they broke--it looks like we normally don't do this, since things like linting require them. I _assume_ Vercel will handle pruning correctly when it comes to actual deployments.

**NOTE x5:** Unless we disable it, pnpm will now manage the version of pnpm based on the `packageManager` field similarly to how corepack has historically done it. We can turn this off via the `manage-packer-manager-versions` setting if desired.

**Follow Up Issue:** https://linear.app/inngest/issue/EXE-676/pin-pnpm-with-integrity-sha-in-packagemanager-and-ci

## Motivation

The goal is to reduce the likelihood of exposure to a supply chain attack by making sure that very recently published packages cannot be installed.

## Review guidance

I've added numerous comments to the PR and analyzed the changelog diff in the [ticket](https://linear.app/inngest/issue/EXE-585/investigate-usingstandardizing-on-minimumreleaseage-dependency) itself.

- [x] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -760,11 +760,13 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		runSpanRef       *tracing.DroppableSpan
 		discoverySpanRef *tracing.DroppableSpan
 	)
-	defer func() {
-		// Always attempt to send the spans. They may have already been dropped
-		// (e.g. if the queue item already exists), in which case `Send()` will
-		// no-op
 
+	// Send spans to the history store (ClickHouse). If not called, we'll drop
+	// the spans and not send them. There's a variety of scenarios where the run
+	// ends up not scheduling so we don't want to add it to the history store.
+	// Some scenarios are happy path (e.g.  queue idempotency) and some are sad
+	// path (e.g. Executor borked)
+	sendSpans := func() {
 		if runSpanRef != nil {
 			err := runSpanRef.Send()
 			if err != nil {
@@ -786,11 +788,10 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 				)
 			}
 		}
-	}()
+	}
 
-	// We may need to drop the spans later, like if the queue item already
-	// exists
-	dropSpans := func() {
+	// Handle span dropping. The drops will be noops if the spans were sent
+	defer func() {
 		if runSpanRef != nil {
 			runSpanRef.Drop()
 		}
@@ -798,7 +799,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		if discoverySpanRef != nil {
 			discoverySpanRef.Drop()
 		}
-	}
+	}()
 
 	status := enums.StepStatusQueued
 	if req.SkipReason() != enums.SkipReasonNone {
@@ -828,6 +829,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 
 	// If this is paused, immediately end just before creating state.
 	if skipped := req.SkipReason(); skipped != enums.SkipReasonNone {
+		sendSpans()
 		return e.handleFunctionSkipped(ctx, req, metadata, evts, skipped)
 	}
 
@@ -889,9 +891,6 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 						l.ReportError(err, "error canceling singleton run")
 					}
 				default:
-					// If the event is being rejected because there is another singleton run already in progress,
-					//  we can safely ignore the span for this schedule request
-					dropSpans()
 					// Immediately end before creating state
 					return nil, ErrFunctionSkipped
 				}
@@ -1047,8 +1046,6 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		// If the item already exists in the queue, we can safely ignore this
 		// entire schedule request; it's basically a retry and we should not
 		// persist this for the user.
-		dropSpans()
-
 		return nil, state.ErrIdentifierExists
 
 	case redis_state.ErrQueueItemSingletonExists:
@@ -1056,15 +1053,13 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		if err != nil {
 			l.ReportError(err, "error deleting function state")
 		}
-		// If the event is being rejected because there is another singleton run already in progress,
-		// we can safely ignore the span for this schedule request
-		dropSpans()
 		return nil, ErrFunctionSkipped
 
 	default:
 		return nil, fmt.Errorf("error enqueueing source edge '%v': %w", queueKey, err)
 	}
 
+	sendSpans()
 	for _, e := range e.lifecycles {
 		go e.OnFunctionScheduled(context.WithoutCancel(ctx), metadata, item, req.Events)
 	}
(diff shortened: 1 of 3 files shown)


tests/js/.npmrc
```diff
@@ -0,0 +1,2 @@
+enable-pre-post-scripts=false
```
doverdbng, line 1: In my testing, this is consistent with the behavior we had with pnpm 8. The default value for this is now `true`...which matches the npm default but seems strange to me. This being set to `true` would imply that running the script `foo` would also run `prefoo` and `postfoo` if they exist. I think not having that magic is a good thing?

ui/.npmrc
```diff
@@ -1 +1,3 @@
+enable-pre-post-scripts=false
```
doverdbng, line 1: In my testing, this is consistent with the behavior we had with pnpm 8. The default value for this is now true...which matches the npm default but seems strange to me. This being set to true would imply that running the script foo would also run prefoo and postfoo if they exist. I think not having that magic is a good thing?

ui/.npmrc
```diff
@@ -1 +1,3 @@
...
+minimum-release-age=1440
```
doverdbng, line 2: I don't have a strong opinion on the correct value here. This is the one in their docs. It will block any dependency versions published in the last 24 hours.

tests/js/.npmrc
```diff
@@ -0,0 +1,2 @@
...
+minimum-release-age=1440
```
doverdbng, line 2: I don't have a strong opinion on the correct value here. This is the one in their docs. It will block any dependency versions published in the last 24 hours.

ui/package.json
```diff
@@ -81,6 +81,9 @@
       "nodemon": ">=3.0.0",
       "nodemon>minimatch": "3.1.2",
       "nodemon>brace-expansion": "1.1.11"
-    }
+    },
+    "onlyBuiltDependencies": [
```
doverdbng, line 85: With the bump to v10, pnpm no longer runs lifecycle scripts for dependencies by default. `@sentry/cli`, `@clerk/shared`, `core-js-pure`, and `esbuild` all want to be able to run lifecycle scripts. Note that, prior to this bump, all of these packages have been permitted to run lifecycle scripts on install and have been doing so.

Based on referencing a recent run without these changes, here are the dependency lifecycle scripts that have been running:

```
.../node_modules/@sentry/cli install$ node ./scripts/install.js
.../node_modules/@clerk/shared postinstall$ node ./scripts/postinstall.mjs
.../node_modules/core-js-pure postinstall$ node -e "try{require('./postinstall')}catch(e){}"
.../esbuild@0.25.2/node_modules/esbuild postinstall$ node install.js
```

---

`.../node_modules/@sentry/cli install$ node ./scripts/install.js`

[[source]](https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js) `@sentry/cli` is **approved** because it will otherwise just fetch the same binary during build time anyway.

---

`.../node_modules/@clerk/shared postinstall$ node ./scripts/postinstall.mjs`

[[source]](https://github.com/clerk/javascript/blob/1236c745fd58020e0972938ca0a9ae697a24af02/packages/shared/scripts/postinstall.mjs) This seems to just print out a warning about clerk collecting telemetry in dev. Unless it becomes a problem, it can likely be **rejected**.

---

`.../node_modules/core-js-pure postinstall$ node -e "try{require('./postinstall')}catch(e){}"`

[[source]](https://github.com/zloirock/core-js/blob/146178f437c5b71b4fbdc22b02b69e959a59f767/packages/core-js/postinstall.js) This seems to just print out a request for donating. Unless it becomes a problem, it can likely be **rejected**.

---

`.../esbuild@0.25.2/node_modules/esbuild postinstall$ node install.js`

[[source]](https://github.com/evanw/esbuild/blob/6b7c4f2dcbcaa1238bca0f7a4a1d95918296c82e/lib/npm/node-platform.ts#L57) This one makes sure the correct binary is available for the platform. It's probably unnecessary, strictly speaking, because pnpm should be able to figure this all out through `optionalDependencies` and other means. However, it's probably reasonable to **approve** this `postinstall` given the core nature of the `esbuild` dependency and the existence of our lockfile and `minimumReleaseAge` approaches to mitigate risk, especially for high visibility dependencies.

doverdbng:
Another question. A message like the following has been printing out on installs for as long as I recall:

```
The field "pnpm.overrides" was found in <...>/package.json.
This will not take effect.
You should configure "pnpm.overrides" at the root of the workspace instead.
```

Does anyone have background here? From looking at the generated lockfiles, it does seem like (before and after this pnpm bump) that the overrides have been respected, but, for some reason, this message still gets printed out.

doverdbng:
One more note: it looks like we should always get `--frozen-lockfile` behavior on CI (and via Vercel as a result), provided they're correctly setting the `CI` environment variable. However, we could consider overriding the install command to `pnpm install --frozen-lockfile` explicitly to make sure of this.

.github/workflows/release.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: 8.15.8
+          version: 10.18.2
```
doverdbng, line 28: Not sure if I should touch this or not? `10.18.2` all the things? Also not sure if there's a way to test it practically.

jacobheric (approved):
This looks good to me. I've verified that clerk and sentry work as expected when deployed from this so I think we are ok with regard to the lifecycle script changes. And, I eyeballed the bundle size in vercel for the branch and it's nearly identical to our current prod deploy.

PR #3128, opened 2025-10-15
Fix Schedule missing some necessary span drop spots

amh4r:
## Description
Fix an issue where we weren't dropping spans when we don't actually enqueue the function run.

We'd previously fixed an issue where we didn't drop spans when hitting queue idempotency (#3115). However, there are more code paths where we don't schedule.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -1059,8 +1059,8 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) 
 		return nil, fmt.Errorf("error enqueueing source edge '%v': %w", queueKey, err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/execution/executor/executor.go
```diff
@@ -1063,6 +1060,7 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 	}
 
 	for _, e := range e.lifecycles {
+		sendSpans()
```
jpwilliams: It won't send many times, but can we move this outside of the loop so it's not happening many times?

amh4r: Good call

PR #3126, opened 2025-10-15
Add new capacity manager interface

BrunoScheufler:
## Description

Supersedes https://github.com/inngest/inngest/pull/3125 by separating changes

This PR is fully isolated from the current system.

## Type of change (choose one)
- [ ] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [ ] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #3142, opened 2025-10-17
Constraint API execution changes

BrunoScheufler:
## Description

Supersedes https://github.com/inngest/inngest/pull/3127 which accidentally got merged into the stacked base branch (oops).

This PR integrates the new capacity manager interface with Schedule.

This PR is stacked on top of https://github.com/inngest/inngest/pull/3126.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/constraints.go
+++ b/pkg/execution/state/redis_state/constraints.go
@@ -135,7 +135,7 @@ func (q *queue) backlogRefillConstraintCheck(
 			return nil, fmt.Errorf("could not enforce constraints and acquire lease: %w", err)
 		}
 
-		// Attempt to fall back to BacklogRefill -- ignore GCRA with fallbackIdempofallbackIdempotencyKey
+		// Attempt to fall back to BacklogRefill -- ignore GCRA with fallbackIdempotencyKey
 		return &backlogRefillConstraintCheckResult{
 			itemsToRefill:          itemIDs,
 			fallbackIdempotencyKey: idempotencyKey,


pkg/execution/executor/executor.go
```diff
@@ -577,9 +596,29 @@ func (e *executor) createEagerCancellationForTimeout(ctx context.Context, since
...
+// Execute loads a workflow and the current run state, then executes the
+// function's step via the necessary driver.
+//
 // If this function has a debounce config, this will return ErrFunctionDebounced instead
 // of an identifier as the function is not scheduled immediately.
-func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest) (*sv2.Metadata, error) {
+func (e *executor) schedule(
+	ctx context.Context,
+	req execution.ScheduleRequest,
+	performChecks bool,
```
lkasinathan, line 654: nit: Can we document what performChecks and fallbackIdempotencyKey are here?

pkg/execution/state/redis_state/queue_processor.go
```diff
@@ -849,6 +850,23 @@ func (q *queue) scanContinuations(ctx context.Context) error {
 // randomOffset allows us to peek jobs out-of-order, and occurs when we hit concurrency key issues
 // such that we can attempt to work on other jobs not blocked by heading concurrency key issues.
 func (q *queue) processPartition(ctx context.Context, p *QueuePartition, continuationCount uint, randomOffset bool) error {
+	if p.AccountID != uuid.Nil && q.capacityManager != nil && q.useConstraintAPI != nil {
```
BrunoScheufler: Context: Previously, we checked account, function concurrency in PartitionLease. If we didn't have sufficient capacity, we requeued the partition a couple seconds in the future.

With the Constraint API, we can perform the same checks, without leasing capacity yet. Since we haven't peeked items yet, leasing like in BacklogRefill isn't possible.

Ultimately, we want to move to key queues, where backlogs are refilled. 

~~In that case, partition lease should not perform checks because items have already been checked. Running constraint checks before leasing the partition could lead to long delays for low concurrency limits.~~ Edit: Refills update active sets, not in progress sets. We can still check concurrency (in progress) in partition lease. BUT: With the Constraint API, concurrency capacity is already consumed. We cannot run the same check again.

So assuming the Constraint API is ready, do we really want to run this check?
- This is another roundtrip in the critical path (every time we peek a partition, we need to run this)
- This won't work with refilled items with leases anyway (unnecessary to check)
- We don't know the dynamic keys, so we can only check account/function concurrency
- We can't acquire a lease yet since we don't have the peeked items here

What would be the benefits of checking here?
- If there's no capacity, we can simply requeue
- If there's reduced capacity, we can reduce the peek size (number of items)

What happens if we don't do this?
- We simply lease the partition (partitionLease gets much more simple)
- We peek items
- For each item, if no capacity lease is present, we acquire a capacity lease
  - These roundtrips may be batched locally (this can be invisible to the queue and an implementation detail in the interface chain)

pkg/execution/executor/executor.go
BrunoScheufler: Done!

PR #3145, opened 2025-10-17
Fix runs page filter logic

AndyInternet:
## Description

Bug in runs page causes unexpected results when filters are applied and multiple pages are loaded.

The bug occurred when the app was trying to manage two separate data requests that could get out of sync. When you changed a filter while scrolled down, it would mix old unfiltered results with new filtered results. The fix was to combine those two requests into one simple hook that automatically clears all the old data whenever you change a filter, so you always see consistent results.

## Motivation
Bug bash Oct 2025

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/hooks/useSearchParam.ts
+++ b/ui/packages/components/src/hooks/useSearchParam.ts
@@ -118,18 +118,21 @@ export function useStringArraySearchParam(
     router.replace((pathname + '?' + params.toString()) as Route);
   }, [name, pathname, router, searchParams]);
 
-  let value = undefined;
   const rawValue = searchParams.get(name);
-  if (typeof rawValue === 'string') {
-    const parsed: unknown = JSON.parse(rawValue);
+  const value = useMemo(() => {
+    let value = undefined;
+    if (typeof rawValue === 'string') {
+      const parsed: unknown = JSON.parse(rawValue);
 
-    if (isStringArray(parsed)) {
-      value = parsed;
-    } else {
-      // This means the query param value is the wrong type
-      console.error(`invalid type for search param ${name}`);
+      if (isStringArray(parsed)) {
+        value = parsed;
+      } else {
+        // This means the query param value is the wrong type
+        console.error(`invalid type for search param ${name}`);
+      }
     }
-  }
+    return value;
+  }, [rawValue]);
 
   return [value, upsert, remove];
 }


ui/apps/dashboard/src/components/Runs/Runs.tsx
```diff
@@ -93,56 +90,50 @@ export const Runs = forwardRef<RefreshRunsRef, Props>(function Runs(
...
+      environmentID: environment.id,
+      functionSlug: functionSlug ?? null,
+      startTime: calculatedStartTime.toISOString(),
+      endTime: endTime ?? null,
+      status: filteredStatus.length > 0 ? filteredStatus : null,
+      timeField,
+      celQuery: search,
+    }),
+    [
+      appIDs,
```
amh4r, line 105: I think `appIDs` will be different on every render because I didn't do a good job with `useStringArraySearchParam`. We might need this change too:

```diff
diff --git a/ui/packages/components/src/hooks/useSearchParam.ts b/ui/packages/components/src/hooks/useSearchParam.ts
index 08b453194..8e23a91fa 100644
--- a/ui/packages/components/src/hooks/useSearchParam.ts
+++ b/ui/packages/components/src/hooks/useSearchParam.ts
@@ -118,18 +118,21 @@ export function useStringArraySearchParam(
     router.replace((pathname + '?' + params.toString()) as Route);
   }, [name, pathname, router, searchParams]);
 
-  let value = undefined;
   const rawValue = searchParams.get(name);
-  if (typeof rawValue === 'string') {
-    const parsed: unknown = JSON.parse(rawValue);
-
-    if (isStringArray(parsed)) {
-      value = parsed;
-    } else {
-      // This means the query param value is the wrong type
-      console.error(`invalid type for search param ${name}`);
+  const value = useMemo(() => {
+    let value = undefined;
+    if (typeof rawValue === 'string') {
+      const parsed: unknown = JSON.parse(rawValue);
+
+      if (isStringArray(parsed)) {
+        value = parsed;
+      } else {
+        // This means the query param value is the wrong type
+        console.error(`invalid type for search param ${name}`);
+      }
     }
-  }
+    return value;
+  }, [rawValue]);
 
   return [value, upsert, remove];
 }
```

What are your thoughts?

AndyInternet: I think you're right. I can make this change.

PR #3181, opened 2025-10-23
Backlog refill items peeked ahead of time

BrunoScheufler:
## Description

This PR updates BacklogRefill to support the upcoming Constraint API

### Motivation

- With the Constraint API, we'll add capacity leases
- For key queues, we will check constraints in one `Acquire` API call and create one lease per queue item to be refilled
- To ensure items properly handle idempotency in case the capacity lease expires before processing starts, we need to set a proper idempotency key
- The most suitable idempotency key is the item ID
- Acquire runs before BacklogRefill
- Currently, BacklogRefill does not know the items to be refilled ahead of time -- we peek within backlogRefill.lua

The proposal is to sequentially peek items to be refilled, then pass the IDs to BacklogRefill.

This should not change any guarantees
- Only one worker can process the backlog for refills. This guarantees exclusivity for sequentially refilling backlogs from the earliest items.
- There may be concurrent edits to the backlog: Some items may be dequeued, other items may be added (with earlier scores)
- This means extracting a peek before BacklogRefill can cause some race conditions
   - Items peeked initially may be deleted by the time we get to BacklogRefill. This can be mitigated by checking for existence (which we already do) before refilling items to the ready queue and ignoring missing items
   - The other question is FIFO guarantees: What if items are inserted inbetween existing items? I would argue that for strict FIFO, you can only append to the queue. Adding items with earlier timestamps should not fall under strict FIFO.

### Changes

- Instead of passing an upper bound of items to refill, we now perform a peek before BacklogRefill and pass the IDs of peeked items into BacklogRefill
- BacklogRefill will continue to check constraints, ignore missing items, and update state as before

The only behavioral difference after this PR is that instead of having backlogRefill.lua peek items, we now do this in a separate step.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/backlog.go
+++ b/pkg/execution/state/redis_state/backlog.go
@@ -41,7 +41,6 @@ type BacklogManager interface {
 	ItemShadowPartition(ctx context.Context, i osqueue.QueueItem) QueueShadowPartition
 }
 
-
 type PartitionConstraintConfig struct {
 	FunctionVersion int `json:"fv,omitempty,omitzero"`
 
@@ -938,6 +937,9 @@ func (q *queue) BacklogPeek(ctx context.Context, b *QueueBacklog, from time.Time
 	return q.backlogPeek(ctx, b, from, until, limit, opts...)
 }
 
+// backlogPeek peeks item from the given backlog.
+//
+// Pointers to missing items will be removed from the backlog.
 func (q *queue) backlogPeek(ctx context.Context, b *QueueBacklog, from time.Time, until time.Time, limit int64, opts ...PeekOpt) ([]*osqueue.QueueItem, int, error) {
 	ctx = redis_telemetry.WithScope(redis_telemetry.WithOpName(ctx, "backlogPeek"), redis_telemetry.ScopeQueue)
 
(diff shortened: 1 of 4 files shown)


pkg/execution/state/redis_state/backlog.go
```diff
@@ -595,7 +604,7 @@ type BacklogRefillResult struct {
 	RetryAt           time.Time
 }
 
-func (q *queue) BacklogRefill(ctx context.Context, b *QueueBacklog, sp *QueueShadowPartition, refillUntil time.Time, latestConstraints PartitionConstraintConfig) (*BacklogRefillResult, error) {
+func (q *queue) BacklogRefill(ctx context.Context, b *QueueBacklog, sp *QueueShadowPartition, refillUntil time.Time, refillItems []string, latestConstraints PartitionConstraintConfig) (*BacklogRefillResult, error) {
```
darwin67, line 606: the list of arguments is getting long. okay for now, but I'd like us to keep a hard limit of 6 args for a function as a code guideline.

nothing solifiied yet, but just sharind ahead of time.

darwin67 (approved):

lkasinathan:
Test failures _seem_ relevant, PTAL before merging.

pkg/execution/state/redis_state/backlog.go
BrunoScheufler: Agree!

BrunoScheufler: We should probably move to options, I'm mostly thinking how to avoid forgetting required args

BrunoScheufler:
> Test failures _seem_ relevant, PTAL before merging.

Thanks for highlighting, fixed in 5d5f60572d9668b551aa4af0cf71383d34d450b9

This happened because peek is filtering out missing items, which some tests were expecting BacklogRefill to gracefully ignore/clean up.

PR #3194, opened 2025-10-24
Dev Server MCP

jakobevangelista:
## Description

Allows send event, list events, poll for event status, as well as embed the docs into the dev server so the LLM has the ability to search the docs

The docs are copy and pasted into the repo, the core changes are in `mcp.go`.

I mainly took this direction because I wanted grepping the docs to be fast, an alternative would be to host an api for grepping and reading docs, which we could do as well. But introduces latency with network calls. But since you have to be online to be talking with an LLM, for the most part, not a bad idea to host an api for this

`internal/embeddeddocs` is just our docs copy and pasted, need to think more about how to keep this updated, but should be fine for now

The vendor is most of line additions, the main changes are in `pkg/devserver/mcp.go`

Growth [PR](https://github.com/inngest/growth/pull/108)

## Motivation
Users are able to have a seamless experience building with LLMs

- [ ] Docs

## Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,6 +26,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v5
@@ -95,6 +97,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v5
@@ -147,6 +151,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v5
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -21,6 +21,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v5
@@ -38,6 +40,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
       - uses: DeterminateSystems/nix-installer-action@main
       - uses: nicknovitski/nix-develop@v1.2.1
       - name: 'Check for dirty Git state after running "make gen protobuf"'
@@ -56,6 +60,8 @@ jobs:
     name: "Test / OS: ${{ matrix.os }}, key-queues: ${{ matrix.experimentalKeyQueues }}"
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Go
         uses: actions/setup-go@v5
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          submodules: true
       - name: Install Go
         uses: actions/setup-go@v5
         with:
@@ -67,6 +68,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          submodules: true
       - name: Set up Node/npm
         uses: actions/setup-node@v4
         with:


pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+	}
+
+	// Parse the event ID to get run IDs
+	eventULID, err := ulid.Parse(evtID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse event ID: %w", err)
+	}
+
+	// Wait a moment for the event to be processed and runs to be created
+	time.Sleep(eventProcessingDelay)
```
jpwilliams: This is flimsy, but it makes sense. 😄

I guess what would be nice for you here would be an API that tells you when an event has been processed so you could at least poll that.

For now, the Dev Server has a "tick" that dictates how often the queue will poll for available partitions; we could use that (multipled by a few times) to be a little more confident that we've grabbed the job.

The tick is available here, above where you create the MCP route, so something like `opts.Tick * 3` should be fine.

https://github.com/inngest/inngest/blob/dd362337fe8b7f9bae10c1fdca72f77f8f6a2ee9/pkg/devserver/devserver.go#L176-L178

pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+	case string:
+		var parsed map[string]any
+		if err := json.Unmarshal([]byte(data), &parsed); err == nil {
+			return parsed
+		}
+		return map[string]any{fallbackKey: data}
+	default:
+		return map[string]any{fallbackKey: input}
+	}
+}
```
jpwilliams, line 153: Is it expected that this happens? I imagine that this isn't what the user would actually want if they tried to send an event and it just came through as one big string in a key?

pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+
+// pollRunStatus handles the poll_run_status tool
+func (h *MCPHandler) pollRunStatus(ctx context.Context, req *mcp.CallToolRequest, args PollRunStatusArgs) (*mcp.CallToolResult, any, error) {
+	// Track MCP tool usage
+	metadata := tel.NewMetadata(ctx)
+	metadata.Context["tool"] = "poll_run_status"
+	metadata.Context["run_count"] = len(args.RunIDs)
+	metadata.Context["timeout"] = args.Timeout
+	metadata.Context["poll_interval"] = args.PollInterval
+	tel.SendEvent(ctx, "cli/mcp.tool.executed", metadata)
```
jpwilliams, line 664: Oh hell yeah 🥳

pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+// NewMCPHandler creates a new MCP handler for the dev server
+func NewMCPHandler(events api.EventHandler, data cqrs.Manager) http.Handler {
+	h := &MCPHandler{
+		events: events,
+		data:   data,
+	}
+
+	// Create a streamable HTTP handler that returns the same server for all requests
+	return mcp.NewStreamableHTTPHandler(func(r *http.Request) *mcp.Server {
+		return h.createMCPServer()
```
jpwilliams: Is this creating a new MCP server for every request? Could we use `sync.Once()` or create this server outside of this function?

pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+		Content: []mcp.Content{
+			&mcp.TextContent{
+				Text: text,
+			},
+		},
+	}, result, nil
+}
+
+// listDocs handles the list_docs tool
+func (h *MCPHandler) listDocs(ctx context.Context, req *mcp.CallToolRequest, _ struct{}) (*mcp.CallToolResult, any, error) {
```
jpwilliams, line 1076: Could this be done once and then cached?

pkg/devserver/mcp.go
```diff
@@ -0,0 +1,1096 @@
...
+		Content: []mcp.Content{
+			&mcp.TextContent{
+				Text: text,
+			},
+		},
+	}, result, nil
+}
+
+// readDoc handles the read_doc tool
+func (h *MCPHandler) readDoc(ctx context.Context, req *mcp.CallToolRequest, args ReadDocArgs) (*mcp.CallToolResult, any, error) {
```
jpwilliams, line 1013: Could this be done once per `args.Path` and then cached?

jpwilliams (changes requested):
I wonder if we could submodule `internal/embeddocs` instead of having it copy/pasted? Even that's not ideal, but at least then we could see what commit we're on etc.

I'd love to see the tick rate fix before we ship! Everything else can be iteration!

pkg/devserver/mcp.go
jakobevangelista: https://github.com/inngest/inngest/pull/3194/commits/16ad8c00520ef45ba0c84cd53076a66b9b782f27

pkg/devserver/mcp.go
jakobevangelista: https://github.com/inngest/inngest/pull/3194/commits/16f8b7fdb54a05159ebe7c81ff85c5679c19a6f3

pkg/devserver/mcp.go
jakobevangelista: https://github.com/inngest/inngest/pull/3194/commits/b0faf1d4a3c9d32025da29f80251ba25d1e8124b

pkg/devserver/mcp.go
jakobevangelista: https://github.com/inngest/inngest/pull/3194/commits/b0faf1d4a3c9d32025da29f80251ba25d1e8124b

PR #3196, opened 2025-10-25
[SYS-370] Add Connect Worker Concurrency

mar-inngest:
## Description

Adding support for connect worker concurrency

- [ ] Docs

## Checklist

## New Redis Keys and TTLs
All keys have TTLs and Expiration, so in the worst case they should be deleted within 4*20 seconds by redis.

* `{<env_id>}:worker-capacity:<instance_id>`: Integer value, only set where worker capacity is limited.
* `{<env_id>}:worker-requests-set:<instance_id>`: Sorted set of strings value with expiration timestamp as score, only set where worker capacity is limited.
*  `{<env_id>}:request-worker:<lease_id>`: String value, only set where worker capacity is limited.
*  The `ConnectWorkerRequestToWorkerMappingTTL` is the TTL/Expiration of `%s:request-worker:%s` and the data inside `%s:worker-requests-set:%s`
* `ConnectWorkerCapacityManagerTTL` is the TTL value of `%s:worker-capacity:%s` and the `%s:worker-requests-set:%s` itself

## Behaviour
* Concurrency is associated with an `InstanceID`
* Concurrency is associated to a `request_id`, not a `lease_id`. Each request usually has multiple leases. There are two TTLs. 
  *  15 minutes - worker's total capacity and available capacity is stored 
  * 2 minutes - A request's mapping to a worker is stored (20 second is the duration for lease renewal, and 5 second is the time for heart beats). The 2 minutes is mostly to allow for gateway restarts, even though the worker should send a new connection request.
* No-capacity-limit is considered the faster/default path, all the capacity checks are generally performed after lease (by the executor). 
* On concurrency limit, we use the Always Retryable Error interface to keep retrying. This means that a job is showed as "Running" in the UI, always.

## Cases Tested
* Two workers with same Instance ID share the capacity between them
  * If one crashes, lease is transferred to other worker since from the perspective of this pull request, there's no change
* Long Running Worker With Capacity Enabled - Heartbeat Messages refresh TTL. Capacity is cleaned on next assign operation.
* Worker Crashes
* Gateway Crashes
* Multiple leases contending for the same Instance ID (gracefully fails, and returns a retryable error)

### DIFF
--- a/pkg/connect/routing/router.go
+++ b/pkg/connect/routing/router.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"slices"
 	"time"
 
@@ -407,6 +408,20 @@ type checkCapacityRes struct {
 	hasWorkerCapacity bool
 }
 
+type activeLeasesLogger struct {
+	stateManager       state.StateManager
+	ctx                context.Context
+	envID              uuid.UUID
+	instanceID         string
+	workerCapUnlimited bool
+}
+
+// this is a redis call that we only want enabled at trace level
+func (a activeLeasesLogger) LogValue() slog.Value {
+	activeLeases, _ := a.stateManager.GetAllActiveWorkerRequests(a.ctx, a.envID, a.instanceID, a.workerCapUnlimited)
+	return slog.AnyValue(activeLeases)
+}
+
 func checkCapacity(ctx context.Context, stateManager state.StateManager, envID uuid.UUID, conn *connectpb.ConnMetadata, log logger.Logger) *checkCapacityRes {
 
 	// Check worker capacity
@@ -424,11 +439,9 @@ func checkCapacity(ctx context.Context, stateManager state.StateManager, envID u
 	}
 	if workerCap.IsAtCapacity() {
 		// Worker has a capacity limit set and is at capacity
-		if log.Level() <= logger.LevelTrace {
-			// get active leases only if trace level is enabled
-			activeLeases, _ := stateManager.GetAllActiveWorkerRequests(ctx, envID, conn.InstanceId, workerCap.IsUnlimited())
-			log.Trace("worker at capacity", "instance_id", conn.InstanceId, "worker_total_capacity", workerCap.Total, "worker_available_capacity", workerCap.Available, "worker_active_leases", activeLeases)
-		}
+		log.Trace("worker at capacity", "instance_id", conn.InstanceId, "worker_total_capacity", workerCap.Total, "worker_available_capacity", workerCap.Available,
+			"worker_active_leases", activeLeasesLogger{stateManager: stateManager, ctx: ctx, envID: envID, instanceID: conn.InstanceId, workerCapUnlimited: workerCap.IsUnlimited()})
+
 		return &checkCapacityRes{hasWorkerCapacity: false}
 	}
 


pkg/connect/state/lua/extend_lease.lua
```diff
@@ -32,9 +37,58 @@ end
 -- If new lease expiry is in the past, drop the lease
 if decode_ulid_time(newLeaseID) - currentTime <= 0 then
 	redis.call("DEL", keyRequestLease)
+	-- If worker capacity is limited, also clean up the counter and mapping
+	if isWorkerCapacityLimited == "true" then
```
mar-inngest: TODO: Refactor this

pkg/connect/grpc/proxy.go
```diff
@@ -377,6 +378,30 @@ func (i *grpcConnector) Proxy(ctx, traceCtx context.Context, opts ProxyOpts) (*c
 			return nil, fmt.Errorf("failed to route message: %w", err)
 		}
 
+		// Get connection metadata to retrieve instance ID for capacity tracking
+		conn, err := i.stateManager.GetConnection(ctx, opts.EnvID, route.ConnectionID)
```
mar-inngest: Use `route` to return the instanceId

pkg/connect/grpc/proxy.go
```diff
@@ -377,6 +378,30 @@ func (i *grpcConnector) Proxy(ctx, traceCtx context.Context, opts ProxyOpts) (*c
...
+		if err != nil {
+			span.RecordError(err)
+			l.ReportError(err, "could not get connection metadata")
+		} else if conn != nil && conn.InstanceId != "" {
```
BrunoScheufler: Do we want to return the error in case GetConnection doesn't work? Or do we actually want to continue here?

e.g. 
```suggestion
		if err != nil {
			span.RecordError(err)
			l.ReportError(err, "could not get connection metadata")
			return fmt.Errorf("could not retrieve connection to route request to: %w", err)
		
		if conn != nil && conn.InstanceId != "" {
```

Also, should the information you need not be in the `route`? We might be able to modify that to match, unless you need specific attributes! That way we could save the additional roundtrip!

mar-inngest: Yeah, I had a comment about it above :D

pkg/connect/state/request.go
```diff
@@ -111,6 +113,9 @@ func (r *redisConnectionStateManager) ExtendRequestLease(ctx context.Context, en
 		newLeaseID.String(),
 		fmt.Sprintf("%d", int(keyExpiry.Seconds())),
```
mar-inngest, line 113: This is about 45 minutes.
TODO: discuss if we should increase the time for the duration of storing the worker-requests map and the request-worker information

pkg/execution/queue/queue.go
```diff
@@ -192,7 +192,8 @@ type alwaysRetry struct {
 func (a alwaysRetry) AlwaysRetryable() {}
 
 func IsAlwaysRetryable(err error) bool {
-	return errors.Is(err, alwaysRetry{})
```
mar-inngest, line 200: `errors.Is` didn't generally work for the class of error (since we usually embed an error inside this struct). Should I move this to a separate pull request?

pkg/connect/grpc/proxy.go
```diff
@@ -388,6 +397,30 @@ func (i *grpcConnector) Proxy(ctx, traceCtx context.Context, opts ProxyOpts) (*c
 			return nil, fmt.Errorf("failed to route message: %w", err)
 		}
 
+		routedInstanceID = route.InstanceID
+		// Assign the request lease to the worker for capacity tracking
+		if err := i.stateManager.AssignRequestToWorker(ctx, opts.EnvID, route.InstanceID, opts.Data.RequestId); err != nil {
```
BrunoScheufler:
```suggestion
		if err := i.stateManager.AssignRequestToWorker(ctx, opts.EnvID, routedInstanceID, opts.Data.RequestId); err != nil {
```

Tiny: I guess you can use the variable here as well or remove it since it looks like only one occasion uses that?

pkg/connect/grpc/proxy.go
```diff
@@ -461,9 +510,37 @@ func (i *grpcConnector) Proxy(ctx, traceCtx context.Context, opts ProxyOpts) (*c
...
+
 		return nil, syscode.Error{
 			Code:    syscode.CodeConnectWorkerStoppedResponding,
 			Message: "The worker stopped responding to the request.",
 		}
 	}
 }
+
+func cleanupWorkerRequestOrLogError(ctx context.Context, stateManager state.StateManager, envID uuid.UUID, instanceID string, requestID string, l logger.Logger, message string) {
+	//
```
BrunoScheufler: are we missing a comment here? 😄

pkg/connect/routing/router.go
```diff
@@ -121,9 +140,25 @@ func getSuitableConnection(ctx context.Context, rnd *util.FrandRNG, stateMgr sta
 	}
 
 	healthy := make([]connWithGroup, 0, len(conns))
```
BrunoScheufler:
```suggestion
	candidates := make([]connWithGroup, 0, len(conns))
```

Since we're using this to represent connections which are both 1) healthy and 2) have capacity, I think we can rename this to avoid confusion. Then your variables below make perfect sense 👍

pkg/connect/state/lua/decr_worker_requests.lua
```diff
@@ -0,0 +1,62 @@
...
+end
+
+-- Remove the specific request ID from the set
+redis.call("ZREM", workerRequestsKey, requestID)
+
+-- Check if set is now empty and delete it
+local remainingCount = redis.call("ZCARD", workerRequestsKey)
+if remainingCount == 0 then
+  -- Set is empty, delete it and the mapping
+  redis.call("DEL", workerRequestsKey)
```
BrunoScheufler: When you remove the last element, the set will be cleaned up automatically so you can omit this first delete, as the set has already been confirmed to be empty 👍

pkg/execution/queue/queue.go
BrunoScheufler: That would be great!

pkg/connect/state/state.go
```diff
@@ -91,6 +97,34 @@ type RequestStateManager interface {
...
+	// AssignRequestToWorker increments the active lease count for a worker instance.
+	// Returns an error if the worker is at capacity.
+	AssignRequestToWorker(ctx context.Context, envID uuid.UUID, instanceID string, requestID string) error
+
+	// DeleteRequestFromWorker decrements the active lease count for a worker instance.
+	DeleteRequestFromWorker(ctx context.Context, envID uuid.UUID, instanceID string, requestID string) error
+
+	// WorkerCapcityOnHeartbeat refreshes the TTL on the worker capacity key.
+	// Called on heartbeat to keep the capacity limit alive while worker is active.
+	WorkerCapcityOnHeartbeat(ctx context.Context, envID uuid.UUID, instanceID string) error
```
BrunoScheufler:
```suggestion
	WorkerCapacityOnHeartbeat(ctx context.Context, envID uuid.UUID, instanceID string) error
```

pkg/connect/state/state.go
```diff
@@ -375,3 +409,21 @@ func (g *WorkerGroup) Sync(ctx context.Context, groupManager WorkerGroupManager,
 
 	return nil
 }
+
+type WorkerCapacity struct {
+	Total         int64
+	Available     int64
+	CurrentLeases []string
```
BrunoScheufler: Do we currently access this? I'm wondering because this could get really large for some users 🤞 We may want to omit this and just return the counts.

pkg/connect/state/state.go
```diff
@@ -80,6 +83,9 @@ type RequestStateManager interface {
 	// GetExecutorIP retrieves the IP of the executor that owns the request's lease.
 	GetExecutorIP(ctx context.Context, envID uuid.UUID, requestID string) (net.IP, error)
 
+	// GetRequestWorkerInstanceID retrieves the instance ID of the worker that is assigned to the request.
+	GetRequestWorkerInstanceID(ctx context.Context, envID uuid.UUID, requestID string) (string, error)
```
BrunoScheufler: Tiny: We could also name this

```suggestion
	GetAssignedWorkerID(ctx context.Context, envID uuid.UUID, requestID string) (string, error)
```

pkg/connect/state/state.go
```diff
@@ -69,7 +71,8 @@ type RequestStateManager interface {
 
 	// ExtendRequestLease attempts to extend a lease for the given request. This will fail if the lease expired (ErrRequestLeaseExpired) or
 	// the current lease does not match the passed leaseID (ErrRequestLeased).
-	ExtendRequestLease(ctx context.Context, envID uuid.UUID, requestID string, leaseID ulid.ULID, duration time.Duration) (newLeaseID *ulid.ULID, err error)
+	// It also refreshes the worker instance's lease by updating the worker instance's last heartbeat.
```
BrunoScheufler, line 74: WIP: What do we use isWorkerCapacityLimited for?

pkg/connect/state/lua/extend_lease.lua
```diff
@@ -29,12 +54,44 @@ if requestItem.leaseID ~= leaseID and decode_ulid_time(requestItem.leaseID) > cu
 	return -2
 end
 
+-- this field is only set if worker capacity is limited
```
BrunoScheufler: We can add a note that worker capacity will be enabled or disabled for a given lease, and it can't change in between because the worker would be terminated (or is it possible to update worker concurrency without restarting?).

pkg/connect/state/lua/heartbeat_worker_capacity.lua
```diff
@@ -0,0 +1,37 @@
...
+end
+
+-- Refresh capacity key TTL
+redis.call("EXPIRE", workerTotalCapacityKey, workerTotalCapacityTTL)
+
+-- Refresh set key TTL
+--local setExists = redis.call("EXISTS", workerRequestsKey)
+-- if setExists == 1 then
+redis.call("EXPIRE", workerRequestsKey, workerRequestsSetTTL) -- incase the set doesn't exist, it' returns -2 but we ignore output
+--end
```
BrunoScheufler: Is this comment still needed or should it be removed?

pkg/connect/state/lua/incr_worker_requests.lua
```diff
@@ -0,0 +1,67 @@
...
+if capacity == nil or capacity == 0 or capacity == false or capacity == "0" then
+  return 0
+end
+
+capacity = tonumber(capacity)
+
+-- Get current time to filter out expired leases
+-- previous second, this makes us very sensitive to time changes
+-- Should we use the logical clock instead?
+local currentTimeUnix = tonumber(currentTime) / 1000
```
BrunoScheufler: You could also go into the past (e.g. clean up items older than 1 heartbeat interval)

BrunoScheufler: We do this in GC: Clean up connections which haven't reported in a multiple heartbeat intervals

pkg/connect/state/redis.go
```diff
@@ -824,3 +836,270 @@ func (r *redisConnectionStateManager) GetAllGatewayIDs(ctx context.Context) ([]s
...
+func (r *redisConnectionStateManager) GetWorkerCapacities(ctx context.Context, envID uuid.UUID, instanceID string) (*WorkerCapacity, error) {
+	totalCapacity, err := r.GetWorkerTotalCapacity(ctx, envID, instanceID)
+	if err != nil {
+		return nil, err
+	}
+
+	if envID == uuid.Nil {
+		return nil, ErrEnvIDCannotBeNil
+	}
+	if strings.TrimSpace(instanceID) == "" {
```
BrunoScheufler, line 912: Is this just a sanity check?

pkg/connect/state/redis.go
```diff
@@ -824,3 +836,270 @@ func (r *redisConnectionStateManager) GetAllGatewayIDs(ctx context.Context) ([]s
...
+	capacity, err := r.GetWorkerTotalCapacity(ctx, envID, instanceID)
+	if err != nil {
+		return err
+	}
+	if capacity <= 0 {
+		return nil
+	}
+
+	// refresh the TTL on the worker requests set (ignore the error for this)
+	_ = r.WorkerCapcityOnHeartbeat(ctx, envID, instanceID)
```
BrunoScheufler: Should we still log the error (just not return it)? What are the implications of this call failing?

pkg/connect/state/lua/extend_lease.lua
```diff
@@ -29,12 +54,44 @@ if requestItem.leaseID ~= leaseID and decode_ulid_time(requestItem.leaseID) > cu
...
+if workerCapUnlimitedBool == false then
+	local workerInstanceID = redis.call("GET", requestWorkerKey)
+	if workerInstanceID ~= instanceID then
```
BrunoScheufler, line 62: Is this a protection against an old worker which is no longer processing this request extending the original lease?

pkg/connect/state/lua/extend_lease.lua
```diff
@@ -29,12 +54,44 @@ if requestItem.leaseID ~= leaseID and decode_ulid_time(requestItem.leaseID) > cu
...
+if workerCapUnlimitedBool == true then
+	return 1
+end
+
+-- Add to the new lease to the worker's set
+redis.call("ZADD", workerRequestsKey, currentTime + requestLeaseDuration, requestID)
+redis.call("EXPIRE", workerRequestsKey, workerRequestsSetTTL)
+
+-- Update the request-worker mapping
+redis.call("SET", requestWorkerKey, instanceID, "EX", requestWorkerMappingTTL)
```
BrunoScheufler: Would this be equivalent to an expire call?

pkg/connect/grpc/proxy.go
mar-inngest: Thanks! There's 2 other places. I'll fix there as well.

pkg/connect/state/state.go
mar-inngest: Yeah, I initially had counts. My biggest concern with this memory fragmentation issues. We only use this information for trace logging (to make debugging easier). 
It made sense to have this information during initial development (to avoid bugs and errors). Now, we can move it to a separate function, maybe. Let me circle back on this later.

pkg/connect/state/state.go
mar-inngest: We use this information for an early escape from the request/avoid extra redis calls. We can change the name. 
Also, good catch. It's supposed to be `isWorkerCapacityUnlimited`

pkg/connect/state/lua/heartbeat_worker_capacity.lua
mar-inngest: I think we can remove this comment now, I was initially doing a redis call and then updating but updating without checking saves a redis call, especially since we don't care about the return code. If it doesn't exist, then we have other problems (but it's already long enough that it should exist)

pkg/connect/state/redis.go
mar-inngest: Yeah, this is just a sanity check. I haven't seen it happen, yet

pkg/connect/state/redis.go
mar-inngest: The implications are relatively straightforward, i.e. a worker dies. We can do a trace log here. We already do a heartbeat check in two other places, so we can even skip this if you want. In the original implementation, the redis TTL was very short, and it made sense. In the current implementation, the redis TTL for the capacity is 2 hour long so we can safely remove this

pkg/connect/state/lua/extend_lease.lua
mar-inngest: Yeah, this is a safeguard against old instance, in case where we have multiple workers

pkg/connect/state/lua/extend_lease.lua
mar-inngest: Good catch! This is a remnant of my initial implementation when I was deleting leases and then adding them again. I'll change it to expire for clarity.

pkg/connect/state/lua/decr_worker_requests.lua
mar-inngest: Thanks! I'm going to refactor and then remove this if condition completely in that case. That would simplify reading the script.

pkg/connect/state/lua/extend_lease.lua
mar-inngest: Yeah, let me update the note. I was using worker capacity limited vs unlimited right now

pkg/connect/state/state.go
mar-inngest: Updated

pkg/connect/state/lua/incr_worker_requests.lua
mar-inngest: Yeah, I guess that approach has some advantages. Since we're already going to delete the finished elements from the set, this would usually happen if a worker lost connection for a while. However, this also means, that if a worker lost connection for a while, a new worker won't be able to get a request.
Currently all of our logic is based the 'now' (for some definition of now). `now()-20 seconds` is doable, but that means that we'll have to update the `GetWorkerCapacity` and `GetAllActiveWorkerRequests` functions to also consider the same value as the ground truth.

One more thing is that when we add an item, it's not added with 20 seconds (i.e. the TTL), we add it with `consts.ConnectWorkerRequestToWorkerMappingTTL` which is currently 2 minutes (different from the 2 hour TTL for total capacity). So, in this case, we will already be waiting 2 minutes (if a worker goes away for some reason) and adding an extra 20 seconds might not benefit us too much.


Good point about using `Unix()` instead of `UnixMilli()`. I'll fix that.

pkg/consts/consts.go
```diff
@@ -148,10 +148,15 @@ const (
...
 	ConnectWorkerRequestGracePeriod   = 5 * time.Second
 
+	// ConnectWorkerNoConcurrencyLimitForRequests is used to indicate that a worker has no capacity limit.
+	ConnectWorkerNoConcurrencyLimitForRequests = -1
+
 	KafkaMsgTooLargeError = "MESSAGE_TOO_LARGE"
 )
 
 var (
+	ConnectWorkerRequestToWorkerMappingTTL  = 6 * ConnectWorkerRequestLeaseDuration  // 2 minutes so in case of gateway failure, we don't lose the mapping for too long
```
mar-inngest, line 158: @BrunoScheufler should we decrease this from 2 minutes to something else? This is what we store inside the worker-requests-set, so we don't force delete a data until after 2 minutes of a worker dying/disappearing. My reason was mostly around gateway restarts.

The value below this one is the one we use for totalcapacity, and also computing the length (i.e. the duration the set is kept around)

pkg/connect/gateway.go
```diff
@@ -160,21 +161,20 @@ func (c *connectGatewaySvc) Handler() http.Handler {
 		}
 
 		closeReason := connectpb.WorkerDisconnectReason_UNEXPECTED.String()
-		var closeReasonLock sync.Mutex
-		setCloseReason := func(reason string) {
-			closeReasonLock.Lock()
-			defer closeReasonLock.Unlock()
+		var closeReasonPtr atomic.Pointer[string]
+		closeReasonPtr.Store(&closeReason)
```
mar-inngest, line 165: This was a data race condition that kept failing some tests. Pointers are usually lighter than locks.

pkg/connect/state/lua/decr_worker_requests.lua
mar-inngest: Done

pkg/connect/state/lua/extend_lease.lua
mar-inngest: Updated the comments

darwin67 (approved):
Looks good in general.

While this looks rather complete as an implementation, there's no data emitted to provide visibility to users on what's happening.
So that'll need to be a follow up.

pkg/consts/consts.go
BrunoScheufler: So just to make sure I understand this correctly, is this like a grace period before cleaning up the sorted set holding requests for the worker?

Would this mean we wait for up to 2 minutes of a worker no longer sending heartbeats before we dropped the state for the worker (is state just the sorted set of requests here?)?

BrunoScheufler: One thing I wonder about is, in the case of a worker disconnecting and then reconnecting, it would continue to extend the lease, so unless the request lease expired in the meantime, you should be able to recreate the state even if it was dropped before? So does this ultimately just affect routing decisions (or not even those because the connections are removed when the worker disconnects)?

mar-inngest: Yeah, this is a worker-level grace period.  On heartbeats, we _don't_ update the request information inside sorted sets (i.e. the request information about all the requests), we only update the TTL of the sorted set itself. 
We rely on the ExtendRequestLease function to arrive (since it should happen 5-6 times) to update the request-specific information of a worker (i.e. inside the sorted set).

We can recreate the previous state in case of failures, but due to executor sending the request to a different worker, it's possible that this worker might not get the same request.

----

When a worker comes to extend a leases, we first check the capacity of the worker. If the worker isn't assigned to the worker it claims to be, we return a SocketError, before we extend the lease. See: https://github.com/inngest/inngest/blob/652f2f8c816ad264ce0ae39d21b284836f4a19d8/pkg/connect/gateway.go#L833

If a worker disconnects, two cases I know are; process restart (i.e. the worker had multiple connections - relatively simple since it would ), and network timeout. For network timeout, 
* if the worker comes back (within 2 minutes), and tries to get the lease back, it should be fine, if there are no other workers. 
* if it comes after 2 minutes, and then tries to extend the lease, we should throw an error (`ErrRequestWorkerDoesNotExist`)
* if it comes back within 2 minutes, but the request was assigned to another worker, in this case, the worker that just came back won't be at full capacity till the end of two minutes (because our system would say that the worker has existing requests).

---

One byzantine case is that if the worker comes and tries to hijack the request of another worker, in such a case, we shouldn't delete our state. That's why we try not to delete anything based on what the customer sent. We only create based on worker, but delete based on our TTLs.

----

In case of a connection cleanup, we try not to touch the worker capacity. That's because of multiple concurrent connections from a single worker. Also, it keeps the concurrency logic simpler and avoids the case of some possible data races (like one connection deleting worker information as the last worker but a new worker creating it).

PR #3213, opened 2025-10-28
Pause block flushing in a system job

KiKoS0:
## Description

Support pause block flushing from the system job. While testing, I found a few issues in related code paths, so this PR includes fixes and adds missing tests.

When a pause is removed from the block store we need its creation timestamp to know which blocks might contain it. Block boundaries are based on that timestamp but this information only exists in the indexes. Once a pause is flushed and removed from the buffer we lose that timestamp.

This PR embeds the creation timestamp directly into newly created pauses. For existing pauses it introduces a temporary workaround in the Redis iterator to add their creation timestamp when fetching pauses for block flushing. This is implemented via `PausesSinceWithCreatedAt`, which also maintains ordered iteration so that block boundaries are correct (`[Pauses[0].CreatedAt, Pauses[last].CreatedAt]`)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/redis_state.go
+++ b/pkg/execution/state/redis_state/redis_state.go
@@ -1862,7 +1862,7 @@ func (i *keyIter) init(ctx context.Context, keys []string, scores []float64, chu
 	i.keys = keys
 	i.chunk = chunk
 	i.scores = scores
-	i.hasScores = len(scores) > 0
+	i.hasScores = len(scores) == len(keys)
 	err := i.fetch(ctx)
 	if err == errScanDone {
 		return nil


pkg/execution/pauses/block.go
```diff
@@ -234,6 +272,17 @@ func (b blockstore) flushIndexBlock(ctx context.Context, index Index) error {
 		// from iterators often;  this only happens in a race where the iterator
 		// has pauses in-memory which are then deleted while iterating, leading to
 		// a race condition.
+
+		// Another case where this could happen is during rollout when we are not
+		// deleting pauses from the buffer after flushing to a block which will
+		// make it that we trigger flush jobs because the buffer is always at the
+		// threshold but we can't find enough pauses since the last block.
```
darwin67: would be nice to have some metrics on each type of scenario to perhaps help with identifying further optimizations later.

pkg/execution/pauses/block.go
```diff
@@ -461,7 +531,7 @@ func (b *blockstore) blockIDForTimestamp(ctx context.Context, idx Index, ts time
 	score := strconv.Itoa(int(ts.UnixMilli()))
 	ids, err := b.rc.Do(
 		ctx,
-		b.rc.B().Zrange().Key(blockIndexKey(idx)).Min("("+score).Max("+inf").Byscore().Limit(0, 1).Build(),
```
darwin67, line 464: this means inclusive of the minimum score value, expected?

darwin67 (approved):
some minor comments and questions. otherwise looks good.

pkg/execution/pauses/block.go
KiKoS0: yeah it is wrote a test for it too, otherwise the pause having a timestamp == latest block boundary cannot be deleted.

PR #3221, opened 2025-10-29
Check score on backlogRefill

BrunoScheufler:
If score is missing, item was removed from backlog between peek and refill.

## Description

#3181 switched from peeking within the script to peeking ahead of time. We already checked the case of items getting dequeued (missing), but not the case where items were removed from the backlog, simply not the entire queue. This could happen if the item is refilled concurrently (which should not happen due to locks, but there may be a case where it still happens).

The solution is to check the score. If it's nil, the item is no longer in the backlog.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/queue_test.go
+++ b/pkg/execution/state/redis_state/queue_test.go
@@ -7365,10 +7365,12 @@ func TestInvalidScoreOnRefill(t *testing.T) {
 	backlog := q.ItemBacklog(ctx, qi)
 	sp := q.ItemShadowPartition(ctx, qi)
 
-	r.ZRem(
+	removed, err := r.ZRem(
 		defaultShard.RedisClient.kg.BacklogSet(backlog.BacklogID),
 		qi.ID,
 	)
+	require.NoError(t, err)
+	require.True(t, removed)
 
 	res, err := q.BacklogRefill(
 		ctx,


pkg/execution/state/redis_state/queue_test.go
```diff
@@ -7277,3 +7277,116 @@ func zcard(t *testing.T, rc rueidis.Client, key string) int {
...
+	require.NoError(t, err)
+
+	qi2, err := q.EnqueueItem(ctx, defaultShard, item2, clock.Now(), osqueue.EnqueueOpts{})
+	require.NoError(t, err)
+
+	backlog := q.ItemBacklog(ctx, qi)
+	sp := q.ItemShadowPartition(ctx, qi)
+
+	fmt.Println(r.Dump())
+	fmt.Println("AFTER")
```
darwin67: remove these

pkg/execution/state/redis_state/queue_test.go
```diff
@@ -7277,3 +7277,116 @@ func zcard(t *testing.T, rc rueidis.Client, key string) int {
...
+
+	r.ZRem(
+		defaultShard.RedisClient.kg.BacklogSet(backlog.BacklogID),
+		qi.ID,
+	)
+
+	fmt.Println(r.Dump())
```
darwin67: this as well

lkasinathan (approved):

PR #3226, opened 2025-10-30
[Insights]: Introduce transformJSONSchema to Simplify UI Schema Rendering

doverdbng:
## Description

This PR adds a `transformJSONSchema` function to convert a JSONSchema to a simplified `SchemaNode` representation.

**NOTE:** Because we are deriving these schemas from concrete event instances, we can make assumptions about JSONSchema features like `additionalProperties` and boolean (accept everything/nothing) schemas not appearing.

**NOTE:** The expectations of this PR are essentially that enough machinery is in place to process (and later render) the common event schema (ignoring the variable "data"). `transformJSONSchema` should not be considered complete as of this PR.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/vercel/error.tsx
+++ b/ui/apps/dashboard/src/app/(organization-active)/(dashboard)/settings/integrations/vercel/error.tsx
@@ -1,9 +1,12 @@
 'use client';
 
 import { useEffect } from 'react';
+import { useRouter } from 'next/navigation';
+import { Button } from '@inngest/components/Button';
 import * as Sentry from '@sentry/nextjs';
 
 import { FatalError } from '@/components/FatalError';
+import { pathCreator } from '@/utils/urls';
 
 type VercelIntegrationErrorProps = {
   error: Error & { digest?: string };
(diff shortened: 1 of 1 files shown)


ui/packages/components/src/SchemaViewer/transform/kind.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (schema.items) return 'array';
+    return 'value';
+  }
+
+  if (Array.isArray(t)) {
+    const set = new Set(t);
+    if (set.has('object') && (set.size === 1 || (set.size === 2 && set.has('null'))))
+      return 'object';
+    if (set.has('array') && (set.size === 1 || (set.size === 2 && set.has('null')))) return 'array';
+    if (set.size === 1 && t.length > 0) return mapSingleType(t[0] as string);
```
AndyInternet: This code is unreachable.
if t is an array and has a size of 1 and contains 'array' or 'object', it's already covered by the two conditionals above it.

ui/packages/components/src/SchemaViewer/transform/kind.ts
```diff
@@ -0,0 +1,33 @@
+import type { JSONSchema } from '../types';
+
+export function determineKind(schema: JSONSchema): 'array' | 'object' | 'value' {
+  const t = schema.type;
+
```
AndyInternet, line 5: Isn't it possible for JSONSchema.type to be a single value as well like 'string' or 'object'. Currently this isn't covered. Should we add a line here like:
`if (typeof t === 'string') return mapSingleType(t);`

doverdbng: That should be the call to `mapSingleType(t)` at the end.

AndyInternet (approved):

PR #3231, opened 2025-10-31
ignore fv in checkpoint

tonyhb:
This fix ignores FVs when comparing config hashes.  This, unfortunately, isn't an ideal fix: it increases allocations, GC pressure, CPU, etc.

Instead, in the future we should have the ability to checksum a function version when upserting data.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/apiv1/checkpoint.go
+++ b/pkg/api/apiv1/checkpoint.go
@@ -654,7 +654,9 @@ func (a checkpointAPI) upsertSyncData(ctx context.Context, auth apiv1auth.V1Auth
 	// XXX: It would be good to add an upsert method to the function CQRS layer.
 	fn, err := a.FunctionReader.GetFunctionByInternalUUID(ctx, fnID)
 	if err == nil && fn != nil {
-		if string(fn.Config) == config {
+
+		out := removeFunctionVersion(ctx, fn.Config)
+		if config == string(out) {
 			return // no need to update
 		}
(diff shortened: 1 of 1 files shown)


pkg/api/apiv1/checkpoint.go
```diff
@@ -736,3 +738,21 @@ func (a checkpointAPI) fn(ctx context.Context, fnID uuid.UUID) (*inngest.Functio
 	}
 	return cfn.InngestFunction()
 }
+
+func removeFunctionVersion(ctx context.Context, input json.RawMessage) []byte {
+	// Unfortunately for us, there's a function version inside this config.  We need to ensure
+	// that this is set to "1".
+	// TODO: remove function version from marshalled config.
+	//
+	// It's safe to ignore errors here;  this only resultss in
```
albertchae, line 747: what a cliffhanger

pkg/api/apiv1/checkpoint.go
```diff
@@ -736,3 +738,21 @@ func (a checkpointAPI) fn(ctx context.Context, fnID uuid.UUID) (*inngest.Functio
...
+	found := inngest.Function{}
+	if err := json.Unmarshal(input, &found); err != nil {
+		logger.StdlibLogger(ctx).Warn("error unmarshalling fn config in checkpoint upsert", "error", err)
```
albertchae, line 750: hm are you saying it's safe to continue here because we'd end up serializing `inngest.Function{FunctionVersion: 1}` which would eventually fail some validation during upsert?

PR #3249, opened 2025-11-04
[Insights]: Update transformJSONSchema for Arbitrary Schemas

doverdbng:
## Description

This expands `transformJSONSchema` to handle various JSONSchema structures that our BE parser might derive from concrete events. It also adds a test harness and a number of tests.

The big picture here is that we have the following pipeline: `JSONSchema` (from BE) -> `SchemaNode` (tree) -> UI tree representation. During that first hop, we essentially squash out any optionality, at least for now, to make it practical to represent the schema in a tree format. The justification for squashing out optionality is twofold:

- there shouldn't be optionality at all given that we're deriving schemas from a concrete event
  - doesn't hold up perfectly, as the parser currently represents heterogeneous arrays via `oneOf` instead of a tuple
- it's not practical to represent in a tree

**NOTE:** The test script is currently commented out in `ui/packages/components/package.json` due to a dev server issue, but you can temporarily replace it with `vitest` to run these tests via `pnpm test`.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/cron/cron_test.go
+++ b/pkg/execution/cron/cron_test.go
@@ -1,6 +1,7 @@
 package cron
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


albertchae (approved):
thanks for adding tests! in the future we should consider enhancing it with some test script that can run an event through the backend code to get a json schema. That way we can tie each test case to an example event that would generate it

PR #3265, opened 2025-11-07
SYS-375: Implement rate limiting in pure Lua

BrunoScheufler:
## Description

This PR implements rate limiting in pure Lua, fully compatible/equivalent with the existing implementation. This is a preparation for the upcoming Constraint API, which will have to handle fallbacks in the existing logic, as well as using the same implementation for Acquire.

## Motivation

- For the Constraint API, we need to add changes to rate limiting
  - To the existing implementation, we need to add fallbacks so that previously-granted requests are not rate limited during retries
  - For the new Acquire method in the Constraint API, we need constraint enforcement logic that mirrors the existing implementation

For these reasons, moving to a common Lua implementation first seems useful. We can reuse code for Acquire and add fallback handling to the new Lua script.

## Test coverage

This PR contains extensive test coverage for the changes

- Tests for Lua rate limiting logic (all base cases, burst handling)
- Tests to ensure backward and forward compatibility
- Side by side tests to expect the exact same outcomes for each operation

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/ratelimit/ratelimit_lua_test.go
+++ b/pkg/execution/ratelimit/ratelimit_lua_test.go
@@ -67,7 +67,7 @@ func TestLuaRateLimit_BasicFunctionality(t *testing.T) {
 		_, rc, _, clock := initRedis(t)
 		defer rc.Close()
 
-		limiter := newLuaGCRARateLimiter(ctx, rc, "test:")
+		limiter := newLuaGCRARateLimiter(ctx, rc, "{rl}:")
 
 		config := inngest.RateLimit{
 			Limit:  1,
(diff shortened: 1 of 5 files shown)


darwin67 (approved):

pkg/execution/ratelimit/ratelimit_lua_test.go
```diff
@@ -0,0 +1,1325 @@
...
+				// Both should have same semantics now
+				require.Equal(t, throttledLimitedStatus, luaLimitedStatus,
+					"request %d: limited status should match (throttled_limited=%v, lua_limited=%v)",
+					i+1, throttledLimitedStatus, luaLimitedStatus)
+
+				// If rate limited, both should have similar retry times (within tolerance)
+				if throttledAllowed && luaAllowed {
+					// Both are rate limited - compare retry times
+					// throttledRetry might be -1 if not rate limited, so check for positive values
+					if throttledRetry > 0 && luaRetry > 0 {
```
lkasinathan: nano-nit:  could we unconditionally test the retry value diff? If not, could we explicitly set it in the testcase whether or not to perform this check to avoid bugs in retry value silently stopping us from running the following assertion.

lkasinathan (approved):

pkg/execution/ratelimit/ratelimit_lua_test.go
BrunoScheufler: Done!

PR #3270, opened 2025-11-07
refactor: store AppID and FunctionID on userland traces

Linell:
## Description

These attributes are required for the rollup to work correctly in cloud.

## Motivation

Enabling userland traces to display in cloud, a la [EXE-850: Change Default OTEL Entitlement to True](https://linear.app/inngest/issue/EXE-850/change-default-otel-entitlement-to-true)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/apiv1/traces.go
+++ b/pkg/api/apiv1/traces.go
@@ -151,7 +151,7 @@ func (a router) convertOTLPAndSend(ctx context.Context, auth apiv1auth.V1Auth, r
 						}
 					}()
 
-					err := a.commitSpan(res, ss.Scope, s)
+					err := a.commitSpan(ctx, auth, res, ss.Scope, s)
 					if err != nil {
 						l.Error("failed to commit span with", "error", err)
 						errs.Add(1)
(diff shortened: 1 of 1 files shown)


pkg/api/apiv1/traces.go
```diff
@@ -207,6 +208,28 @@ func (a router) commitSpan(res *resource.Resource, scope *commonv1.Instrumentati
...
+			}
+			break
+		}
+	}
+
+	if functionID == uuid.Nil {
+		return fmt.Errorf("missing function ID in span attributes")
+	}
+
+	fn, err := a.opts.FunctionReader.GetFunctionByInternalUUID(ctx, functionID)
```
rhino1998, line 216: Maybe we should use `a.opts.FunctionRunReader.GetFunctionRun()` to get the function ID from the run ID so that we're respecting workspace/account IDs rather than just trusting the attribute from the span. Also, then we're validating that the runID is real and that the authed user has access.

Linell: 👍  you're right, changed!

rhino1998 (approved):

jpwilliams (approved):

PR #3282, opened 2025-11-11
Fix Step Input/Output displaying as byte arrays in PostgreSQL

orionmiz:
## Description

Fixes a display bug where Step Input and Step Output fields showed Go struct representations (e.g., `{[110 117 108 110] true}`) instead of proper JSON format when using PostgreSQL as the database backend. The issue was caused by `fmt.Append()` converting `pqtype.NullRawMessage` structs to their string representation rather than extracting the underlying JSON bytes. This fix uses type assertion to directly access the `RawMessage` field from the struct, properly extracting the JSON data.

## Motivation

Users running self-hosted Inngest with PostgreSQL were unable to view step input/output data in the dashboard, as the data appeared as byte arrays wrapped in struct notation rather than readable JSON. This made debugging and monitoring step execution significantly harder. The `pqtype.NullRawMessage` type is a struct containing both the JSON data (`RawMessage []byte`) and a validity flag (`Valid bool`). By using type assertion instead of generic string conversion, we can access these fields directly and extract only the JSON bytes, resulting in proper display in the UI.

### Testing

Before:
- Step Output: `{[34 123 92 34 100 97 116 97 92 34 58 110 117 108 108 125 34] true}`

After:
- Step Output: `null`

Test Environment:
- Self-hosting with external PostgreSQL service
- Verified via dashboard UI: Run → Span → Step Output

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/cqrs/base_cqrs/cqrs.go
+++ b/pkg/cqrs/base_cqrs/cqrs.go
@@ -1628,30 +1628,51 @@ func (w wrapper) GetSpanOutput(ctx context.Context, opts cqrs.SpanIdentifier) (*
 			// Extract bytes from pqtype.NullRawMessage using type assertion
 			if nullMsg, ok := row.Input.(pqtype.NullRawMessage); ok && nullMsg.Valid {
 				so.Input = []byte(nullMsg.RawMessage)
+
+				// PostgreSQL-specific handling for double-encoded JSON and null values
+				// First, try to unmarshal as a string (in case it's double-encoded)
+				var jsonStr string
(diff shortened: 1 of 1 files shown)


jacobheric:
Thanks a ton for catching this and submitting a fix @orionmiz! I implemented a similar fix at the postgresql/sqlite normalization level to keep the code above db agnostic: https://github.com/inngest/inngest/pull/3291. I'll ping you in discord when it's released.

orionmiz:
@jacobheric Totally agreed. It seems better to fix it at the normalization level.

PR #3285, opened 2025-11-11
john/pdes 33 fix ux issues with click to copy timestamps in clickable

whatupjohnb:
- **changed the hover functionality, hand for links pointer for non-links. Clicking the icon only is what copies to clipboard. Links now navigate to prper path if they are a timestamp**
- **Make icon larger. Fix layout shift on hover of timestamps.**
- **added copy prop to time.tsx, removed copy on event-type details page, webhooks table selector and event keys selector**

## Description

Made time stamp click to copy on icon only
As a result, link cannibalism when there is a tooltip on a timestamp is fixed.
Made icon larger because target is smaller.
Changed the icon show/hide mechanism to prevent layout shift
Removed copy on /eventkeys. /event-type/details, /webhooks

image
image
image

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/AppRoot/globals.css
+++ b/ui/packages/components/src/AppRoot/globals.css
@@ -494,13 +494,3 @@ input[type='number'] {
   scrollbar-width: 10px;
   scrollbar-gutter: 10px;
 }
-
-/* Time elements inside links should show pointer cursor */
-/* Higher specificity to override cursor-default class */
-/* Account for TooltipTrigger wrapper between link and time element */
-a time.cursor-default,
-a * time.cursor-default,
-a .cursor-default time,
-a * .cursor-default time {
-  cursor: pointer !important;
-}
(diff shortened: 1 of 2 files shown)


ui/packages/components/src/AppRoot/globals.css
```diff
@@ -494,3 +494,13 @@ input[type='number'] {
   scrollbar-width: 10px;
   scrollbar-gutter: 10px;
 }
+
```
anafilipadealmeida: Why this instead of just adding the cursor-pointer on copyable items, and let their parents decide on the cursor in other situations?

albertchae: Good point! We didn't realize `<TooltipTrigger>` was adding a `<button>` to the DOM which would get `cursor: pointer` from tailwind v3 defaults, which then blocked inheriting the cursor from the parent. I changed it to use `asChild` so there's no more intermediary `<button>`. This also seems reasonable semantically since this PR is reducing the clickable area from the whole `<time>` to just the icon inside. Maybe we should wrap that icon inside a button now 🤔

anafilipadealmeida: That's a good call! Yep, semantically it would be ideal to wrap the icon in a <button> element now that the clickable area is just the icon

PR #3298, opened 2025-11-13
EXE-907: Create new bulk re-run endpoint

djfarrelly:
## Description

An endpoint that accepts multiple run ids and re-runs them. Each re-run may potentially fail, but it doesn't fail the entire endpoint.  Current limit of 50 runs per request.

```
curl -XPOST http://localhost:8288/api/v2/bulk/rerun \
  -H "content-type: application/json" -d '{ 
    "runIds": ["01K9XTMJ7CNG4J853W5Z8K694V", "01K9XTMJ798MZ6Z589W94H9AX7"] 
  }'
{
  "data": [
    {
      "originalRunId": "01K9XTMJ7CNG4J853W5Z8K694V",
      "newRunId": "01K9XTPXGPCZFDJNXXN7F46800",
      "success": true
    },
    {
      "originalRunId": "01K9XTMJ798MZ6Z589W94H9AX7",
      "newRunId": "01K9XTPXGTNFVQ95YM7M89PT2R",
      "success": true
    }
  ],
  "metadata": {
    "fetchedAt": "2025-11-13T05:23:45.820203Z",
    "cachedUntil": null
  }
}
```

See related #3297

## Motivation
EXE-907

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/v2/endpoints.go
+++ b/pkg/api/v2/endpoints.go
@@ -7,7 +7,13 @@ import (
 
 	"github.com/inngest/inngest/pkg/api/v2/apiv2base"
 	"github.com/inngest/inngest/pkg/consts"
+	"github.com/inngest/inngest/pkg/event"
+	"github.com/inngest/inngest/pkg/execution"
+	"github.com/inngest/inngest/pkg/run"
+	"github.com/inngest/inngest/pkg/util"
 	apiv2 "github.com/inngest/inngest/proto/gen/api/v2"
+	"github.com/oklog/ulid/v2"
+	"go.opentelemetry.io/otel/attribute"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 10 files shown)


proto/api/v2/service.proto
```diff
@@ -912,6 +912,78 @@ service V2 {
...
+        key: "200"
+        value: {
+          description: "Bulk rerun results"
+          schema: {
+            json_schema: {
+              ref: "#/definitions/v2BulkRerunResponse"
+            }
+          }
+        }
+      }
```
djfarrelly, line 940: Should this be a 202 Accepted instead since this is async?

darwin67:
no tests?

proto/api/v2/service.proto
darwin67: if this is sequential, why not just return all the new runIDs that were scheduled, and just keep it as 200?

also, how does this handle partial failures? what status code is that going to be?

darwin67:
seems like a start but 50 is kinda low as a limit.
if it's async, it seems better to create an inngest job to handle the reruns, and this API just pass on the provided runIDs as a list in the event, and return a JobID or something that the consumer can check the status on.

pkg/api/v2/endpoints.go
```diff
@@ -277,3 +283,113 @@ func (s *Service) ListWebhooks(ctx context.Context, req *apiv2.ListWebhooksReque
...
+
+	if len(req.RunIds) > 50 {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Cannot rerun more than 50 runs at once")
+	}
+
+	accountID := consts.DevServerAccountID
+	workspaceID := consts.DevServerEnvID
+
+	// Process each run ID
+	results := make([]*apiv2.RerunResult, len(req.RunIds))
```
djfarrelly, line 301: Some of this could be refactored to a cqrs interface or similar to share code between here and the GraphQL API. I'd like to get this out, then plan the interface out for other operations like Replay, cancel, etc.

PR #3299, opened 2025-11-13
Ignore singletons during normalization

BrunoScheufler:
## Description

Singleton state is already set so we can safely ignore checking it during normalization.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/backlog_normalization_test.go
+++ b/pkg/execution/state/redis_state/backlog_normalization_test.go
@@ -2,10 +2,12 @@ package redis_state
 
 import (
 	"context"
+	"crypto/rand"
 	"testing"
 	"time"
 
 	"github.com/inngest/inngest/pkg/util"
+	"github.com/oklog/ulid/v2"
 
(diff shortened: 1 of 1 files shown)


pkg/execution/state/redis_state/lua/queue/enqueue.lua
```diff
@@ -77,7 +77,7 @@ if redis.call("HSETNX", queueKey, queueID, queueItem) == 0 then
 end
 
 -- Check if the item is a singleton and if an existing item already exists
-if exists_without_ending(singletonKey, ":singleton:-") then 
+if exists_without_ending(singletonKey, ":singleton:-") and not is_normalize then
```
BrunoScheufler, line 80: Singleton state would have been added during the first enqueue. The enqueue during normalization attempts to write the existing item with potentially new concurrency/throttle keys to a different backlog. The singleton key and run lock should meanwhile stay the same, as the queue item ID, singleton key, or run ID does not change.

KiKoS0 (approved):

PR #3302, opened 2025-11-13
Block Compaction

KiKoS0:
## Description

Implements pause block compaction with:
- Proper leasing, I reused the same block leaser for flushing it should be good to reuse with a different prefix.
- Add new compaction feature flag, when disabled it would just do a dry run of how many blocks it would have tried to compact.
- Add metrics for compactions.
- Tests

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/pauses/block.go
+++ b/pkg/execution/pauses/block.go
@@ -37,7 +37,10 @@ const (
 	DefaultCompactionSample = 0.1
 
 	// DefaultCompactionLeaseRenewInterval is the lease renewal period for compaction.
-	DefaultCompactionLeaseRenewInterval = 25 * time.Second
+	DefaultCompactionLeaseRenewInterval = 15 * time.Second
+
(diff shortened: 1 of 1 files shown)


pkg/execution/pauses/block.go
```diff
@@ -569,16 +593,222 @@ func (b *blockstore) PauseByID(ctx context.Context, index Index, pauseID uuid.UU
...
+		},
+		func(ctx context.Context, leaseID ulid.ULID) error {
+			return b.leaser.Revoke(ctx, index, leaseID)
+		},
+		func(ctx context.Context) error {
+			// Call this function and block, renewing leases in the background
+			// until this function is done.
+			return b.compact(ctx, index)
+		},
+		25*time.Second,
```
darwin67: should this interval be configurable?
I see you added a histogram, so I assume you'd want to tweak this duration after observing how it's doing after some time?

KiKoS0: Yeah it definitely should be

PR #3304, opened 2025-11-14
[SYS-370] - Total Capacity For Connect Workers Concurrency

mar-inngest:
## Description

Metrics For Connect Workers Concurrency. 

## Motivation

* Instance-level metrics for connect-worker concurrency
* We display total capacity, and percentage usage (0% when there's no concurrency). We currently _do not_ show current usage (since we currently don't record it when worker concurrency is disabled). 

- [ ] Docs
- [x] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist

## Change
* Added an `Int64` data type to GraphQL
* Updating an SQLite/Postgres table schema (hence breaking change) for worker connections
* On every heartbeat, check worker capacity

### DIFF
--- a/pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
+++ b/pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
@@ -165,7 +165,7 @@ CREATE TABLE worker_connections (
     instance_id VARCHAR NOT NULL,
     status INT NOT NULL,
     worker_ip VARCHAR NOT NULL,
-	max_worker_concurrency INT NOT NULL,
+    max_worker_concurrency INT NOT NULL,
 
     connected_at INT NOT NULL,
     last_heartbeat_at INT,


mar-inngest:
No Limit. Note: The code was moved to this PR: https://github.com/inngest/inngest/pull/3341
image

mar-inngest:
Concurrency Limit. Note: The code was moved to this PR: https://github.com/inngest/inngest/pull/3341
image

mar-inngest:
CHECKPOINT: Now the UI shows the value of max worker concurrency for the worker.
Note: The UI code was moved to this PR: https://github.com/inngest/inngest/pull/3341

mar-inngest:
Working UI, but need to move it behind a feature flag and refactor some of the backend code now.

Note: The UI code was moved to this PR: https://github.com/inngest/inngest/pull/3341
image

mar-inngest:
Note: The UI code was moved to this PR: https://github.com/inngest/inngest/pull/3341
image

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
```diff
@@ -165,6 +165,7 @@ CREATE TABLE worker_connections (
     instance_id VARCHAR NOT NULL,
     status INT NOT NULL,
     worker_ip VARCHAR NOT NULL,
+	max_worker_concurrency INT NOT NULL,
```
darwin67: weird indentation

pkg/consts/consts.go
```diff
@@ -148,8 +148,9 @@ const (
 	ConnectWorkerRequestLeaseDuration = 20 * time.Second
 	ConnectWorkerRequestGracePeriod   = 5 * time.Second
 
-	// ConnectWorkerNoConcurrencyLimitForRequests is used to indicate that a worker has no capacity limit.
-	ConnectWorkerNoConcurrencyLimitForRequests = -1
+	// ConnectWorkerCapacityForNoConcurrencyLimit is used to indicate that a worker has no capacity limit.
+	// Due to integer overflows of using negative numbers, we use 0 to indicate no limit.``
+	ConnectWorkerCapacityForNoConcurrencyLimit = 0
```
darwin67, line 153: any concerns for changing this?

mar-inngest: That was the original implementation. There's a `WorkerCapacity` struct that uses it. I reworked that as well. We already had tests for it from the previous implementation.

Previously, we would use `-1` as the current capacity to show it was not used, now, we use `total capacity==0` to check that

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
mar-inngest: Spaces vs tabs, fixed

pkg/coreapi/gql.schema.graphql
```diff
@@ -697,6 +698,7 @@ type ConnectV1WorkerConnection {
   gatewayId: ULID!
   instanceId: String!
   workerIp: String!
+  maxWorkerConcurrency: Int64!
```
amh4r, line 701: What's the reason for adding a new `Int64` scalar instead of using `Int`? JS doesn't have an int64 type (or an int type at all 😆)

mar-inngest: Yeah, I had this in the configuration. It's mostly useful for other languages that try to convert it into an Int64 datatype, and then an Int. There was also a `Uint` in gqlgen.yml already. Should I change it to `Int` or is it okay?

https://gqlgen.com/reference/scalars/#int64
"Since the GraphQL spec identifies Int as a signed 32-bit integer, gqlgen provides an Int64 scalar to represent 64-bit integers. Anywhere you want to use a Go int and not deal with overflows, use Int64 in your schema."

amh4r (approved):

PR #3328, opened 2025-11-19
Insights query formatting and autocomplete

AndyInternet:
## Description
- Add SQL formatting (Prettify) support to SQLEditor component
  - keyboard shortcut
  - context menu option
  - kebob menu option
- Add autocomplete in editor
  - autocomplete event names
  - autocomplete data properties (when available)
  - autocomplete available clickhouse functions
- Add context menu (right-click menu) on query editor
  - Prettify SQL
  - Cut
  - Copy
  - Paste
  - Run query
  - Save query

## Motivation
It's got to be easy to use!

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/hooks/useSQLCompletionConfig.ts
+++ b/ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/hooks/useSQLCompletionConfig.ts
@@ -36,7 +36,7 @@ const KEYWORDS = [
 const TABLES = ['events'] as const;
 
 // Common columns available on the events table
-const COLUMNS = ['name', 'data'] as const;
+const COLUMNS = ['name', 'data', 'id', 'ts', 'v'] as const;
 
 // Convert ClickHouse functions to the format expected by autocomplete
 const CLICKHOUSE_FUNCTIONS = availableClickhouseFunctions.map((name) => ({


ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContextMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+        </DropdownMenuItem>
+        <div className="border-subtle my-1 border-t" />
+        <DropdownMenuItem
+          className="text-basis px-4 outline-none data-[disabled]:cursor-not-allowed data-[disabled]:opacity-50"
+          disabled={!contextMenu.hasSelection}
+          onSelect={onCut}
+        >
+          <span>Cut</span>
+          <span className="ml-auto">
+            <KeyboardShortcut color="text-gray-500" keys={['cmd', 'ctrl', 'X']} />
```
anafilipadealmeida: `text-gray-500` is not part of our color tokens, it should be using one from our [palette](https://github.com/inngest/inngest/blob/main/ui/packages/components/tailwind.config.ts) instead

AndyInternet: Thank you @anafilipadealmeida I'm adventure learning my way through our react setup right now. I updated with `text-muted`

ui/packages/components/src/SQLEditor/hooks/availableClickhouseFunctions.ts
```diff
@@ -0,0 +1,542 @@
+export const availableClickhouseFunctions = [
```
albertchae, line 1: do we no longer need https://github.com/inngest/inngest/tree/main/ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/functions then? This was added in https://github.com/inngest/inngest/pull/3079

FYI but there was some previous discussion on this https://inngest.slack.com/archives/C0953R5EP0R/p1759343860169569?thread_ts=1759268907.069619&cid=C0953R5EP0R
and this ticket https://linear.app/inngest/issue/EXE-367/improve-generation-of-clickhouse-function-allowlist

but TLDR copying from backend is fine, I don't think we ever want to serve these from an API. But I do need to make the process of scraping supported functions from ClickHouse into our backend parser and then into this frontend autocomplete repeatable. Also, at some point we might want to consider only autocompleting on a subset of functions since CH has a LOT of functions

ui/packages/components/src/SQLEditor/hooks/useSQLCompletions.ts
```diff
@@ -24,8 +24,58 @@ export function useSQLCompletions(config: SQLCompletionConfig) {
...
+                insertText: prop.name,
+                label: prop.name,
+                range,
+                detail: prop.type,
+              });
+            }
+          });
+          // Return early - only show data properties in this context
+          return { suggestions };
+        }
```
albertchae, line 75: aren't these mutually exclusive? might be clearer if the conditions are joined with an else

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/InsightsSQLEditor.tsx
```diff
@@ -3,13 +3,103 @@
...
+
+  const handlePrettifySQL = () => {
+    const editor = editorRef.current;
+    if (!editor) return;
+    editor.getAction('editor.action.formatDocument')?.run();
+  };
+
+  const handleRunQuery = () => {
+    runQuery();
+  };
```
albertchae, line 96: probably a question for @sanjanaladdha but it could be weird experience compared to other similar products to run the entire query on a highlight instead of just the selected query

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContextMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <DropdownMenuContent
+        align="start"
+        style={{
+          position: 'fixed',
+          left: `${contextMenu.x}px`,
+          top: `${contextMenu.y}px`,
+        }}
+      >
+        <DropdownMenuItem className="text-basis px-4 outline-none" onSelect={onPrettifySQL}>
+          <span>Prettify SQL</span>
```
albertchae, line 75: nit: retool, clickhouse, supabase use Format for this action. Supabase is also inconsistent and uses Prettify SQL in one of their buttons, but Format document for the context menu. Neon surprisingly doesn't have this feature afaict?

ui/packages/components/src/SQLEditor/constants.ts
```diff
@@ -36,12 +36,19 @@ export const EDITOR_OPTIONS: editor.IEditorOptions = {
...
+    comments: false,
+    strings: true,
+  },
   suggest: {
     showFunctions: true,
     showKeywords: true,
     showSnippets: true,
     showWords: false,
   },
+  suggestOnTriggerCharacters: true,
```
albertchae, line 50: this is what we're using to trigger without ctrl+space right?

AndyInternet: yes

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContextMenu.tsx
AndyInternet: I had it as Format until I saw the designs. Let's roll it out with Prettify and get some feedback.

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/InsightsSQLEditor.tsx
AndyInternet: This was my bad. I threw it in at the last minute so I just triggered it instead of running the selected portion. That's a fantastic idea though! We could also do something similar to TablePlus and have two keyboard shortcuts, one for run all and one for run selected.

sanjanaladdha:
> weird that pnpm-lcok.yaml changed so much with the smallest package.json change 🤔
> 
> Nothing blocking but some cleanup stuff we can follow up on.
> 
> Also it looks like we don't show the kebab menu until the query is saved. Maybe we should revisit that now that we have a format action @sanjanaladdha ?

Yeah this is a good catch. Now that we have the format action there, it can stay

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContextMenu.tsx
sanjanaladdha: I am okay with either in this case!

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/InsightsSQLEditor.tsx
sanjanaladdha: I didn't want to complicate the menu bar because we were focusing more on the formatting side of things. but if we wanna go deeper into this, makes sense to have run highlighted and run all separated. Seems like this is an existing pattern as well

albertchae: maybe we should just make it clear it's "run all" in the right click menu for now and have a separate ticket for the run highlighted interaction

PR #3330, opened 2025-11-19
Add metadata spans, REST API, and opcode plumbing

rhino1998:
## Description
This adds new span metadata spans, an endpoint to add requests, and plumbing for extracting metadata from extended trace spans.

This also includes new queries to retrieve spans by runID, stepID, step attempt, and span ID.

This also includes basic plumbing to put metadata behind a feature flag.

The feature flag defaults to `false` which makes this safe to merge & then add the featureflag handling in monorepo later.

## Motivation
This allows users to tag runs/steps/step executions/extended trace spans with custom metadata as well as the basic plumbing necessary to automatically extract metadata from extended trace spans.

- [ ] Docs

## Checklist

EXE-942, EXE-943

### DIFF
--- a/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
+++ b/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
@@ -466,7 +466,7 @@ SELECT
   )) AS span_fragments
 FROM spans
 WHERE run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND account_id = sqlc.arg(account_id) AND (parent_span_id IS NULL OR parent_span_id = '0000000000000000')
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 ORDER BY start_time ASC
 LIMIT 1;
 
@@ -491,12 +491,12 @@ WHERE span_id IN (
     parent_span_id
   FROM spans execSpans
   WHERE execSpans.run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND execSpans.account_id = sqlc.arg(account_id)
-  GROUP BY dynamic_span_id
-  HAVING SUM(attributes->>'$."_inngest.step.id"' = sqlc.arg(step_id)::text) > 0
-  ORDER BY start_time
+  GROUP BY dynamic_span_id, parent_span_id
+  HAVING SUM(((attributes#>>'{}')::json->>'_inngest.step.id' = sqlc.arg(step_id)::text)::int) > 0
+  ORDER BY MIN(start_time)
   LIMIT 1
 )
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 ORDER BY start_time ASC
 LIMIT 1;
 
@@ -517,11 +517,11 @@ SELECT
   )) AS span_fragments
 FROM spans
 WHERE run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND account_id = sqlc.arg(account_id)
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 HAVING
-  SUM(attributes->>'$."_inngest.step.id"' = sqlc.arg(step_id)::text) > 0
+  SUM(((attributes#>>'{}')::json->>'_inngest.step.id' = sqlc.arg(step_id)::text)::int) > 0
 AND
-  SUM(attributes->>'$."_inngest.step.attempt"' = sqlc.arg(step_attempt)::bigint) > 0
+  SUM(((((attributes#>>'{}')::json->>'_inngest.step.attempt')::bigint) = sqlc.arg(step_attempt)::bigint)::int) > 0
 ORDER BY start_time ASC
 LIMIT 1;
 
@@ -542,9 +542,9 @@ SELECT
   )) AS span_fragments
 FROM spans b
 WHERE b.run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND b.account_id = sqlc.arg(account_id)
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 HAVING
-  SUM(attributes->>'$."_inngest.step.id"' = sqlc.arg(step_id)::text) > 0
+  SUM(((attributes#>>'{}')::json->>'_inngest.step.id' = sqlc.arg(step_id)::text)::int) > 0
 ORDER BY start_time DESC
 LIMIT 1;
 
@@ -565,6 +565,6 @@ SELECT
   )) AS span_fragments
 FROM spans
 WHERE run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND span_id = sqlc.arg(span_id) AND account_id = sqlc.arg(account_id)
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 ORDER BY start_time ASC
 LIMIT 1;
--- a/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql.go
+++ b/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql.go
@@ -559,11 +559,11 @@ SELECT
   )) AS span_fragments
 FROM spans
 WHERE run_id = CAST($1 AS CHAR(26)) AND account_id = $2
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 HAVING
-  SUM(attributes->>'$."_inngest.step.id"' = $3::text) > 0
+  SUM(((attributes#>>'{}')::json->>'_inngest.step.id' = $3::text)::int) > 0
 AND
-  SUM(attributes->>'$."_inngest.step.attempt"' = $4::bigint) > 0
+  SUM(((((attributes#>>'{}')::json->>'_inngest.step.attempt')::bigint) = $4::bigint)::int) > 0
 ORDER BY start_time ASC
 LIMIT 1
 `
@@ -1025,9 +1025,9 @@ SELECT
   )) AS span_fragments
 FROM spans b
 WHERE b.run_id = CAST($1 AS CHAR(26)) AND b.account_id = $2
-GROUP BY dynamic_span_id
+GROUP BY dynamic_span_id, run_id, trace_id, parent_span_id
 HAVING
-  SUM(attributes->>'$."_inngest.step.id"' = $3::text) > 0
+  SUM(((attributes#>>'{}')::json->>'_inngest.step.id' = $3::text)::int) > 0
 ORDER BY start_time DESC
 LIMIT 1
 `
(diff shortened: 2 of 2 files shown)


pkg/devserver/devserver.go
```diff
@@ -295,6 +296,7 @@ func start(ctx context.Context, opts StartOpts) error {
 		Continuations: true,
 	}
 	enableKeyQueues := os.Getenv("EXPERIMENTAL_KEY_QUEUES_ENABLE") == "true"
+	enableStepMetadata := os.Getenv("EXPERIMENTAL_STEP_METADATA") == "true"
```
jpwilliams, line 303: I think this is good for the first merge to be safe, but noting that we should be able to fast follow by always enabling this in local dev - Cloud can still be protected though.

pkg/execution/executor/executor.go
```diff
@@ -396,6 +397,24 @@ type ExecutorRealtimeConfig struct {
 	PublishURL string
 }
 
+// AllowStepMetadata determines if key queues should be enabled for the account
```
jpwilliams: Comment 🙂

pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
```diff
@@ -447,3 +447,122 @@ FROM spans
 WHERE span_id IN (SELECT UNNEST(sqlc.slice('ids')::TEXT[]))
 LIMIT 2;
 
+-- name: GetRunSpanByRunID :one
+SELECT
+  run_id,
+  trace_id,
+  dynamic_span_id,
+  MIN(start_time) as start_time,
+  MAX(end_time) AS end_time,
```
jpwilliams, line 458: Are these still the start/end times? Now that we have metadata that can update things after completion, would this skew the end time?

pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
```diff
@@ -447,3 +447,122 @@ FROM spans
...
+  parent_span_id,
+  json_agg(json_build_object(
+    'name', name,
+    'attributes', attributes,
+    'links', links,
+    'output_span_id', CASE WHEN output IS NOT NULL THEN span_id ELSE NULL END,
+    'input_span_id', CASE WHEN input IS NOT NULL THEN span_id ELSE NULL END
+  )) AS span_fragments
+FROM spans
+WHERE run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND account_id = sqlc.arg(account_id) AND (parent_span_id IS NULL OR parent_span_id == '0000000000000000')
```
jpwilliams:
```suggestion
WHERE run_id = CAST(sqlc.arg(run_id) AS CHAR(26)) AND account_id = sqlc.arg(account_id) AND (parent_span_id IS NULL OR parent_span_id = '0000000000000000')
```

Double `=`?

Interesting that `make queries` still works 🤔

pkg/execution/executor/executor.go
```diff
@@ -1487,6 +1508,28 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 			return nil, err
 		}
 
+		if e.allowStepMetadata.Enabled(ctx, instance.Metadata().ID.Tenant.AccountID) {
+			for _, opcode := range resp.Generator {
+				for _, metadata := range opcode.Metadata {
+					if err := metadata.Validate(); err != nil {
+						l.Warn("invalid metadata in driver response", "error", err)
```
jpwilliams, line 1520: A note is that if we could add something like `l.With("step_metadata", true)` to all logs around this feature, it'd be a nice easy way to spot issues/warnings as we roll out.

pkg/execution/executor/executor.go
```diff
@@ -1487,6 +1508,28 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
+						continue
+					}
+
+					// TODO: validate metadata kinds & sizes
```
jpwilliams: Is this not done with `metadata.Validate()` above?

pkg/telemetry/metrics/counter.go
```diff
@@ -698,3 +698,12 @@ func IncrPausesLegacyDeletionCounter(ctx context.Context, opts CounterOpt) {
 		Tags:        opts.Tags,
 	})
 }
+
+func IncrMetadataSpansTotal(ctx context.Context, opts CounterOpt) {
+	RecordCounterMetric(ctx, 1, CounterOpt{
+		PkgName:     opts.PkgName,
+		MetricName:  "metadata_spans_total",
```
jpwilliams, line 705: Nice!

pkg/tracing/meta/serializers.go
```diff
@@ -253,6 +254,24 @@ func TruncatedStringAttr(key string, length int) attr[*string] {
 	}
 }
 
+func StringishAttr[T ~string](key string) attr[*T] {
```
jpwilliams, line 262: Can we have a comment for why this exists?

pkg/tracing/meta/serializers.go
```diff
@@ -488,6 +507,35 @@ func StepOpAttr(key string) attr[*enums.Opcode] {
 	}
 }
 
+func TextAttr[T any, TPtr interface {
```
jpwilliams, line 518: Same - comment for why this exists 🙂

pkg/tracing/meta/serializers_test.go
```diff
@@ -139,6 +142,26 @@ func TestAddAttr(t *testing.T) {
 			require.True(t, found, "StringAttr should be serialized correctly")
 		})
 
+		t.Run("StringishAttr", func(t *testing.T) {
```
jpwilliams, line 145: 🙌

pkg/tracing/metadata/kind.go
```diff
@@ -0,0 +1,43 @@
...
+	return string(k)
+}
+
+func (k Kind) IsInngest() bool {
+	return strings.HasPrefix(string(k), KindPrefixInngest)
+}
+
+func (k Kind) IsUser() bool {
+	return strings.HasPrefix(string(k), KindPrefixUserland)
+}
```
jpwilliams, line 35: 🙌

pkg/tracing/metadata/warning.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (e *WarningError) Error() string {
+	return e.Err.Error()
+}
+
+type Warnings map[string]error
+
+func (wm Warnings) Kind() Kind {
+	return KindInngestWarnings
+}
```
jpwilliams, line 23: Interesting, so is the thought here that every kind would have its own file here and then we'd have `Extract*()` methods?

pkg/api/apiv1/metadata.go
```diff
@@ -0,0 +1,187 @@
...
+
+		if target.StepAttempt == nil {
+			span, err = a.opts.TraceReader.GetStepSpanByStepID(ctx, runID, stepID, auth.AccountID(), auth.WorkspaceID())
+		} else if *target.StepAttempt < 0 {
+			span, err = a.opts.TraceReader.GetLatestExecutionSpanByStepID(ctx, runID, stepID, auth.AccountID(), auth.WorkspaceID())
+		} else {
+			span, err = a.opts.TraceReader.GetExecutionSpanByStepIDAndAttempt(ctx, runID, stepID, *target.StepAttempt, auth.AccountID(), auth.WorkspaceID())
+		}
+	default:
+		span, err = a.opts.TraceReader.GetSpanBySpanID(ctx, runID, *target.StepID, auth.AccountID(), auth.WorkspaceID())
```
jpwilliams: Should this be using `*target.StepID`?

pkg/tracing/tracer.go
```diff
@@ -58,6 +58,9 @@ type CreateSpanOptions struct {
 	EndTime            time.Time
 
 	Seed []byte
+
+	// DynamicSeed is optional and used for CreateOrUpdate operations
+	DynamicSeed []byte
```
jpwilliams, line 65: I think Tony added `Seed` here too - are they used for the same thing? Can we unify?

pkg/tracing/metadata/kind.go
```diff
@@ -0,0 +1,43 @@
...
+
+const (
+	MaxKindLength = 128
+
+	KindPrefixInngest  = "inngest."
+	KindPrefixUserland = "userland."
+
+	KindInngestAI       Kind = "inngest.ai"
+	KindInngestHTTP     Kind = "inngest.http"
+	KindInngestWarnings Kind = "inngest.warnings"
```
jpwilliams, line 22: Nice! I'd love to see these defined once somewhere alongside types etc

pkg/tracing/metadata/warning.go
rhino1998: Most extractors will for extended trace spans, so they'll implement `SpanExtractor`, but yeah, although placing the other extractors in a subpackage might make more sense to prevent circular dependency issues.

pkg/tracing/tracer.go
rhino1998: Seed generates both the span ID and the trace ID while DynamicSeed leaves trace ID to the parent and span ID random but generates a deterministic dynamic span ID (so that we can upsert spans)

pkg/execution/executor/executor.go
rhino1998: All `metadata.Validate` does is check Kind length so far. It doesn't check for if the kind is allowed or any sort of size limit checks.

pkg/api/apiv1/metadata.go
rhino1998: Good catch [a7effe5](https://github.com/inngest/inngest/pull/3330/commits/a7effe52fa4a547eac6b4592eec8e6b67e62b2b2)

pkg/execution/executor/executor.go
rhino1998: I added a more explicit TODO [271712c](https://github.com/inngest/inngest/pull/3330/commits/271712c0b92eba46a5ce34699d4e81cb614f29e7)

pkg/execution/executor/executor.go
rhino1998: [a4b8d95](https://github.com/inngest/inngest/pull/3330/commits/a4b8d95e00a161fc95c781008c369c890fab1247)

pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
rhino1998: In all these queries, we're only checking for a single span grouped by dynamic span ID, so unless a new EXTEND span (a call to UpdateSpan) is created at the relevant level, that shouldn't be an issue.

pkg/tracing/meta/serializers.go
rhino1998: [b9a4687](https://github.com/inngest/inngest/pull/3330/commits/b9a46871cc6761cdc8cc012c3d0f84e725293ca0)

pkg/tracing/meta/serializers.go
rhino1998: [b9a4687](https://github.com/inngest/inngest/pull/3330/commits/b9a46871cc6761cdc8cc012c3d0f84e725293ca0)

pkg/tracing/metadata/kind.go
rhino1998: I'll move them in #3346

pkg/tracing/tracer.go
rhino1998: Added [72dab9e](https://github.com/inngest/inngest/pull/3330/commits/72dab9e8d8119bb87b9b3f341b069821a181e319) to explain the distinction

PR #3344, opened 2025-11-24
Distinguish between self hosted and dev mode

jakobevangelista:
## Description

This pr introduces some changes to distinguish between self-hosted mode and dev server mode

## Motivation
There is some user reports that causes confusion because when the dev server is in self hosted mode, there are minimal differences, this pr adds more diffeerences to distinguish between the 2 modes

- [ ] Docs

## Checklist


cmanou:
I believe this breaks feature flags in self hosted mode since it was using that url to fetch them https://github.com/inngest/inngest/blob/main/ui/apps/dev-server-ui/src/hooks/useFeatureFlags.ts

PR #3345, opened 2025-11-24
Add step metadata GQL plumbing

rhino1998:
## Description

This adds span metadata rollup & plumbing through GQL.

Currently this plumbs metadata effectively as a `map[string]json.RawMessage` in GQL. Should we have full schemas for the typed metadata so that individual subfields can be queried directly in GQL? Or is having an appropriate type in Typescript sufficient? We can represent the structured metadata with a sufficiently complex union type like so 

```typescript
export type SpanMetadata =
  | SpanMetadataInngestAI
  | SpanMetadataUserland;

export type SpanMetadataInngestAI = {
  kind: 'inngest.ai';
  values: {
    url: string;
    model: string;
    system: string;
    operation_name: string;

    id?: string;
    tokens_in?: number;
    tokens_out?: number;
  };
}

export type SpanMetadataUserland = {
  kind: `userland.${string}`;
  values: Record<string, any>;
}

```

## Motivation

We want to expose metadata in the UI eventually and we're already getting all the metadata spans during the run trace query, so just roll up metadata spans and attach them to their parent spans.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/coreapi/gql.schema.graphql
+++ b/pkg/coreapi/gql.schema.graphql
@@ -11,6 +11,8 @@ scalar UUID
 scalar Bytes
 scalar Unknown
 scalar Int64
+scalar SpanMetadataKind
+scalar SpanMetadataValues
 
 "The pagination information in a connection."
 type PageInfo {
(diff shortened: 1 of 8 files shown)


pkg/coreapi/gql.schema.graphql
```diff
@@ -11,6 +11,8 @@ scalar UUID
 scalar Bytes
 scalar Unknown
 scalar Int64
+scalar SpanMetadataKind
```
rhino1998, line 14: Should this just be a string?

PR #3351, opened 2025-11-25
andy/insights fix and tweak bundle

AndyInternet:
## Description

- **EXE-1016: fix autocomplete with forward slash by replacing monaco default word boundary detection with custom logic**
- **PDES-36: Load all event types in autocomplete menu - add pagination support to pull more than 40 - dynamically call GQL endpoint with search string using nameSearch param - add useCache hook to cache all results - add debouncer so GQL is only called 200ms after final keypress**
- **EXE-895: copy schema explorer value - right-justify type to bring inline with design - add row hover with copy icon - onClick copies to clipboard - toast to validate copy success**
- **EXE-1020: fix HistoryButton hover bug**
- **EXE-889: update Resizable component to have 1px borders to match design**
- **EXE-1008: make kebob menu always visible, change prettify to format sql.**

## Motivation
Fixes and tweaks for insights

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/SQLEditor/hooks/useCache.ts
+++ b/ui/packages/components/src/SQLEditor/hooks/useCache.ts
@@ -1,4 +1,4 @@
-import { useRef } from 'react';
+import { useMemo, useRef } from 'react';
 
 export interface CacheEntry<T> {
   data: T;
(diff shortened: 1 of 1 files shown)


ui/packages/components/src/SQLEditor/hooks/useCache.ts
```diff
@@ -0,0 +1,90 @@
...
+ * }
+ *
+ * // Clear all cache entries
+ * cache.clear();
+ */
+export function useCache<T>(options: CacheOptions): Cache<T> {
+  const { ttl } = options;
+  const cacheRef = useRef<Map<string, CacheEntry<T>>>(new Map());
+
+  return {
```
anafilipadealmeida: We should probably memoize the return object, something like `return useMemo(() => ({`

PR #3356, opened 2025-11-29
Use emission interval to compute variance instead of original period_ms

albertchae:
## Description

Comparing our math here to throttled/throttled
https://github.com/throttled/throttled/blob/d5ee505c72a471702dfb7eca8e156d592bf9a68c/rate.go#L159-L161

If `variance` in our system means the same thing as `delayVariationTolerance`
in throttled, I believe we are not calculating it correctly because we are
using the raw period_ms instead of the normalized value computed off period and limit, e.g.
https://github.com/throttled/throttled/blob/d5ee505c72a471702dfb7eca8e156d592bf9a68c/rate.go#L117-L118

An example scenario that seems to be computing it incorrectly

Config A: 100 requests over 10s, burst 1
Config B: 10 requests over 1s, burst 1

Computing emission:
A: 10000 ms / 100 = 100ms
B:  1000 ms /  10 = 100ms

However, under the previous math, our variance would look like
A: 10000 ms * 1 = 10000 ms
B:  1000 ms * 1 =  1000 ms

(We also seem to be interpreting burst differently from throttled, in our case it's the total burstable requests allowed
whereas in throttled it's additional burstable requests, but that's of less consequence)

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/ratelimit/ratelimit_lua_test.go
+++ b/pkg/execution/ratelimit/ratelimit_lua_test.go
@@ -16,7 +16,7 @@ import (
 
 const prefix = "{rl}:"
 
-// initRedis creates both miniredis/rueidis for Lua, throttled store, and fake clock
+// initRedis creates miniredis/rueidis for Lua and fake clock
 func initRedis(t *testing.T) (*miniredis.Miniredis, rueidis.Client, clockwork.FakeClock) {
 	r := miniredis.RunT(t)
 


pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -7,7 +7,7 @@ local function gcra(key, now_ms, period_ms, limit, burst)
 
 	local emission  = period_ms / math.max(limit, 1)   -- how frequently we can admit new requests
 	local increment = emission * cost         -- this request's time delta
-	local variance  = period_ms * (math.max(burst, 1)) -- variance takes into account bursts
+	local variance  = emission * (math.max(burst, 1)) -- variance takes into account bursts
```
albertchae, line 10: it looks like we call this by a different name and compute this more closely to how throttled does it here (including the way they define burst as additional burst)

https://github.com/inngest/inngest/blob/1c6b030e0a2eb755e548ef33ef76e7a7bafa4c11/pkg/execution/state/redis_state/lua/includes/gcra.lua#L61

https://github.com/inngest/inngest/blob/1c6b030e0a2eb755e548ef33ef76e7a7bafa4c11/pkg/constraintapi/lua/check.lua#L167

albertchae: hm but I see the existing code using period matches what is posted about here https://blog.ian.stapletoncordas.co/2018/12/understanding-generic-cell-rate-limiting 🤔

albertchae: verified both variance and burst here https://github.com/inngest/inngest/pull/3383

albertchae:
Using this bug report repro https://github.com/Spydr-Digital/inngest-throttling-test
## Before
Screenshot 2025-12-05 at 9 43 57 AM



## After
image

PR #3384, opened 2025-12-05
EXE-1045: tanstack compatible versions of our shared components

jacobheric:
## Description

Like we did for `dev-server-ui` conversion, I've created "New" tanstack compatible versions of our shared components that currently rely on next. That way we can merge these in ahead of time without disrupting production and make the dashboard conversion PR a bit more manageable.  Once the conversion is complete I'll follow up with a PR that merges the "New" tanstack components back onto the old next ones and update imports. 

## Motivation
Phased conversion role out to minimize disruption and reduce PR size. 

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/utils/colors.ts
+++ b/ui/packages/components/src/utils/colors.ts
@@ -8,6 +8,15 @@ export const rgbToHex = (r: number, g: number, b: number): string =>
     .join('');
 
 export const resolveColor = (
+  colorValue: any,
+  isDark: boolean,
+  defaultColor: string = '#f6f6f6' // carbon 50
+): string =>
+  !colorValue || typeof colorValue !== 'string'
+    ? defaultColor
+    : resolver(colorValue, isDark, defaultColor);
+
+export const resolver = (
   colorValue: string,
   isDark: boolean,
   defaultColor: string = '#f6f6f6' // carbon 50
--- a/ui/packages/components/src/utils/monaco.ts
+++ b/ui/packages/components/src/utils/monaco.ts
@@ -1,6 +1,4 @@
-'use-client';
-
-import { languages } from 'monaco-editor';
+import type { languages } from 'monaco-editor';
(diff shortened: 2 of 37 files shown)


ui/packages/components/src/utils/colors.ts
```diff
@@ -8,6 +8,15 @@ export const rgbToHex = (r: number, g: number, b: number): string =>
     .join('');
 
 export const resolveColor = (
+  colorValue: any,
```
jacobheric, line 11: These last two files contain the only inline changes to existing components. This one here is a backwards compatible change.

ui/packages/components/src/utils/monaco.ts
```diff
@@ -1,6 +1,4 @@
-'use-client';
```
jacobheric, line 1: Safe to delete "use-client" as it's not a valid next directive. Type import change is for our more modern/strict typescript settings in tanstack.

amh4r:
Instead of prefixing with `New`, what are your thoughts on prefixing the old stuff with `Old`? This isn't a blocking suggestion by any means, but with big refactors it can be nice to introduce the new stuff in its "final state". That way you don't need more changes to update the new stuff when the old stuff can be deleted: the new stuff is already good to go

amh4r (approved):

jacobheric:
> Instead of prefixing with `New`, what are your thoughts on prefixing the old stuff with `Old`? This isn't a blocking suggestion by any means, but with big refactors it can be nice to introduce the new stuff in its "final state". That way you don't need more changes to update the new stuff when the old stuff can be deleted: the new stuff is already good to go

It's a good suggestion, but I did already establish this convention as part of the `dev-server-ui` conversion, which shipped. So it would mean a pause now with a decent lift to do that refactor and regression test and get it through first. I'd rather avoid that side quest if I can help it. Big lift coming with the merge of the dashboard components and routes!

PR #3412, opened 2025-12-10
Add new GCRA implementation

BrunoScheufler:
## Description

This PR recreates the GCRA implementation from scratch based on https://github.com/throttled/throttled/blob/master/rate.go#L267-L271

This includes capacity, retry calculation, etc. out of the box, so we can replace gcraCapacity and gcraUpdate with read-only and read-write calls to it.

This should fix any outstanding bugs with the future standard throttle implementation.

What still needs to be done
- [x] More tests, particularly around compatibility mode
- [x] Figure out why throughput dropped when compatibility mode was enabled
     - Because the arrival rate in the queue is not uniform and GCRA request rate smoothing would reject requests in the same emission interval. Increasing the internal burst used by GCRA to admit more than a single request (similar to key queues/Constraint API) fixed this
- [x] Add tests to ensure we never exceed rate with new implementation, with and without burst

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/lua_test.go
+++ b/pkg/execution/state/redis_state/lua_test.go
@@ -150,7 +150,7 @@ func TestNewGCRAScript(t *testing.T) {
 		limit := 10
 		burst := 0
 
-		// Read initial capacity
+		// First request should be admitted
 		res := runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -160,6 +160,7 @@ func TestNewGCRAScript(t *testing.T) {
 			quantity: 1,
 		})
 
+		require.False(t, res.Limited)
 		require.Equal(t, (6 * time.Second).Milliseconds(), res.EmissionInterval)
 		require.Equal(t, res.TAT, clock.Now().UnixMilli())
 		require.Equal(t, res.NewTAT, clock.Now().Add(6*time.Second).UnixMilli())
@@ -199,6 +200,7 @@ func TestNewGCRAScript(t *testing.T) {
 			quantity: 2,
 		})
 
+		require.False(t, res.Limited)
 		require.Equal(t, (6 * time.Second).Milliseconds(), res.EmissionInterval)
 		require.Equal(t, res.TAT, clock.Now().UnixMilli())
 		require.Equal(t, res.NewTAT, clock.Now().Add(2*6*time.Second).UnixMilli())
@@ -229,7 +231,7 @@ func TestNewGCRAScript(t *testing.T) {
 		limit := 10
 		burst := 0
 
-		// Read initial capacity
+		// First request should be allowed
 		res := runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -242,10 +244,12 @@ func TestNewGCRAScript(t *testing.T) {
 		require.Equal(t, res.TAT, clock.Now().UnixMilli())
 		require.Equal(t, res.NewTAT, clock.Now().Add(1*6*time.Second).UnixMilli())
 
+		require.False(t, res.Limited)
 		require.Equal(t, 1, res.Limit)
 		require.Equal(t, 0, res.Remaining)
 		require.Equal(t, clock.Now().Add(6*time.Second).UnixMilli(), res.RetryAtMS)
 
+		// Second request should be limited
 		res = runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -263,6 +267,7 @@ func TestNewGCRAScript(t *testing.T) {
 		require.Equal(t, clock.Now().Add(2*6*time.Second).Add(-6*time.Second).UnixMilli(), res.AllowAt)
 		require.Equal(t, -(6 * time.Second).Milliseconds(), res.Diff)
 
+		require.True(t, res.Limited)
 		require.Equal(t, 1, res.Limit)
 		require.Equal(t, 0, res.Remaining)
 		// Accounts for burst
@@ -276,7 +281,7 @@ func TestNewGCRAScript(t *testing.T) {
 
 		clock := clockwork.NewFakeClock()
 
-		_, rc := initRedis(t)
+		r, rc := initRedis(t)
 		defer rc.Close()
 
 		key := "test"
@@ -285,7 +290,7 @@ func TestNewGCRAScript(t *testing.T) {
 		limit := 1
 		burst := 0
 
-		// Read initial capacity
+		// First request should work
 		res := runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -295,6 +300,7 @@ func TestNewGCRAScript(t *testing.T) {
 			quantity: 1,
 		})
 
+		require.False(t, res.Limited)
 		require.Equal(t, res.TAT, clock.Now().UnixMilli())
 		require.Equal(t, res.NewTAT, clock.Now().Add(24*time.Hour).UnixMilli())
 
@@ -302,6 +308,7 @@ func TestNewGCRAScript(t *testing.T) {
 		require.Equal(t, 0, res.Remaining)
 		require.Equal(t, clock.Now().Add(24*time.Hour).UnixMilli(), res.RetryAtMS)
 
+		// Second request should be limited
 		res = runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -319,11 +326,42 @@ func TestNewGCRAScript(t *testing.T) {
 		require.Equal(t, clock.Now().Add(24*time.Hour).Add(1*24*time.Hour).Add(-24*time.Hour).UnixMilli(), res.AllowAt)
 		require.Equal(t, -(24 * time.Hour).Milliseconds(), res.Diff)
 
+		require.True(t, res.Limited)
 		require.Equal(t, 1, res.Limit)
 		require.Equal(t, 0, res.Remaining)
 		require.Equal(t, 24*time.Hour, time.Duration(res.ResetAfterMS)*time.Millisecond)
 		require.Equal(t, 24*time.Hour, time.Duration(res.RetryAfterMS)*time.Millisecond)
 		require.Equal(t, clock.Now().Add(24*time.Hour).UnixMilli(), res.RetryAtMS)
+
+		// Waiting should reduce ttl but still reject
+
+		clock.Advance(4 * time.Hour)
+		r.FastForward(4 * time.Hour)
+		r.SetTime(clock.Now())
+
+		res = runScript(t, rc, gcraScriptOptions{
+			key:      key,
+			now:      clock.Now(),
+			period:   period,
+			limit:    limit,
+			burst:    burst,
+			quantity: 1,
+		})
+
+		require.Equal(t, (24 * time.Hour).Milliseconds(), res.EmissionInterval)
+		require.Equal(t, res.TAT, clock.Now().Add(20*time.Hour).UnixMilli())
+		require.Equal(t, res.NewTAT, clock.Now().Add(20*time.Hour+24*time.Hour).UnixMilli())
+		require.Equal(t, (24 * time.Hour * 1).Milliseconds(), res.DVT)
+		require.Equal(t, (1 * 24 * time.Hour).Milliseconds(), res.Increment)
+		require.Equal(t, clock.Now().Add(20*time.Hour).Add(1*24*time.Hour).Add(-24*time.Hour).UnixMilli(), res.AllowAt)
+		require.Equal(t, -(20 * time.Hour).Milliseconds(), res.Diff)
+
+		require.True(t, res.Limited)
+		require.Equal(t, 1, res.Limit)
+		require.Equal(t, 0, res.Remaining)
+		require.Equal(t, 20*time.Hour, time.Duration(res.ResetAfterMS)*time.Millisecond)
+		require.Equal(t, 20*time.Hour, time.Duration(res.RetryAfterMS)*time.Millisecond)
+		require.Equal(t, clock.Now().Add(20*time.Hour).UnixMilli(), res.RetryAtMS)
 	})
 
 	t.Run("3000 requests every minute should work", func(t *testing.T) {
@@ -340,7 +378,7 @@ func TestNewGCRAScript(t *testing.T) {
 		limit := 3000
 		burst := 0
 
-		// Read initial capacity
+		// First request should be allowed
 		res := runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
@@ -361,6 +399,7 @@ func TestNewGCRAScript(t *testing.T) {
 		require.Equal(t, time.Duration(0).Milliseconds(), res.Diff)
 
 		// Since we don't have a burst, only one request will be allowed every 20ms
+		require.False(t, res.Limited)
 		require.Equal(t, 1, res.Limit)
 		require.Equal(t, 0, res.Remaining)
 
@@ -383,7 +422,7 @@ func TestNewGCRAScript(t *testing.T) {
 		limit := 3000
 		burst := 1
 
-		// Read initial capacity
+		// First request should be allowed
 		res := runScript(t, rc, gcraScriptOptions{
 			key:      key,
 			now:      clock.Now(),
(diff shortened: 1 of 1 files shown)


pkg/execution/state/redis_state/lua_test.go
```diff
@@ -251,30 +1061,28 @@ func TestLuaGCRA(t *testing.T) {
 					capacityBefore:  8, // assume 1 item got refilled
 					consumeCapacity: 0,
 					capacityAfter:   8,
-					retryAt:         2 * time.Minute,
+					retryAt:         6 * time.Minute,
```
BrunoScheufler, line 1014: retryAt is now calculated as the time the next unit of capacity is refilled assuming that all capacity is consumed

Because we have capacity > 0, this now becomes the emission interval 6m.

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -251,30 +1061,28 @@ func TestLuaGCRA(t *testing.T) {
...
 				},
 				{
 					delay:           0,
 					capacityBefore:  9,
 					consumeCapacity: 9,
 					capacityAfter:   0,
-					// we are 15mins in (10 + 5 above) without consuming, so the next
-					// refill is expected in (6 + 6 + 6) - (10 + 5) = 3 mins
-					retryAt: 3 * time.Minute,
+					retryAt:         3 * time.Minute,
```
BrunoScheufler, line 1028: When there's a lack of capacity, we correctly calculate the retryAt as the time until the next request can be sent which depends on the current time and current tat (basically the diff between now and allow_at)

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
 
-  -- Convert the remaining time budget back into a number of tokens.
-  local capacity = math.floor(time_capacity_remain / emission)
+	return { res["allowed"], used_burst }
+end
 
-  -- The capacity cannot exceed the defined limit + burst.
-  local final_capacity = math.min(capacity, limit + burst)
+local function gcraUpdate(key, now_ms, period_ms, limit, burst, quantity)
+	local allowRefillingAll = limit + burst - 1
```
BrunoScheufler, line 118: For key queues, we need to be able to consume all capacity in a single backlogRefill call.

This does impact request smoothing within the period, but ensures we don't reject items even though there is capacity in the period, which is the higher priority.

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
-  -- calculate total capacity in time units
-  local total_capacity_time = emission * (limit + burst)
+-- performs gcra rate limiting for a given key.
+--
+-- Returns true on success, false if the key has been rate limited.
+local function gcra(key, now_ms, period_ms, limit, burst)
+	-- NOTE: we need to admit more than a single item every emission interval, as the queue
+	-- does not follow a uniform arrival rate and we would throttle the majority of queue items,
+	-- leading to significantly lower queue throughput
+	local maxBurst = limit + burst - 1
```
BrunoScheufler, line 108: GCRA with the default internal burst of 1 means only one request can be admitted every emission interval. Any additional requests will be rejected until we enter the next period.

The queue does not mirror the arrival rate of events, but instead groups multiple items into Lease calls sent around a similar time. If the window between distinct items is <= the emission interval, this may lead to rate smoothing throttling queue items. This then reduces the throughput, as we reject most items in a batch, requeue them 1s in the future (even though they could run in a couple milliseconds based on the emission interval), and retry.

To prevent reduced throughput after shipping the dvt calculation fix, we now allow a higher burst. While this impacts smoothing, users care most about hitting the limit every period, and not going over it. Enforcing uniform request arrival is a different challenge which we can expose as an explicit option in the future.

pkg/constraintapi/lua/acquire.lua
```diff
@@ -259,75 +259,106 @@ end
...
 		tat = tonumber(tat)
 	end
 
-	-- remaining capacity in time units
-	local time_capacity_remain = now_ms + total_capacity_time - tat
+	result["tat"] = tat
 
-	-- Convert the remaining time budget back into a number of tokens.
-	local capacity = math.floor(time_capacity_remain / emission)
+	local origQuantity = quantity
```
albertchae: worth copying over the comment from throttled?

https://github.com/throttled/throttled/blob/d5ee505c72a471702dfb7eca8e156d592bf9a68c/rate.go#L210

pkg/constraintapi/lua/acquire.lua
```diff
@@ -259,75 +259,106 @@ end
 ---@param period_ms integer
 ---@param limit integer
 ---@param burst integer
----@return integer[]
-local function throttleCapacity(key, now_ms, period_ms, limit, burst)
-	-- TODO: Reuse shared script
+---@param quantity integer
+local function throttle(key, now_ms, period_ms, limit, burst, quantity)
+	---@type { allowed: boolean, limit: integer?, retry_after: integer?, reset_after: integer?, remaining: integer? }
```
albertchae: how come this isn't just calling `applyGCRA`, it looks like all the implementation is the same?

pkg/constraintapi/lua/check.lua
```diff
@@ -173,41 +173,106 @@ end
 ---@param period_ms integer
 ---@param limit integer
 ---@param burst integer
----@return integer[]
-local function throttleCapacity(key, now_ms, period_ms, limit, burst)
-	-- TODO: Reuse shared script
+---@param quantity integer
+local function throttle(key, now_ms, period_ms, limit, burst, quantity)
+	---@type { allowed: boolean, limit: integer?, retry_after: integer?, reset_after: integer?, remaining: integer? }
```
albertchae: same comment about calling `applyGCRA`, looks like same implementation

pkg/execution/state/redis_state/lua/test/gcra_ratelimit.lua
```diff
@@ -3,11 +3,10 @@ local currentTime = tonumber(ARGV[2])
...
 local limit = tonumber(ARGV[4])
 local burst = tonumber(ARGV[5])
-local enableFix = tonumber(ARGV[6]) == 1
 
 -- $include(gcra.lua)
 
-local throttleResult = gcra(throttleKey, currentTime, period_ms, limit, burst, enableFix)
+local throttleResult = gcra(throttleKey, currentTime, period_ms, limit, burst)
 
 -- not allowed
```
albertchae, line 11: we could probably delete this file now that `pkg/execution/state/redis_state/lua/test/gcra.lua` exists and use that from the go test instead

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
-    -- NOTE: This fixes the delay variation tolerance calculation
-    variance = emission * (math.max(burst, 1))
-  end
-
-	-- fetch the theoretical arrival time for equally spaced requests
-	-- at exactly the rate limit
+-- applyGCRA runs GCRA
+local function applyGCRA(key, now_ms, period_ms, limit, burst, quantity)
+	---@type { allowed: boolean, limit: integer?, retry_after: integer?, reset_after: integer?, remaining: integer? }
+	local result = {}
```
albertchae: any possible performance concerns returning a table with more fields than the 2 element sequential table before? (maybe memory usage in lua?) 

also looks like the type annotation doesn't capture all the possible keys being set on the table, is it only capturing the ones that matter for actual use?

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
+
+	-- limit defines the maximum number of requests that can be admitted at once (irrespective of current usage)
+	result["limit"] = burst + 1
```
albertchae, line 7: somewhat confusing that `limit` in this result table isn't necessarily the value of the parameter `limit` (which is coming from function config, right?). I see the parameter  `burst` is being set to `limit + burst` from the function config by the caller. Can we name the key here something different?

Also whatever we name it to, we can make a variable and reuse it on the `dvt` calculation line below `local dvt = emission * (burst + 1)`

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
-	local new_tat = math.max(tat, now_ms) + increment -- add the request's cost to the theoretical arrival time.
-	local allow_at_ms = new_tat - variance            -- handle bursts.
-	local diff_ms = now_ms - allow_at_ms
+	result["tat"] = tat
 
-	if diff_ms < 0 then
-		return {false, false}
+	local origQuantity = quantity
+	if quantity == 0 then
+		quantity = 1
```
albertchae, line 34: is it worth copying over the comment from throttled? I remember being confused reading that code before seeing the comment

https://github.com/throttled/throttled/blob/d5ee505c72a471702dfb7eca8e156d592bf9a68c/rate.go#L210

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
+	-- limit defines the maximum number of requests that can be admitted at once (irrespective of current usage)
+	result["limit"] = burst + 1
+
+	-- emission defines the window size
+	local emission = period_ms / math.max(limit, 1)
+	result["ei"] = emission
+
+	-- retry_at is always computed under the assumption that all
+	-- remaining capacity is consumed
+	result["retry_at"] = now_ms + emission
```
albertchae, line 15: is this true? if we only increment by one emission isn't that only 1 "unit"? Also didn't see an analogue to retry_at in throttled's code so couldn't compare

pkg/execution/state/redis_state/lua/includes/gcra.lua
albertchae: oh I see, it was matching the `Limit` field in the throttled implementation

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
-	local variance  = period_ms * (math.max(burst, 1)) -- variance takes into account bursts
-  if enableThrottleFix then
-    -- NOTE: This fixes the delay variation tolerance calculation
-    variance = emission * (math.max(burst, 1))
-  end
-
-	-- fetch the theoretical arrival time for equally spaced requests
-	-- at exactly the rate limit
+-- applyGCRA runs GCRA
+local function applyGCRA(key, now_ms, period_ms, limit, burst, quantity)
```
albertchae, line 2: compared this line by line to https://github.com/throttled/throttled/blob/d5ee505c72a471702dfb7eca8e156d592bf9a68c/rate.go#L158-L273, besides slightly different ways we refer to period and limit, I am convinced this is a correct Lua reimplementation!

pkg/execution/state/redis_state/lua/includes/gcra.lua
albertchae: We discussed in DMs but posting here for posterity, this also means that configurations that previously resulted in the same emission interval no longer behave the same, e.g.

limit: 1000, period: 10s
limit: 100, period: 1s

whereas the intent before was to smooth these to the same rate, they will have different bursty behavior. And I agree with this
> Enforcing uniform request arrival is a different challenge which we can expose as an explicit option in the future.

pkg/execution/state/redis_state/lua/includes/gcra.lua
```diff
@@ -1,103 +1,128 @@
...
+			result["ttl"] = ttl
+		end
+
+		if origQuantity > 0 then
+			-- if we did want to update, we got limited
+			local next = dvt - ttl
+			result["next"] = next
+			if next > -emission then
+				result["remaining"] = math.floor(next / emission)
+			end
```
albertchae: if we got limited, won't remaining always be 0?

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -16,6 +17,814 @@ import (
...
+		require.Equal(t, (6 * time.Second).Milliseconds(), res.EmissionInterval)
+		require.Equal(t, res.TAT, clock.Now().UnixMilli())
+		require.Equal(t, res.NewTAT, clock.Now().Add(6*time.Second).UnixMilli())
+		require.Equal(t, (6 * time.Second).Milliseconds(), res.DVT)
+		require.Equal(t, (6 * time.Second).Milliseconds(), res.Increment)
+		require.Equal(t, clock.Now().UnixMilli(), res.AllowAt)
+		require.Equal(t, int64(0), res.Diff)
+
+		require.Equal(t, 1, res.Limit)
+		require.Equal(t, 1, res.Remaining)
```
albertchae, line 133: I'm somewhat surprised these aren't both 10, (or 9?) because we didn't use any quantity

albertchae: oh looks like `res.Limit` will always be 1? because this test is just testing the direct behavior copied from throttle which computes 1 per emission interval

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -16,6 +17,814 @@ import (
...
+		// Read initial capacity
+		res := runScript(t, rc, gcraScriptOptions{
+			key:      key,
+			now:      clock.Now(),
+			period:   period,
+			limit:    limit,
+			burst:    burst,
+			quantity: 1,
+		})
+
```
albertchae, line 162: should you check `require.False(t, res.Limited)` here? some of the tests below aren't checking for it either

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -16,6 +17,814 @@ import (
...
+		res = runScript(t, rc, gcraScriptOptions{
+			key:      key,
+			now:      clock.Now(),
+			period:   period,
+			limit:    limit,
+			burst:    burst,
+			quantity: 1,
+		})
+
```
albertchae, line 261: would it be `require.True(t, res.Limited)` here

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -16,6 +17,814 @@ import (
...
+		require.Equal(t, res.TAT, clock.Now().UnixMilli())
+		require.Equal(t, res.NewTAT, clock.Now().Add(24*time.Hour).UnixMilli())
+
+		require.Equal(t, 1, res.Limit)
+		require.Equal(t, 0, res.Remaining)
+		require.Equal(t, clock.Now().Add(24*time.Hour).UnixMilli(), res.RetryAtMS)
+
```
albertchae, line 310: don't you need to `clock.Advance` and set the redis clock here?

pkg/execution/state/redis_state/lua_test.go
```diff
@@ -16,6 +17,814 @@ import (
...
+		require.Equal(t, 5, res.Limit)
+		require.Equal(t, 0, res.Remaining)
+
+		require.Equal(t, int64(0), res.RetryAfterMS)
+		require.Equal(t, time.Hour.Milliseconds(), res.TTL)
+		require.Equal(t, time.Hour.Milliseconds(), res.ResetAfterMS)
+	})
+
+	t.Run("compatibility mode should have incorrectly large dvt", func(t *testing.T) {
+		t.Skip("compatibility mode was removed")
```
albertchae: remove?

pkg/constraintapi/lua/acquire.lua
BrunoScheufler: The plan is to make the Constraint API become the single source of truth for this, and it's fully isolated/in a different directory, so we cannot easily include other scripts the way we do for the queue

BrunoScheufler: Sharing Lua scripts is done in a pretty ~hacky~ brilliant way using go:embed flags, so currently this only works for all Lua scripts for the same codebase area (queue, constraint api, connect).

pkg/constraintapi/lua/check.lua
BrunoScheufler: Answered above

pkg/execution/state/redis_state/lua/test/gcra_ratelimit.lua
BrunoScheufler: This is still nice because we can directly test all behavior changes made in gcra() to verify the "legacy" implementation is correctly applying the default burst + limit, etc.!

BrunoScheufler: We can remove this when we drop all constraint handling from the queue

pkg/execution/state/redis_state/lua/includes/gcra.lua
BrunoScheufler: There will be higher memory usage in Lua but since we only use the table internally and do _not_ send it to the client, this should be fairly performant.

BrunoScheufler: Good note on the type annotation, forgot to update it along the way.

pkg/execution/state/redis_state/lua/includes/gcra.lua
BrunoScheufler: Done, good idea!

pkg/execution/state/redis_state/lua/includes/gcra.lua
BrunoScheufler: This is more in line with how we handled the key queue implementation and works as long as there is remaining capacity. If there's none, the retry after value may be smaller (e.g. 2 out of 5 seconds until the next request can be sent have already passed, so retry after should be 3 seconds)

pkg/execution/state/redis_state/lua/includes/gcra.lua
BrunoScheufler: Good catch, yeah!

pkg/execution/state/redis_state/lua_test.go
BrunoScheufler: Yes, exactly

pkg/execution/state/redis_state/lua_test.go
BrunoScheufler: Yep!

pkg/execution/state/redis_state/lua_test.go
BrunoScheufler: Added one more case to cover this

pkg/constraintapi/lua/acquire.lua
albertchae: cool i assumed something like this was the reason

PR #3430, opened 2025-12-16
Switch rate limit to use new GCRA implementation (with nanosecond precision)

BrunoScheufler:
## Description

This PR updates the rate limit implementation to the new GCRA implementation introduced in #3412, adjusted for nanosecond precision. This way, we can maintain two mostly equal implementations for different levels of precision.

Since the previous rate limit implementation was mostly aligned with the throttled package, the behavior should not change.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/constraintapi/lua/helper/gcra.lua
+++ b/pkg/constraintapi/lua/helper/gcra.lua
@@ -5,6 +5,7 @@
 ---@param burst integer
 ---@param quantity integer
 local function rateLimit(key, now_ns, period_ns, limit, burst, quantity)
+	-- Ensure limit is always >= 1
 	limit = math.max(limit, 1)
 
 	---@type { limit: integer, ei: number, retry_at: number, dvt: number, tat: number, inc: number, ntat: number, aat: number, diff: number, retry_after: integer?, u: number, next: number?, remaining: integer?, reset_after: integer?, limited: boolean? }
@@ -117,6 +118,9 @@ end
 ---@param burst integer
 ---@param quantity integer
 local function throttle(key, now_ms, period_ms, limit, burst, quantity)
+	-- Ensure limit is always >= 1
+	limit = math.max(limit, 1)
+
 	---@type { limit: integer, ei: number, retry_at: number, dvt: number, tat: number, inc: number, ntat: number, aat: number, diff: number, retry_after: integer?, u: number, next: number?, remaining: integer?, reset_after: integer?, limited: boolean? }
 	local result = {}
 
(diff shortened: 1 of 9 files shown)


pkg/execution/ratelimit/ratelimit_lua_test.go
```diff
@@ -908,15 +832,20 @@ func TestLuaRateLimit_EdgeCases(t *testing.T) {
...
 		// So we test that our Lua implementation gracefully handles zero limits
-		// by immediately rate limiting (which is the logical behavior)
+		// by falling back to 1
 		r.SetTime(clock.Now())
 		res, err := limiter.RateLimit(ctx, "test-key", config, WithNow(clock.Now()))
 		require.NoError(t, err)
 		t.Logf("Lua with zero limit: limited=%v, retry=%v", res.Limited, res.RetryAfter)
 
 		// Zero limit should immediately rate limit
+		require.False(t, res.Limited)
```
albertchae, line 842: delete/edit comment above?

pkg/constraintapi/lua/helper/gcra.lua
```diff
@@ -0,0 +1,221 @@
+---@param key string
+---@param now_ns integer
+---@param period_ns integer
+---@param limit integer
+---@param burst integer
+---@param quantity integer
+local function rateLimit(key, now_ns, period_ns, limit, burst, quantity)
+	limit = math.max(limit, 1)
```
albertchae, line 9: besides the ns vs ms precision, it looks like the main difference between this `rateLimit` function and `throttle` below in the same file is that here `limit` gets shadowed, so later calculations of `u`/`used_tokens` can differ

`local used_tokens = math.min(math.ceil(ttl / emission), limit)`

Just wondering if that is intentional

pkg/execution/ratelimit/lua/ratelimit.lua
```diff
@@ -26,170 +26,140 @@ local burst = tonumber(ARGV[4])
...
-	end
-
-	local raw_tat = tonumber(tat)
-	if not raw_tat then
-		return now_ns -- Reset if tonumber failed
-	end
+---@param limit integer
+---@param burst integer
+---@param quantity integer
+local function rateLimit(key, now_ns, period_ns, limit, burst, quantity)
```
albertchae, line 35: 👍 looks like the same changes as in the constraintapi lua

I think we are also using this for Rest API rate limiting, but since all of those are currently configured with a period of 5s, it probably isn't a huge impact to change the smoothing behavior

pkg/execution/ratelimit/ratelimit_lua_test.go
BrunoScheufler: Oops yeah, updated comment

pkg/constraintapi/lua/helper/gcra.lua
BrunoScheufler: I'll add a comment to this as well, the actual reason for this is to ensure limit is always at least one. This is to avoid dividing by 0 and raising other issues. Validation should prevent this upfront but this just adds another safeguard on the script level

PR #3516, opened 2026-01-08
Insights AI agent updates

AndyInternet:
## Description

- Implement prompt observability
- Update event-matcher and query-generator to handle updating existing prompt in addition to new query generation
- Add "Fix with AI" workflow when query results in error
- Open Insights AI drawer by default
- Replace "Thinking..." with more fun option (current limitation of agent-kit is we don't get information on what agent is running so I went the Claude Code route)
- Stability improvements and bug fixes

## Motivation
Better Insights AI experience

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ErrorState.tsx
+++ b/ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ErrorState.tsx
@@ -1,5 +1,6 @@
 import { Banner } from '@inngest/components/Banner/Banner';
 import { Button } from '@inngest/components/Button';
+import * as Sentry from '@sentry/tanstackstart-react';
 
 import { useInsightsAIHelper } from '../../InsightsAIHelperContext';
 import { useInsightsStateMachineContext } from '../../InsightsStateMachineContext/InsightsStateMachineContext';
@@ -24,7 +25,7 @@ export function ErrorState() {
         '\n```\n\nPlease provide a corrected version of the query that fixes this error.';
       await aiHelper.openAIHelperWithPrompt(prompt);
     } catch (error) {
-      console.error('Failed to open AI helper:', error);
+      Sentry.captureException(error);
       // Error is logged but doesn't prevent user interaction
     }
   };
--- a/ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContext.tsx
+++ b/ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContext.tsx
@@ -6,6 +6,7 @@ import {
   type ReactNode,
 } from 'react';
 import { flushSync } from 'react-dom';
+import * as Sentry from '@sentry/tanstackstart-react';
 
 import type { SQLEditorInstance } from '@inngest/components/SQLEditor/SQLEditor';
 
@@ -73,7 +74,7 @@ export function SQLEditorProvider({ children }: SQLEditorProviderProps) {
         try {
           runQuery();
         } catch (error) {
-          console.error('Failed to run query:', error);
+          Sentry.captureException(error);
         }
       });
     },


ui/apps/dashboard/src/lib/inngest/functions/agents/observability.ts
```diff
@@ -0,0 +1,89 @@
...
+>(
+  network: Network<InsightsAgentState>,
+  agent: T,
+  defaults: NonNullable<InsightsAgentState['observability']>[T],
+): NonNullable<NonNullable<InsightsAgentState['observability']>[T]> {
+  if (!network.state.data.observability) {
+    network.state.data.observability = {};
+  }
+  if (!network.state.data.observability[agent]) {
+    network.state.data.observability[agent] = defaults as any;
```
jakobevangelista: do we need a single generic `ensureObservability` function? 

each agent only ever calls it for itself (event matcher always passes 'eventMatcher' etc), so the generic doesn't provide value, it just creates type complexity that requires `as any` to work around

since observability data is only serialized/logged and never read back in typed code, we could simplify to:

```
type AgentObservability = {
  promptContext?: unknown;
  output?: unknown;
};

export function setObservability(
  network: Network<InsightsAgentState>,
  agent: 'eventMatcher' | 'queryWriter' | 'summarizer',
  data: AgentObservability,
) {
  if (!network.state.data.observability) {
    network.state.data.observability = {};
  }
  const obs = network.state.data.observability;
  obs[agent] = { ...obs[agent], ...data };
}
```

this removes the `as any` while keeping one simple function. `unknown` is type-safe (unlike any) and it accepts any value but would force validation if we ever tried to read it back

ui/apps/dashboard/src/lib/inngest/functions/run-network.ts
```diff
@@ -61,14 +65,44 @@ export const runAgentNetwork = inngest.createFunction(
...
+      const summaryOutput = summarizerResult?.output.find(
+        (msg) => msg.type === 'text' && msg.role === 'assistant',
+      );
+      if (
+        summaryOutput &&
+        'content' in summaryOutput &&
+        typeof summaryOutput.content === 'string'
+      ) {
+        const obs = ensureObservability(
+          { state: networkRun.state } as any,
```
jakobevangelista:
```suggestion
          networkRun,
```
can we just pass `networkRun` here instead of casting this `as any`?

ui/apps/dashboard/src/components/Insights/InsightsDataTable/states/ErrorState.tsx
```diff
@@ -1,28 +1,51 @@
...
+    try {
+      const prompt =
+        'The query execution failed with the following error:\n\n' +
+        errorMessage +
+        '\n\nThe query that failed was:\n\n```sql\n' +
+        query +
+        '\n```\n\nPlease provide a corrected version of the query that fixes this error.';
+      await aiHelper.openAIHelperWithPrompt(prompt);
+    } catch (error) {
+      console.error('Failed to open AI helper:', error);
```
jakobevangelista: should we report this to sentry?

ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsHelperPanel/features/InsightsChat/InsightsChat.tsx
```diff
@@ -69,40 +100,37 @@ export function InsightsChat({ agentThreadId, className }: InsightsChatProps) {
...
 
-  // When active, auto-apply latest generated SQL if newer
-  const lastAppliedSqlRef = useRef<string | null>(null);
+  // When active, auto-apply latest generated SQL whenever version changes
   useEffect(() => {
     if (currentThreadId !== agentThreadId) return;
+    if (!editorActions) return; // Not in query tab context
     const latest = getLatestGeneratedSql(agentThreadId);
     if (!latest) return;
-    if (lastAppliedSqlRef.current === latest) return;
```
jakobevangelista, line 88: we removed a guard here that prevented react from re-running the same sql on re-renders (component remount, tab switch etc) 

should we re-introduce this guard?

ui/apps/dashboard/src/components/Insights/InsightsSQLEditor/SQLEditorContext.tsx
```diff
@@ -0,0 +1,124 @@
...
+      flushSync(() => {
+        onChange(formattedSql);
+      });
+
+      // Run the query immediately - state is now guaranteed to be synced
+      queueMicrotask(() => {
+        try {
+          runQuery();
+        } catch (error) {
+          console.error('Failed to run query:', error);
```
jakobevangelista: should we report this to sentry as well?

ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsHelperPanel/features/InsightsChat/InsightsChat.tsx
```diff
@@ -12,29 +11,62 @@ import { AssistantMessage } from './messages/AssistantMessage';
...
-  if (toolName) {
-    switch (toolName) {
-      case 'select_events':
-        return 'Analyzing events…';
-      case 'generate_sql':
-        return 'Generating query...';
-      default:
-        return 'Thinking...';
+// Fun technical phrases that rotate while the agent is working
+const LOADING_PHRASES = [
```
jakobevangelista, line 15: i like this 😆

ui/apps/dashboard/src/lib/inngest/functions/agents/observability.ts
AndyInternet: Good idea! Much simpler

ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsHelperPanel/features/InsightsChat/InsightsChat.tsx
AndyInternet: This was intentional. If the agent generates the same query that has already run, but has cleared the results it sits and doesn't do anything. I ran into this edge case a few times which is why I removed the guard.

jakobevangelista: ah, that makes sense, i figured there was a few counter-edge cases i wasn't considering 

thanks for the context

PR #3531, opened 2026-01-13
Constraint API pre-release fixes

BrunoScheufler:


pkg/constraintapi/lua.go
```diff
@@ -326,3 +332,66 @@ func strSlice(args []any) ([]string, error) {
...
+	// Track success
+	metrics.IncrConstraintAPILuaScriptExecutionCounter(ctx, 1, metrics.CounterOpt{
+		PkgName: pkgName,
+		Tags: map[string]any{
+			"operation": name,
+			"status":    "success",
+		},
+	})
+	return rawRes, nil
+}
```
tonyhb, line 371:
```suggestion

	// Mark error as retriable in case of timeouts
	status, retry := luaError(err)

	// Report duration, which also has a counter.
	metrics.HistogramConstraintAPILuaScriptDuration(ctx, clock.Since(start), metrics.HistogramOpt{
		PkgName: pkgName,
		Tags: map[string]any{
			"operation": name,
			"status":   status,
		},
	})

	if err != nil {
		return nil, errs.Wrap(0, retry, "%s script %s: %w", name, status, err)
	}

	return rawRes, nil
}

func luaError(err error) (status string, retry bool) {
	if isTimeout(err) {
		return "timeout", true
	}
	if err != nil {
		return "error", false
	}
	return "success", false
}
```

This simplifies the code a bunch.  We don't need separate counters for errors - histogram embeds a counter.  By using a util func for the status and retry value we reduce a lot of branching.

BrunoScheufler: This is nice! Added in https://github.com/inngest/inngest/pull/3531/changes/bf57720451ef9ef831f8333a86d9cc2093ad6868

PR #3550, opened 2026-01-16
Fix potential processor iterator race condition

djfarrelly:
## Description

This aims to fix a potential race condition possible with the new queue interface introduced in #3480.

The new test introduced did validate that there is a race condition which highlighted losing up to 12 out of 100 items.

### How to reproduce

Check out the first commit which adds the new test then run this in the `queue` dir:
```
go test -race -run TestProcessorIteratorCounterRaceCondition
```

## Motivation
When investigating another potential issue today, I unleashed Claude to explore #3480 for any potential issues related to parallelization. It looked like a real potential issue given our volume.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/queue/processor_iterator.go
+++ b/pkg/execution/queue/processor_iterator.go
@@ -50,9 +50,9 @@ type ProcessorIterator struct {
 	// These flags are used to handle partition requeueing.
 	// These counters must be accessed atomically as they may be incremented
 	// concurrently when Parallel=true.
-	CtrSuccess     int32
-	CtrConcurrency int32
-	CtrRateLimit   int32
+	CtrSuccess     atomic.Int32
+	CtrConcurrency atomic.Int32
+	CtrRateLimit   atomic.Int32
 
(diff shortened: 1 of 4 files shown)


pkg/execution/queue/processor_iterator.go
```diff
@@ -46,22 +47,25 @@ type ProcessorIterator struct {
 	// Parallel processing breaks best effort fifo but increases throughput.
 	Parallel bool
 
-	// These flags are used to handle partition rqeueueing.
+	// These flags are used to handle partition requeueing.
+	// These counters must be accessed atomically as they may be incremented
+	// concurrently when Parallel=true.
 	CtrSuccess     int32
 	CtrConcurrency int32
 	CtrRateLimit   int32
```
tonyhb: This is a great PR!  Should we make this `atomic.Int32` so that people are _forced_ to use atomics and can never use regular oeprators?

djfarrelly: Yeah - good idea, I should have done it that way!

PR #3553, opened 2026-01-16
Extract metrics middleware

djfarrelly:
## Description

Extract the fairly generic metrics middleware into the root of the API package, re-using it in the v2 package and m

## Motivation
We are lacking metrics on v0 and v2 and this re-uses this code.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/v2/service.go
+++ b/pkg/api/v2/service.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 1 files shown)


pkg/api/metrics.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/inngest/inngest/pkg/telemetry/metrics"
 )
 
+const (
+	metricsPkgName = "api.inngest"
```
djfarrelly, line 12: We're not using the previous package name in our upstream telemetry (e.g. not using in Datadog tags), so making this more generic to `api.inngest` should not affect anything.

PR #3568, opened 2026-01-20
Add range picker option corresponding to account history limit

rhino1998:
## Description

This adds a range picker quick relative time option corresponding the the account history limit.

Before
image

After
image

## Motivation

Currently legacy Pro plan users have 14 days of history but there is no time range picker quick option for `"Last 14 days"`. This makes it unclear that 14 days of history are actually available.

- [ ] Docs

## Checklist

### DIFF
--- a/ui/packages/components/src/DatePicker/RangePicker.tsx
+++ b/ui/packages/components/src/DatePicker/RangePicker.tsx
@@ -66,10 +66,12 @@ const planRelatives = (daysAgoMax?: number) => {
 
   const unsorted = { ...RELATIVES, [`${daysAgoMax}d`]: `Last ${daysAgoMax} days` };
 
+  const now = new Date();
+
   return Object.fromEntries(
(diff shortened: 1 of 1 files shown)


ui/packages/components/src/DatePicker/RangePicker.tsx
```diff
@@ -50,6 +51,32 @@ const RELATIVES = {
...
+  const now = new Date();
+
+  return Object.fromEntries(
+    Object.entries(unsorted).sort(([a], [b]) => {
+      const aTime = subtractDuration(now, parseDuration(a));
+      const bTime = subtractDuration(now, parseDuration(b));
+      return bTime.getTime() - aTime.getTime();
+    })
+  );
+};
```
albertchae, line 78: I thought this might be overkill since we can assume RELATIVES is already sorted except for the last element, but TIL JS uses timsort which does well in these cases https://dev.to/bekmurzintimur/how-arrayprototypesort-works-3kcn

PR #3574, opened 2026-01-22
Introduce Kafka fallback producer

BrunoScheufler:
## Description

This PR introduces a new kafka utility package containing a fallback producer, which will accept a list of producers and try each until a produce operation is successful.

This also includes a Kafka producer which is based on franz-go. This will call TryProduce and fail if the batch is full, the operation times out, brokers are unavailable, or for any other reason that would fail the produce process. We can then continue with the next processor using the new fallback mechanism.

This PR also contains comprehensive tests to ensure this implementation is solid, covers all cases, and doesn't introduce regressions in future changes.

- [ ] Docs

## Checklist


pkg/util/kafka/producer.go
```diff
@@ -0,0 +1,113 @@
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err, open := <-errChan:
+		if !open {
+			return nil
+		}
+		return err
+	}
+}
```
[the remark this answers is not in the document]
BrunoScheufler:
```
// TryProduce is similar to Produce, but rather than blocking if the client
// currently has MaxBufferedRecords or MaxBufferedBytes buffered, this fails
// immediately with ErrMaxBuffered. See the Produce documentation for more
// details.
func (cl *Client) TryProduce(
	ctx context.Context,
	r *Record,
	promise func(*Record, error),
) {
	cl.produce(ctx, r, promise, false)
}
```

This is the full method, it does not return an error.

pkg/telemetry/metrics/histogram.go
```diff
@@ -516,3 +516,14 @@ func HistogramConstraintAPIRequestLatency(ctx context.Context, latency time.Dura
 		Boundaries:  DefaultBoundaries,
 	})
 }
+
```
mar-inngest, line 519: If we're going for a TryProduce, should we have a metric for keeping track of the number records in the backlog that haven't been sent to kafka (as it can keep growing)?

pkg/util/kafka/producer.go
```diff
@@ -0,0 +1,136 @@
...
+	producers []Producer
+}
+
+// String implements Producer.
+func (f *fallbackProducer) String() string {
+	return "fallback"
+}
+
+// Produce implements Producer.
+func (f *fallbackProducer) Produce(ctx context.Context, r *Record) error {
```
mar-inngest: non-blocking: at some point, it might interesting to have a benchmark test with and without this interface since one of the advantages of franz over other kafka libraries is that it's more lightweight

pkg/util/kafka/record.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+// RecordHeader contains extra information that can be sent with Records.
+type RecordHeader struct {
+	Key   string
+	Value []byte
+}
+
+// Record is a library-agnostic Kafka record for producing messages.
```
mar-inngest: non-blocking: mostly playing devil's advocate here

One of the advantages of using `kgo.Record` data type here than creating our own `Record` data type would be that (while we're locked in to an external library) we won't have to perform any extra data transformation to and from the kgo.Record and, we could write our own custom functions for this data type that we can use.
This way, we can skip copying the data to and from this format.

pkg/telemetry/metrics/histogram.go
BrunoScheufler: Good idea, added a gauge for the buffer size!

pkg/util/kafka/record.go
BrunoScheufler: I think both sides are fine. I initially used kgo.Record but was wondering whether we want to avoid lock-in. Alternatively we can start with kgo.Record since this package is scoped around Kafka anyway, so we can assume we'll use franz-go for some time. We can still swap this with another struct in the future.

PR #3575, opened 2026-01-22
SYS-585: Shard group leasing for dynamic executor-to-shard assignment

lkasinathan:
## Description

Shard group leasing for dynamic executor-to-shard assignment

**Summary**
- Introduce a shard lease mechanism that allows executors to dynamically claim and renew leases on queue shards within a configured shard group at startup, enabling elastic executor scaling without static shard assignment
- Add a Lua-based Redis implementation (shardLease.lua) that manages a set of leases per shard group key with automatic expiry cleanup and max-lease enforcement
- In addition to name and connection urls, each shard also has a `ShardAssignmentConfig` denoting the shard group and the desired number of executors for each shard. This config is used at executor startup time to figure out which shard within the assigned shard group each executor should consume queue items from.
- SYS-631: Metrics: Track capacity of shard leases available and actual shard leases claimed as guages.

**Testing**
- Unit tests for ShardLease in queue_test.go covering: claim, renewal, expiry, max-lease cap, and invalid lease  scenarios
- **All this code is disabled by default in OSS.**

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/queue/enqueue.go
+++ b/pkg/execution/queue/enqueue.go
@@ -55,7 +55,6 @@ func (q *queueProcessor) Enqueue(ctx context.Context, item Item, at time.Time, o
 		"env_id", item.WorkspaceID,
 		"app_id", item.Identifier.AppID,
 		"fn_id", item.Identifier.WorkflowID,
-		"queue_shard", q.primaryQueueShard.Name(),
 	)
 
 	if item.QueueName == nil && qi.FunctionID == uuid.Nil {


pkg/execution/state/redis_state/queue.go
```diff
@@ -52,13 +56,14 @@ func (q *queue) Client() *QueueClient {
 	return q.RedisClient
 }
 
-func NewQueueShard(name string, queueClient *QueueClient, opts ...osqueue.QueueOpt) RedisQueueShard {
+func NewQueueShard(name string, queueClient *QueueClient, shardAssignmentConfig osqueue.ShardAssignmentConfig, opts ...osqueue.QueueOpt) RedisQueueShard {
```
BrunoScheufler: I don't have strong opinions on this but I was wondering whether we could make the shard assignment part of the queue options, or is that not feasible?

pkg/execution/state/redis_state/queue_test.go
```diff
@@ -2174,6 +2174,87 @@ func TestQueueLeaseSequential(t *testing.T) {
 	})
 }
 
+func TestQueueShardLease(t *testing.T) {
```
BrunoScheufler: Do you want to move the new shard logic and test into its own test file? I was trying to break up this huge test file previously

pkg/execution/queue/processor.go
```diff
@@ -402,6 +408,16 @@ func (q *queueProcessor) ResetAttemptsByJobID(ctx context.Context, shardName str
 }
 
 func (q *queueProcessor) Run(ctx context.Context, f RunFunc) error {
+	// claimShardLease will block until a shard lease is obtained to process the primaryQueueShard.
+	l := logger.StdlibLogger(ctx)
+	if len(q.runMode.ShardGroup) != 0 {
+		l.Info("Executor started in ShardGroup mode, attempting to claim a shard lease", "shard_group", q.runMode.ShardGroup)
+		q.claimShardLease(ctx)
```
BrunoScheufler, line 437: Would this need to be started as a goroutine?

pkg/execution/queue/processor.go
```diff
@@ -402,6 +408,16 @@ func (q *queueProcessor) ResetAttemptsByJobID(ctx context.Context, shardName str
...
+	} else {
+		l.Info("Executor started in assignedQueueShard Mode", "queue_shard", q.primaryQueueShard.Name())
+	}
+	q.initializeShardAssignment(q.primaryQueueShard)
```
BrunoScheufler: I'm not sure I fully get how this will work: We're acquiring shard leases, how do we know which queue to work on?

lkasinathan: Before, we set queueProcessor.primaryAssignedShard when creating the queueProcessor.
Now, we set queueProcessor.primaryAssignedShard after successfully grabbing a shard lease for any shard in the assigned group 
https://github.com/inngest/inngest/blob/lakshmi/sys-585-shard-groups/pkg/execution/queue/shard_lease.go#L99

All dequeue operations in queue.Run() should still target primaryAssignedShard, just that now it is initialized a bit later, i.e., after getting a shard lease.

pkg/execution/queue/processor.go
lkasinathan: Not really, when q.runMode.ShardGroup is set, none of the following steps (sequential scan lease, config/instrumentation leases) can proceed without successfully obtaining a shard lease and setting it's own primaryQueueShard so the Run call will block until we are able to get a shard lease.

PR #3577, opened 2026-01-23
Add go struct to ts type generator

AndyInternet:
## Description

- New Go tool using go/ast that scans for `//tygo:generate` annotations and collects type definitions into a single barrel file
- Configuration for tygo with frontmatter defining SpanMetadataKind union type derived from generated constants
- New package containing `doc.go` (with `go:generate` directive) and generated `types_gen.go` barrel file
- Added annotations to `AIMetadata`, `HTTPMetadata`, `Warnings` types and their Kind constants in source files
- Generated TypeScript output with all types in a single file
- Documentation explaining usage and why the collector was needed

All details for code gen are documented here: https://github.com/inngest/inngest/blob/andy/exe-1198-install-and-configure-tygo/docs/TYGO_TYPE_GENERATION.md

## Motivation
🏎️ Speed Tracer

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/tracing/metadata/types/types_gen.go
+++ b/pkg/tracing/metadata/types/types_gen.go
@@ -3,6 +3,14 @@
 
 package types
 
+// From warning.go
+const (
+	KindInngestWarnings = "inngest.warnings"
+)
+
+// From warning.go
+type Warnings map[string]error
+
 // From ai.go
 const (
 	KindInngestAI = "inngest.ai"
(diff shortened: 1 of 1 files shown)


cmd/tygo-collect/main_test.go
jakobevangelista, line 1: test coverage looks great!

should we use testify to assert things rather than doing manual assertions?

this would align more with existing repo conventions + tests

.gitignore
```diff
@@ -35,3 +35,6 @@ go.work
 go.work.sum
 
 pkg/execution/state/redis_state/testdata
+.specify/
+specs/
+CLAUDE.md
```
jakobevangelista, line 40: should claude files be gitignored?

AndyInternet: I added this because we don't have a comprehensive, standardized claude.md and I use it all the time and just want to avoid accidentally checking one in that's only for the feature I'm currently working on.

cmd/tygo-collect/main_test.go
AndyInternet: I updated these to use testify.

PR #3578, opened 2026-01-23
Fix initial run RequestVersion edge case

rhino1998:
## Description

Ignore explicit 0 values on initial state creation.

## Motivation

Currently there is an edge case on deploy where we end up w/ 0 in state on runs queued from versions older than #3565 and then erroneously use that instead of `-1`.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/state/redis_state/redis_state.go
+++ b/pkg/execution/state/redis_state/redis_state.go
@@ -277,6 +277,10 @@ func (m shardedMgr) New(ctx context.Context, input state.Input) (state.State, er
 	}
 
 	rv := consts.RequestVersionUnknown
+	// XXX: The 0-check here is to handle an edge case with older
+	// executors that didn't have RequestVersion set,
+	// thus defaulting to 0, causing the affected runs
+	// to be forced to run using executor v0.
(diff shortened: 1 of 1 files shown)


pkg/execution/state/redis_state/redis_state.go
```diff
@@ -277,7 +277,11 @@ func (m shardedMgr) New(ctx context.Context, input state.Input) (state.State, er
 	}
 
 	rv := consts.RequestVersionUnknown
-	if input.RequestVersion != nil {
+	// XXX: The 0-check here is to handle an edge case with older
+	// executors that didn't have RequestVersion set,
+	// thus defaulting to 0, causing the affected runs
+	// to be forced to run using executor v0.
```
amh4r, line 283:
```suggestion
	// XXX: The 0-check here is to handle an edge case with older
	// executors that didn't have RequestVersion set,
	// thus defaulting to 0, causing the affected runs
	// to be forced to run using executor v0.
	//
	// It's safe to ignore v0 because we never want to
	// explicitly start a run on v0. It's OK if the SDK later
	// switches us to v0, but we initially don't want v0.
```

rhino1998:
Obsoleted by #3579

Issue #3315, opened 2025-11-17
In dev server CEL queries are not applied


dan-massey:
**Describe the bug**

**To Reproduce**
1. Run the dev server
2. Complete some runs and navigate to the /runs view
3. Write a query that should filter the results (e.g. `event.data.id == "foo"`)

**Expected behavior**
In app.inngest.com the list of runs updates and only matching runs are shown.
In the dev server, the filter does not apply

**Screenshots / Stack trace dump**

**System info**
- inngest-cli version:  1.13.6
- OS: macOS
- Version: 26.1

alessandroseni:
i'm experiencing the same

ramadanomar:
This also applies to self hosted versions

ramadanomar:
After a bit of research, the CEL Filtering is commented out in preview mode. (when dev server UI runs traces-preview flag is unset so it defaults to true which causes preview:true to be passed to the graphql api. This triggers GetSpanRuns instead of GetTraceRuns)

There's also a TODO at line 2933 inside pkg/cqrs/base_cqrs (Event ID filtering based on CEL filtering)

albertchae:
This was implemented in https://github.com/inngest/inngest/pull/3640

PR #3601, opened 2026-01-30
bug: CEL expression filtering don't apply on self hosted / dev server

ramadanomar:
## Description
Enable CEL (Common Expression Language) query filtering in GetSpanRuns for the dev server and self-hosted versions.
Previously, CEL query filtering was commented out in the GetSpanRuns function, which is used when the UI's traces-preview flag is enabled (the default). This meant that CEL search queries in the Runs page would return unfiltered results instead of properly filtered results.

Changes made:
- Uncommented and enabled the CEL expression handler in GetSpanRuns (pkg/cqrs/base_cqrs/cqrs.go)
- Added event ID filtering logic to filter runs based on CEL-matched events
- Added output expression matching to filter runs based on output data
- Added output field extraction from spans to support output-based filtering
- 

image
image

## Motivation

Users running the dev server (inngest dev) or self-hosted version (inngest start) were unable to filter runs using CEL queries. The search functionality appeared to work (no errors), but queries like event.name == "user/created" or event.data.userId == "123" would return all runs instead of filtered results.

This was caused by the traces-preview feature flag defaulting to true in the UI, which routes queries through GetSpanRuns instead of GetTraceRuns. The CEL filtering code in GetSpanRuns was commented out with a TODO note.

- [ ] Docs

## Checklist
- [x] I've linked any associated issues to this PR. #3315 

### DIFF
--- a/pkg/cqrs/base_cqrs/cqrs.go
+++ b/pkg/cqrs/base_cqrs/cqrs.go
@@ -2825,22 +2825,22 @@ func (w wrapper) GetSpanRuns(ctx context.Context, opt cqrs.GetTraceRunOpt) ([]*c
 	l := logger.StdlibLogger(ctx)
 
 	// use evtIDs as post query filter
-	// evtIDs := []string{}
-	// expHandler, err := run.NewExpressionHandler(ctx,
-	// 	run.WithExpressionHandlerBlob(opt.Filter.CEL, "\n"),
-	// )
-	// if err != nil {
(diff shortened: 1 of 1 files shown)


ramadanomar:
image

albertchae:
Hi, thanks for your contribution but we are moving away from in memory aggregation and filtering to SQL based in https://github.com/inngest/inngest/pull/3640

PR #3616, opened 2026-02-03
Add batching to batching

tonyhb:
circular:  high throughput batches are created then initialized fast. this is great, but can also be handled in-memory within ~100ms, reducing the total # of calls to the buffering store.

this commit introduces a bunch of changes to allow in-memory buffering of batches prior to committing to the state store to reduce QPS under many events.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/batch/buffer.go
+++ b/pkg/execution/batch/buffer.go
@@ -15,7 +15,7 @@ import (
 const (
 	// DefaultMaxBufferDuration is the default max time to buffer events in-memory
 	// before flushing.  must be less than the event ack deadline
-	DefaultMaxBufferDuration = 2 * time.Second
+	DefaultMaxBufferDuration = 500 * time.Millisecond
 
 	// DefaultMaxBufferSize is the default max events per buffer key before flush
 	DefaultMaxBufferSize = 100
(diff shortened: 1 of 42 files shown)


BrunoScheufler (approved):
Should we add some metrics to the buffer? Would be nice to see how this is utilized!

pkg/execution/batch/buffer.go
```diff
@@ -0,0 +1,415 @@
...
+		// non-blocking send (channel has buffer of 1)
+		select {
+		case p.resultCh <- result:
+		default:
+			// Channel already has a result or is closed
+		}
+	}
+
+	// handle scheduling based on result
+	if err == nil && bulkResult != nil {
```
BrunoScheufler: What happens if new-runs crashes inbetween returning the result to the client, unblocking the caller, and acknowledging the message and running handleScheduling?

The batch is safely stored in Redis but how do we know to run it? Is there a timeout job that was added? I think that just happens in handleScheduling, so does that mean we acknowledged but the batch is in the air?

BrunoScheufler: Should we run handleScheduling _before_ sending the result to the client? This way, we wouldn't acknowledge before we are fully done (and if the new-runs instance died, we could simply retry the event which is idempotently added to the batch and then idempotently scheduled)

tonyhb: indeed, just moved that - thanks!

PR #3628, opened 2026-02-05
Prepare Capacity Manager for full cutover rollout

BrunoScheufler:
## Description

This PR refactors the Constraint API in preparation for the rollout. This requires two parts

- Horizontally-scalable shards: The Capacity Manager has been refactored to accept a single provided client. There will be one instance per shard, and we will implement shard management externally. This makes the interface a lot nicer. All keys have the same prefix, so Redis Clustering will not move them around.
- Completely remove backward compatibility with queue/schedule: We are now aiming for a hard cutover, where old constraint state will be dismissed. This works because queue scavenging already uses a different index that is populated regardless. As we no longer need to be compatible, we can remove a lot of assumptions and clean up both the Constraint API and queue code.
- In the new rollout model, we will simply switch over to using the Constraint API for an account, which then means we will acquire leases and skip constraint checks and updates in Schedule and queue processing.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/executor/constraints.go
+++ b/pkg/execution/executor/constraints.go
@@ -255,7 +255,13 @@ func getScheduleConstraints(ctx context.Context, req execution.ScheduleRequest) 
 	// Throttle + concurrency constraints are checked in the queue.
 	if req.Function.RateLimit != nil {
 		rateLimitKey, err := ratelimit.RateLimitKey(ctx, req.Function.ID, *req.Function.RateLimit, req.Events[0].GetEvent().Map())
-		if err != nil {
+		switch err {
+		case ratelimit.ErrNotRateLimited:
+			// no rate limit configured, do not return constraints
+			return nil, nil
+		case nil:
+			// enforce rate limit
+		default:
 			return nil, fmt.Errorf("could not get rate limit key: %w", err)
 		}
 


pkg/execution/executor/constraints.go
```diff
@@ -151,7 +140,7 @@ func WithConstraints[T any](
 		metrics.IncrScheduleConstraintsCheckFallbackCounter(ctx, enums.ScheduleConstraintCheckFallbackReasonMissingLease.String(), metrics.CounterOpt{
 			PkgName: pkgName,
 		})
-		return fn(ctx, true)
+		return zero, fmt.Errorf("constraint API did not return lease ID")
```
lkasinathan: Might make sense to remove this counter here and instead increment an error counter that we can track since we do not fall back here.

pkg/execution/executor/constraints.go
```diff
@@ -107,23 +105,14 @@ func WithConstraints[T any](
...
 		req,
 		idempotencyKey,
-		fallback,
 		constraints,
 	)
 	if err != nil {
 		l.Error("failed to check constraints", "err", err)
 		metrics.IncrScheduleConstraintsCheckFallbackCounter(ctx, enums.ScheduleConstraintCheckFallbackReasonConstraintAPIError.String(), metrics.CounterOpt{
 			PkgName: pkgName,
 		})
```
lkasinathan, line 113: Might make sense to remove this counter here and instead increment an error counter that we can track since we do not fall back here anymore.

pkg/execution/state/redis_state/constraints.go
```diff
@@ -80,25 +79,13 @@ func (q *queue) BacklogRefillConstraintCheck(
...
 		logger.StdlibLogger(ctx).Error("acquiring capacity lease failed", "err", err, "method", "backlogRefillConstraintCheck", "functionID", *shadowPart.FunctionID)
-
-		if !fallback {
-			return nil, fmt.Errorf("could not enforce constraints and acquire lease: %w", err)
-		}
-
-		// Attempt to fall back to BacklogRefill -- ignore GCRA with constraint check idempotency
 		metrics.IncrBacklogRefillConstraintCheckFallbackCounter(ctx, enums.BacklogRefillConstraintCheckFallbackReasonConstraintAPIError.String(), metrics.CounterOpt{
 			PkgName: pkgName,
 		})
```
lkasinathan: same here

pkg/execution/state/redis_state/constraints.go
```diff
@@ -266,23 +249,13 @@ func (q *queue) ItemLeaseConstraintCheck(
...
 		l.Error("acquiring capacity lease failed", "err", err, "method", "itemLeaseConstraintCheck", "constraints", constraints, "item", item, "function_id", *shadowPart.FunctionID)
-
-		if !fallback {
-			return osqueue.ItemLeaseConstraintCheckResult{}, fmt.Errorf("could not enforce constraints and acquire lease: %w", err)
-		}
-
-		// Fallback to Lease (with idempotency)
 		metrics.IncrQueueItemConstraintCheckFallbackCounter(ctx, enums.QueueItemConstraintFallbackReasonConstraintAPIError.String(), metrics.CounterOpt{
 			PkgName: pkgName,
 		})
```
lkasinathan: same here

lkasinathan:
Looks good, left some minor comments around metrics.
Also, lint failures seem related.

lkasinathan (approved):

pkg/execution/executor/constraints.go
BrunoScheufler: Good point, I renamed the counter since there's no concept of fallbacks anymore, anyway.

proto/constraintapi/v1/service.proto
```diff
@@ -114,7 +114,6 @@ message ConcurrencyConstraint {
   ConstraintApiConcurrencyScope scope = 2;
   string key_expression_hash = 3;
   string evaluated_key_hash = 4;
-  string in_progress_item_key = 5; 
 }
```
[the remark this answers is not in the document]
BrunoScheufler: It's not used at all right now, so we're fine

PR #3640, opened 2026-02-10
add CEL filtering support for runs using trace refactor (i.e. spans table)

albertchae:
**Best reviewed by commits**, the earlier ones are small setup and the last one is the bulk of this PR

## Description
CEL filtering did work on the legacy traces table but was mostly in memory after fetching every run. 
On the new spans table, CEL didn't work and we did all the aggregation into runs in application (also fetching every run).

This change moves to limit the number of runs we fetch, which requires doing filtering and aggregation in SQL.

For aggregation, if our spans looks like this
```
┌─────────┬────────┬────────────────────┬────────────┬──────────┬───────────┐
│ span_id │ run_id │        name        │ start_time │ end_time │ event_ids │
├─────────┼────────┼────────────────────┼────────────┼──────────┼───────────┤
│ span_1  │ run_A  │ executor.run       │ 10:00      │ 10:05    │ [evt_1]   │
├─────────┼────────┼────────────────────┼────────────┼──────────┼───────────┤
│ span_2  │ run_A  │ executor.step      │ 10:01      │ 10:03    │ NULL      │
├─────────┼────────┼────────────────────┼────────────┼──────────┼───────────┤
│ span_3  │ run_A  │ executor.execution │ 10:02      │ 10:03    │ NULL      │
├─────────┼────────┼────────────────────┼────────────┼──────────┼───────────┤
│ span_4  │ run_B  │ executor.run       │ 10:10      │ 10:15    │ [evt_2]   │
└─────────┴────────┴────────────────────┴────────────┴──────────┴───────────┘
```

We get this after GROUP BY run_id:

```
┌────────┬─────────────────────────────┬─────────────────────────────┬─────────────────────────────┐
│ run_id │         start_time          │          end_time           │          event_ids          │
├────────┼─────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ run_A  │ MIN(10:00, 10:01, 10:02)    │ MAX(10:05, 10:03, 10:03)    │ MAX([evt_1], NULL, NULL)    │
│        │ = 10:00                     │ = 10:05                     │ = [evt_1]                   │
├────────┼─────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ run_B  │ MIN(10:10)                  │ MAX(10:15)                  │ MAX([evt_2])                │
│        │ = 10:10                     │ = 10:15                     │ = [evt_2]                   │
└────────┴─────────────────────────────┴─────────────────────────────┴─────────────────────────────┘
```

For filtering, it's straightforward if we're only looking at output, but for filtering on event data, compared to our cloud ClickHouse setup, events are not denormalized into spans/runs so we have to perform a join to make them filterable in SQL. This could be problematic at scale for Postgres because each span is getting joined to the event data before the final group by, so we might need to move the join after an initial group by and then group by again.

e.g. filtering by `event.data.foo == "bar"`, we unpack event_ids and join before grouping:

```
spans × UNNEST(event_ids):
┌────────┬─────────┬──────────┐
│ run_id │ span_id │ event_id │
├────────┼─────────┼──────────┤
│ run_A  │ span_1  │ evt_1    │
├────────┼─────────┼──────────┤
│ run_A  │ span_1  │ evt_2    │
├────────┼─────────┼──────────┤
│ run_A  │ span_2  │ evt_1    │
├────────┼─────────┼──────────┤
│ run_A  │ span_2  │ evt_2    │
├────────┼─────────┼──────────┤
│ run_B  │ span_3  │ evt_3    │
├────────┼─────────┼──────────┤
│ run_B  │ span_3  │ evt_4    │
├────────┼─────────┼──────────┤
│ run_B  │ span_4  │ evt_3    │
├────────┼─────────┼──────────┤
│ run_B  │ span_4  │ evt_4    │
└────────┴─────────┴──────────┘
```
JOIN events: (note the unneccessary duplication of event_data)

```
┌────────┬──────────┬──────────────────┐
│ run_id │ event_id │    event_data    │
├────────┼──────────┼──────────────────┤
│ run_A  │ evt_1    │ {"foo": "bar"} ✓ │
├────────┼──────────┼──────────────────┤
│ run_A  │ evt_2    │ {"foo": "bar"} ✓ │
├────────┼──────────┼──────────────────┤
│ run_A  │ evt_1    │ {"foo": "bar"} ✓ │
├────────┼──────────┼──────────────────┤
│ run_A  │ evt_2    │ {"foo": "bar"} ✓ │
├────────┼──────────┼──────────────────┤
│ run_B  │ evt_3    │ {"foo": "baz"} ✗ │
├────────┼──────────┼──────────────────┤
│ run_B  │ evt_4    │ {"foo": "baz"} ✗ │
├────────┼──────────┼──────────────────┤
│ run_B  │ evt_3    │ {"foo": "baz"} ✗ │
├────────┼──────────┼──────────────────┤
│ run_B  │ evt_4    │ {"foo": "baz"} ✗ │
└────────┴──────────┴──────────────────┘
```
After WHERE + GROUP BY run_id:

```
┌────────┐
│ run_id │
├────────┤
│ run_A  │
└────────┘
```

It should be fine for SQLite dev server since the data is usually ephemeral and we shouldn't have too much data to worry about.

Because spans and events in Postgres and SQLite are mostly the same
except for some slight differences in column types as well as SQL dialects,
the implementation is a mostly common query base with db specific adapters.

The main differences in the adapters are:
- syntax for extracting JSON fields with JSON paths
- syntax for unnesting event_ids
- how event_ids is parsed. It is stored as a string of event_ids, which
  is just a text column in SQLite but gets double encoded in Postgres JSONB
- spans.output is also doubly encoded in Postgres JSONB
- timestamp is a go time.Time string in SQLite and a timestamptz in Postgres

We should probably fix the doubly encoded Postgres JSON columns in the future, as it negates any benefit of storing those as JSONB rather than text.

It would also be nice to fix the SQLite timestamp serialization to use a more standard format.

## Motivation

Fix CEL search for self hosted

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/run/cel_sql.go
+++ b/pkg/run/cel_sql.go
@@ -15,6 +15,10 @@ import (
 // Supports: event.id, event.name, event.ts, event.v
 //
 // This applies to searching the events table, either by itself or when joined to runs for CEL search.
+//
+// N.B. we intentionally do not worry too much about SQL injection because this code should only be used
+// - dev server, which is not secure by design
+// - self hosted OSS, so a single tenant deployment and the self hosted dashboard ui should not be exposed externally
 func EventFieldConverter(ctx context.Context, n *expr.Node) ([]sq.Expression, error) {
 	if !n.HasPredicate() {
 		return []sq.Expression{}, nil


pkg/run/cel_sql.go
```diff
@@ -0,0 +1,233 @@
...
+	switch n.Predicate.Ident {
+	case "event.id":
+		id, ok := literal.(string)
+		if !ok {
+			return nil, fmt.Errorf("expects 'event.id' to be a string: %v", literal)
+		}
+		field := "event_id"
+		switch n.Predicate.Operator {
+		case operators.Equals:
+			f = sq.C(field).Eq(id)
```
rhino1998, line 39: Is this also vulnerable to injection? (Not necessarily a blocker (The existing system was also vulnerable))

At least a TODO comment somewhere could be good. We could conceivably move some of the insights CEL -> goqu logic into OSS to handle this. We'd need to split out the CH-specific stuff into an operator handler interface of some sort, but otherwise should be doable.

pkg/run/cel_sql.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// handleSQLiteJSONFilter creates SQL filters for JSON field access in SQLite.
+// Uses json_extract with dot-notation paths (e.g., $.field.subfield).
+func handleSQLiteJSONFilter(expr, fieldPath string, literal interface{}, op string) ([]sq.Expression, error) {
+	jsonPath := fmt.Sprintf("$.%s", fieldPath)
+	jsonExpr := fmt.Sprintf("json_extract(%s, '%s')", expr, jsonPath)
+
+	switch v := literal.(type) {
+	case string:
```
rhino1998, line 175: What about bools or explicit null checks? (literal type bool or nil)

pkg/run/cel_sql.go
```diff
@@ -0,0 +1,233 @@
...
+	case int64, float64:
+		numExpr := fmt.Sprintf("(%s)::numeric", jsonExpr)
+		return handleNumericOp(numExpr, v, op)
+	case bool:
+		// PostgreSQL JSONB stores booleans as 'true'/'false' strings when extracted with #>>
+		boolStr := "false"
+		if v {
+			boolStr = "true"
+		}
+		return handleStringOp(jsonExpr, boolStr, op)
```
rhino1998, line 216: should also have a case for null (literal type nil) checks

tests/golang/function_runs_test.go
```diff
@@ -287,20 +291,25 @@ func TestFunctionRunList(t *testing.T) {
...
-			assert.False(t, pageInfo.HasNextPage, "Failed with IDs: %s (%s)", fnIDs, fmt.Sprintf("fn-run-err-%s", failedEventName))
-			assert.Equal(t, failureTotal, total)
-			assert.Equal(t, remain, len(edges), "Got %#v and page info %#v", edges, pageInfo)
+			assert.False(ct, pageInfo.HasNextPage, "Failed with IDs: %s (%s)", fnIDs, fmt.Sprintf("fn-run-err-%s", failedEventName))
+			assert.Equal(ct, failureTotal, total)
+			assert.Equal(ct, remain, len(edges), "Got %#v and page info %#v", edges, pageInfo)
 		}, 10*time.Second, 2*time.Second)
 	})
 
 	t.Run("filter with event CEL expression", func(t *testing.T) {
```
rhino1998, line 300: How does this all handle dynamic fields with mismatched types? (IE trying to compare a json string field as a number)?

pkg/run/cel_sql.go
albertchae: It probably is, but I don't think it's priority for us since the OSS use cases are

- dev server, which is not secure by design
- self hosted, a single tenant deployment and the self hosted dashboard ui should not be exposed externally

added a comment about ⬆️ 

Similarly, I don't think we necessarily have to worry about perfect consistency of CEL search behavior between OSS and cloud or making it as feature rich, since self hosted users have direct access to their database if necessary.

pkg/run/cel_sql.go
albertchae: good call, added

pkg/run/cel_sql.go
albertchae: added

tests/golang/function_runs_test.go
albertchae: I sent an event trigger with this payload
```
  "data": {
    "x": "5"
  },
 ```
 
`event.data.x > 4` seemed to find that run
even `event.data.x < 10` worked, which I could see being an issue if they were being compared as strings.

I also tried this on cloud and it seems consistent

PR #3590, opened 2026-01-27
Batching: Return new status when first event in a batch is retried

lkasinathan:
## Description

When we hit transient errors (like i/o timeouts) on the first event in a batch, we skip scheduling the batch for execution. This change returns a New status when we retry the first event in a batch which should ensure that we don't skip scheduling the batch for execution after the timeout.

## Motivation
https://inngest.slack.com/archives/C06Q9JP3FQE/p1769438213885799

## Type of change (choose one)
- [ ] Chore (refactors, upgrades, etc.)
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [x] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #3675, opened 2026-02-12
Fix unscheduled batches in buffered bulk append

lkasinathan:
## Description

If there were duplicate events in the buffered batch, schedule the batch for execution after the batch timeout.

This is necessary for cases where the first event in a new batch fails due to transient issues like i/o timeouts writing to redis,  we might still write the event to a redis batch and return an error, which leads to not scheduling the batch for execution ever. This results in stuck batches. 
See https://inngest.slack.com/archives/C09KAK9AK0S/p1770673139099469  

To avoid that, we always schedule the batch for execution when any of the events are duplicates. While this scheduling attempt is only required if the retried event was the first event in a new batch, it is hard to distinguish that case because we bulk append events. So we just schedule a job every time there are _any_ duplicate elements in a batch. This is safe because batcher.ScheduleExecution is idempotent for a given batchID, so if a job already exists, the schedule call is a no-op.

This exact same issue was fixed for the non-buffered append in #3590

## Motivation
https://inngest.slack.com/archives/C09KAK9AK0S/p1770673139099469 

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/execution/batch/buffer.go
+++ b/pkg/execution/batch/buffer.go
@@ -368,15 +368,25 @@ func (ab *appendBuffer) handleScheduling(result *BulkAppendResult, fn inngest.Fu
 
 	ctx := context.Background()
 
+	// For new batches, schedule an execution after the batch timeout.
+	//
+	// If there were duplicate events in the buffered batch, also schedule an execution after the batch timeout.
+	// This is necessary for cases where the first event in a new batch fails due to transient issues like i/o timeouts writing to redis,
+	// we might still write the event to a redis batch and return an error, which leads to not scheduling the batch for execution ever.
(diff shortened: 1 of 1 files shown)


pkg/execution/batch/lua/bulk_append.lua
```diff
@@ -78,11 +78,6 @@ redis.call("EXPIRE", batchIdempotenceKey, idempotenceSetTTL)
 
 -- If all events were duplicates, return early
 if #eventsToAdd == 0 then
-  local currentLen = redis.call("LLEN", batchKey)
```
lkasinathan, line 81: this is dead code

BrunoScheufler (approved):

PR #3694, opened 2026-02-17
Insights: updated data table

sanjanaladdha:
## Description
- Improve Insights data table UX: add type-based column sizing and truncation so wide columns don’t blow out the layout.
- Add a cell detail helper panel: clicking a cell opens a right-hand panel showing the full value in a Monaco-based code editor, with JSON pretty-printing.
- Harden shared components: extend the shared Table to support cell clicks and column sizing

## Motivation
Improve the insights table

## Type of change (choose one)
- [x] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [ ] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #3726, opened 2026-02-20
Try using NewCodeBlock for Insights CellDetailContext

albertchae:
## Description

Another follow up to https://github.com/inngest/inngest/pull/3694

I believe we want to move to NewCodeBlock in general, so trying it here

- [ ] Docs

## Checklist

### DIFF
--- a/ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsHelperPanel/features/CellDetail/CellDetailView.tsx
+++ b/ui/apps/dashboard/src/components/Insights/InsightsTabManager/InsightsHelperPanel/features/CellDetail/CellDetailView.tsx
@@ -79,24 +79,21 @@ function CellValueDisplay({
     return { content: String(value ?? 'null'), language: 'plaintext' };
   }, [columnType, value]);
 
-  if (columnType === 'date') {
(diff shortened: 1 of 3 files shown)


albertchae:
@sanjanaladdha this PR tries using NewCodeBlock instead (which is what we're using on the current RunDetails and EventDetails pages)

a few things to note, there's a pill that we can put content into. It currently isn't hidden if it's empty but we can change that if we think we shouldn't put anything in there
image

for json columns we can actually show a parsed view based on this toggle
image
image

sanjanaladdha:
- can't think of any information we can show in the pill at the moment. maybe the data type but we already have that covered above. so maybe we just hide it? wdyt?
- i really like the parsed view but the toggle works weirdly

sanjanaladdha:
I have a couple of design comments to make just to clean this up a bit. let me know if there's something i can do to help with it.
- can we make this hug content (like it's doing in the current implementation)
Screenshot 2026-02-25 at 18 55 56

- can we add a border around the codeblock so it looks a bit more defined like in the image below
Screenshot 2026-02-25 at 19 48 02

Rest looks and works well for me!

PR #3731, opened 2026-02-22
V2 Invoke API

tonyhb:
## Description

Adds an endpoint for invoking functions to the V2 API.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/api/v2/http_test.go
+++ b/pkg/api/v2/http_test.go
@@ -412,9 +412,9 @@ func TestHTTPGateway_InvokeFunction(t *testing.T) {
 	handler, err := newTestHTTPHandler(ctx, ServiceOptions{}, opts)
 	require.NoError(t, err)
 
-	t.Run("POST /api/v2/functions/{id}/invoke with mode=async in request body", func(t *testing.T) {
-		body := `{"data": {"message": "Hello, World!"}, "mode": "async", "idempotencyKey": "test-123"}`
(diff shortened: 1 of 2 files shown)


pkg/devserver/function_provider.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	for _, fn := range fns {
+		if fn.Slug == identifier || fn.ID.String() == identifier {
+			inngestFn, err := fn.InngestFunction()
+			if err != nil {
+				return inngest.DeployedFunction{}, err
+			}
+			return inngest.DeployedFunction{
+				ID:            fn.ID,
+				Slug:          fn.Slug,
```
djfarrelly, line 36: Some users have multiple apps with the same fn slug - that's a case we need to handle. See other comment

proto/api/v2/service.proto
```diff
@@ -912,6 +913,144 @@ service V2 {
       }
     };
   }
+
+  rpc InvokeFunction(InvokeFunctionRequest) returns (InvokeFunctionResponse) {
+    option (google.api.http) = {
+      post: "/functions/{function_id}/invoke",
```
djfarrelly: Given the other input, I think we need to include the app id here. Devs and agents will know these 2 IDs, but won't really ever know the uuid. I know looks less nice with the app nesting, but I think it handles conflicts of re-used fn slugs across apps and makes it easy for an agent/dev/SDK to map what's written in their code/SDK to call the FN:
```
/apps/{app_id}/functions/{function_id}
```

tonyhb: done, ty

PR #3741, opened 2026-02-23
Add expand/contract all buttons

jakobevangelista:
## Description

Adds expand/contract all buttons
image

## Motivation

EXE-1315
EXE-1314

- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds "Expand all" and "Collapse all" buttons to the Timeline's root bar row. The feature bulk-updates the `expandedBars` Set by collecting all expandable IDs recursively via `collectExpandableIds`, preserving root bar IDs on collapse. The final commit adds a `vi.mock('../Button', ...)` shim to fix the vitest path alias and adds three tests covering expand-all, collapse-all, and round-trip behavior.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit e95dc6da283a27e1d0bc4c9a1499ee9db107c667.</sup>

### DIFF
--- a/ui/packages/components/vitest.config.ts
+++ b/ui/packages/components/vitest.config.ts
@@ -1,14 +1,8 @@
-import path from 'path';
 import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
   plugins: [react()],
-  resolve: {
-    alias: {
-      '@inngest/components': path.resolve(__dirname, 'src'),
-    },
-  },
   test: {
     environment: 'jsdom',
   },
--- a/ui/packages/components/src/RunDetailsV4/Timeline.test.tsx
+++ b/ui/packages/components/src/RunDetailsV4/Timeline.test.tsx
@@ -5,11 +5,28 @@
 
 import type { ReactNode } from 'react';
 import { cleanup, fireEvent, render, screen } from '@testing-library/react';
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import { TooltipProvider } from '../Tooltip/Tooltip';
 import { Timeline } from './Timeline';
 import type { TimelineData } from './TimelineBar.types';
+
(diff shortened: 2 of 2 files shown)


ui/packages/components/src/RunDetailsV4/Timeline.tsx
```diff
@@ -7,10 +7,10 @@
  * - Expansion state for all expandable bars
  * - Recursive rendering of nested steps
  * - Column resize handling
- */
-
-import { useCallback, useState } from 'react';
+ */ import { useCallback, useMemo, useState, type ReactNode } from 'react';
```
AndyInternet: nit - `*/` and `import` should be on separate lines

ui/packages/components/src/RunDetailsV4/Timeline.tsx
```diff
@@ -464,6 +503,43 @@ export function Timeline({ data, onSelectStep }: Props): JSX.Element {
     [onSelectStep]
   );
 
+  const handleExpandAll = useCallback(() => {
+    const rootBarIds = bars.filter((bar) => bar.isRoot).map((bar) => bar.id);
```
AndyInternet: Minor but rootBarIds = bars.filter().map() is duplicated. Could be extracted to useMemo() and referenced in both locations.

ui/packages/components/vitest.config.ts
```diff
@@ -1,8 +1,14 @@
+import path from 'path';
 import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
   plugins: [react()],
+  resolve: {
```
AndyInternet: The changeset doesn't strictly require this change because the alias isn't used. Should either leverage the alias or hold off on this change until its needed.

jakobevangelista: the alias was needed because our `Timeline` component now imports `Button`, and `Button.tsx` internally uses self referencing `@inngest/components/Tooltip` imports that vitest can't resolve

but i just saw your tests for rundetailsV4 which mocks button and I've switched to that the test instead

PR #3750, opened 2026-02-24
Improve postgres performance for runs counts

jakobevangelista:
## Description

Improve postgres performance for runs counts. We can do a count in SQL rather than an in-memory `len`

- [ ] Docs

## Checklist


pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -2184,24 +2184,75 @@ func newRunsQueryBuilder(ctx context.Context, opt cqrs.GetTraceRunOpt) *runsQuer
...
+// getTraceRunsCount returns the count of trace runs matching the given filters.
+// Uses len(GetTraceRuns) because CEL event filtering is applied in Go post-query.
+func (w wrapper) getTraceRunsCount(ctx context.Context, opt cqrs.GetTraceRunOpt) (int, error) {
+	opt.Items = 0
+	res, err := w.GetTraceRuns(ctx, opt)
 	if err != nil {
 		return 0, err
 	}
-
 	return len(res), nil
```
albertchae, line 2201: honestly we need to just delete the opt.Preview branch and this method, but fine to leave for now

pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -2184,24 +2184,75 @@ func newRunsQueryBuilder(ctx context.Context, opt cqrs.GetTraceRunOpt) *runsQuer
...
 }
 
+// getSpanRunsCount returns the count of span-based runs matching the given filters
+// using a SQL COUNT(DISTINCT run_id) query instead of fetching all rows.
+func (w wrapper) getSpanRunsCount(ctx context.Context, opt cqrs.GetTraceRunOpt) (int, error) {
```
albertchae, line 2206: i think this could be rewritten as building the query from `GetSpanRuns`, then overwriting the select fragment to `COUNT(DISTINCT run_id)` and removing the limit. That way it's really obvious how this query differs from that one. or maybe factor out the common part of the query building to its own sub method that both reuse

pkg/cqrs/base_cqrs/migrations/postgres/000019_add_perf_indexes.up.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE INDEX IF NOT EXISTS idx_spans_run_dynamic_endtime_status
+    ON spans(run_id, dynamic_span_id, end_time DESC) INCLUDE (status);
+CREATE INDEX IF NOT EXISTS idx_spans_start_time ON spans(start_time DESC);
+
+-- Trace Runs Table
+CREATE INDEX IF NOT EXISTS idx_trace_runs_queued_at ON trace_runs(queued_at DESC, run_id);
+CREATE INDEX IF NOT EXISTS idx_trace_runs_started_at ON trace_runs(started_at DESC, run_id);
+CREATE INDEX IF NOT EXISTS idx_trace_runs_ended_at ON trace_runs(ended_at DESC, run_id);
+CREATE INDEX IF NOT EXISTS idx_trace_runs_app_id ON trace_runs(app_id);
+CREATE INDEX IF NOT EXISTS idx_trace_runs_function_id ON trace_runs(function_id);
```
albertchae: we should probably add concurrently to all of these and also split out the migrations into their own PR i think. it might also make sense to put each index in its own migration file since golang-migrate can be pretty finicky

albertchae (approved):

PR #3757, opened 2026-02-25
Fix/run spans query

### LINKED ISSUE #3612: Runs view very slow on large data sets
**Is your feature request related to a problem? Please describe.**
When trying to view the runs on large tables on self hosted, the queries take a very long time to run.

The query that seems to run very slow is this: https://github.com/inngest/inngest/blob/05dc3a4287bbcb596f535ede82ae335d7c4cb680/pkg/cqrs/base_cqrs/cqrs.go#L2850-L2877
```sql
SELECT "run_id", "account_id", "app_id", "function_id", "trace_id", "dynamic_span_id", "start_time", "end_time", "status", "span_id", "name", "attributes", "links", "output", "event_ids", "input" 
FROM "spans" 
WHERE (("dynamic_span_id" IN ((SELECT DISTINCT "dynamic_span_id" FROM "spans" WHERE ("name" = $1)))) AND ("debug_run_id" IS NULL) AND ("start_time" >= $2) AND ("start_time" < $3)) 
ORDER BY "start_time" DESC, "run_id" ASC
```

The only indexes I see we have are here: https://github.com/inngest/inngest/blob/05dc3a4287bbcb596f535ede82ae335d7c4cb680/pkg/cqrs/base_cqrs/migrations/postgres/000013_span_tables.up.sql#L24-L25

**Describe the solution you'd like**
Could we add these 2 indexes to our speed things up, they seem to improve the queries a lot:
```sql
CREATE INDEX idx_spans_name_dynamic ON spans(name, dynamic_span_id);  -- For spans_1 subquery
CREATE INDEX idx_spans_debug_start ON spans(debug_run_id, start_time); -- For main spans filter
```

**Describe alternatives you've considered**
A lot larger database could help but its not a lot of data but really the query could be more efficient.

**Additional context**

zia-r:
## Description

Fixes #3612 - I think an attempt was made with https://github.com/inngest/inngest/pull/3652  but that wont actually fix the underlying behavior.                                                                                                                                                                                                    

  Optimizes the `GetSpanRuns` query path for PostgreSQL, addressing the runs list page being unusable on large self-hosted datasets. Three changes:

  1. **Database indexes** (migration 000019): A partial index on root spans for LIMIT pushdown, and a covering index for correlated subquery lookups. `IF NOT EXISTS` is used so operators can optionally run these
  manually with `CONCURRENTLY` before applying the migration to avoid write locks on large tables.

  2. **Query rewrite** (PostgreSQL only): The original query used `GROUP BY` over all spans with `MIN()`/`MAX()` aggregates and a correlated status subquery per group. With `GROUP BY + LIMIT`, Postgres must fully
  aggregate all groups before applying LIMIT — making the query O(total_runs) regardless of page size. The rewrite queries root spans (`name='executor.run'`) directly, applies ORDER BY + LIMIT early, then runs
  correlated subqueries only on the result set (~40 rows). The SQLite path is unchanged.

  3. **Singleflight cache** for `App()` and `Function()` field resolvers: gqlgen resolves fields concurrently, causing 40+ duplicate DB calls per field per page. A per-request cache with channel-based deduplication
   ensures one DB call per unique key. Falls back to direct DB calls if the cache is not in context.

I structured them as three separate commits in case you would like me to resolve them 

  ### Behavioral notes

  - The PostgreSQL query path activates when `w.isPostgres()` is true and `opt.Preview` is true. SQLite behavior is completely unchanged.
  - The status filter now applies to the root span's status rather than any span in the run group. The original behavior could return runs whose displayed status didn't match the filter (the filter applied
  pre-GROUP BY on individual span rows, but the displayed status came from the latest span). The new behavior is more consistent.
  - `GetTraceRunsCount` uses `COUNT(DISTINCT run_id)` on PostgreSQL instead of fetching all rows and calling `len()`.
  - Migration number 000019 is currently free on main — if there's a collision with an in-flight PR, happy to renumber.

### Tests

I didn't see any existing tests for postgres specific queries and didn't see any unit testing for the dataloaders so I didn't add any.

## Motivation

On a self-hosted deployment with ~1.2M spans / ~88K runs (PostgreSQL on RDS), the runs list page took 40-90 seconds to load. After this change it completes in <100 ms. This both fixes initial page load and allows for the infinite scroll in pagination to actually work seamlessly.

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/coreapi/graph/loaders/cache.go
+++ b/pkg/coreapi/graph/loaders/cache.go
@@ -1,0 +1,81 @@
+package loader
+
+import (
+	"context"
+	"sync"
+
+	"github.com/google/uuid"
+)
+
+// LookupCache provides per-request caching with singleflight deduplication
+// for frequently looked-up entities. When 100 concurrent goroutines request
+// the same key, only one DB call is made and all others wait for the result.
+type LookupCache struct {
+	mu      sync.Mutex
+	data    map[string]interface{}
+	flights map[string]*flight
+}
+
+type flight struct {
+	done chan struct{}
+	val  interface{}
+	err  error
+}
+
+type lookupCacheKey struct{}
+
+func WithLookupCache(ctx context.Context) context.Context {
+	return context.WithValue(ctx, lookupCacheKey{}, &LookupCache{
+		data:    make(map[string]interface{}),
+		flights: make(map[string]*flight),
+	})
+}
+
+func GetLookupCache(ctx context.Context) *LookupCache {
+	if c, ok := ctx.Value(lookupCacheKey{}).(*LookupCache); ok {
+		return c
+	}
+	return nil
+}
+
+// GetOrLoad returns a cached value or calls the loader function exactly once
+// per key, even under concurrent access. All concurrent callers for the same
+// key block until the single loader completes.
+func (c *LookupCache) GetOrLoad(kind string, id uuid.UUID, loader func() (interface{}, error)) (interface{}, error) {
+	key := kind + ":" + id.String()
+
+	c.mu.Lock()
+	// Check cache first
+	if v, ok := c.data[key]; ok {
+		c.mu.Unlock()
+		return v, nil
+	}
+	// Check if there's an in-flight request
+	if f, ok := c.flights[key]; ok {
+		c.mu.Unlock()
+		<-f.done
+		return f.val, f.err
+	}
+	// Start a new flight
+	f := &flight{done: make(chan struct{})}
+	c.flights[key] = f
+	c.mu.Unlock()
+
+	// Execute the loader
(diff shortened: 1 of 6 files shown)


vipulpandey7860:
+1 on this. I'm hitting the same performance issue on my self-hosted setup with PostgreSQL. The runs list page is essentially unusable — taking 40+ seconds to load or timing out entirely.

karlisl:
+1! We are encountering the same issue.

image

image

Alexandredc:
Same issue here !
thank a lot @zia-r for your work ! 
Huge improvement 90s -> 100ms !

Alexandredc:
Any new regarding this issue ?

PR #3767, opened 2026-02-26
Check before peek

BrunoScheufler:
- [ ] Docs

## Checklist


BrunoScheufler (commented):
There are some downsides to this
- We likely can't cache Check() requests, at least not the remaining capacity. We _can_ cache exhausted constraints, but this won't help us constrain peek size, only whether we can skip scanning
- If we ran a Check for every leased partition, we would have _a lot_ of throughput, and latency may grow too large, slowing down the entire system

PR #3777, opened 2026-02-27
EXE-1382: Prevent duplicate apps being synced to the Inngest server

djfarrelly:
## Description

Prevent duplicate apps from being synced to the dev server or self-hosted Inngest. Related to #3361.

## Motivation
EXE-1382

- [ ] Docs

## Checklist

### DIFF
--- a/pkg/config/registration/registration.go
+++ b/pkg/config/registration/registration.go
@@ -9,6 +9,7 @@ import (
 	"github.com/inngest/inngest/pkg/execution/queue"
 	"github.com/inngest/inngest/pkg/execution/state"
 	"github.com/inngest/inngest/pkg/telemetry/trace"
+	"github.com/inngest/inngest/pkg/util/rueidisconn"
 )
 
 var (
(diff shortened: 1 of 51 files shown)


pkg/devserver/service.go
```diff
@@ -748,7 +754,7 @@ func upsertErroredApp(
 				String: pingError.Error(),
 				Valid:  true,
 			},
-			ID:  inngest.DeterministicAppUUID(appURL),
+			ID:  inngest.DeterministicAppUUID(util.NormalizeAppURL(appURL, false)),
```
rhino1998, line 757: Will this result in different IDs for existing apps that have what would now considered unnormalized app urls (ie `util.NormalizeAppURL(appURL, false) != appURL`)?

djfarrelly: I think in this function here `upsertErroredApp`, it's only called during auto-discovery where an app placeholder is created in a placeholder state.

PR #3797, opened 2026-03-04
Use `_inngest.response.headers` attribute rather than `inngest.response_headers` metadata

rhino1998:
## Description

This redacts the same sensitive headers from the response headers metadata and improves the attribute storage by compacting single-value headers into a string instead of `[]string`

## Motivation

EXE-1419

- [ ] Docs

## Checklist


ui/apps/dashboard/src/queries/useGetRun.ts
```diff
@@ -59,6 +59,10 @@ const traceDetailsFragment = graphql(`
         timedOut
       }
     }
+    response {
```
rhino1998: Maybe I need to split this into a separate PR to merge after the monorepo PR?

pkg/headers/compact.go
```diff
@@ -0,0 +1,82 @@
...
+func (h Compact) MarshalGQLContext(ctx context.Context, w io.Writer) error {
+	return json.NewEncoder(w).Encode(h)
+}
+
+func (h *Compact) UnmarshalGQLContext(ctx context.Context, v any) error {
+	vm, ok := v.(map[string]any)
+	if !ok {
+		return fmt.Errorf("cannot unmarshal %T as RawMetadata", v)
+	}
+
```
AndyInternet: Should we initialize the map in case `*h` is nil like you have in `UnmarshalJSON()`?
```suggestion
*h = make(Compact)
```

pkg/coreapi/coreapi.go
```diff
@@ -107,6 +108,7 @@ func NewCoreApi(o Options) (*CoreAPI, error) {
 			RequireKeys:     o.RequireKeys,
 		}
 		srv := handler.NewDefaultServer(generated.NewExecutableSchema(generated.Config{Resolvers: a.resolver}))
+		srv.Use(extension.FixedComplexityLimit(300))
```
AndyInternet: Magic number, what is 300?

rhino1998: Ah, I might be able to remove that actually. I'm not sure what the default GQL complexity limit is in gqlgen. In monorepo I needed to bump the limit slightly because adding the response field put us over the limit.

rhino1998: Removed

ui/apps/dashboard/src/queries/useGetRun.ts
rhino1998: Split in [6abf9c3](https://github.com/inngest/inngest/pull/3797/commits/6abf9c3b9bd57d5f689ef7cfbbc885d134179b96)

PR #3804, opened 2026-03-05
fix: userland OTel span parenting w/checkpointing

Linell:
## Description

Userland spans created inside checkpointed steps were invisible in the UI because the executor didn't create matching step/execution spans for them to parent under, and the GQL trace loader was silently dropping leaf userland spans during collapse.

This fixes things by:
- Create deterministic executor.step and executor.execution spans in checkpoint sync/async paths using seeded span IDs
- Reparent userland spans via inngest.step.parentSpanId attribute in span ingestion
- Fix leaf userland span collapse in GQL trace conversion

## Motivation
https://linear.app/inngest/issue/EXE-1408/extended-traces-not-working

- [ ] Docs

## Checklist

## Related

* https://github.com/inngest/inngest-js/pull/1359

### DIFF
--- a/pkg/coreapi/graph/loaders/trace.go
+++ b/pkg/coreapi/graph/loaders/trace.go
@@ -25,6 +25,10 @@ const (
 	DiscoveryStepSpanName    = "Discovery step"
 	GenericExecutionSpanName = "Execution"
 	FinalizationSpanName     = "Finalization"
+
+	// SDKExecutionSpanName is an alias for meta.SDKExecutionSpanName
+	// used locally for readability.
+	SDKExecutionSpanName = meta.SDKExecutionSpanName
 )
 
 var ErrSkipSuccess = fmt.Errorf("skip success span")
@@ -403,11 +407,14 @@ func (tr *traceReader) convertRunSpanToGQL(ctx context.Context, span *cqrs.OtelS
 				}
 			case meta.SpanNameStepDiscovery, meta.SpanNameStep:
 				{
-					// Don't copy status from userland spans to preserve step status
-					if !child.IsUserland {
-						gqlSpan.EndedAt = child.EndedAt
-						gqlSpan.Status = child.Status
+					// Userland spans don't carry step execution metadata;
+					// so skip all parent-property propagation for them.
+					if child.IsUserland {
+						break
 					}
+
+					gqlSpan.EndedAt = child.EndedAt
+					gqlSpan.Status = child.Status
 
 					if isFirstChild {
 						isFirstChild = false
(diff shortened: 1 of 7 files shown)


pkg/coreapi/graph/loaders/trace.go
```diff
@@ -455,16 +452,26 @@ func (tr *traceReader) convertRunSpanToGQL(ctx context.Context, span *cqrs.OtelS
...
+		// We only collapse when the child is specifically the SDK execution
+		// wrapper span. Other userland spans with children (e.g., spans
+		// within checkpointed steps) must be preserved in the tree.
+		//
+		// If the execution wrapper has no children (i.e., it is a leaf
+		// node), we skip the collapse to avoid silently dropping it.
+		if !gqlSpan.IsUserland && len(gqlSpan.ChildrenSpans) == 1 && gqlSpan.ChildrenSpans[0].IsUserland && gqlSpan.ChildrenSpans[0].Name == SDKExecutionSpanName {
+			if len(gqlSpan.ChildrenSpans[0].ChildrenSpans) > 0 {
+				gqlSpan.ChildrenSpans = gqlSpan.ChildrenSpans[0].ChildrenSpans
+			}
```
rhino1998:
```suggestion
			gqlSpan.ChildrenSpans = gqlSpan.ChildrenSpans[0].ChildrenSpans
```

NIT: The meaningful change is the name check above, we can leave the rest of the condition the same as it was so we still elide the inngest.execution span if one with no children is recieved.

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -190,6 +193,31 @@ func (c checkpointer) CheckpointSyncSteps(ctx context.Context, input SyncCheckpo
 				l.Error("error saving span for checkpoint op", "error", err)
 			}
 
+			// Create an executor.execution child span so the UI renders "Inngest" /
```
rhino1998: I think it might be safer to tweak the UI to show the new sections for `executor.step` spans than to create a phantom executor.execution span (also various other callsites (like metadata) depend on the convention of executor.execution spans only being under step discovery spans.

Also this duplicates a bunch of attribute data & emits extra spans (which have a perf cost) (also some tests are failing because of this)

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -436,6 +466,29 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 				// We should never hit a blocker creating a span.  If so, warn loudly.
 				l.Error("error saving span for checkpoint op", "error", err)
 			}
+
+			// Create an executor.execution child span so the UI renders
+			// correctly, matching the non-checkpointed structure.
+			if stepSpanRef != nil {
```
rhino1998: Same issues as above, the UI should be tweaked to handle this rather than changing the underlying data to fix the UI

PR #3806, opened 2026-03-05
EXE-1422: lots more ux defense around neon/supabase failure scenarios

jacobheric:
## Description

a lot of neon/supabase integration scenarios failed without giving the user any indication what happened or what to do about it. This adds some human readable error message and let's them know to remove and retry when it fails. 

there is a companion monorepo and dbcap pr that makes retries much more resilient. I've tested it out locally e2e for a lot of common failure scenarios and it's now able to recover.

https://github.com/inngest/dbcap/pull/7
https://github.com/inngest/monorepo/pull/6487

- [ ] Docs

## Checklist

---

> [!NOTE]
> This PR improves UX for Neon/Supabase integration failure scenarios by: surfacing actual error messages instead of null, supporting multiple publications per integration (filter instead of find), adding status-aware UI labels/colors, and adding a broken integration warning banner.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 18c2d301e9f7e40381da18eeb4222475cfd24c16.</sup>

### DIFF
--- a/ui/packages/components/src/PostgresIntegrations/types.ts
+++ b/ui/packages/components/src/PostgresIntegrations/types.ts
@@ -47,22 +47,6 @@ export type IntegrationPageContent = {
   url: string;
 };
 
-export function parseConnectionString(connectionString: string) {
-  const regex = /postgresq?l?:\/\/([\w.\-]+):([^@]+)@([^/]+)/;
-  const match = connectionString.match(regex);
-
-  if (match) {
-    const [, _username, _password, host] = match;
-    return {
-      name: `Neon-${host}`,
-      engine: 'postgresql',
-      adminConn: connectionString,
-    };
-  }
-
-  return null;
-}
-
 export type Publication = {
   id: string;
   name: string;


ui/packages/components/src/PostgresIntegrations/utils.ts
```diff
@@ -1,4 +1,4 @@
-const regex = /postgresq?l?:\/\/([\w-]+):([^@]+)@([^/]+)/;
+const regex = /postgresq?l?:\/\/([\w.\-]+):([^@]+)@([^/]+)/;
```
jacobheric, line 1: This regex is for supabase direct connect only...which *does not* work with ipv4 at all. I've altered it to also accept the pool uri which works with ipv6 (had to do this even to test it).

albertchae: maybe we should share the regex with `PostgresIntegrations/types.ts`, unless we separated them intentionally (they diverged before but your edits make them the same)

jacobheric: great catch. missed that. the other one is actually dead code. I removed it.

PR #3840, opened 2026-03-12
Add metadata span size limits and enable step metadata by default

AndyInternet:
## Description

- Add per-metadata-span size validation (64 KB) and per-run cumulative metadata size tracking/validation (1 MB) to prevent abuse before enabling step metadata by default
- Add `Values.Size()` method, `ErrMetadataSpanTooLarge` and `ErrRunMetadataSizeExceeded` error sentinels, and `MaxMetadataSpanSize`/`MaxRunMetadataSize` constants to the data model layer
- Enforce both size limits in `executor.createMetadataSpan` — the single chokepoint for all metadata spans (user-submitted and internally-generated) — with in-memory counter updates for intra-step accuracy
- Add `MetadataSize` field to `RunMetrics` for persistent cross-step cumulative tracking
- Both limits use non-fatal rejection: spans are dropped with warning logs but the step and run continue

## Test plan

- [x] Unit test: `Values.Size()` correctly sums key lengths + raw JSON value lengths
- [x] Unit test: `createMetadataSpan` returns `ErrMetadataSpanTooLarge` for spans > 64 KB
- [x] Unit test: `createMetadataSpan` accepts spans at exactly 64 KB
- [x] Unit test: `createMetadataSpan` returns `ErrRunMetadataSizeExceeded` when cumulative metadata exceeds 1 MB
- [x] Unit test: sequential spans within a single call correctly accumulate against the 1 MB limit
- [x] E22 test: function run exceeding 1 MB cumulative metadata has later spans rejected but completes successfully
image
image

- [x] E2E test: single metadata span > 64 KB is rejected, step continues
image
image

- [x] E2E test: happy path
image

## Notes on metadata dropped on checkpointing bug:
When a function got checkpointed, suspended and resumed later, we were silently dropping metadata spans.

The checkpoint process was calling CreateMetadataSpan, which internally calls spanMetadata.Serialize(). By the time the checkpoint process actually ran, the underlying metadata struct was already consumed. Serialize() would return empty or error out, and we'd lose everything attached during that step.

The fix was to serialize the metadata eagerly, grab the values up front while they're still valid, then create the span. The executor was already doing this (it needs the serialized size for limit enforcement), but the checkpoint codepath wasn't. We introduced `CreateMetadataSpanFromValues()` to accept pre-serialized values instead of the Structured interface, and wired the checkpoint process to use it.

To summarize: Serialize first, create the span second. The checkpoint path now mirrors what the executor does, and metadata now makes its way through checkpointed runs.

## Motivation
Allow step metadata to be enabled for all

- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds per-span (64 KB) and per-run cumulative (1 MB) metadata size limits enforced at the `createMetadataSpan` chokepoint. Introduces `Values.Size()`, two error sentinels, two constants, and mutex-backed size tracking on `RunMetrics`. Fixes metadata spans silently dropped in checkpoint paths by wiring `AllowStepMetadata` through the checkpoint layer and using pre-serialized values. Replaces a global `sync.Map` mutex store with an embedded `*sync.Mutex` pointer on `RunMetrics`. Moves `inngest.usage` metadata bytes computation from write-time to read-time in `mapRootSpansFromRows`. The `LoadMetadata` fallback bug from previous reviews is now fixed.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit efbe7bd9d2b314d44186f269cfdeb29fc82fbb7b.</sup>

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -3402,6 +3402,30 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
 		return err
 	}
 
+	// Extract AI metadata from step output before saving so the cumulative
+	// metadata size delta is accurate when persisted alongside the step.
+	if e.allowStepMetadata.Enabled(ctx, runCtx.Metadata().ID.Tenant.AccountID) {
+		// Calculate step duration in milliseconds
+		stepDurationMs := gen.Timing.B / 1_000_000
+
+		md := metadata.WithWarnings(extractors.ExtractAIOutputMetadata(
+			[]byte(output),
+			stepDurationMs,
+		))
+		for _, m := range md {
+			_, err := e.createMetadataSpan(
+				ctx,
+				runCtx,
+				"executor.handleGeneratorStep.aiOutput",
+				m,
+				enums.MetadataScopeStepAttempt,
+			)
+			if err != nil {
+				e.log.Warn("error creating AI output metadata span", "error", err)
+			}
+		}
+	}
+
 	// Persist the cumulative metadata size delta alongside the step output.
 	// SwapMetadataSizeDelta atomically reads the delta and advances the
 	// loaded baseline so that concurrent handlers in handleGeneratorGroup
@@ -3425,31 +3449,6 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
 	// the next step is enqueued and accounting is handled.
 	if err := runCtx.ReleaseCapacityLease(); err != nil {
 		logger.StdlibLogger(ctx).ReportError(err, "could not release capacity lease early")
-	}
-
-	// Extract AI metadata from step output.
-	// This attempts to detect and parse AI SDK responses (e.g. Vercel AI SDK)
-	// from any step output, using a cheap byte-level pre-check to skip non-AI outputs.
-	if e.allowStepMetadata.Enabled(ctx, runCtx.Metadata().ID.Tenant.AccountID) {
-		// Calculate step duration in milliseconds
-		stepDurationMs := gen.Timing.B / 1_000_000
-
-		md := metadata.WithWarnings(extractors.ExtractAIOutputMetadata(
-			[]byte(output),
-			stepDurationMs,
-		))
-		for _, m := range md {
-			_, err := e.createMetadataSpan(
-				ctx,
-				runCtx,
-				"executor.handleGeneratorStep.aiOutput",
-				m,
-				enums.MetadataScopeStepAttempt,
-			)
-			if err != nil {
-				e.log.Warn("error creating AI output metadata span", "error", err)
-			}
-		}
 	}
 
 	// Steps can be batched with checkpointing!  Imagine an SDK that opts into checkpointing,


pkg/api/apiv1/metadata.go
```diff
@@ -115,6 +123,31 @@ func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID
...
+			EnvID:     auth.WorkspaceID(),
+			AccountID: auth.AccountID(),
+		},
+	}
+
+	var stateMetadata *statev2.Metadata
+	if a.opts.State != nil {
+		md, err := a.opts.State.LoadMetadata(ctx, stateID)
+		if err != nil {
+			logger.StdlibLogger(ctx).Warn("failed to load run metadata for size limit check, falling back to request-only limit",
```
rhino1998: How stale does the run have to be before it is no longer in the state store? 

If we want the ability to also enforce per-run metadata limits for very old completed runs, we can use CH-side info to do it (for the fallback case) with a new metrics query that sums metadata spans for the run.

pkg/execution/executor/finalize.go
```diff
@@ -48,6 +50,31 @@ func (e *executor) Finalize(ctx context.Context, opts execution.FinalizeOpts) er
 		)
 	}
 
+	// Emit a run-scoped usage metadata span so the cumulative metadata size
+	// is visible in the UI alongside other metadata.  We pass nil for
+	// stateMetadata to skip the size-limit check — this is a tiny system span
+	// created at teardown.
+	if opts.Metadata.Metrics.MetadataSize > 0 {
```
rhino1998: Won't this fall out of sync when metadata is added after run end (or be entirely absent if the first metadata span is added after run end)?

Whole-run metadata size could be computed during trace fetch [here](http://github.com/inngest/monorepo/blob/2aa27b8742cf279b33721d74e4ade3be8f91881d/pkg/service/metrics/function_run.go#L2093) where we flatten metadata by summing the metadata values attribute sizes and added as another field plumbed through the metrics service (also saves a span).

pkg/api/apiv1/metadata.go
jakobevangelista: +1 

i also think in `pkg/api/apiv1/metadata.go` we fall back to request limit, but when `LoadMetadata` fails we pass `stateMetadata=nil` into `CreateMetadataSpan`, so run cumulative enforcement is skipped in this path

i think we should consider implementing a real request local fallback cap

AndyInternet: Good catches both. I've addressed the immediate gap, when LoadMetadata fails, we now create a zero-value Metadata as a request-local fallback so CreateMetadataSpan still enforces the 1MB cumulative limit within the request. This prevents unbounded writes to old runs. We also log a warning when this fallback path is hit.

For the CH-based approach: I'll create a ticket for it. We can watch the logs to see if this fallback path is actually getting hit in practice before investing in the additional plumbing. Thoughts?

pkg/execution/executor/finalize.go
AndyInternet: Good feedback Riley! Metadata size is now computed at read time during trace fetch instead of being written as a span.

pkg/api/apiv1/metadata.go
rhino1998: That seems reasonable

PR #3847, opened 2026-03-16
[NO-TKT] Adding Plan To State Proto

mar-inngest:
- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds a `plan` field to the `Tenant` proto message and Go struct, then wires a `PlanSelector` function through `Opts` into both the `checkpointer` and `executor`. An `enrichID` helper is introduced in both structs to stamp the plan onto every state ID before passing it to the state store. The re-enrichment after `LoadMetadata` is intentional — the redis adapter reconstructs `Metadata.ID` from storage without the `Plan` field, so it must be re-applied on every subsequent state call.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit db2a0aa386b9cbcd0180be6592803643322443a6.</sup>

### DIFF
--- a/pkg/api/apiv1/apiv1.go
+++ b/pkg/api/apiv1/apiv1.go
@@ -1,6 +1,7 @@
 package apiv1
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/execution/state/state.go
```diff
@@ -87,6 +87,8 @@ type Identifier struct {
 	WorkspaceID uuid.UUID `json:"wsID"`
 	// AppID represents the app ID for this run
 	AppID uuid.UUID `json:"appID"`
+	// Plan for this run
+	Plan int32 `json:"plan,omitempty"`
```
KiKoS0: this will end up in the queue too right? wonder if we can avoid that

mar-inngest: Good call, lemme see if I can move things around and avoid modifying this

PR #3853, opened 2026-03-18
fix: prevent double JSON encoding of span attributes in postgres

### LINKED ISSUE #3632: Postgres: inngest.spans.attributes stored as JSON string (jsonb_typeof='string'); trace_runs.is_debounce never true (can't audit debounced runs)
We’re using Inngest debounce successfully (behavior works), but we can’t audit/confirm debounced runs from the backing Postgres DB because debounce-related flags/attributes don’t appear to be persisted in a queryable way.

### What we expected
For runs involving debounced execution, we expected either:
- `inngest.trace_runs.is_debounce = true`, and/or
- a span attribute like `sys.debounce.timeout` (per OSS constant `consts.OtelSysDebounceTimeout = "sys.debounce.timeout"`), making it possible to identify debounced runs in the DB.

### What we see instead (prod Postgres)
1. `inngest.trace_runs.is_debounce` is always false:
```sql
select count(*) as debounce_true_total
from inngest.trace_runs
where is_debounce is true
  and to_timestamp(started_at/1000.0) > now() - interval '14 days';
-- returns 0
```

2. `inngest.spans.attributes` appears to be stored as a JSON *string*, not an object:
```sql
select jsonb_typeof(attributes) as type, count(*)::bigint as count
from inngest.spans
where start_time > now() - interval '14 days'
group by 1;

-- returns: type='string' for all/most rows
```

Example row shape (attributes is a quoted JSON blob):
```json
"{\"_inngest.run.id\":\"...\",\"_inngest.function.id\":\"...\", ... }"
```

Because of this, JSONB operators don’t work:
```sql
select count(*)
from inngest.spans
where attributes ? 'sys.debounce.timeout';
-- returns 0 (and generally attributes key queries aren’t possible)
```

### Evidence debounce is happening (behavior)
- Functions are configured with debounce in `inngest.functions.config` (we can see `"debounce": {...}` blocks).
- We see system-initiated cancellations consistent with debounce resets (`inngest.history` has `FunctionCancelled` rows where `cancel_request.UserID` is NULL).
- In dev, we built a minimal debounced “probe” function and confirmed only the last event triggers a run after the debounce window.

### Question / Ask
- Is it expected that `inngest.spans.attributes` is stored as a JSON string in Postgres? If so, what’s the intended way to query attributes (and specifically debounce markers) from SQL?
- If not expected: is this a bug in span normalization/persistence (JSON encoding happening twice)?
- What is the canonical DB-level signal to identify debounced runs (period-triggered or timeout-triggered)?

### Environment
- Inngest Self-hosted
- Postgres backing store (prod)
- Debounced functions exist and behavior works; missing/opaque DB audit signals is the issue.

majiayu000:
Fixes #3632

## Problem

Span attributes are stored as JSONB string type instead of JSONB object in PostgreSQL. This happens because `toNullRawMessage()` calls `json.Marshal()` on a value that is already a JSON string, double-encoding it. As a result, JSONB operators (`?`, `->`, `@>`) fail on string-typed JSONB values, and `trace_runs.is_debounce` is never true.

**Root cause chain:**
- `tracer_sqlc.go:188` — `json.Marshal(attrs)` produces JSON bytes
- `tracer_sqlc.go:230` — `string(attrsByt)` converts to Go string
- `db_normalization.go:861` — passes string to `toNullRawMessage()`
- `normalization.go:451` — `json.Marshal(v)` re-encodes the already-JSON string
- PostgreSQL stores the result as JSONB string type, not object type

## Changes

**Write path** (`normalization.go` + `db_normalization.go`):
- Add `rawJSONToNullRawMessage()` that detects pre-serialized JSON strings/bytes and wraps them directly as `json.RawMessage` without re-marshaling
- Switch `Attributes`, `Links`, and `EventIds` fields in `InsertSpan()` to use the new function
- Keep `Output` and `Input` using `toNullRawMessage()` since they receive raw Go values that need marshaling

**Read path** (`cqrs.go`):
- In `parseSpanFromRow()` and `rollupSpanMetadataFromFragments()`, use a type switch for attributes:
  - `string` → unmarshal (handles legacy double-encoded rows)
  - `map[string]any` → use directly (handles new properly-encoded JSONB objects)
- Move `output_span_id` / `input_span_id` fragment processing outside the attributes type guard

## Test Plan
- Unit tests for `rawJSONToNullRawMessage` covering valid JSON strings/bytes, empty/nil inputs, invalid JSON fallback, and map fallback
- Explicit double-encoding prevention test
- Backward compatibility test for `toNullRawMessage` with Go values
- CI matrix covers both SQLite and PostgreSQL

---

> [!NOTE]
> Fixes double JSON encoding of span attributes in PostgreSQL by introducing `rawJSONToNullRawMessage` that detects pre-serialized JSON strings/bytes and uses them directly without re-marshaling. Updates the read path with a type switch to handle both legacy string-encoded attributes (SQLite/old rows) and new `map[string]any` JSONB objects (PostgreSQL). Also fixes a latent bug where `output_span_id`/`input_span_id` processing was incorrectly gated on attributes being a string type.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 2951f3e9c38bc1b78c2481e71c62ebd26da77b2b.</sup>

### DIFF
--- a/pkg/cqrs/base_cqrs/cqrs.go
+++ b/pkg/cqrs/base_cqrs/cqrs.go
@@ -429,31 +429,33 @@ fragmentLoop:
 			}
 		}
 
-		if attrs, ok := fragment["attributes"].(string); ok {
+		switch attrs := fragment["attributes"].(type) {
+		case string:
 			fragmentAttr := map[string]any{}
 			if err := json.Unmarshal([]byte(attrs), &fragmentAttr); err != nil {
 				logger.StdlibLogger(ctx).Error("error unmarshalling span attributes", "error", err)
 				return nil, err
 			}
-
 			maps.Copy(newSpan.RawOtelSpan.Attributes, fragmentAttr)
+		case map[string]any:
+			maps.Copy(newSpan.RawOtelSpan.Attributes, attrs)
+		}
 
-			if outputRef, ok := fragment["output_span_id"].(string); ok && info != nil {
-				outputSpanID = &outputRef
-				if io, ok := info.dynamicRefs[dynamicSpanID.String]; ok && io != nil {
-					io.OutputRef = outputRef
-				} else {
-					info.dynamicRefs[dynamicSpanID.String] = &IODynamicRef{OutputRef: outputRef}
-				}
+		if outputRef, ok := fragment["output_span_id"].(string); ok && info != nil {
+			outputSpanID = &outputRef
+			if io, ok := info.dynamicRefs[dynamicSpanID.String]; ok && io != nil {
+				io.OutputRef = outputRef
+			} else {
+				info.dynamicRefs[dynamicSpanID.String] = &IODynamicRef{OutputRef: outputRef}
 			}
+		}
 
-			if inputRef, ok := fragment["input_span_id"].(string); ok && info != nil {
-				inputSpanID = &inputRef
-				if io, ok := info.dynamicRefs[dynamicSpanID.String]; ok && io != nil {
-					io.InputRef = inputRef
-				} else {
-					info.dynamicRefs[dynamicSpanID.String] = &IODynamicRef{InputRef: inputRef}
-				}
+		if inputRef, ok := fragment["input_span_id"].(string); ok && info != nil {
+			inputSpanID = &inputRef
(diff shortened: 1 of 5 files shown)


JJZFIVE:
+1

PR #3856, opened 2026-03-18
Pause processing considers event receive time

amh4r:
## Description

When comparing an event to pauses, skip pauses that were created after the event was received

## Motivation

If there's a system error that forces a reprocess (e.g. `i/o timeout` against Redis), then we want to ensure that the event doesn't match pauses created after the initial process

---

> [!NOTE]
> Adds a time-based guard to pause processing so that when an event is reprocessed (e.g. after a Redis i/o timeout), pauses created more than 5 seconds after the event's received time are skipped. Introduces `GetReceivedAt()` on the `TrackedEvent` interface, with `cqrs.Event` returning the real timestamp and `InternalEvent` now carrying a `ReceivedAt` field. All other implementations (`BaseTrackedEvent`, `BatchItem`, `DebounceItem`) return zero time, making the filter a no-op for those types.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 11963ebf94249d7c8289206ba50ea1529176f4e6.</sup>

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -200,6 +200,9 @@ type InternalEvent struct {
 	WorkspaceID uuid.UUID `json:"workspace_id"`
 	// Event is the underlying event received.
 	Event Event `json:"event"`
+
+	// ReceivedAt is the time that our system received the event.
+	ReceivedAt time.Time `json:"received_at"`
 }
 
 func (i InternalEvent) GetAccountID() uuid.UUID {
@@ -219,7 +222,7 @@ func (i InternalEvent) GetEvent() Event {
 }
 
 func (i InternalEvent) GetReceivedAt() time.Time {
-	return time.Time{}
+	return i.ReceivedAt
 }
 
 func IsCron(evtName string) bool {


pkg/event/event.go
tonyhb, line 202: We should/could probably add a ReceivedAt field directly to this internal event, too, so that we can normalize OSS and cloud here

pkg/execution/executor/executor.go
```diff
@@ -2451,6 +2460,14 @@ func (e *executor) handleAggregatePauses(ctx context.Context, evt event.TrackedE
 	for _, i := range evals {
 		// Copy pause into function
 		pause := *i
+
+		// Skip pauses created after the event. They can't be a valid match.
+		// CreatedAt may be zero for older pauses; allow those through. Small
+		// fudge factor, just in case.
+		if !evtTime.IsZero() && !pause.CreatedAt.IsZero() && pause.CreatedAt.After(evtTime.Add(5*time.Second)) {
```
tonyhb: This is done twice - maybe we could add a func to pauses?

```
func (p Pause) CreatedAfterEvent(e event.Event) bool {
    evtTime := e.ReceivedAt()
    if p.CreatedAt.IsZero() || evtTime.IsZero() {
        return false
    }
    return p.CreatedAt.After(evtTime.Add(5 * time.Second))
}
```

PR #3863, opened 2026-03-20
Add Durable Endpoints streaming support

amh4r:
The primary goal of this PR is to add support for Durable Endpoint streaming. Along the way, we noticed issues with the broadcaster implementation and fixed them.

## Durable Endpoint streaming

Add support for Durable Endpoints to stream data to clients (e.g. browsers) through the Inngest backend via Redis pub/sub. When a run goes async, the checkpoint response includes a short-lived realtime JWT so the client can subscribe to the stream via SSE.

- `POST /v1/realtime/publish/tee` accepts signing key auth (in addition to JWT), with request body size limits
- Checkpoint new-run response includes a `realtime_token` for client-side subscription. This JWT has a 15 minute expiry, to match the max duration of a Durable Endpoint streaming.

## Broadcaster unification

Collapse the two-layer broadcaster design (in-memory + Redis wrapper) into a single Redis-backed `broadcaster`. All message delivery (structured, raw bytes, and streaming chunks) flows through Redis pub/sub.

- `Write` publishes to topic key `envID:xxhash(channel):$stream` (where `channel` is the run ID for Durable Endpoints).
- `startTopic` returns a ready channel so `Subscribe` confirms the Redis subscription is active before returning.
- Dev Server uses miniredis when no Redis URI is provided.

## Bug fixes

- Double delivery in `Write` (local + Redis).
- `Write` silently dropped data when no local subscribers existed (broke multi-instance).
- SSE response headers not flushed on connect (clients blocked until first data write or keepalive).
- Stream-start `Data` field was raw bytes instead of a JSON string, causing chunk routing to fail.
- Unbounded request body size on `/publish/tee`.
- `RUnlock()` not hit in broadcaster `keepalive()`.
- TOCTOU race in broadcaster `Close()`.
- Wrap nil error in `PostPublish()`.

---

> [!NOTE]
> Adds Durable Endpoint streaming by routing HTTP response data through the Inngest server to waiting clients via Redis pub/sub. Unifies the two-layer broadcaster into a single Redis-backed implementation, adds signing-key auth fallback to `POST /realtime/publish/tee`, includes a `realtime_token` in checkpoint responses, and fixes several bugs (double delivery, silent data drop, SSE header flushing, stream-start JSON encoding, goroutine leak on partial subscription failure, TOCTOU race in `Close()`).
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 40e2d30771169673be039b33b989858abe14598b.</sup>


pkg/execution/realtime/broadcaster.go
```diff
@@ -176,9 +211,61 @@ func (b *broadcaster) subscribe(
 		Tags:    map[string]any{},
 	})
 
+	b.l.Unlock()
+
+	// Wait for Redis subscription confirmations outside the lock. Do this
+	// outside the lock in case any of the topics take a long time to be ready.
+	var anyErr error
+	for _, pt := range pendingTopics {
+		if err := <-pt.ready; err != nil {
```
rhino1998: I'm always afraid of bare channel receives/sends (outside of a select). Tracing the rest of these changes, I think if runTopic panics then this never receives? Maybe do this w/ a selec with the context too? So at the very least it doesn't hang indefinitely if the context is canceled.

pkg/execution/realtime/broadcaster.go
```diff
@@ -602,6 +703,214 @@ func (b *broadcaster) keepalive(ctx context.Context, subID uuid.UUID) {
...
+	for i := 0; i < redisPublishAttempts; i++ {
+		if i > 0 {
+			metrics.IncrRealtimeRedisOpsTotal(ctx, metrics.CounterOpt{
+				PkgName: "realtime",
+				Tags: map[string]any{
+					"op":     "publish",
+					"status": "retry",
+				},
+			})
+			<-time.After(redisRetryInterval)
```
rhino1998: Could (maybe should?) use a single timer that we reset here instead of time.After and also use a select w/ ctx.Done() so that if the context is canceled during a retry we don't wait for the retry interval sleep.

pkg/execution/realtime/broadcaster.go
```diff
@@ -51,30 +77,49 @@ func newBroadcaster() *broadcaster {
...
-	// TopicStop is called when the last subscriber disconnects from a topic.
-	TopicStop func(ctx context.Context, t Topic) error
+	// pubc is the client connected to Redis for publishing messages.
+	pubc rueidis.Client
+
+	// subc is the client connected to Redis for subscribing to pub-sub streams.
+	subc rueidis.Client
+
+	// redisMu protects topicCancelFuncs.
+	redisMu sync.Mutex
```
rhino1998: NIT: to make this relationship more obvious, maybe call this topicCancelMu or topicCancelLock?

pkg/execution/realtime/broadcaster.go
```diff
@@ -331,92 +408,114 @@ func (b *broadcaster) Close(ctx context.Context) error {
...
 				logger.StdlibLogger(ctx).Warn("error closing realtime subscription", "error", err)
 			}
 		}
+		b.l.RUnlock()
+
+		// Cancel topics
+		b.redisMu.Lock()
+		for key, cancel := range b.topicCancelFuncs {
+			cancel()
+			delete(b.topicCancelFuncs, key)
```
rhino1998: exteremely minor NIT: maybe just `clear(b.topicCancelFuncs)` after the loop? (so that the map doesn't need to track all the intermediate deletes?)

pkg/execution/realtime/broadcaster.go
```diff
@@ -176,9 +211,61 @@ func (b *broadcaster) subscribe(
...
+			delete(as.Topics, pt.topic.String())
+			if len(as.Topics) == 0 {
+				delete(b.subs, s.ID())
+			}
+		}
+	}
+	b.l.Unlock()
+
+	for _, t := range topicsToStop {
+		b.stopTopic(t)
```
rhino1998, line 270: If `redisMu` is removed, then `stopTopic` should be moved into the locked region above.

amh4r: I think I'm gonna keep that lock (now renamed to `topicCancelMu`, per your other comment). While we could restructure other code to ensure `topicCancelFuncs` stuff only happens within the other lock, I'm a bit concerned about the risk that would create. The fact that we're in a "refactor consumer code to ensure it's locked" scenario now makes me think there's a risk of regressing back into it later with another code change.

In other words, while `topicCancelMu` might technically be superfluous, it seems safe to ensure it has its own lock just in case consumers do something wonky

pkg/devserver/devserver.go
```diff
@@ -1020,6 +1023,48 @@ func PartitionConstraintConfigGetter(dbcqrs cqrs.Manager) queue.PartitionConstra
...
+		}
+		subc, err := connectToOrCreateRedis(redisURI)
+		if err != nil {
+			pubc.Close()
+			return nil, nil, fmt.Errorf("error creating realtime redis sub client: %w", err)
+		}
+		return realtime.NewRedisBroadcaster(pubc, subc), func() { pubc.Close(); subc.Close() }, nil
+	}
+
+	r := miniredis.NewMiniRedis()
```
tonyhb: Why would we want this here?  We **must** have some connection to a backend in order to bring this up.

tonyhb: Or, in other words, if you bring up many replicas (self hosting?) you go back to in-memory broadcasting.  Should probably have a way to pass in the creds for this from DI.

PR #3868, opened 2026-03-21
Fix silent message loss in Connect gateway Forward path

amh4r:
Fixes intermittent `connect_worker_stopped_responding` errors during Gateway drains.

`Forward()` returned `Success: true` after the channel send, before the WebSocket write. If the write then failed (e.g. context cancelled during drain), the message was silently dropped and the executor waited until the lease expired, producing `connect_worker_stopped_responding`.

Now `Forward()` blocks until the WebSocket write completes and returns the actual result.

---

> [!NOTE]
> Fixes silent message loss in the Connect gateway Forward path by introducing a `forwardMessage` wrapper that pairs each request with a result channel. `Forward()` now blocks until the WebSocket write completes before returning success/failure to the executor, preventing the silent drop that occurred when writes failed during gateway drains. The latest commit adds structured logging fields (account/app/env/fn/req IDs) to the error and timeout log lines in `Forward()`.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 4ac4e049ec28020bba1a517b67a4a8fd6a876677.</sup>

### DIFF
--- a/pkg/connect/rpc.go
+++ b/pkg/connect/rpc.go
@@ -38,14 +38,29 @@ func (c *connectGatewaySvc) Forward(ctx context.Context, req *pb.ForwardRequest)
 			select {
 			case err := <-resultCh:
 				if err != nil {
-					l.Error("failed to write message to websocket", "err", err)
+					l.Error("failed to write message to websocket",
+						"err", err,
+						"account_id", req.Data.AccountId,
+						"app_id", req.Data.RunId,
+						"conn_id", req.ConnectionID,
+						"env_id", req.Data.EnvId,
+						"fn_id", req.Data.FunctionId,
+						"req_id", req.Data.RequestId,
+					)
 					return &pb.ForwardResponse{Success: false}, nil
 				}
 				return &pb.ForwardResponse{Success: true}, nil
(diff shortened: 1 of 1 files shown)


pkg/connect/rpc.go
```diff
@@ -8,20 +8,46 @@ import (
...
 
-			return &pb.ForwardResponse{Success: true}, nil
+		select {
+		case msgChan <- msg:
+			// Block until the WebSocket write completes. Only ack success to
+			// the executor after confirmed delivery to the worker.
+			select {
+			case err := <-resultCh:
+				if err != nil {
+					l.Error("failed to write message to websocket", "err", err)
```
KiKoS0: do we want to include any ids in these errors?

amh4r: Seems like the connect ID is the only ID we have access to here. Is that true? Ideally we'd also have run ID, but I'm not sure if that's possible

KiKoS0: we can peek inside the `req.Data` for that can't we?

amh4r: Oops, you're right. I added them

PR #3881, opened 2026-03-25
Semaphores, Fn concurrency, Worker concurrency

tonyhb:
TODO:

- [x] Add semaphores as fundamental primitive to constraint svc
- [x] Plumb semaphore mgr into executor for schedule, finalize
- [x] Add fn level concurrency to config
- [x] Update fn config with app level concurrency via workers

- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds semaphores as a first-class constraint primitive to power worker and function concurrency. Semaphores use Redis INCRBY/DECRBY counters for O(1) capacity checks, with two release modes: auto (per-step) and manual (held for the run lifetime). The PR wires fn-level and worker-level concurrency through the semaphore system, adds connect lifecycle hooks to adjust capacity on worker connect/disconnect, and updates protos accordingly.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a74681c1e2cb6ad2489640e73a237b8f3972c00c.</sup>

### DIFF
--- a/pkg/constraintapi/constraints_test.go
+++ b/pkg/constraintapi/constraints_test.go
@@ -1420,6 +1420,50 @@ func TestConstraintEnforcement(t *testing.T) {
 				require.Equal(t, 0, resp.AvailableCapacity)
 			},
 		},
+
+		{
+			name: "semaphore auto release",
+			config: ConstraintConfig{
+				FunctionVersion: 1,
+				Semaphores: []Semaphore{
+					{ID: "app:test-sem", Weight: 1, Release: SemaphoreReleaseAuto},
+				},
+			},
+			constraints: []ConstraintItem{
+				{
+					Kind: ConstraintKindSemaphore,
+					Semaphore: &SemaphoreConstraint{
+						ID:      "app:test-sem",
+						Weight:  1,
+						Release: SemaphoreReleaseAuto,
+					},
+				},
+			},
+			amount:              1,
+			expectedLeaseAmount: 1,
+			beforeAcquire: func(t *testing.T, deps *deps) {
+				// Set semaphore capacity to 5
+				sem := deps.constraints[0].Semaphore
+				capKey := sem.CapacityKey(accountID)
+				require.NoError(t, deps.r.Set(capKey, "5"))
+			},
+			afterAcquire: func(t *testing.T, deps *deps, resp *CapacityAcquireResponse) {
+				// Semaphore usage should be 1
+				sem := deps.constraints[0].Semaphore
+				usageKey := sem.UsageKey(accountID)
+				val, err := deps.r.Get(usageKey)
+				require.NoError(t, err)
+				require.Equal(t, "1", val)
+			},
+			afterRelease: func(t *testing.T, deps *deps, resp *CapacityReleaseResponse) {
+				// After release, semaphore usage should be decremented (auto-release)
+				sem := deps.constraints[0].Semaphore
+				usageKey := sem.UsageKey(accountID)
+				val, err := deps.r.Get(usageKey)
+				require.NoError(t, err)
+				require.Equal(t, "0", val)
+			},
+		},
 	}
 
 	for _, test := range testCases {
--- a/pkg/constraintapi/convert_test.go
+++ b/pkg/constraintapi/convert_test.go
@@ -178,6 +178,11 @@ func TestConstraintKindConversion(t *testing.T) {
 			input:    ConstraintKindThrottle,
 			expected: pb.ConstraintApiConstraintKind_CONSTRAINT_API_CONSTRAINT_KIND_THROTTLE,
 		},
+		{
+			name:     "semaphore",
+			input:    ConstraintKindSemaphore,
+			expected: pb.ConstraintApiConstraintKind_CONSTRAINT_API_CONSTRAINT_KIND_SEMAPHORE,
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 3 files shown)


pkg/constraintapi/lua/acquire.lua
```diff
@@ -236,6 +236,17 @@ for index, value in ipairs(constraints) do
 		local throttleRes = throttle(value.t.k, nowMS, value.t.p, value.t.l, maxBurst, 0)
 		constraintCapacity = throttleRes["remaining"] or 0
 		constraintRetryAt = toInteger(throttleRes["retry_at"]) -- already in ms
+	elseif value.k == 4 then
+		-- semaphore
+		local usage = tonumber(call("GET", value.sem.k)) or 0
+		local capacity = tonumber(call("GET", value.sem.ck)) or 0
```
BrunoScheufler, line 242: Should we store capacity as a key or supply it as a number in the call like we do with existing constraints?

pkg/constraintapi/lua/semaphore_set_capacity.lua
```diff
@@ -0,0 +1,21 @@
+--
+-- semaphore_set_capacity
+-- Idempotently sets total capacity for a named semaphore.
+--
+
+local keyCapacity = KEYS[1]
```
BrunoScheufler, line 6: Should we merge this with adjust capacity? Do we need both?

pkg/execution/queue/backlog.go
```diff
@@ -247,6 +251,11 @@ func ItemBacklog(ctx context.Context, i QueueItem) QueueBacklog {
 		}
 	}
 
+	// Propagate semaphore constraints from item to backlog
+	if len(i.Data.Semaphores) > 0 {
+		b.Semaphores = i.Data.Semaphores
```
BrunoScheufler: We might have to be careful here
- Can items for the same backlog have different semaphores?
- If so, do we need to group those items by semaphore when it comes to backlogs?
- If so, we'll have to make sure the backlog ID generation is adjusted

BrunoScheufler:
On codecov, let's make sure to update tests for conversion, add Constraint API tests for handling the semaphore during acquire, integration tests for the semaphore manager, and queue tests to ensure semaphores are checked when present on the item.

pkg/constraintapi/lua/acquire.lua
tonyhb: Store it so that lookups don't exist on the hot path.  It also needs to be stored due to the mutations required for worker concurrency.

pkg/execution/queue/backlog.go
tonyhb: > Can items for the same backlog have different semaphores?
> If so, do we need to group those items by semaphore when it comes to backlogs?
> If so, we'll have to make sure the backlog ID generation is adjusted

This should have nothing to do with backlogs, unless im missing something?

tonyhb: We've actually removed everything from backlogs and instead store directly on queue items, which is required - because queue items themselves dictate semaphore flows.

pkg/constraintapi/semaphore_manager.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// AdjustCapacity atomically adjusts capacity by delta (e.g., +N on worker connect, -N on disconnect).
+	AdjustCapacity(ctx context.Context, accountID uuid.UUID, name, idempotencyKey string, delta int64) error
+
+	// GetCapacity returns current capacity and usage for a named semaphore.
+	GetCapacity(ctx context.Context, accountID uuid.UUID, name, usageValue string) (capacity int64, usage int64, err error)
+
+	// ReleaseSemaphore decrements the usage counter for a manual-release semaphore.
+	// Called on run finalization for function concurrency. Must be idempotent.
+	ReleaseSemaphore(ctx context.Context, accountID uuid.UUID, name, usageValue, idempotencyKey string, weight int64) error
```
lkasinathan, line 32: nano-nit: s/usageValue/usageKey perhaps.

I was initially a bit confused by usageValue here before I realized it's the key for usage counter.

pkg/constraintapi/lua/semaphore_set_capacity.lua
tonyhb: We need both when it comes time for changing these via an API (internal or public)

pkg/connect/lifecycles/semaphore.go
```diff
@@ -0,0 +1,102 @@
...
+
+	l := logger.StdlibLogger(ctx)
+
+	for _, group := range conn.Groups {
+		if group.AppID == nil {
+			continue
+		}
+
+		semID := constraintapi.SemaphoreIDApp(*group.AppID)
+		idempotencyKey := fmt.Sprintf("connect-%s", conn.ConnectionId)
```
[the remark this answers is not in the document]
tonyhb: connection IDs aren't reused.

pkg/inngest/concurrency.go
```diff
@@ -110,18 +147,87 @@ func hashConcurrencyKey(key string) string {
...
+// carry the constraint.
+type FnConcurrency struct {
+	Limit int                `json:"limit"`
+	Scope FnConcurrencyScope `json:"scope,omitempty"` // defaults to "fn"
+	Key   *string            `json:"key,omitempty"`   // optional expression for the semaphore name
+
+	// ID represents the pre-computed semaphore ID for this concurrency limit.
+	// Set internally during registration (e.g., "app:<appID>" for connect apps).
+	// For fn-scoped limits without a pre-set ID, evaluateFnConcurrency computes it.
+	ID string `json:"id,omitempty"`
```
lkasinathan, line 190: I don't fully understand why we expose `ID` here.

- For app scoped limit, we inject it internally with appID and pass that through when calling schedule. 
- For fn scoped limit, we always ignore this ID and just call [evaluateFnConcurrency](https://github.com/inngest/inngest/blob/187cf4b30a33e1ef83e3f51be1911b39a04cd554/pkg/execution/executor/constraints.go#L310)

For the former, we can just set the Semaphore ID to `constraintapi.SemaphoreIDApp(appID)` when creating the sempahore during schedule, and for the latter, it seems unused -- so we can probably remove this field altogether?

pkg/connect/lifecycles/semaphore.go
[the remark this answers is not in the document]
tonyhb: Conns always get new IDs.  Not valid.

PR #3890, opened 2026-03-26
Fix Vercel serve path configuration deletion by making the rows controlled with stable keys

albertchae:
## Description

User reported a bug where deleting a row would visually make it look
like the last row was deleted. Saving the change actually sent the right
data over the network but this was still a bad UI bug.

It turns out these rows were uncontrolled inputs too because they only
set defaultValue and not value, so change them to fully controlled.

The react key with array index is precisely the pitfall here since we
allow deletions
https://react.dev/learn/rendering-lists#why-does-react-need-keys

- [ ] Docs

## Checklist

---

> [!NOTE]
> Fixes a UI bug where deleting a serve path row would visually appear to delete the last row instead of the selected one. The fix replaces array-index-based React keys with stable UUIDs and converts uncontrolled inputs (`defaultValue`) to controlled inputs (`value`).
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 07e83d8040222db27844ae81bd3618694930fda5.</sup>

### DIFF
--- a/ui/apps/dashboard/src/routes/_authed/settings/integrations/vercel/configure/$id/index.tsx
+++ b/ui/apps/dashboard/src/routes/_authed/settings/integrations/vercel/configure/$id/index.tsx
@@ -31,6 +31,19 @@ import { useVercelIntegration } from '@/queries/useVercelIntegration';
 
 const defaultPath = '/api/inngest';
 
+type ServePathRow = {
+  id: string;
+  value: string;
+};
+
+function createServePathRow(value: string): ServePathRow {
+  return {
+    // Generate a stable key for each row using a random number converted to a base-36 string.
+    id: Math.random().toString(36).slice(2),
+    value,
+  };
+}
+
 export const Route = createFileRoute(
   '/_authed/settings/integrations/vercel/configure/$id/',
 )({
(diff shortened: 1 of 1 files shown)


ui/apps/dashboard/src/routes/_authed/settings/integrations/vercel/configure/$id/index.tsx
```diff
@@ -252,16 +269,19 @@ function VercelConfigure() {
...
                     className="flex flex-row items-center justify-start"
                   >
                     <div className="mr-2 w-full">
                       <Input
-                        defaultValue={path}
+                        value={path.value}
                         className="text-basis h-10 w-full px-2 py-2 text-base"
                         onChange={({ target: { value } }) => {
-                          setPaths(paths.map((p, n) => (i === n ? value : p)));
-                          setProject(project);
```
albertchae, line 264: setProject is not needed now that the row is controlled

ui/apps/dashboard/src/routes/_authed/settings/integrations/vercel/configure/$id/index.tsx
```diff
@@ -252,16 +269,19 @@ function VercelConfigure() {
...
+                          setPaths((prevPaths) =>
```
albertchae, line 280: using functional form of setPaths since it's based on the previous state https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

PR #3893, opened 2026-03-26
Fix connect graceful shutdown draining

BrunoScheufler:
## Description

This PR adds some small changes to the connect gateway, including
- removing the context from message handling, which may prevent connection status update/message forward operations from being canceled when draining starts
- marking the connection as draining in memory when the gateway itself starts draining (ensuring all subsequent updates don't flip the connection back to healthy)

- [ ] Docs

## Checklist

---

> [!NOTE]
> This PR fixes connect gateway graceful shutdown by: removing the request context from `handleIncomingWebSocketMessage` and using `context.Background()` for all state updates to prevent cancellation during drain; setting `ch.draining.Store(true)` before the Redis status update to close the race window where heartbeats could reset status back to READY; adding a 5s `wsWriteTimeout` for all WebSocket writes; and fixing the router to not delete connections on a draining gateway. Includes comprehensive drain scenario tests covering heartbeat status preservation, new connection rejection, reply/ack/lease-extension processing during drain.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 6fbe256f5ffb3d8c38c9dc62b84bdea6a6f0a2cb.</sup>

### DIFF
--- a/pkg/connect/routing/router.go
+++ b/pkg/connect/routing/router.go
@@ -397,11 +397,6 @@ func isHealthy(ctx context.Context, stateManager state.StateManager, envID uuid.
 				shouldDeleteUnhealthyGateway:    true,
 			}
 		}
-
-		// Drop associated connection
-		return isHealthyRes{
-			shouldDeleteUnhealthyConnection: true,
-		}
 	}
 
 	return isHealthyRes{


pkg/connect/gateway.go
```diff
@@ -264,6 +264,9 @@ func (c *connectGatewaySvc) Handler() http.Handler {
 			// If gateway is shutting down, we must immediately start the draining process
 			ch.log.Debug("context done, starting draining process")
 
+			// Mark connection as draining to prevent subsequent heartbeats from marking it as ready
```
BrunoScheufler, line 273: Not sure why we didn't do this before but when the gateway itself is draining, we can make sure that all further connection updates mark it as draining too.

pkg/connect/gateway.go
```diff
@@ -498,7 +503,7 @@ func (c *connectGatewaySvc) Handler() http.Handler {
 					Tags:    tags,
 				})
 
-				serr := ch.handleIncomingWebSocketMessage(ctx, &msg)
+				serr := ch.handleIncomingWebSocketMessage(&msg)
```
BrunoScheufler, line 514: Passing the context here is misleading since we likely want to allow most operations even when draining. The context is canceled when draining starts.

pkg/connect/gateway.go
```diff
@@ -687,22 +705,33 @@ func (c *connectionHandler) handleIncomingWebSocketMessage(ctx context.Context,
...
+		c.log.Debug("marked worker as ready")
+
 		return nil
 	case connectpb.GatewayMessageType_WORKER_HEARTBEAT:
-		if c.svc.isDraining.Load() {
-			return &ErrDraining
+		// Always allow heartbeats
+
+		// Don't reset status to READY if the connection or service is draining.
+		status := connectpb.ConnectionStatus_READY
```
BrunoScheufler, line 723: This was a little weird. We should always allow heartbeats from connections but when draining, those should be marked as draining.

PR #3932, opened 2026-04-03
EXE-1546: Add per-function Prometheus metrics

djfarrelly:
## Description
Add new Prometheus metrics which can be scraped per-function using in-memory counters. This expands the metrics available. 

High cardinality of function slugs could cause memory issues, but those totals would have to be very high.

Set `INNGEST_EXPERIMENTAL_PROM_METRICS=1` to enable new per-function metrics, unset or use `INNGEST_EXPERIMENTAL_PROM_METRICS=0` to disable.

NOTE - This is forked from the latest 1.13.x release here: https://github.com/inngest/inngest/tree/v1.13.7 - if this is stable and accepted, we can release a stable 1.13.8 release then the head of the repo to bring it into main. This will create a fork for the time being.

## Motivation
EXE-1546

Built on the changes in #3912, but back ported for `v1.13.x` and refactored with fixes and with additional tests for the lifecycle methods. This will be brought to future releases after beta testing.

- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds per-function Prometheus lifecycle metrics (function run and SDK step counters labeled by function slug) behind an `INNGEST_EXPERIMENTAL_PROM_METRICS` env var toggle. Refactors the metrics registry from instance-level to package-level so lifecycle metrics share the same registry as the queue depth gauge. Removes the signing-key auth requirement from `/metrics` in the dev server. The latest commit also fixes goreleaser to publish pre-release Docker images without tagging them as `latest`.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 11e62b87ea524c6b6efe9f6b18d83db1a36ccc1a.</sup>

### DIFF
--- a/pkg/metrics/lifecycle.go
+++ b/pkg/metrics/lifecycle.go
@@ -66,6 +66,15 @@ func NewPrometheusLifecycleListener() *PrometheusLifecycleListener {
 	return &PrometheusLifecycleListener{}
 }
 
+// NewLifecycleListeners returns the Prometheus lifecycle listener if the
+// experimental metrics feature is enabled, or an empty slice otherwise.
+func NewLifecycleListeners() []execution.LifecycleListener {
+	if !ExperimentalPromMetricsEnabled() {
+		return nil
+	}
+	return []execution.LifecycleListener{NewPrometheusLifecycleListener()}
+}
+
 func fnLabel(md statev2.Metadata) string {
 	if slug := md.Config.FunctionSlug(); slug != "" {
 		return slug
--- a/pkg/metrics/lifecycle.go
+++ b/pkg/metrics/lifecycle.go
@@ -1,0 +1,158 @@
+package metrics
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/inngest/inngest/pkg/event"
+	"github.com/inngest/inngest/pkg/execution"
+	"github.com/inngest/inngest/pkg/execution/queue"
+	statev1 "github.com/inngest/inngest/pkg/execution/state"
+	statev2 "github.com/inngest/inngest/pkg/execution/state/v2"
+	"github.com/inngest/inngest/pkg/inngest"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	functionRunScheduled = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "inngest_function_run_scheduled_total",
+		Help: "The total number of function runs scheduled",
+	}, []string{"fn"})
+
+	functionRunStarted = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "inngest_function_run_started_total",
+		Help: "The total number of function runs started",
+	}, []string{"fn"})
+
+	functionRunEnded = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "inngest_function_run_ended_total",
+		Help: "The total number of function runs ended",
+	}, []string{"fn", "status"})
+
+	sdkReqScheduled = prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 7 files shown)


pkg/metrics/lifecycle.go
```diff
@@ -0,0 +1,158 @@
...
+	}, []string{"fn"})
+
+	sdkReqEnded = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "inngest_sdk_req_ended_total",
+		Help: "The total number of SDK invocation/step execution ended",
+	}, []string{"fn", "status"})
+)
+
+func init() {
+	registry.MustRegister(
```
djfarrelly, line 49: Uses a local registry instead of the global default registry as it could cause duplicate registration or collisions with metrics. This is safer and cleaner.

pkg/metrics/lifecycle.go
```diff
@@ -0,0 +1,158 @@
...
+
+func NewPrometheusLifecycleListener() *PrometheusLifecycleListener {
+	return &PrometheusLifecycleListener{}
+}
+
+func fnLabel(md statev2.Metadata) string {
+	if slug := md.Config.FunctionSlug(); slug != "" {
+		return slug
+	}
+	return "unknown"
```
djfarrelly, line 82: Use "unknown" instead of the uuid as that could just pollute the tags. This should never happen given every function has a slug, but this is a cleaner fallback to group all unknown metrics than have hundreds or thousands of tags.

pkg/metrics/metrics.go
```diff
@@ -91,35 +93,49 @@ type Opts struct {
 	QueueManager   QueueManager
 }
 
+// ExperimentalPromMetricsEnabled reports whether the experimental Prometheus
+// lifecycle metrics are enabled via the INNGEST_EXPERIMENTAL_PROM_METRICS
+// environment variable. The feature is off by default; set to "1", "true",
+// "yes", or any truthy value to enable. "0" always disables.
+func ExperimentalPromMetricsEnabled() bool {
```
AndyInternet, line 100: Should we document this in the CLI help output? This will become tribal knowledge that's easy to lose track of if we don't document it somewhere.

pkg/metrics/lifecycle.go
```diff
@@ -0,0 +1,167 @@
...
+	md statev2.Metadata,
+	_ queue.Item,
+	_ []json.RawMessage,
+	resp statev1.DriverResponse,
+) {
+	status := "Completed"
+	if resp.Err != nil {
+		status = "Failed"
+	}
+	functionRunEnded.WithLabelValues(fnLabel(md), status).Inc()
```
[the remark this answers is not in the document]
AndyInternet: I agree with thin finding. Also, prometheus labels are lowercase as a convention

pkg/metrics/metrics.go
```diff
@@ -91,35 +93,49 @@ type Opts struct {
...
+	v := strings.TrimSpace(os.Getenv("INNGEST_EXPERIMENTAL_PROM_METRICS"))
+	if v == "" || v == "0" || strings.EqualFold(v, "false") || strings.EqualFold(v, "no") {
+		return false
+	}
+	return true
```
AndyInternet, line 105: This logic seems inverted. If we're defaulting to on, we should change this to
	if v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes") || strings.EqualFold(v, "on") {
		return true
	}
	return false

AndyInternet (approved):

AndyInternet (approved):

pkg/metrics/metrics.go
djfarrelly: Good call - maybe we should make it opt-in by default and move out of experimental, like `INNGEST_PROM_METRICS_ENABLED=1` - wdyt?

PR #3945, opened 2026-04-03
refactor(base_cqrs): switch migrations to goose baselines

darwin67:
## What changed

- Switches `pkg/cqrs/base_cqrs` runtime migrations from `golang-migrate` to goose baselines for both SQLite and Postgres.
- Moves the baseline SQL under `pkg/db/{sqlite,postgres}/migrations` and keeps the legacy migration files so the legacy-to-goose transition can be tested.
- Adds dumped schemas under `pkg/db/{sqlite,postgres}/schema.sql` and updates `make schema-dump` to write both the existing `base_cqrs/sqlc` schema files and the new `pkg/db` schema files.
- Aligns the goose baselines with the dumped schemas from `main`, including the goose metadata table.
- Adds migration and regression coverage for schema parity, defaults, round-trips, and legacy-to-goose baseline behavior.
- Fixes follow-on issues found during rebase/testing:
  - SQLite `span_fragments` scan handling
  - Postgres `InsertEventBatch` ULID binding via a local `TextULID` sqlc wrapper
  - schema test expectations around `goose_db_version`
  - Go SDK test assertion for replayed trailing `"end"`

## Why

This consolidates runtime schema management around goose baselines while preserving compatibility with databases that were already migrated with the legacy flow. The branch also tightens schema verification so future drift between baselines, dumps, and generated sqlc schemas is caught directly in tests.

## Validation

```bash
go test ./pkg/db -count=1
go test ./pkg/cqrs/base_cqrs -count=1
TEST_DATABASE=postgres go test ./pkg/db -count=1
TEST_DATABASE=postgres go test ./pkg/cqrs/base_cqrs -count=1
make schema-dump
go test ./tools/schema-dump -count=1
```

Targeted follow-up checks run while fixing CI regressions:

```bash
go test ./pkg/db -run 'TestSchemaColumnsMatchSqlc|TestCrossDialectSchemaParity' -count=1
TEST_DATABASE=postgres go test ./pkg/db -run 'TestSchemaColumnsMatchSqlc|TestCrossDialectSchemaParity' -count=1
go test ./pkg/cqrs/base_cqrs -run 'Test(LegacyMigrationThenGooseBaselineIsNoopSQLite|SchemaMatchesSqlcSQLite|BaselineOnFreshSQLite)' -count=1
TEST_DATABASE=postgres go test ./pkg/cqrs/base_cqrs -run 'Test(LegacyMigrationThenGooseBaselineIsNoopPostgres|SchemaMatchesSqlcPostgres|BaselineOnFreshPostgres)' -count=1
```

---

> [!NOTE]
> Switches `pkg/cqrs/base_cqrs` runtime migrations from golang-migrate to goose baselines for SQLite and Postgres, adds dumped schemas under `pkg/db/{sqlite,postgres}/schema.sql`, aligns baselines with the dumped schemas, and adds comprehensive migration/regression test coverage (schema parity, defaults, round-trips, legacy-to-goose noop checks). Also adds a plan tracking script and updates the plan doc with status metadata.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 695b81c63b2401eafbc2eac2dfc9d9f6f064af60.</sup>


pkg/cqrs/base_cqrs/cqrs.go
```diff
@@ -3139,7 +3148,7 @@ func (w wrapper) convertSpanRunRows(
 
 		traceRun := &cqrs.TraceRun{
 			AccountID:   accountUUID,
-			WorkspaceID: accountUUID,
+			WorkspaceID: workspaceUUID,
```
albertchae, line 3151: 😱  although shouldn't matter since we don't have more than one account/workspace anyway

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE apps (
 	status VARCHAR NOT NULL,
 	error TEXT,
 	checksum VARCHAR NOT NULL,
-	created_at TIMESTAMP NOT NULL,
+	created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
albertchae, line 19: hm if you're going to add this default this we might end up having some issues because of inconsistent SQLite timestamp handling, I wrote about some of it in https://github.com/inngest/inngest/pull/2684#issue-3228659128

I believe CURRENT_TIMESTAMP returns format `2026-04-16 21:53:12`, but we store go time as string and goqu does it as RFC3339

pkg/db/postgres/querier.go
```diff
@@ -229,11 +230,23 @@ func (pq *pgQuerier) GetEventsIDbound(ctx context.Context, arg db.GetEventsIDbou
...
+		arg.AccountID,
+		arg.WorkspaceID,
+		arg.AppID,
+		arg.WorkflowID,
+		arg.RunID.String(),
+		arg.StartedAt,
+		arg.ExecutedAt,
+		arg.EventIds,
+	)
+	return err
```
albertchae: did AI inline this? I thought we wanted to manage all these queries with sqlc

pkg/cqrs/base_cqrs/migration_test.go
```diff
@@ -0,0 +1,540 @@
...
+	defer cleanup()
+
+	require.NoError(t, up(db, opts))
+
+	actual := readApplicationSchemaSnapshot(t, db, migrationDialectPostgres)
+	expected := expectedTableColumns(t, migrationDialectPostgres)
+	require.Equal(t, expected, actual.Tables)
+
+	expectedIdx := expectedIndexNames(t, migrationDialectPostgres)
+	actualIdx := actualIndexNames(actual.Indexes)
```
albertchae: interesting that this passes because when I tested this locally I saw several indexes from the golang migrate migrations that were missing from the goose one

https://albertchae.github.io/inngest-schema/postgres-schema-diff.html

e.g. `idx_events_internal_id` was added in https://github.com/inngest/inngest/blob/bf3bf91fed2163cab66ae6326703b451d9fa5d50/pkg/cqrs/base_cqrs/migrations/postgres/000012_add_pg_indexes.up.sql#L11

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE apps (
 	status VARCHAR NOT NULL,
```
albertchae, line 16: how did you dump this schema file? Unless I'm missing something it looks like we didn't have a canonical way to dump db schema in this repo before, we should probably add one and add it to the `make gen` target. Probably good to have one for postgres as well

albertchae: I think there were several changes to the sqlite schema that aren't captured here. e.g.
https://albertchae.github.io/inngest-schema/sqlite-schema-diff.html

looks like we changed a good number of BLOB columns previously for ULIDs to CHAR(26) which is good (if we're choosing to store them as strings rather than bytes). The ones where we change CHAR(36) to UUID is weird because of sqlite's weird type affinity system https://www.sqlite.org/datatype3.html

Since UUID isn't an actual type in sqlite, the type affinity would match it to NUMERIC. But then if we insert those UUIDs as strings anyway, it would actually be stored as a TEXT since those UUIDs wouldn't be a valid INTEGER or REAL.

You can see this behavior by going to https://sqlime.org/

then run
```
CREATE TABLE test (
  a UUID,
  b CHAR(36),
  c TEXT
);

-- Check the declared types and affinities
PRAGMA table_info(test);

-- Insert the same UUID into all three
INSERT INTO test VALUES (
  '550e8400-e29b-41d4-a716-446655440000',
  '550e8400-e29b-41d4-a716-446655440000',
  '550e8400-e29b-41d4-a716-446655440000'
);
```

Then running `SELECT typeof(a), typeof(b), typeof(c) FROM test;` should return text for all 3 anyway. But you could see that the UUID column has NUMERIC affinity by doing `INSERT INTO test VALUES ('123', '123', '123');` and then checking the types again and you'd see it parsed it as an Integer first.

Anyway that is a long explanation for why I think labeling those types as UUID could be confusing.

albertchae (commented):
mostly looks good but I think we're missing some postgres indices and changed some of the types in SQLite weirdly (although I don't think those would impact the code at runtime)

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
darwin67: Gonna rebase this PR against `main` and recalibrate now that https://github.com/inngest/inngest/pull/4019 is merged.

pkg/db/postgres/querier.go
darwin67: yeah I think so, missed this. thanks.

darwin67: fixed this with a ULID text wrapper for now since the postgres id field is a string type instead of a binary type...

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
darwin67: recalibrated the schema dump.
it nows dump to the original location and `pkg/db` but it references the baseline sql now.

the only diffs should be the goose db version data model, but everything else should still stay the same.

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
darwin67: with the schema dump tweaks, we can confirm that this is the actual schema in the db because when running the schema-dump tool, there are no diffs on the original `sqlite/schema.sql` besides the newly added goose migration tables.

pkg/cqrs/base_cqrs/base_cqrs.go
```diff
@@ -131,85 +127,27 @@ func New(ctx context.Context, opts BaseCQRSOptions) (*sql.DB, error) {
 	return db, err
 }
 
-// FS contains the filesystem of the stdlib, containing all migrations in subdirs
-// relative to this package.
-//
-//go:embed **/**/*.sql
-var FS embed.FS
```
darwin67, line 138: with this deletion, there should be no migration being ran from this pkg anymore.

schemas here are only kept for full backward compatibility check purposes, and they can be wiped after this PR is merged.

pkg/cqrs/base_cqrs/migration_test.go
darwin67: This should match now.

pkg/db/postgres/sqltypes/text_ulid.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/oklog/ulid/v2"
+)
+
+//
+// NOTE
+// This file exists because some of the ID columns in the tables that are
+// supposed to be storing ULID data are not in binary types but in text instead.
+// So unfortunately, we need a wrapper to handle the conversion since the ULID
+// package assume the data is in binary format already.
```
albertchae, line 14: 👍

pkg/db/sqlite/schema.sql
```diff
@@ -0,0 +1,212 @@
+CREATE TABLE goose_db_version (
+		id INTEGER PRIMARY KEY AUTOINCREMENT,
+		version_id INTEGER NOT NULL,
+		is_applied INTEGER NOT NULL,
+		tstamp TIMESTAMP DEFAULT (datetime('now'))
```
albertchae: similar question as for postgres, what's the difference between this dump and pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql ?

pkg/cqrs/base_cqrs/base_cqrs.go
albertchae: cool I was wondering why we had duplicate schema dumps

docs/plans/001-db-adapter-refactor.md
albertchae, line 1: are you going to delete this now??

albertchae (approved):
LGTM, looks like nothing is missing now and much easier to verify with the schema work

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
albertchae: sounds good, it matches the schema as is currently. we can fix the timestamp inconsistency later maybe

pkg/cqrs/base_cqrs/sqlc/sqlite/schema.sql
albertchae: awesome

docs/plans/001-db-adapter-refactor.md
darwin67: it got renamed to the 001 plan. just want the plans to be numbered.

albertchae: you think it's worth storing them though? i think it's fine to delete once we've implemented it

darwin67: oh I wasn't really planning to delete them. it's good context to leave around as well

darwin67: I might tweak and update it a little since I have a format that I'm working with now

darwin67: updated plan status and a script to check on plans
https://github.com/inngest/inngest/pull/3945/changes/695b81c63b2401eafbc2eac2dfc9d9f6f064af60

Liooo:
@darwin67 @albertchae 

just a friendly report, this revision probably breaks app when updated from pre https://github.com/inngest/inngest/pull/3032 versions, because the spans.event_ids column added there is squashed into initial migration as `CREATE TABLE IF NOT EXISTS spans (...` in this PR, which causes error on incoming requests.

darwin67:
> @darwin67 @albertchae
> 
> just a friendly report, this revision probably breaks app when updated from pre #3032 versions, because the spans.event_ids column added there is squashed into initial migration as `CREATE TABLE IF NOT EXISTS spans (...` in this PR, which causes error on incoming requests.

報告ありがとうございます。

While I don't think there's much we could've done differently in terms of technical approaches, we could've done better with the communication via changelogs.
I'm taking this back internally and try to get a proper process going so these issues are surfaced and is more obvious going forward.

PR #3984, opened 2026-04-11
EXE-1605: add phase 1 agentic api contract

jacobheric:
## Description

Adds the v2 contract for:

`GET /v2/runs/{run_id}` (delegates to oss)
`GET /v2/runs/{run_id}/trace`

I pared back a lot of what we return based on initial review comments below. See response in the monorepo and oss impl PRs: l: https://github.com/inngest/monorepo/pull/6824 and https://github.com/inngest/inngest/pull/3985. 

- [ ] Docs

## Checklist

---

> [!NOTE]
> Implements the Phase 1 agentic API endpoints (`GetFunctionRun`, `GetFunctionTrace`) in OSS, mapping internal `cqrs`/`models` types to the v2 proto contract. Includes a custom JSON marshaler to strip verbose enum prefixes, recursive trace span tree building, optional output loading, and comprehensive unit tests.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 948854e3e4457ed064e1c2ca1f2cc2639bb5354d.</sup>


pkg/api/v2/endpoints_function.go
```diff
@@ -169,6 +169,18 @@ func (s *Service) InvokeFunction(ctx context.Context, req *apiv2.InvokeFunctionR
 	return nil, s.base.NewError(http.StatusInternalServerError, apiv2base.ErrorInternalError, "There was an error invoking your function")
 }
 
+func (s *Service) InvokeFunctionBySlug(ctx context.Context, req *apiv2.InvokeFunctionBySlugRequest) (*apiv2.InvokeFunctionResponse, error) {
```
djfarrelly: How is this one different from `InvokeFunction`? would this use the full app id + fn slug? do we need that?

pkg/api/v2/util.go
```diff
@@ -9,10 +9,6 @@ import (
 )
 
 func validateInvokeRequest(ctx context.Context, req *apiv2.InvokeFunctionRequest) error {
-	if req.AppId == "" {
-		return apiv2base.NewError(http.StatusBadRequest, apiv2base.ErrorMissingField, "App ID is required")
-	}
```
djfarrelly, line 14: I think we need this one - the existing invoke requires an app id in the URL and has the fn slug match the id you use in your codebase.

proto/api/v2/service.proto
```diff
@@ -1051,6 +1083,23 @@ service V2 {
...
+      summary: "Invoke function by slug"
+      description: "Invokes a function using its public slug instead of requiring app and function IDs"
+      security: {
+        security_requirement: {
+          key: "BearerAuth"
+          value: {}
+        }
+      }
+    };
+  }
```
djfarrelly: see other comments

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  TRACE_SPAN_STATUS_COMPLETED = 2;
+  TRACE_SPAN_STATUS_FAILED = 3;
+  TRACE_SPAN_STATUS_WAITING = 4;
+}
+
+enum TraceStepOp {
+  TRACE_STEP_OP_UNSPECIFIED = 0;
+  TRACE_STEP_OP_RUN = 1;
+  TRACE_STEP_OP_SLEEP = 2;
+  TRACE_STEP_OP_WAIT_FOR_EVENT = 3;
```
djfarrelly, line 1268: Do we also have a wait for signal op?

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  TRACE_STEP_OP_INVOKE = 4;
+  TRACE_STEP_OP_SEND_EVENT = 5;
+  TRACE_STEP_OP_AI_GATEWAY = 6;
+}
+
+message FunctionRef {
+  string id = 1;
+  string slug = 2;
```
djfarrelly: The user's code always sees "id" in reference to the slug, I think we should rename this so the `id` matches the user's idea of the id (without the app id), which should be the slug, we can remove the uuid I think.

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  string name = 3;
+}
+
+message AppRef {
+  string id = 1;
+  string external_id = 2;
+  string name = 3;
```
djfarrelly: Similar comment to fn ids, which is "external id" here? I think we may only need app id from this codebase.  do we have a name to include here or is that more future thinking?

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+}
+
+message RunTrigger {
+  repeated string event_ids = 1;
+  optional string event_name = 2;
+  bool is_batch = 3;
+  optional string batch_id = 4;
+  optional google.protobuf.Timestamp batch_created_at = 5;
```
djfarrelly: Probably doesn't need to expose this to users right now

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  optional string cron_schedule = 6;
+  optional string source_id = 7;
```
djfarrelly: Let's redact this for now, maybe we can provide this type of info when we add source endpoints that they can look things up - and also we could return this data on a fetch event endpoint instead I think. Theoretically also different events from a single batch could come from different places.

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+}
+
+message FunctionRun {
+  string id = 1;
+  string trace_id = 2;
```
djfarrelly: Maybe redact for now - I want to reduce the non-essential fields from the first release because it will be hard/impossible to remove them.

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  FunctionRef function = 3;
+  AppRef app = 4;
+  FunctionRunStatus status = 5;
+  google.protobuf.Timestamp queued_at = 6;
+  optional google.protobuf.Timestamp started_at = 7;
+  optional google.protobuf.Timestamp ended_at = 8;
+  uint64 duration_ms = 9;
+  RunTrigger trigger = 10;
+  optional google.protobuf.Struct output = 11;
+  bool has_ai = 12;
```
djfarrelly: Let's remove this - this is just for the UI design in our dashboard

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+message FunctionRun {
+  string id = 1;
+  string trace_id = 2;
+  FunctionRef function = 3;
+  AppRef app = 4;
+  FunctionRunStatus status = 5;
+  google.protobuf.Timestamp queued_at = 6;
+  optional google.protobuf.Timestamp started_at = 7;
+  optional google.protobuf.Timestamp ended_at = 8;
+  uint64 duration_ms = 9;
```
djfarrelly: will this duration be an increasing number from queued ad? or will it be 0 when it's not present? I'm curious how this will work

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  int32 status_code = 1;
+  map<string, string> headers = 2;
+}
+
+message TraceSpan {
+  string span_id = 1;
+  string name = 2;
+  TraceSpanStatus status = 3;
+  optional TraceStepOp step_op = 4;
+  optional string step_id = 5;
```
djfarrelly: The end user might not know the diff between span id and step id - should we just use one for now? Or is there a benefit to having both? I'm open minded

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  uint64 duration_ms = 6;
+  google.protobuf.Timestamp queued_at = 7;
+  optional google.protobuf.Timestamp started_at = 8;
+  optional google.protobuf.Timestamp ended_at = 9;
+  map<string, string> attributes = 10;
```
djfarrelly: Is this metadata or something else?

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  optional google.protobuf.Struct input = 11;
+  optional google.protobuf.Struct output = 12;
+  optional TraceResponse response = 13;
+  bool is_root = 14;
+  bool is_userland = 15;
```
djfarrelly: I wonder if we should redact this for now, I'm not sure on the "userland" naming and how we'd use the root

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+  repeated TraceSpan children = 16;
```
djfarrelly: How deep can this go? What do you think? An alternative is make the API return a flat list with parent id and "is child" references. Not sure what the right approach would be for REST in this sense.

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+}
+
+message FunctionTrace {
+  string run_id = 1;
+  string trace_id = 2;
+  TraceSpan root_span = 3;
```
djfarrelly: Similar questions to above, curious what these might be used for from end uses in keeping things minimal. Open minded too if you have thoughts!

proto/api/v2/service.proto
```diff
@@ -1084,6 +1133,120 @@ message ResponseMetadata {
...
+}
+
+message GetFunctionTraceRequest {
+  string run_id = 1;
+  optional bool include_output = 2;
```
djfarrelly, line 1345: Do you think there are times when someone wouldn't want output? Like a fast status check or something?

djfarrelly (commented):
Leaving lots of ideas and open questions  in my mind - this is in the right direction, just want to think about the API contract that we first roll out!

pkg/api/v2/endpoints_function.go
jacobheric: I was confused by the function id naming in that one. Assumed it was an internal id. I'll remove this one.

pkg/api/v2/util.go
jacobheric: good catch, will restore.

pkg/api/v2/endpoints_function.go
jacobheric: I've removed this!

pkg/api/v2/util.go
jacobheric: restored

proto/api/v2/service.proto
jacobheric: removed

proto/api/v2/service.proto
jacobheric: added

proto/api/v2/service.proto
jacobheric: I thought about this a bit and added a max_depth with a default to 5 and a max of 10. I think this makes it `just work` for most runs without the user having to think about paging. I've also added the ability to get the rest by span id as a way of paging, see it in action (with example response data) here: https://github.com/inngest/monorepo/pull/6824

proto/api/v2/service.proto
jacobheric: Yea I've seen users doing fairly high volume status checking stuff and since outputs can be big it seemed like a good option. Let me know if you think it's overkill.

proto/api/v2/service.proto
jacobheric: done

proto/api/v2/service.proto
jacobheric: done

proto/api/v2/service.proto
jacobheric: done

proto/api/v2/service.proto
jacobheric: It's calculated only when ended at is present (start -> end) and omitted otherwise. I can remove it if you think it's overkill.

proto/api/v2/service.proto
jacobheric: removed it

proto/api/v2/service.proto
jacobheric: I remove trace_id, but kept root_span since we now return a bounded tree of children and allow them to then fetch the remainder by span. the root indicator is a nice convenience for that scenario.

proto/api/v2/service.proto
jacobheric: removed

proto/api/v2/service.proto
```diff
@@ -1178,6 +1227,124 @@ message ResponseMetadata {
...
+  string id = 1;
+}
+
+message RunTrigger {
+  repeated string event_ids = 1;
+  optional string event_name = 2;
+  bool is_batch = 3;
+  optional string batch_id = 4;
+  optional string cron_schedule = 6;
+}
```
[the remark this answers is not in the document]
jacobheric: I'll update

proto/gen/api/v2/service.pb.go
```diff
@@ -25,6 +25,183 @@ const (
...
+
+// Enum value maps for FunctionRunStatus.
+var (
+	FunctionRunStatus_name = map[int32]string{
+		0: "FUNCTION_RUN_STATUS_UNSPECIFIED",
+		1: "FUNCTION_RUN_STATUS_QUEUED",
+		2: "FUNCTION_RUN_STATUS_RUNNING",
+		3: "FUNCTION_RUN_STATUS_COMPLETED",
+		4: "FUNCTION_RUN_STATUS_FAILED",
+		5: "FUNCTION_RUN_STATUS_CANCELLED",
```
djfarrelly, line 47: question: what was the idea behind these longer strings like `FUNCTION_RUN_STATUS_COMPLETED` instead of just `COMPLETED`? You may have some thought behind it so I'm curious.

proto/gen/api/v2/service.pb.go
```diff
@@ -25,6 +25,183 @@ const (
...
+var (
+	TraceSpanStatus_name = map[int32]string{
+		0: "TRACE_SPAN_STATUS_UNSPECIFIED",
+		1: "TRACE_SPAN_STATUS_RUNNING",
+		2: "TRACE_SPAN_STATUS_COMPLETED",
+		3: "TRACE_SPAN_STATUS_FAILED",
+		4: "TRACE_SPAN_STATUS_WAITING",
+	}
+	TraceSpanStatus_value = map[string]int32{
+		"TRACE_SPAN_STATUS_UNSPECIFIED": 0,
```
djfarrelly: I'm guessing this is a catch all if it's unknown? I wonder if there is a better name for it that follows our patterns/DX 🤔

proto/gen/api/v2/service.pb.go
```diff
@@ -25,6 +25,183 @@ const (
...
+var (
+	TraceStepOp_name = map[int32]string{
+		0: "TRACE_STEP_OP_UNSPECIFIED",
+		1: "TRACE_STEP_OP_RUN",
+		2: "TRACE_STEP_OP_SLEEP",
+		3: "TRACE_STEP_OP_WAIT_FOR_EVENT",
+		4: "TRACE_STEP_OP_INVOKE",
+		5: "TRACE_STEP_OP_SEND_EVENT",
+		6: "TRACE_STEP_OP_AI_GATEWAY",
+		7: "TRACE_STEP_OP_WAIT_FOR_SIGNAL",
```
djfarrelly, line 170: Similar question on the prefixes here vs just `SEND_EVENT` for example

proto/api/v2/service.proto
```diff
@@ -1178,6 +1227,124 @@ message ResponseMetadata {
...
+  TraceSpanStatus status = 3;
+  optional TraceStepOp step_op = 4;
+  optional string step_id = 5;
+  optional uint64 duration_ms = 6;
+  google.protobuf.Timestamp queued_at = 7;
+  optional google.protobuf.Timestamp started_at = 8;
+  optional google.protobuf.Timestamp ended_at = 9;
+  optional google.protobuf.Struct input = 10;
+  optional google.protobuf.Struct output = 11;
+  optional TraceResponse response = 12;
```
djfarrelly: I can see this making sense it include for debugging purposes or redact it and make it only available in the `/trace/spans` endpoint to return it there. I'm just thinking if it results in more information than typically needed in the top level list. It also would reduce the data returned from clickhouse (I believe) in prod.

One other thing that could be useful (not required yet) in the `/trace/spans` endpoint could be step metadata attached. We can leave that out for now.

proto/gen/api/v2/service.pb.go
jacobheric: good call, I'll shorten

jacobheric: so proto enums are scoped at the package level, so that's whey the are fully qualified here. let me see how easy it is to shorten just for json.

jacobheric: I've added a small proto enum marshaller that drops the verbose namespace in favor of the plain "COMPLETE" etc value for all these.I use it in both oss and monorepo. You can see example data in the impls:
https://github.com/inngest/monorepo/pull/6824
https://github.com/inngest/inngest/pull/3985

proto/gen/api/v2/service.pb.go
jacobheric: Switched to UNKNOWN. This will never actually get set as it's narrowed upstream to a know set of statuses. But zero value is required in the proto.

proto/api/v2/service.proto
jacobheric: I removed response altogether rather than just conditionally up top just so we can keep the return shape consistent/shared between the two. Note that we're leveraging the existing trace span fetch and map (because it's so complicated) so we won't actually save on clickhouse fetching here. But, it will make it nicer for the end user. 

I added step metadata to both monorepo and oss, see example shapes in the imple PRs.

proto/gen/api/v2/service.pb.go
jacobheric: cleaned up in the json serializer.

jacobheric: see other comment, added a json serializer to remove these verbose proto prefixes.

proto/api/v2/service.proto
```diff
@@ -1205,6 +1254,126 @@ message ResponseMetadata {
...
 }
 
+enum FunctionRunStatus {
+  FUNCTION_RUN_STATUS_UNSPECIFIED = 0;
+  FUNCTION_RUN_STATUS_QUEUED = 1;
+  FUNCTION_RUN_STATUS_RUNNING = 2;
+  FUNCTION_RUN_STATUS_COMPLETED = 3;
+  FUNCTION_RUN_STATUS_FAILED = 4;
+  FUNCTION_RUN_STATUS_CANCELLED = 5;
+}
```
djfarrelly, line 1252: Do we need the prefixes on the protobuf enum? I know we're stripping them out in [`enum_json.go`](https://github.com/inngest/inngest/pull/3984/changes#diff-59da9e8d5a711a77957418388a14670f445a1b561dc7ec4b1b54fa730301b13a), but I was wondering if we should adjust the protobuf enum instead and reduce the need to strip them, or is there a benefit to having these enums be more verbose?

darwin67 (approved):

albertchae (approved):

proto/api/v2/service.proto
```diff
@@ -1205,6 +1242,112 @@ message ResponseMetadata {
...
+
+message GetFunctionTraceRequest {
+  string run_id = 1;
+  optional bool include_output = 2;
+}
+
+message GetFunctionTraceResponse {
+  FunctionTrace data = 1;
+  ResponseMetadata metadata = 2;
+}
```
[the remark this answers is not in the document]
jacobheric: a basic guard was added on the monorepo side: https://github.com/inngest/monorepo/pull/6824/changes#diff-49e15444c15023c5ec9318beb945daaa05ebac4c51309d79f12c8be6f05125daR134

djfarrelly (approved):

proto/api/v2/service.proto
jacobheric: these enum values are at api v2 package scope. so if we collapse them to "FAILED", "CANCELLED" they will collide with similar values from other enums. I know it's super verbose but I think it's the most idiomatic to name re-usable enum values in proto.

proto/api/v2/service.proto
```diff
@@ -1205,6 +1242,112 @@ message ResponseMetadata {
...
+  optional TraceStepOp step_op = 4;
+  optional string step_id = 5;
+  optional uint64 duration_ms = 6;
+  google.protobuf.Timestamp queued_at = 7;
+  optional google.protobuf.Timestamp started_at = 8;
+  optional google.protobuf.Timestamp ended_at = 9;
+  optional google.protobuf.Struct input = 10;
+  optional google.protobuf.Struct output = 11;
+  repeated TraceSpanMetadata metadata = 12;
+  repeated TraceSpan children = 13;
```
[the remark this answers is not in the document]
jacobheric: FYI, I've added arbitrary depth protection to our internal impl here: https://github.com/inngest/monorepo/pull/6824/changes#diff-49e15444c15023c5ec9318beb945daaa05ebac4c51309d79f12c8be6f05125daR22

PR #3988, opened 2026-04-12
fix(start): set Poll=true so --poll-interval actually re-syncs apps

### LINKED ISSUE #3791: `inngest start` does not set `Poll: true`, making `--poll-interval` / `INNGEST_POLL_INTERVAL` ineffective
**Describe the bug**

`inngest start` accepts a `--poll-interval` flag (and `INNGEST_POLL_INTERVAL` env var) but never sets `StartOpts.Poll = true`. This means the poll loop in `pollSDKs()` skips all healthy apps due to this guard:

```go
// pkg/devserver/service.go
if !d.Opts.Poll && len(app.Error.String) == 0 {
    continue
}
```

`inngest dev` correctly sets `Poll: !noPoll` (defaults to true), but `inngest start` omits it entirely so `Poll` stays at the Go zero value `false`.

The result is that apps which synced successfully once are never re-synced, even with `--poll-interval 60` and `--sdk-url` configured. The only apps that get re-polled are ones in an error state.

**To Reproduce**

1. Run `inngest start --signing-key <key> --event-key <key> --sdk-url https://myapp.com/api/inngest --poll-interval 60 --postgres-uri <uri> --redis-uri <uri>`
2. App syncs successfully on first boot
3. Update the app's functions (e.g. rename or add a function) and redeploy
4. Wait well beyond the poll interval
5. The Inngest server still shows the old function list

**Expected behavior**

Setting `--poll-interval` should cause the server to periodically re-sync all apps, not just errored ones. The behavior should match `inngest dev` which sets `Poll: true` by default.

**System info**

- inngest version: v1.17.1 (also confirmed on `main` at HEAD)
- OS: Linux (Docker container on AWS ECS)

**Additional context**

The fix looks straightforward. In `cmd/start/start.go`, the `StartOpts` construction should include `Poll: true` (or add a `--no-poll` flag like `inngest dev` has). Ref:

- `cmd/devserver/devserver.go` sets `Poll: !noPoll`
- `cmd/start/start.go` never sets `Poll`

Yanhu007:
## Summary

Fixes #3791

`inngest start` accepts `--poll-interval` and `INNGEST_POLL_INTERVAL` but never sets `StartOpts.Poll = true`. The `pollSDKs()` loop guards on this flag:

```go
if !d.Opts.Poll && len(app.Error.String) == 0 {
    continue
}
```

Since `Poll` stays at the Go zero value (`false`), only errored apps get re-polled. Healthy apps are never re-synced, making `--poll-interval` and `--sdk-url` ineffective.

## Fix

One-line change: set `Poll: true` in `cmd/start/start.go`'s `StartOpts` construction, matching `cmd/devserver/devserver.go`'s behavior.

| Command | Before | After |
|---------|--------|-------|
| `inngest dev` | `Poll: !noPoll` (true by default) ✅ | unchanged |
| `inngest start` | `Poll:` not set (false) ❌ | `Poll: true` ✅ |

---

> [!NOTE]
> One-line fix that sets `Poll: true` in `inngest start`'s `StartOpts`, enabling the `--poll-interval` flag to actually re-sync healthy apps. Previously, `Poll` defaulted to `false`, causing `pollSDKs()` to skip all non-errored apps.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 888c8c4ac53e6e26686336ee8c2e516bf71849e1.</sup>

### DIFF
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -129,6 +129,7 @@ func action(ctx context.Context, cmd *cli.Command) error {
 		EventKeys:               eventKeys,
 		NoUI:                    localconfig.GetBoolValue(cmd, "no-ui", false),
 		Persist:                 true,
+		Poll:                    true,
 		PollInterval:            localconfig.GetIntValue(cmd, "poll-interval", devserver.DefaultPollInterval),
 		PostgresConnMaxIdleTime: cmd.Int("postgres-conn-max-idle-time"),
 		PostgresConnMaxLifetime: cmd.Int("postgres-conn-max-lifetime"),


JJZFIVE:
+1

PR #4000, opened 2026-04-15
[e13n phase 2] emit inngest.experiment metadata from opcode opts

AndyInternet:
## Summary

When the SDK runs a `group.experiment()` variant callback, it spreads experiment context fields (`experimentName`, `variant`, `selectionStrategy`) onto the variant sub-step's `OutgoingOp.opts`. This change makes the executor read those fields from opts and emit an `inngest.experiment` step-scoped metadata span itself, rather than requiring the SDK to issue an explicit `addMetadata()` call.

## Why

The SDK-side approach (see [inngest-js#1458](https://github.com/inngest/inngest-js/pull/1458)) has two downsides:

1. **Requires an SDK upgrade** before end users can aggregate variant output metrics in a single ClickHouse row — anyone on an older SDK version silently loses experiment observability.
2. **Fragments the metadata contract across SDK languages.** Every SDK (JS, Python, Go) would need to implement the same `addMetadata()` call. By emitting from the server side we get consistent behaviour for free once any SDK sends the opts fields.

The SDK already spreads the experiment context onto `op.opts`, so the executor has everything it needs.

### New extractor — [pkg/tracing/metadata/extractors/experiment_opts.go](pkg/tracing/metadata/extractors/experiment_opts.go)

Parses flat `experimentName` / `variant` / `selectionStrategy` fields from a `GeneratorOpcode.Opts` value (accepting decoded `map[string]any`, `json.RawMessage`, or raw `[]byte`). Returns an `ExperimentMetadata` ready for `createMetadataSpan`, or `nil` when the step carries no experiment fields. Reuses the existing `ExperimentMetadata` struct (already used by the OTel span extractor), so no new types or kinds are introduced.

### Executor wiring — [pkg/execution/executor/executor.go](pkg/execution/executor/executor.go)

`handleGeneratorStep` now calls a small helper `emitExperimentMetadataFromOpts` in the same `allowStepMetadata` block that already extracts AI output metadata and timing metadata. Errors are logged and swallowed — experiment metadata emission must never break step execution.

### Checkpoint wiring — [pkg/execution/checkpoint/checkpoint.go](pkg/execution/checkpoint/checkpoint.go)

Recent SDK versions checkpoint by default ([inngest/inngest-js#1453](https://github.com/inngest/inngest-js/pull/1453) added `checkpointingMaxRuntime`), which bypasses `handleGeneratorStep`. `processMetadata` now also extracts experiment metadata from opts and feeds it through the same path as SDK-supplied metadata, so variant sub-steps get their metadata span on both code paths.

## Testing

- **Extractor unit tests** — 6 cases covering full fields, name-only (older SDK), nil/empty opts, raw JSON bytes, and empty-name handling.
- **Executor method test** — verifies the end-to-end emission path via a spy executor, including the older-SDK-missing-strategy case.
- **Checkpoint tests** — asserts that a step with experiment opts produces an extra metadata span, and that non-variant steps do not.
- **End-to-end** — built the modified binary, spun up a dev server, registered a function using `group.experiment()` via the slim companion SDK ([inngest/inngest-js#1463](https://github.com/inngest/inngest-js/pull/1463)), triggered it, and confirmed via the GraphQL trace query that the variant sub-step carries `inngest.experiment` metadata with `experiment_name`, `variant`, and `selection_strategy`. Full trace dump below.

```
  my-exp      [inngest.experiment, inngest.http.timing, inngest.timing]   // selector (SDK-side addMetadata)
  alpha-step  [inngest.experiment]                                        // variant sub-step (new executor emission)
```

## Companion change

The SDK slim follow-up to [inngest-js#1458](https://github.com/inngest/inngest-js/pull/1458) lives in [inngest/inngest-js#1463](https://github.com/inngest/inngest-js/pull/1463) — it adds `selectionStrategy` to the ALS context and removes a replay-path guard so opts reach the executor even on memoized selector steps. Neither PR requires the other to merge first: old SDKs get partial coverage (`experimentName` + `variant` only) once this executor change ships; the slim SDK PR adds the `selectionStrategy` field and closes the replay-discovery gap.

## Checklist

- [x] Added unit tests for the extractor
- [x] Added executor method test covering variant and non-variant paths
- [x] Added checkpoint tests covering variant and non-variant paths
- [x] Ran the combined stack end-to-end against the companion SDK PR

## Related

- EXE-1617
- Companion SDK PR: inngest/inngest-js#1463
- Original SDK-side attempt: inngest/inngest-js#1458

---

> [!NOTE]
> Adds server-side emission of `inngest.experiment` metadata spans from opcode opts in both the executor (`handleGeneratorStep`) and checkpoint (`processMetadata`) paths. A new `ExtractExperimentOptsMetadata` extractor normalizes opts from `map[string]any`, `[]byte`, or `json.RawMessage` into an `ExperimentMetadata` struct, reusing the existing type. The change is guarded by the existing `AllowStepMetadata` feature flag and errors are swallowed to avoid interrupting step execution.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 049b07fa545178600712186654871f6614d11bdc.</sup>

### DIFF
--- a/pkg/execution/checkpoint/checkpoint.go
+++ b/pkg/execution/checkpoint/checkpoint.go
@@ -25,6 +25,8 @@ import (
 	"github.com/inngest/inngest/pkg/logger"
 	"github.com/inngest/inngest/pkg/tracing"
 	"github.com/inngest/inngest/pkg/tracing/meta"
+	"github.com/inngest/inngest/pkg/tracing/metadata"
+	"github.com/inngest/inngest/pkg/tracing/metadata/extractors"
 )
 
 type Checkpointer interface {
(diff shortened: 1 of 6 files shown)


AndyInternet:
Replaced by #4002 after a branch rename to match ticket EXE-1617.

PR #4001, opened 2026-04-15
use a per event idempotency key instead of a large zset per workflow id

KiKoS0:
## Description
it would still be on the same shard but the idea is to avoid ZADD and
ZREMRANGEBYSCORE which are respectively O(log(N)) and O(log(N+M))
N being the number of elements in the sorted set.

- [ ] Docs

## Checklist

---

> [!NOTE]
> Replaces the per-workflow-ID sorted set (`ZADD`/`ZREMRANGEBYSCORE`, O(log N)) with per-event `SET NX EX` keys (O(1)) for batch idempotency deduplication. Includes a legacy ZSCORE fallback to handle events written to the old sorted set before rollout, with a TODO to remove it after the old keys expire.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a64e1eb47d1441d24d34f69acc1326d54839ecb4.</sup>

### DIFF
--- a/pkg/execution/batch/lua/bulk_append.lua
+++ b/pkg/execution/batch/lua/bulk_append.lua
@@ -38,15 +38,15 @@ end
 
 -- start execution
 local batchID = get_or_create_batch_key(batchPointerKey)
-local isNewBatch = (batchID == newULID)
 
 -- NOTE: these need to be identical to the ones in the queue key generator
 --   * Batch
 --   * BatchMetadata
 local keyfmt = "%s:batches:%s"
-local idempotenceKeyFmt = "%s:batch_idempotence"
+local idemKeyFmt = "%s:batch_idem:%s"
+local legacyIdempotenceKeyFmt = "%s:batch_idempotence"
+local legacyIdempotenceKey = string.format(legacyIdempotenceKeyFmt, prefix)
 local batchKey = string.format(keyfmt, prefix, batchID)
-local batchIdempotenceKey = string.format(idempotenceKeyFmt, prefix)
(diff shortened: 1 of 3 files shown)


mar-inngest (approved):

pkg/execution/batch/lua/bulk_append.lua
```diff
@@ -64,18 +64,19 @@ for i = 1, eventCount do
...
 
-  -- check if event has already been appended
-  local newEvent = redis.call("ZADD", batchIdempotenceKey, "NX", nowUnixSeconds, eventID)
-  if newEvent == 0 then
-    duplicateCount = duplicateCount + 1
-  else
+  local idemKey = string.format(idemKeyFmt, prefix, eventID)
+  local newEvent = redis.call("SET", idemKey, "1", "NX", "EX", idempotenceSetTTL)
+  -- TODO: Remove this legacy check. All old sorted sets should expire within 30 min after deploy.
+  if newEvent and redis.call("ZSCORE", legacyIdempotenceKey, eventID) then
```
mar-inngest, line 71: might want to do a lua false check here as well

https://redis.io/docs/latest/commands/set/

```sh
[RESP2 null bulk reply] and [RESP2 null multi-bulk reply-> Lua false boolean type
```

mar-inngest: Basically, some of these redis operations (I always forget), bulk string nil reply from redis becomes a false in lua.

https://redis.io/docs/latest/commands/zscore/

lkasinathan (approved):

PR #4010, opened 2026-04-16
Add `DeferAdd` and `DeferCancel` Opcodes

Linell:
## Description

Add the `DeferAdd` and `DeferCancel` opcodes discussed in [EXE-1622](https://linear.app/inngest/issue/EXE-1622/add-deferred-run-opcode) and in the linked spec. These allow us to register deferred child runs during execution, optionally cancelled by hashed ID, and converted into `inngest/deferred.schedule` events at finalize.

Fun Parts:
 - **Lazy ops** are a new shape: opcodes that piggyback on a host op in the same SDK response (e.g. `[StepRun, DeferAdd]`) instead of standing alone. The host op drives forward progress; the lazy op only records state. `OpcodeIsLazy` keeps batches like that from looking like parallel steps — `nonLazyOpCount` is the new
  gate for the parallelism / `ForceStepPlan` checks. A bare batch of only lazy ops shouldn't happen, but if it does, `runCtx.OnlyHasLazyOps()` flips and the executor enqueues a discovery step as a safety fallback.
- **Priority ops** generalize the existing `WaitForEvent` pattern via `OpcodeIsPriority`, so `[DeferAdd, RunComplete]` persists the defer before finalize fires.
  Reordering happens in both the executor and the checkpoint paths.
- **Aborted defers are sticky for the run's lifetime.** `saveDefer.lua` checks the existing `meta:<hashedID>` field inside the same invocation; if it's already `Aborted`, the script no-ops. This closes a `SaveDefer → SetDeferStatus(Aborted) → SaveDefer retry` race.
- **The 20-defer cap (`MaxDefersPerRun`) only applies to new hashedIDs.** Re-saves of an existing defer always succeed, so legitimate SDK retransmits don't trip the limit.
- **Meta and Input are split into separate hash fields** (`meta:<hashedID>` + `input:<hashedID>`) so `SetDeferStatus` never round-trips Input through cjson. Input is HSET as raw bytes and never decoded by Lua.
- **Defer emission at finalize is best-effort.** `LoadDefers` runs before `Delete` (state doesn't survive deletion), and load/build/publish errors log and continue rather than failing the run.
- **Executor and checkpoint each have their own DeferAdd/DeferCancel handlers**, but they call the same v2 state methods with the same statuses, so the resulting Redis state is identical regardless of path.
- **SDK request gets a `defers` map** with `{ "abortable": bool }` per entry — `abortable` is true only when status is `AfterRun`. The SDK uses it to skip re-emitting `DeferAdd` for already-registered defers and to know when `DeferCancel` is still possible.
- **Event IDs are deterministic per `(runID, hashedID)`** via a new `util.DeterministicULID`, with the parent run's start time as the ULID time prefix. Duplicate publishes dedupe via the runner's existing event-ID idempotency key.

---

> [!NOTE]
> Adds `OpcodeDeferAdd` and `OpcodeDeferAbort` end-to-end: new lazy/priority opcode classification, executor and checkpoint path handling, atomic Lua scripts for Redis state, priority ordering to drain defers before `RunComplete`, run finalization converting `DeferStatusAfterRun` defers into `inngest/deferred.schedule` events, and SDK request enrichment with a `defers` map carrying cancellability metadata.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 3fe8cccfc3bd2bdc47e187ee9e9d8b52a6823f77.</sup>


pkg/execution/executor/finalize.go
```diff
@@ -133,13 +154,73 @@ func (e *executor) Finalize(ctx context.Context, opts execution.FinalizeOpts) er
...
+					"deferred_run": map[string]any{
+						"companion_id": d.CompanionID,
+					},
+					"parent_run": map[string]any{
+						"fn_slug": fnSlug,
+						"run_id":  opts.Metadata.ID.RunID.String(),
+					},
+				},
+				"input": json.RawMessage(d.Input),
+			},
```
amh4r: Can we turn this into a struct with a `Validate()` method?

pkg/execution/executor/finalize.go
```diff
@@ -133,13 +154,73 @@ func (e *executor) Finalize(ctx context.Context, opts execution.FinalizeOpts) er
...
+		copy(eventID[:], idHash[:16])
+
+		events = append(events, event.Event{
+			ID:        eventID.String(),
+			Name:      "inngest/deferred.start",
+			Timestamp: now.UnixMilli(),
+			Data: map[string]any{
+				"_inngest": map[string]any{
+					"deferred_run": map[string]any{
+						"companion_id": d.CompanionID,
```
amh4r: Let's change this to `fn_slug`. It'll be the "fully-qualified function ID" (a.k.a. the function slug) for the `onDefer` Inngest function

pkg/execution/executor/finalize.go
amh4r: Actually, maybe we should take the same approach as `inngest/function.invoked`: make `event.data` the user's data and then add our special `_inngest` field:
```json
{
  "id": "01KPDXABTEYRJRQCFKSK94Q0K1",
  "name": "inngest/function.invoked",
  "data": {
    "_inngest": {
      "expire": 0,
      "fn_id": "onDefer-rowebrpr8r-fn-defer-process",
      "gid": "",
      "name": "",
      "source_app_id": "",
      "source_fn_id": "",
      "source_fn_v": 0
    },
    "my": {
      "special": "data"
    }
  },
  "ts": 1776435933006
}
```

pkg/execution/executor/executor.go
```diff
@@ -3452,6 +3458,61 @@ func (e *executor) handleGeneratorDiscoveryRequest(ctx context.Context, runCtx e
...
+	}
+
+	// continue execution, re-enqueue the discovery step
+	groupID := uuid.New().String()
+	return e.maybeEnqueueDiscoveryStep(state.WithGroupID(ctx, groupID), runCtx, gen, edge, groupID, hasPendingSteps)
+}
+
+func (e *executor) handleGeneratorDeferCancel(ctx context.Context, runCtx execution.RunContext, gen state.GeneratorOpcode, edge queue.PayloadEdge) error {
+	// Memoize with null data so the SDK doesn't re-emit this opcode on resume.
+	hasPendingSteps, err := e.smv2.SaveStep(ctx, runCtx.Metadata().ID, gen.ID, []byte("null"))
```
tonyhb: Looks like we're piggy backing off of the step state's idempotency here.

Instead of doing 2 round trips for this, can we just make SetDeferStatus idempotent?

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -507,6 +548,67 @@ func (c checkpointer) finalize(ctx context.Context, md state.Metadata, result ap
...
+		}
+	}
+
+	opts, err := op.DeferAddOpts()
+	if err != nil {
+		l.Error("error parsing DeferAdd opts in checkpoint", "error", err)
+		return fmt.Errorf("error parsing DeferAdd opts: %w", err)
+	}
+
+	if err := c.State.SaveDefer(ctx, id, state.Defer{
```
tonyhb: Do we need to run `c.State.SaveStep` before running SaveDever?  We now have two state store ops running in sequence, both handling idempotency, when `c.State.SaveDefer` could potentially do the same thing in one req?

tonyhb: i realize I made this comment in the executor code last week!

pkg/execution/executor/executor.go
amh4r: This might be a mistake. @Linell, what's the reason for saving a step in `handleGeneratorDeferCancel`?

Linell: This looks like something vestigial from before `defers` was added to the SDK request and should be changed.

pkg/execution/checkpoint/checkpoint.go
amh4r: We removed the `SaveStep()`!

amh4r (approved):

PR #4016, opened 2026-04-17
Fix GetSpanOutput missing run ID filter

amh4r:
## Description

Fix `GetSpanOutput` missing run ID filter. Before this PR, changing step output between runs would still show only step output in the new runs. For example, if step `a` in run 1 returned `"foo"` and then returned `"bar"` in run 2, then run 2 showed `"foo"` in the UI

- [ ] Docs

---

> [!NOTE]
> Fixes a bug where `GetSpanOutput` was missing a `run_id` filter, causing step output from a previous run to bleed into subsequent runs when the OTEL span ID was deterministically identical across runs. The fix threads `RunID` through `SpanIdentifier`, encodes it into the output ID, and adds it as a SQL filter in both SQLite and Postgres queries.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit b2ea02d062c1914d477a065b0d37a9fabe0e2f1a.</sup>

### DIFF
--- a/pkg/cqrs/base_cqrs/cqrs.go
+++ b/pkg/cqrs/base_cqrs/cqrs.go
@@ -428,7 +428,7 @@ fragmentLoop:
 
 	// If this span has finished, set a preliminary output ID.
 	if (outputSpanID != nil && *outputSpanID != "") || (inputSpanID != nil && *inputSpanID != "") {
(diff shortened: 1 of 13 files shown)


rhino1998 (approved):
Looks good (other than maybe needing to rerun `make gen`)

PR #4035, opened 2026-04-21
EXE-1484: Gracefully handle long userland attributes

scottnuma:
## Description

- Previously, we did not gracefully handle very long keys or values for userland span attributes

Before:
before

- We now gracefully truncate keys and values
- For keys (which tend to be short), users can copy the entire value out
- For values, users can copy the value out or adjust the width of the entire panel to see more.

After:
after

- We utilize grid to be explicit about our intent to present a grid
- We add min-w-0 to correctly resolve how the spacing and truncation resolves
- I considered adding hovering to reveal a hint with the entire value, but I think this is too low discoverability and can still present a poor viewing experience for very long strings

## Motivation

- Gracefully handle long keys and values for userland span attributes

- [ ] Docs

## Checklist

---

> [!NOTE]
> Fixes overflow of long userland span attribute keys and values by switching from fixed-width flex to CSS grid with `min-w-0 truncate`. The new commits also add `title={key}` tooltips on key cells in both V3 and V4 so users can hover to see full key text.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit ef45df5560f2f7b245bcf49f5a097ac0183d23b8.</sup>

### DIFF
--- a/ui/packages/components/src/RunDetailsV3/UserlandAttrs.tsx
+++ b/ui/packages/components/src/RunDetailsV3/UserlandAttrs.tsx
@@ -14,9 +14,9 @@ export const UserlandAttrs = ({ userlandSpan }: { userlandSpan: UserlandSpanType
   return attrs ? (
     <div className="h-full overflow-y-auto">
       <div className="mb-4 mt-2 flex max-h-full flex-col gap-2">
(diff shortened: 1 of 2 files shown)


scottnuma:
mendral-app review is from an outdated commit where keys had the `title` but values didn't. I outline the work around to retrieve full content in the description.

rhino1998 (approved):
Looks good!

Issue #3556, opened 2026-01-18
[Self-Hosted] `functions` table missing PRIMARY KEY causes duplicate entries and sync failures


ChaseWNorton:
## Summary

The `functions` table in the self-hosted PostgreSQL schema is created without a PRIMARY KEY constraint. This causes duplicate function entries to accumulate over time during sync operations, eventually leading to `mismatched param and argument count` errors from pgx when Inngest attempts to delete functions.

## Environment

- Inngest Server: v1.16.0
- Inngest SDK: v3.48.1
- Database: PostgreSQL (Azure Database for PostgreSQL)
- Deployment: Azure Container Apps

## Current Schema (from migration `000001_create_apps_table.up.sql`)

```sql
CREATE TABLE functions (
    id CHAR(36),           -- No PRIMARY KEY
    app_id CHAR(36),       -- No FOREIGN KEY  
    name VARCHAR NOT NULL,
    slug VARCHAR NOT NULL,
    config VARCHAR NOT NULL,
    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
);
```

## Expected Schema

```sql
CREATE TABLE functions (
    id CHAR(36) PRIMARY KEY,
    app_id CHAR(36) REFERENCES apps(id),
    name VARCHAR NOT NULL,
    slug VARCHAR NOT NULL,
    config VARCHAR NOT NULL,
    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    UNIQUE(app_id, slug)
);
```

## Steps to Reproduce

1. Deploy Inngest self-hosted with PostgreSQL
2. Register an SDK app with functions
3. Wait for multiple sync cycles (or restart the worker multiple times)
4. Query the database: `SELECT slug, COUNT(*) FROM functions WHERE archived_at IS NULL GROUP BY slug HAVING COUNT(*) > 1;`
5. Observe duplicate entries accumulating

## Error Messages

When duplicates accumulate, the following error appears during sync:

```
WARN error registering functions error="mismatched param and argument count"
```

This is related to GitHub Issue #3551 - the `DeleteFunctionsByIDs` query fails when operating on duplicate rows.

## Impact

- **Production outage**: Function registration fails completely
- **Cron jobs stop working**: Scheduled functions cannot execute
- **Manual intervention required**: Must truncate tables and flush Redis to recover

## Workaround

Currently the only workaround is to periodically reset the database:

```sql
[...]

Issue #3551, opened 2026-01-16
[Self-Hosted/Dev] "mismatched param and argument count" when archiving multiple functions (DeleteFunctionsByIDs)


b4ti1c:
# Bug Report: "mismatched param and argument count" when archiving multiple functions

## Summary

When deploying a new version of an app with fewer functions than before, the Inngest dev server fails with `"mismatched param and argument count"` error. Once triggered, the app is stuck in a broken state that cannot be recovered without manual database intervention.

## Environment

- **Inngest Version**: v1.16.0 (Docker: `inngest/inngest:v1.16.0`)
- **Database**: PostgreSQL (via `postgres://postgres:postgres@host.docker.internal:5432/inngest`)
- **SDK**: `inngest` npm package with `connect()` mode

## Steps to Reproduce

1. Deploy an app with multiple functions (e.g., 17 functions)
2. Remove/rename several functions from the codebase
3. Redeploy the app with fewer functions (e.g., 11 functions)
4. The sync fails with `SYNC_FAILED` and the error is persisted

### Detailed Reproduction Method

We successfully reproduced this bug with the following steps:

**Step 1: Start with a clean database**
```bash
# Reset the inngest database
PGPASSWORD=postgres psql -h localhost -U postgres -c "
  SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = 'inngest';
"
PGPASSWORD=postgres psql -h localhost -U postgres -c "DROP DATABASE IF EXISTS inngest; CREATE DATABASE inngest;"
docker restart local-infra-inngest-1
```

**Step 2: Register 5 test stub functions**

Create stub functions in your app:
```typescript
// test-stubs.ts
import {inngestClient} from '../client';

export const testStub1 = inngestClient.createFunction(
  {id: 'test-stub-1'}, {event: 'test-stub-1'}, async () => { throw new Error('Stub'); }
);
export const testStub2 = inngestClient.createFunction(
  {id: 'test-stub-2'}, {event: 'test-stub-2'}, async () => { throw new Error('Stub'); }
);
// ... repeat for testStub3, testStub4, testStub5
export const TestStubFunctions = [testStub1, testStub2, testStub3, testStub4, testStub5];
```

Add them to your function exports and run the worker to register them:
```bash
npm run start
# Verify they're registered:
PGPASSWORD=postgres psql -h localhost -U postgres -d inngest -c "SELECT slug FROM functions WHERE slug LIKE '%test-stub%';"
```

[...]

PR #4037, opened 2026-04-22
Postgres slice bug and functions primary key gap (self-hosting)

djfarrelly:
## Description

Addresses two open self-hosted bugs:

- **#3551** — `GetFunctionRunFinishesByRunIDs` fails on Postgres with `expected 1 arguments, got 3` when queried with 2+ run IDs. The sibling `DeleteFunctionsByIDs` was fixed previously via sqlc regeneration but this call site was missed. The generated query at `pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql.go` still used `WHERE run_id IN ($1)` paired with a runtime `strings.Replace` against a `/*SLICE:run_ids*/?` marker that was never emitted into the template, so the substitution was a no-op.
- **#3556** — `functions` table was created without `PRIMARY KEY (id)`. Duplicate rows accumulated over resyncs, which is how this eventually hit `DeleteFunctionsByIDs` in the wild (and why #3556 was reported alongside #3551).

### Postgres slice fix (#3551)
- Rewrote `GetFunctionRunFinishesByRunIDs` source SQL to use `ANY($1::BYTEA[])` (same shape as the `DeleteFunctionsByIDs` fix) and regenerated sqlc.
- Regression test `TestGetFunctionRunFinishesByRunIDsMultiple` proves the failure on the old query.

### Functions primary key + uniqueness (#3556)
- New goose migrations adding `PRIMARY KEY (id)` to `functions` plus a partial unique index `(app_id, slug) WHERE archived_at IS NULL`:
  - Postgres: `pkg/db/postgres/migrations/000004_functions_primary_key.sql`, `000005_functions_unique_active_slug.sql`
  - SQLite: `pkg/db/sqlite/migrations/000002_functions_primary_key.sql`, `000003_functions_unique_active_slug.sql`
- Dedup in both dialects archives older copies (we can't `DELETE` because `function_runs` / `history` reference function ids).
- Renamed `InsertFunction` → `UpsertFunction` with `ON CONFLICT (id) DO UPDATE` so resyncs don't fail against the new PK.
- `TestForeignKeyAndPrimaryKey/functions.id` proves the missing PK; additional regression tests cover slug-rename collision and `created_at` preservation through upsert.

### Dev server resync correctness
The new uniqueness invariant exposed several latent bugs in the resync flow that this PR also fixes:
- **Resolve function identity by `(app_name, slug)` before insert/update** so URL rotation (which changes the SDK's deterministic function UUID via `Function.URI`) is a no-op — the existing row is reused and the old UUID stays the canonical handle for any `function_runs` / history rows that already reference it.
- **Archive removed functions before the insert/update loop**, guarding against a slug-rename + URL-change cross-product where the predecessor would otherwise still be active when the new row's INSERT runs.
- **Look up by app *name*, not the derived app UUID.** The dev server derives app UUIDs from different inputs across paths (URL for placeholders in `service.go` / `app_mutations.go`, name post-sync in `api.go`), so a UUID-keyed lookup could miss the row even when it exists. `GetActiveFunctionByAppAndSlug` now joins on `apps.name` — the one identifier that's stable across both paths.
- **Propagate non-`sql.ErrNoRows` errors** from the slug lookup. Silently swallowing transient DB errors would leave the SDK-minted UUID in place, archive the persisted row via the early-deletes loop, and insert a fresh row — orphaning every `function_runs` / history record that referenced the old UUID.

### Tooling
- `tools/schema-dump`: strip `pg_dump` version comments from schema dumps so `TestSchemaColumnsMatchSqlc` doesn't churn on minor `pg_dump` upgrades.

## Motivation
Investigate and fix issues related to #3551 and #3556

- [ ] Docs

## Checklist

---

> [!NOTE]
> Fixes two self-hosted bugs: (1) `GetFunctionRunFinishesByRunIDs` failing on Postgres with multi-ID queries due to a broken `sqlc.slice` substitution — switched to `ANY($1::BYTEA[])` and regenerated; (2) `functions` table missing `PRIMARY KEY (id)`, allowing duplicate rows to accumulate across resyncs. Adds migrations for PK + partial unique index on `(app_id, slug) WHERE archived_at IS NULL`, renames `InsertFunction` → `UpsertFunction` with `ON CONFLICT (id) DO UPDATE`, and fixes the dev server resync flow to resolve function identity by `(app_name, slug)` so URL rotation reuses the persisted UUID. The final commit fixes a missed `df.ID` update that left the concurrency semaphore keyed by the stale SDK-minted UUID after URL rotation.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a4fd7a6db52a86f772272614a3c6f381c3b40178.</sup>


pkg/db/postgres/migrations/000004_functions_primary_key.sql
```diff
@@ -0,0 +1,16 @@
...
+-- Deduplicate any rows that accumulated before a PRIMARY KEY existed on
+-- functions.id. For each group of duplicates, keep the physically latest
+-- row (highest ctid) and delete the rest. Referencing tables (function_runs,
+-- history, etc.) store function_id as a value — they stay valid because the
+-- surviving row retains the same id.
+DELETE FROM functions a
+USING functions b
+WHERE a.id = b.id AND a.ctid < b.ctid;
+
+ALTER TABLE functions ADD PRIMARY KEY (id);
```
djfarrelly, line 12: reminder - check if this should still be the pkey - or we should add other unique indexes

pkg/api/apiv1/checkpoint.go
```diff
@@ -476,30 +476,7 @@ func (a checkpointAPI) upsertSyncData(ctx context.Context, auth apiv1auth.V1Auth
...
-			logger.StdlibLogger(ctx).Error("failed to update fn config",
-				"error", err,
-				"app_id", input.AppID(auth.WorkspaceID()),
-			)
-		}
-		return
-	}
-
-	_, err = a.FunctionCreator.InsertFunction(ctx, cqrs.InsertFunctionParams{
+	_, err = a.FunctionCreator.UpsertFunction(ctx, cqrs.UpsertFunctionParams{
```
djfarrelly, line 479: We move to an upsert always now which will allow us to remove the flow above to explicitly look for functions and update them above.

pkg/cqrs/base_cqrs/sqlc/postgres/db_normalization.go
```diff
@@ -286,6 +286,18 @@ func (q NormalizedQueries) GetFunctionBySlug(ctx context.Context, slug string) (
...
+		Name: arg.Name,
+		Slug: arg.Slug,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return function.ToSQLite()
+}
+
```
djfarrelly, line 300: We now use app name (e.g. the "app id" to the user) and the slug (function id) as we enforce these as unique to in cloud and to the user, this is the key way that we want to look them up as well instead of deterministic UUIDs which are being created differently (so therefore not very determinsitic)

pkg/cqrs/base_cqrs/sqlc/postgres/db_normalization.go
```diff
@@ -535,7 +547,11 @@ func (q NormalizedQueries) GetFunctionRunsTimebound(ctx context.Context, params
 }
 
 func (q NormalizedQueries) GetFunctionRunFinishesByRunIDs(ctx context.Context, runIDs []ulid.ULID) ([]*sqlc_sqlite.FunctionFinish, error) {
-	finishes, err := q.db.GetFunctionRunFinishesByRunIDs(ctx, runIDs)
+	byteIDs := make([][]byte, len(runIDs))
+	for i, id := range runIDs {
+		byteIDs[i] = id[:]
+	}
+	finishes, err := q.db.GetFunctionRunFinishesByRunIDs(ctx, byteIDs)
```
djfarrelly, line 554: This fixes an issue similar to #3551

pkg/cqrs/base_cqrs/sqlc/postgres/db_normalization.go
```diff
@@ -286,6 +286,18 @@ func (q NormalizedQueries) GetFunctionBySlug(ctx context.Context, slug string) (
 	return function.ToSQLite()
 }
 
+func (q NormalizedQueries) GetFunctionByAppNameAndSlug(ctx context.Context, arg sqlc_sqlite.GetFunctionByAppNameAndSlugParams) (*sqlc_sqlite.Function, error) {
```
darwin67, line 289: note to myself, need to make sure these are captured properly in https://github.com/inngest/inngest/pull/4045 when rebasing.

pkg/cqrs/base_cqrs/sqlc/sqlite/queries.sql
```diff
@@ -53,10 +53,17 @@ UPDATE apps SET error = ? WHERE id = ? RETURNING *;
...
 
 -- note - this is very basic right now.
--- name: InsertFunction :one
+-- name: UpsertFunction :one
 INSERT INTO functions
 	(id, app_id, name, slug, config, created_at) VALUES
-	(?, ?, ?, ?, ?, ?) RETURNING *;
+	(?, ?, ?, ?, ?, ?)
+ON CONFLICT (id) DO UPDATE SET
+	app_id = excluded.app_id,
```
darwin67, line 61: curious in what cases does the `app_id` will be updated?

pkg/db/postgres/migrations/000004_functions_primary_key.sql
```diff
@@ -0,0 +1,16 @@
+-- +goose Up
+
+-- Deduplicate any rows that accumulated before a PRIMARY KEY existed on
+-- functions.id. For each group of duplicates, keep the physically latest
+-- row (highest ctid) and delete the rest. Referencing tables (function_runs,
+-- history, etc.) store function_id as a value — they stay valid because the
+-- surviving row retains the same id.
+DELETE FROM functions a
+USING functions b
+WHERE a.id = b.id AND a.ctid < b.ctid;
```
darwin67, line 10: might want to put a note on release with this behavior, because it could lock up the migration and boot of the server if the table is large.

generally speaking it shouldn't be a problem unless for someone in self-hosting that's been very active.

darwin67 (approved):
have some minor non blocking comments, questions. otherwise looks good.

pkg/cqrs/base_cqrs/sqlc/sqlite/queries.sql
djfarrelly: There really shouldn't be a case where `app_id` or `slug` are changed at all, but given we're handling the conflict on `id` (new primary key), I think it's safer to always ensure the app id and slug are correctly matching the inserted data to avoid drift/a mismatch if there is a future bug in the application code. The `id` field is usually deterministically created using the `slug` itself (which embeds the app id), so it shouldn't mismatch, but since also the config contains the slug itself now, I think it's safer to always ensure it's updated.

There is a uniqueness on the app id + slug combo so this update would fail if the app id or fn slug exist in another row which would surface a bug which I think is good. 

Given a full re-think about this schema, I might change a few more fundamental things, but it's not urgent

djfarrelly:
Draining test is always failing - it passed earlier - need to fix CI for good

PR #4043, opened 2026-04-23
fix(cqrs/postgres): COALESCE finish columns in function-run LEFT JOINs (#4042)

### LINKED ISSUE #4042: GET /v1/runs/{runID} returns 500 for all in-flight runs on Postgres self-hosted
## Summary

On self-hosted Inngest using the Postgres backend, `GET /v1/runs/{runID}` returns HTTP 500 with body

```json
{"error":"Unable to load function run: <runID>","data":null,"status":500}
```

for every run that has been scheduled but has not yet finished. The run itself executes normally and the same endpoint eventually returns `Completed` once the finish row is persisted, but while the run is in-flight the REST read path is hard-broken.

Sister endpoint `GET /v1/events/{eventID}/runs` works for the same run because its backing query uses `COALESCE` on the joined finish columns.

## Environment

- Self-hosted Inngest (`inngest/inngest` docker image) with `INNGEST_POSTGRES_URI=postgres://...`
- Reproduced on `v1.17.9-ce0021b68` and still present on `main` (commit `04558d7` at time of writing)
- Independent of SDK language; reproducible with the generic `curl` calls below

## Reproduction

```bash
# 1. Fire any event that maps to a function which takes ≳ a few seconds
curl -s -X POST -H 'Content-Type: application/json' \
  "http://localhost:8288/e/$INNGEST_EVENT_KEY" \
  -d '{"name":"demo/long.running","data":{}}' | jq -r '.ids[0]'
# -> <EVENT_ID>

# 2. Grab the run_id (this works fine)
curl -s -H "Authorization: Bearer $INNGEST_SIGNING_KEY" \
  "http://localhost:8288/v1/events/<EVENT_ID>/runs" | jq '.data[0].run_id'
# -> <RUN_ID>

# 3. Query the run while it's still executing
curl -s -w '\nHTTP:%{http_code}\n' \
  -H "Authorization: Bearer $INNGEST_SIGNING_KEY" \
  "http://localhost:8288/v1/runs/<RUN_ID>"
# -> {"error":"Unable to load function run: <RUN_ID>","data":null,"status":500}
#    HTTP:500

# 4. Wait for the run to complete, then re-query
curl -s "http://localhost:8288/v1/runs/<RUN_ID>" ...
# -> HTTP:200, status:"Completed"
```

Minimal Go-level repro (added as a failing test in the linked PR):

```go
func TestFunctionRunWithoutFinish(t *testing.T) {
    // TEST_DATABASE=postgres
    // Insert a function_runs row; do NOT insert a function_finishes row.
    // Then call q.GetFunctionRun(ctx, runID).
    // Expected: err is nil.
    // Actual on main:
    //   sql: Scan error on column index 10, name "status":
    //   converting NULL to string is unsupported
}
```

## Root Cause

Three `sqlc` queries against the Postgres backend embed `function_finishes` via `LEFT JOIN` without `COALESCE`:

- `GetFunctionRun` (backs `GET /v1/runs/{runID}`)
- `GetFunctionRuns`
- `GetFunctionRunsTimebound`

Definitions in `pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql`:

```sql
-- name: GetFunctionRun :one
SELECT sqlc.embed(function_runs), sqlc.embed(function_finishes)
  FROM function_runs
  LEFT JOIN function_finishes ON function_finishes.run_id = function_runs.run_id
  WHERE function_runs.run_id = $1;
```

`function_finishes` rows are only written when a run terminates, so an in-flight run joins to all-NULL finish columns. The Postgres-side `sqlc` generated `FunctionFinish` struct uses bare non-nullable Go types:

```go
// pkg/cqrs/base_cqrs/sqlc/postgres/models.go
type FunctionFinish struct {
    RunID              ulid.ULID
    Status             string      // <- not sql.NullString
    Output             string
    CompletedStepCount int32
    CreatedAt          time.Time
}
```

so `database/sql.Rows.Scan` fails:

```
sql: Scan error on column index 10, name "status":
converting NULL to string is unsupported
```

That error surfaces in `pkg/api/apiv1/runs.go` where the handler wraps every `TraceReader.GetRun` error as HTTP 500:

```go
fr, err := a.opts.TraceReader.GetRun(...)
if err != nil {
    _ = publicerr.WriteHTTP(w, publicerr.Wrapf(err, 500, "Unable to load function run: %s", ...))
    return
}
```

`GetFunctionRunsFromEvents` (same file) already uses the correct pattern, which is why `GET /v1/events/{eventID}/runs` is unaffected:

```sql
-- name: GetFunctionRunsFromEvents :many
SELECT sqlc.embed(function_runs),
    COALESCE(function_finishes.status, '') AS finish_status,
    COALESCE(function_finishes.output, '') AS finish_output,
    COALESCE(function_finishes.completed_step_count, 0) AS finish_completed_step_count,
    COALESCE(function_finishes.created_at, function_runs.run_started_at) AS finish_created_at
FROM function_runs
LEFT JOIN function_finishes ...;
```

## Scope

- **Postgres backend only.** The SQLite sqlc config generates `FunctionFinish.Status` as `sql.NullString` (and the other columns as `sql.NullInt64` / `sql.NullTime`), so scanning a NULL works fine and SQLite users never hit this.
- Not a timing / indexing / caching issue — deterministic for every in-flight run.
- Handler-side: `GET /v1/runs/{runID}` has no `ErrNoRows → 404` branch, so any error from the read becomes a 500. Fixing the SQL side is sufficient; no handler change needed.

## Proposed Fix

Apply the `GetFunctionRunsFromEvents` pattern to the three broken queries. PR coming in a minute.

ivanfuland:
## Summary

Fixes #4042.

Three `sqlc` queries against the Postgres backend (`GetFunctionRun`, `GetFunctionRuns`, `GetFunctionRunsTimebound`) embed `function_finishes` via `LEFT JOIN` without `COALESCE`. The Postgres-side generated `FunctionFinish` struct uses bare non-nullable Go types, so `sql.Rows.Scan` fails on any run that hasn't written a `function_finishes` row yet:

```
sql: Scan error on column index 10, name "status": converting NULL to string is unsupported
```

The error bubbles up through `TraceReader.GetRun` into `pkg/api/apiv1/runs.go::GetFunctionRun`, which wraps every error as HTTP 500 — so `GET /v1/runs/{runID}` returns 500 for every in-flight run on a Postgres-backed self-hosted deploy.

This PR mirrors the existing pattern from `GetFunctionRunsFromEvents` (which already uses `COALESCE` and is unaffected):

```sql
COALESCE(function_finishes.status, '')                                  AS finish_status,
COALESCE(function_finishes.output, '')                                  AS finish_output,
COALESCE(function_finishes.completed_step_count, 0)                     AS finish_completed_step_count,
COALESCE(function_finishes.created_at, function_runs.run_started_at)    AS finish_created_at
```

## Changes

- `pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql` — add `COALESCE(...) AS finish_*` to the three broken queries.
- `pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql.go` — regenerated via `sqlc generate` (sqlc v1.30.0, matches the version recorded in the existing generated files).
- `pkg/cqrs/base_cqrs/sqlc/postgres/normalization.go` — update `GetFunctionRunRow.ToSQLite` / `GetFunctionRunsRow.ToSQLite` / `GetFunctionRunsTimeboundRow.ToSQLite` to rebuild a `FunctionFinish` from the new flat `FinishStatus` / `FinishOutput` / `FinishCompletedStepCount` / `FinishCreatedAt` fields, matching the existing handling of `GetFunctionRunsFromEventsRow.ToSQLite`.
- `pkg/db/postgres/querier.go` — same shape change at the `db.FunctionRunRow` layer for `GetFunctionRun` / `GetFunctionRuns` / `GetFunctionRunsTimebound`.
- `pkg/db/regression_test.go` — adds `TestFunctionRunWithoutFinish` that inserts a `function_runs` row with no matching `function_finishes` and asserts the three queries return cleanly with `Status.Valid=false` / `Output.Valid=false`.

## Scope notes

- **Postgres only.** SQLite's generated `FunctionFinish` already uses `sql.NullString` / `sql.NullInt64` / `sql.NullTime`, so NULL scans cleanly and no SQL change is needed.
- **No handler change.** `pkg/api/apiv1/runs.go::GetFunctionRun` conflates "read error" and "not found" into HTTP 500, but once the underlying query stops erroring, the handler naturally returns a 200 with the in-flight status. A separate change could split 500 vs. 404, but that's out of scope here.

## Test plan

- [x] `go test ./pkg/db/...` under SQLite (default) — `TestFunctionRunWithoutFinish` passes.
- [x] `TEST_DATABASE=postgres go test ./pkg/db/...` — `TestFunctionRunWithoutFinish` passes after the fix; fails on `main` with the exact scan error quoted above.
- [x] `go test ./pkg/cqrs/base_cqrs/...` — green.
- [x] `go vet ./pkg/cqrs/base_cqrs/... ./pkg/db/postgres/...` — clean.
- [x] `sqlc generate` on the patched `queries.sql` produces the committed `queries.sql.go` with zero extra diff.

---

> [!NOTE]
> Fixes a production bug where `GET /v1/runs/{runID}` returns HTTP 500 for every in-flight run on Postgres-backed deployments. Three sqlc queries (`GetFunctionRun`, `GetFunctionRuns`, `GetFunctionRunsTimebound`) used `LEFT JOIN function_finishes` without `COALESCE`, causing `sql.Rows.Scan` to fail on NULL columns when no finish row exists. The fix mirrors the identical pattern already used by `GetFunctionRunsFromEvents`, updates the Go scan targets and normalization layer accordingly, and adds a regression test.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit b5bab88d6d78ba72807f5fee610b9bec19f73a59.</sup>

### DIFF
--- a/pkg/cqrs/base_cqrs/sqlc/postgres/normalization.go
+++ b/pkg/cqrs/base_cqrs/sqlc/postgres/normalization.go
@@ -243,7 +243,14 @@ func (r *GetFunctionRunRow) ToSQLite() (*sqlc.GetFunctionRunRow, error) {
 		return nil, err
 	}
 
-	finish, err := r.FunctionFinish.ToSQLite()
+	pgFinish := FunctionFinish{
+		RunID:              r.FunctionRun.RunID,
+		Status:             r.FinishStatus,
+		Output:             r.FinishOutput,
+		CompletedStepCount: r.FinishCompletedStepCount,
+		CreatedAt:          r.FinishCreatedAt,
+	}
+	finish, err := pgFinish.ToSQLite()
 	if err != nil {
 		return nil, err
 	}
@@ -260,7 +267,14 @@ func (r *GetFunctionRunsTimeboundRow) ToSQLite() (*sqlc.GetFunctionRunsTimebound
 		return nil, err
 	}
 
-	finish, err := r.FunctionFinish.ToSQLite()
+	pgFinish := FunctionFinish{
+		RunID:              r.FunctionRun.RunID,
+		Status:             r.FinishStatus,
+		Output:             r.FinishOutput,
+		CompletedStepCount: r.FinishCompletedStepCount,
+		CreatedAt:          r.FinishCreatedAt,
+	}
+	finish, err := pgFinish.ToSQLite()
 	if err != nil {
 		return nil, err
 	}
@@ -277,7 +291,14 @@ func (r *GetFunctionRunsRow) ToSQLite() (*sqlc.GetFunctionRunsRow, error) {
 		return nil, err
 	}
 
-	finish, err := r.FunctionFinish.ToSQLite()
+	pgFinish := FunctionFinish{
+		RunID:              r.FunctionRun.RunID,
+		Status:             r.FinishStatus,
+		Output:             r.FinishOutput,
+		CompletedStepCount: r.FinishCompletedStepCount,
+		CreatedAt:          r.FinishCreatedAt,
+	}
+	finish, err := pgFinish.ToSQLite()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
+++ b/pkg/cqrs/base_cqrs/sqlc/postgres/queries.sql
@@ -102,17 +102,31 @@ INSERT INTO function_finishes
     ($1, $2, $3, $4, $5);
 
 -- name: GetFunctionRun :one
-SELECT sqlc.embed(function_runs), sqlc.embed(function_finishes)
+SELECT sqlc.embed(function_runs),
+    COALESCE(function_finishes.status, '') AS finish_status,
+    COALESCE(function_finishes.output, '') AS finish_output,
+    COALESCE(function_finishes.completed_step_count, 0) AS finish_completed_step_count,
+    COALESCE(function_finishes.created_at, function_runs.run_started_at) AS finish_created_at
   FROM function_runs
   LEFT JOIN function_finishes ON function_finishes.run_id = function_runs.run_id
   WHERE function_runs.run_id = $1;
 
 -- name: GetFunctionRuns :many
-SELECT sqlc.embed(function_runs), sqlc.embed(function_finishes) FROM function_runs
+SELECT sqlc.embed(function_runs),
+    COALESCE(function_finishes.status, '') AS finish_status,
+    COALESCE(function_finishes.output, '') AS finish_output,
+    COALESCE(function_finishes.completed_step_count, 0) AS finish_completed_step_count,
+    COALESCE(function_finishes.created_at, function_runs.run_started_at) AS finish_created_at
+FROM function_runs
 LEFT JOIN function_finishes ON function_finishes.run_id = function_runs.run_id;
 
 -- name: GetFunctionRunsTimebound :many
-SELECT sqlc.embed(function_runs), sqlc.embed(function_finishes) FROM function_runs
+SELECT sqlc.embed(function_runs),
+    COALESCE(function_finishes.status, '') AS finish_status,
+    COALESCE(function_finishes.output, '') AS finish_output,
+    COALESCE(function_finishes.completed_step_count, 0) AS finish_completed_step_count,
+    COALESCE(function_finishes.created_at, function_runs.run_started_at) AS finish_created_at
+FROM function_runs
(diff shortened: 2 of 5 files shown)


darwin67 (approved):
Looks good.
Thanks for addressing this!

I was planning to audit some of these after doing some refactor of the DB adapters.
Saved me one task.

PR #4067, opened 2026-04-28
add schedule type & original run id to replayed runs

KiKoS0:
## Description
V1 traces added rerun spans that pointed out that a run is replayed along with its original run ID but V2 doesn't so i guess this is more of regression than a new thing. Adds schedule_type which is either "event", "cron", "rerun" and original_run_id to the dynamic attributes of a root v2 span.

- [ ] Docs

## Checklist

---

> [!NOTE]
> Restores V1 tracing behavior for replayed runs in V2 by adding two new span attributes: `run.replay_original_run_id` and `run.schedule_type`. The follow-up commit converts `ScheduleType` from a raw string to a proper `enumer`-generated enum with JSON/GQL/text marshaling support.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a195975014f695f86d63f65d3f83220223e99741.</sup>


pkg/execution/execution.go
```diff
@@ -275,6 +275,8 @@ type ScheduleRequest struct {
 	DebugSessionID *ulid.ULID
 	// DebugRunID is the ID of the debugger run that this function is being scheduled from.
 	DebugRunID *ulid.ULID
+	// ScheduleType describes how this run was triggered, e.g. "event", "cron-queue", "rerun".
+	ScheduleType string
```
albertchae: is there an enum we can use instead? although I think I never added it after adding the instrumentation for scheduling either :(

albertchae (approved):

darwin67 (approved):

PR #4082, opened 2026-04-30
SYS-751: Create a ShardRegisty interface that owns all known shards and shard selector logic

lkasinathan:
## Description

Create a `ShardRegistry` interface that owns all known shards. It collapses the `(shards map, ShardSelector, primary QueueShard)` trio that was threaded independently into `queue.New`, the executor, the singleton store, and api surfaces into a single `ShardRegistry` value.

All callers now use `QueueShardRegistry` instead of individual components

No behavior change — pure plumbing. Test files updated to construct via `NewShardRegistry` / `NewSingleShardRegistry`.

## Motivation

Today, multiple components -  the queue processor, executor, singleton store, api servers are each initialized with their own copy of the shards map plus a `ShardSelector` and primary shard, and operate against that snapshot for their lifetime. 

This is fine while the topology is static, but it doesn't hold up once shards can come up at runtime: a new shard added after boot would be invisible to every component that already captured its own copy. 

- [ ] Docs

## Checklist

---

> [!NOTE]
> Introduces a `ShardRegistry` interface hierarchy (`ShardRegistry` → `QueueShardRegistry` → `ShardRegistryController`) that consolidates the previously scattered `(shards map, ShardSelector, primaryQueueShard)` trio into a single value. All call sites — queue processor, executor, singleton store, debug API, test API, and API v1 — are updated to accept the narrowest interface they need. `NewSingleShardRegistry` covers the single-shard case (devserver, tests); `NewShardRegistry` with `WithShardSelector`/`WithPrimary` options covers multi-shard topologies. The shard lease loop calls `SetPrimary` on the registry after claiming a lease.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 8f09db6f20a1f17b46e0cff33a5f3be8ccd3a20b.</sup>


darwin67 (approved):
👏

PR #4103, opened 2026-05-01
fix(devserver): wire SQL history reader, drop in-memory mirror

### LINKED ISSUE #3892: Memory leak in self-hosted Inngest - RSS grows indefinitely, never freed
**Describe the bug**

Self-hosted Inngest leaks memory continuously. RSS grows linearly over time and is never reclaimed by the Go runtime. The only way to free memory is to restart the container.

- After fresh restart: ~500 MiB RSS
- After 12 hours: ~3 GiB RSS
- After 2-3 days: 10+ GiB RSS → system freeze

Heap profile shows ~1 GiB inuse on the Go heap, but RSS is 10.6 GiB — the Go runtime retains ~9.5 GiB that is no longer actively used. `GODEBUG=madvdontneed=1` does not help. Setting `GOMEMLIMIT=8GiB` causes the GC to spin at 700% CPU without reducing RSS.

The tracing/telemetry subsystem appears to be a significant contributor based on heap profiling:
- `pkg/tracing/tracer_sqlc.go:220` — `dbExporter.ExportSpans`
- `pkg/tracing/execution_processor.go:213` — `executionProcessor.OnEnd`
- `pkg/tracing/tracer.go:334` — `otelTracerProvider.UpdateSpan`
- `pkg/execution/queue/process_partition.go` — `ProcessPartition`
- `pkg/telemetry/metrics/histogram.go:90` — `HistogramProcessPartitionDuration`

**To Reproduce**

1. Deploy self-hosted Inngest with PostgreSQL 17 + Redis 7 (external, not embedded)
2. Register functions with concurrency (100 global, 4 per key) and throttle (50/s)
3. Run sustained workload: bulk event processing sending ~150 events per batch, ~1000+ function runs per day
4. Monitor container RSS over 24-48 hours
5. Observe memory grows continuously and never decreases

**Expected behavior**

Memory should stabilize or decrease after function runs complete. The Go GC should be able to reclaim unused memory and return it to the OS, especially with `GODEBUG=madvdontneed=1`.

Heap profile summary after ~2 days uptime:
heap profile: 17619: 1045123504 [10738482: 70175133832] @ heap/1048576

- Inuse: 17,619 objects / ~1 GiB
- Total allocated: 10,738,482 objects / ~70 GiB cumulative
- RSS: 10.6 GiB

Process stats:
CPU:      680% (GC spinning)
MEM:      10.13 GiB / 12 GiB limit (84%)
Goroutines: 503
VmRSS:   10,607,584 kB
VmPeak:  12,254,428 kB
VmData:  10,996,112 kB

Postgres table sizes after 2 days (traces accumulate fast):
spans:   5.2 GB
history: 3.1 GB
traces:  2.4 GB

Binary pprof heap and CPU profiles attached (no application data, safe to share).

**System info**
- inngest-cli version: 1.17.5-db8cff1a6
- Go version (in image): 1.24.13
- Image: `inngest/inngest:latest` (sha256:e1743e1d87b7678907835dce108a44d108b6eb7bc960f225d9edba4bb57ed281)
- OS: Ubuntu Linux (Docker)
- VM: 16 GB RAM, 8 vCPU

**Additional context**

- Workaround: daily container restart via cron
- `GOMEMLIMIT` is counterproductive — at 8 GiB the GC enters an infinite loop consuming 700% CPU because RSS cannot be brought below the limit
- `GODEBUG=madvdontneed=1` has no measurable effect
- Redis is clean (7-42 MiB, keys expire properly with TTL)
- The leak is in the Inngest Go process itself, not in Redis or Postgres

PS: I could provide the prof files if needed

CSteigstra:
Closes #3892 (or at least the dominant retention path of it).

## Description

The in-memory history mirror in `pkg/history_drivers/memory_store` grows unbounded under sustained load. `Singleton.Data` is never deleted: every run accumulates a `[]history.History` slice with full `Result.Output` strings retained for the entire process lifetime.

The mirror's only consumer is the GraphQL dashboard via `memory_reader.NewReader()`. The same data is already being durably written to Postgres/SQLite by the SQL history driver (`hd`, devserver.go:209), and `base_cqrs.NewHistoryReader` already implements the same `history_reader.Reader` interface — it just wasn't wired up.

This PR wires the GraphQL resolvers to the SQL reader and stops registering the in-memory writer.

## Why this is the right fix

Method-by-method, the SQL reader is at parity or strictly better than the memory reader:

| Method | `memory_reader` | `base_cqrs.reader` |
|---|---|---|
| `CountRuns` | `len(store.Data)` (only the mirror's view) | `q.HistoryCountRuns` (durable count) |
| `GetRun` / `GetFunctionRun` | map lookup | `q.GetFunctionRun` |
| `GetFunctionRunsFromEvents` | linear scan | indexed query |
| `GetRunHistory` | slice from `RunData.History` | `q.HistoryGetRunHistory` over `function_run_history` |
| `GetRunHistoryItemOutput` | linear scan | per-item query |
| `GetRuns` | **ignores `opts`, returns ALL runs** | respects filters |
| `GetRunsByEventID` | linear scan | indexed query |
| `GetSkippedRunsByEventID`, `GetUsage`, `GetReplayRuns`, `CountReplayRuns`, `GetActiveRunIDs`, `CountActiveRuns` | `not implemented` | `not implemented` (parity) |

The behavioural difference users will see: the dashboard's `GetRuns` now respects its `opts` filters instead of returning every cached run unfiltered. That's a correctness improvement, not a regression.

After this change `pkg/history_drivers/memory_{store,reader,writer}` have no remaining consumers in the codebase. They can be deleted in a follow-up PR (kept here for reviewability).

## Why now

Issue #3892 has been open since March with five+ users reporting the same shape, and an unofficial fork (`killfill/inngest`) shipping a FIFO cap as a workaround. Capping is a partial fix — it bounds count but not per-entry size, evicts in-flight runs by insertion order, and is only available as an unofficial Docker image. Removing the redundancy is the proper fix because the SQL store is already the source of truth for durable history; the mirror is just a stale, ever-growing copy.

## Test plan

- Existing `go.yaml` lint and unit tests should pass — no interface changes, no new code.
- Existing dashboard e2e (`e2e.yml`) should pass — same data, same shape.
- Manual: `inngest start` + Postgres, drive load, confirm dashboard `/runs` lists, drill-in, history, and step output panels still work.
- Manual: `inngest dev` with default SQLite, same checks.
- Soak: 24h workload comparable to #3892's repro, watch RSS — should plateau instead of growing linearly.

## Alternative if you'd rather keep the mirror

Happy to rework as either of:

1. Gate the in-memory wiring on `opts.NoUI` (or a new `--in-memory-history` flag), default to SQL — preserves the option for users who want the in-memory speedup.
2. Keep the mirror, FIFO-cap it (similar to https://github.com/killfill/inngest/commit/a0ebc222ab64952acdb73bcd842f45a47d7a1323) — partial fix but smaller behavioural change.

Drop the mirror feels like the architecturally correct choice given it's a redundant cache of data that's already durably stored, but I defer to maintainer preference.

## Results on our machine with high traffic.

So far so good, no climbing memory usage, and UI works fine. Around 10:00 we switched version.

image

## Type of change

---

> [!NOTE]
> Removes the unbounded in-memory history mirror (`memory_writer`/`memory_reader`) from the dev server and wires the existing SQL-backed `base_cqrs.NewHistoryReader` and `hd` (`history.Driver`) in their place. The change is a 2-line addition / 7-line deletion with no new code paths.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 302afb96d39a9c1e48720655e14047855f7f650e.</sup>

### DIFF
--- a/pkg/devserver/devserver.go
+++ b/pkg/devserver/devserver.go
@@ -65,8 +65,6 @@ import (
 	sv2 "github.com/inngest/inngest/pkg/execution/state/v2"
 	"github.com/inngest/inngest/pkg/expressions"
 	"github.com/inngest/inngest/pkg/expressions/expragg"
-	"github.com/inngest/inngest/pkg/history_drivers/memory_reader"
-	"github.com/inngest/inngest/pkg/history_drivers/memory_writer"
 	"github.com/inngest/inngest/pkg/logger"
 	"github.com/inngest/inngest/pkg/metrics"
 	"github.com/inngest/inngest/pkg/pubsub"
@@ -484,8 +482,6 @@ func start(ctx context.Context, opts StartOpts) error {
 		return fmt.Errorf("failed to create publisher: %w", err)
 	}
 
-	hmw := memory_writer.NewWriter(ctx, memory_writer.WriterOptions{DumpToFile: false})
-
 	tp := tracing.NewSqlcTracerProvider(adapter.Q())
 
 	url := opts.Config.CoreAPI.Addr
@@ -517,7 +513,6 @@ func start(ctx context.Context, opts StartOpts) error {
 				history.NewLifecycleListener(
 					nil,
 					hd,
-					hmw,
 				),
 				Lifecycle{
 					Cqrs:       dbcqrs,
@@ -608,7 +603,7 @@ func start(ctx context.Context, opts StartOpts) error {
 	)
 
 	// The devserver embeds the event API.
-	ds := NewService(opts, runner, dbcqrs, pb, stepLimitOverrides, stateSizeLimitOverrides, unshardedRc, hmw, nil)
+	ds := NewService(opts, runner, dbcqrs, pb, stepLimitOverrides, stateSizeLimitOverrides, unshardedRc, hd, nil)
 	ds.State = sm
 	ds.Queue = rq
 	ds.Executor = exec
@@ -631,7 +626,7 @@ func start(ctx context.Context, opts StartOpts) error {
 		Queue:          ds.Queue,
 		EventHandler:   ds.HandleEvent,
 		Executor:       ds.Executor,
-		HistoryReader:  memory_reader.NewReader(),
+		HistoryReader:  base_cqrs.NewHistoryReader(adapter),
 		DisableGraphQL: &opts.NoUI,
 		ConnectOpts: connectv0.Opts{
 			GroupManager:               connectionManager,


Pckool:
Is there someone who can get this merged?? We're running into this issue @tonyhb @djfarrelly?

CSteigstra:
Closing — this was fixed upstream by #4440, which removed `pkg/history_drivers/{memory_reader,memory_store,memory_writer}` entirely and left the SQL-backed reader as the only path. Shipped in **v1.29.0** (2026-06-17).

Flagging for visibility since it wasn't obvious: that landed as `chore(tests): flaky test fix`, so it appears in the changelog as `*(tests)* Flaky test fix (#4440)`, and #3892 is still open. @Pckool that's likely why you couldn't find it — upgrading to >=v1.29.0 should resolve it without this patch.

One caveat: this only covers the in-memory history mirror. The other unbounded-growth source in #3892 is unchanged as of v1.41.1 — `getTracer` still constructs a `SimpleSpanProcessor`/`TracerProvider` per call (`pkg/tracing/tracer.go:98`), and there's no span GC in the product; the only `DELETE FROM spans` in the tree is in `docs/POSTGRES_RETENTION.md`. So #3892 arguably shouldn't be closed on the strength of #4440 alone.

Pckool:
Thanks for the update! Appreciate it!

PR #4105, opened 2026-05-01
Revert "Fix constraint cache silently dropping retries on exhaustion"

djfarrelly:
Reverts inngest/inngest#4084

The goal of this is to test/confirm if the original PR breaks our tests

---

> [!NOTE]
> Reverts PR #4084 ("Fix constraint cache silently dropping retries on exhaustion"), removing the `addedAt`/`RequestTime` cache-bypass logic, the `RequestTime` field from `CapacityAcquireRequest`, and all associated tests. The dev server is also reverted to use the unwrapped `cm` instead of `cachedCM`, and the constraint cache wrapper is removed from the dev server startup path entirely.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 426a347c67a30205ca2a642dc4a2eb5a3d76d0ab.</sup>


BrunoScheufler:
Might be the part of adding it to the dev server, plus maybe some of the tests aren’t waiting. We can try and see if rolling back the dev server change works.

djfarrelly:
Superseded by #4106

PR #4108, opened 2026-05-01
Handle HTTP Timeouts w/Checkpointing

Linell:
## Description

When the executor times out an HTTP dispatch and requeues the queue item, the original SDK process keeps running and POSTs async checkpoints for steps the new dispatch will execute again, producing duplicate step runs.

Mint a DispatchID on Lease, preserve it across ExtendLease, and rotate it on Requeue. The driver echoes it to the SDK in SDKRequestContext; the SDK echoes it back on async checkpoint POSTs. The checkpoint API loads the queue item, compares DispatchIDs, and returns 409 Conflict when they don't match so the stale SDK process aborts before the next step runs.

The connect driver also switches from sending a raw job ID to the base64-encoded shardID::jobID queueref, matching the http driver — the encoded form is required so the server can resolve the shard for the queue item lookup.

## Motivation

[EXE-1552: HTTP timeout leads to duplicate steps with checkpointing](https://linear.app/inngest/issue/EXE-1552/http-timeout-leads-to-duplicate-steps-with-checkpointing)

## Problem

The problem is that something times out the executor -> SDK request. The executor retries, the original SDK process keeps running steps and POSTing to the async-checkpoint API. Two SDK processes execute the same step bodies in parallel. Redis dedups the result writes, but user side effects like sending emails, have already fired multiple times.

### Solution

The way I've gone about fixing this is to mint a per-dispatch ID on every `Lease`, send it to the SDK, and then have the SDK echo it back on every checkpoint POST< and reject (409 CONFLICT) any POST whose ID doesn't match the queue item's current dispatch ID.

- `Lease` mints a fresh `DispatchID` ulid onto the queue item.
- `ExtendLease` leaves it alone because one dispatch can span many extensions.
- `Requeue` rotates it because a requeue is a new dispatch

`DispatchID` is a fence around one SDK invocation. The queue item holds the canonical fence value. The SDK only knows the fence it was handed at dispatch time. If the queue rotates the fence, because we requeued, the old SDK's writes get rejected at the door and the SDK uses that rejection as its signal to die quietly instead of blundering through the rest of the function.

#### Why Not QueueID

We can't use the existing `QueueItem.ID` because it identifies the _job_ and what we need is to identify a _specific delivery of that job_. the ID is stable across requeues, when we actually need to be able to tell that something has changed. The idea for the `DispatchID` is that it's minted on `Lease`, untouched by `ExtendLease`, and rotated on `Requeue`.

- `LeaseID` is not suitable because it rotates on every `ExtendLease`, which happens mid-dispatch as a normal operation adn would falsely reject legit checkpoints from long running steps.
- `Data.Attempt` doesn't bump on `AlwaysRetryable` errors or scavenger requeues, which is the path of this error actually. HTTP timeout -> executor stops extending -> lease expires -> scavenger requeues -> new dispatch leases, while the original SDK is still alive and about to POST a checkpoint.

- [ ] Docs

## Checklist

## Related
https://github.com/inngest/inngest-js/pull/1511
https://github.com/inngest/inngestgo/pull/213

---

> [!NOTE]
> Refactors the stale-dispatch fencing mechanism from a random ULID (`DispatchID`) to a monotonic `GenerationID` counter. The counter starts at 1 on enqueue, is preserved across Lease/ExtendLease, and is incremented on Requeue. The driver passes it to the SDK via `SDKRequestContext`; the SDK echoes it back on async checkpoint POSTs; the checkpoint API rejects mismatches with 409 Conflict to prevent duplicate step execution after HTTP timeouts. The connect driver is also fixed to send the base64-encoded `shardID::jobID` queueref instead of a raw job ID.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 1a836f6952b30fda968437c186b6a6fc20f666f2.</sup>


pkg/execution/checkpoint/checkpoint.go
```diff
@@ -464,6 +474,38 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
...
+	}
+
+	item, err := loader.LoadQueueItem(ctx, ref.ShardID(), ref.JobID())
+	if errors.Is(err, queue.ErrQueueItemNotFound) {
+		return fmt.Errorf("%w: queue item not found", ErrStaleDispatch)
+	}
+	if err != nil {
+		return err
+	}
+	if item.DispatchID == nil || item.DispatchID.String() != input.DispatchID {
```
BrunoScheufler: We will replace this with a "generation", a monotonically-increasing attempt counter on the queue item.

This is required as we don't always increase the attempt count on Requeue operations, in case of transient system errors.

We can then use this "generation" identifier to understand whether the executor has moved on to a new request or is still processing the original request.

TO DO:
- Add generation field to queue items, update on every Requeue, wait until deployed
- Pass through generation to httpdriver/connectdriver to use during Checkpointing in the SDK
- Check if generation has changed for every Checkpoint request
  - Look up current item from queue using ref.ShardID() and ref.JobID()

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -388,6 +394,10 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 		return fmt.Errorf("cannot checkpoint async steps")
 	}
 
+	if err := c.validateAsyncDispatch(ctx, input); err != nil {
```
BrunoScheufler, line 405: Does this introduce problematic latency? We seem to run this for every checkpoint request

Linell:
Closing in favor of https://github.com/inngest/inngest/pull/4139 and https://github.com/inngest/inngest/pull/4140

PR #4125, opened 2026-05-05
Remove per-Acquire constraintUsage from Lua

dlambrig:
## Summary

This removes the per-Acquire `constraintUsage` path from the `inngest` constraint API implementation:

- removes the `Usage` field from the Acquire response path
- deletes the Lua `cu` collection and the related request-state plumbing
- removes the Go-side conversion and lifecycle hooks for that data
- updates the related tests and snapshot

## Why

The per-Acquire usage tracking is no longer needed now that the account concurrency signal comes from the poller path.

## Testing

- `go test ./pkg/constraintapi -run ^`

---

> [!NOTE]
> Removes the per-Acquire `constraintUsage` (`cu`) tracking from the Lua script, Go response structs, proto conversion, lifecycle hooks, and all associated tests. The feature is no longer needed since account concurrency signals now come from the poller path.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 7985dca6f2e5f99a30b5b995114030a88c78dc85.</sup>

### DIFF
--- a/pkg/constraintapi/acquire.go
+++ b/pkg/constraintapi/acquire.go
@@ -166,36 +166,6 @@ type acquireScriptResponse struct {
 	RetryAt              int                       `json:"ra"`
 	Debug                flexibleStringArray       `json:"d"`
 	CacheHit             int                       `json:"ch"`
-	ConstraintUsage      []constraintUsageResponse `json:"cu"`
-}
-
-type constraintUsageResponse struct {
-	Limit int `json:"l"`
-	Usage int `json:"u"`
-}
-
-func buildConstraintUsage(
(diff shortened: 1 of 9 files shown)


darwin67 (changes requested):
don't make the vendored code changes in this repo, and do it in inngestgo instead.

then vendor it in.

dlambrig:
> don't make the vendored code changes in this repo, and do it in inngestgo instead.
> 
> then vendor it in.

Branches were messed up. I’ve split that out now. This PR is just constraintUsage cleanup now. The vendored handler/connect fixes will be done separately (upstreamed in inngestgo first, then re-vendored.)

PR #4130, opened 2026-05-06
fix: gracefully handle checkpoint dupes

Linell:
## Description

Fix a bug where when the SDK reports a step that was already saved by a checkpoint, `handleGeneratorStep` was returning early on the dupe and skipping the discovery enqueue, which stranded the run. This lets it fall through so that the next step gets scheduled.

## Motivation

[EXE-1625: Hanging runs with checkpointing](https://linear.app/inngest/issue/EXE-1625/hanging-runs-with-checkpointing)

- [ ] Docs

## Checklist

---

> [!NOTE]
> Fixes a run-stranding bug where `handleGeneratorStep` returned `nil` early on `ErrDuplicateResponse`/`ErrIdempotentResponse` (step already persisted via the checkpoint path), skipping `maybeEnqueueDiscoveryStep`. The fix clears the error and falls through to the discovery enqueue, ensuring the run advances. Two targeted tests cover the broken path and the duplicate-metric suppression behavior in the checkpoint sync path.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 94f872b471a0bfd8a0bec7dac8dba96b62b4940b.</sup>

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -3717,10 +3717,12 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
 	}
 
 	hasPendingSteps, err := e.smv2.SaveStep(ctx, runCtx.Metadata().ID, gen.ID, []byte(output))
+	// A duplicate or idempotent save means the step was already persisted
+	// (typically via the checkpoint path). Keep the original output and
+	// continue so the run still advances.
 	if errors.Is(err, state.ErrDuplicateResponse) || errors.Is(err, state.ErrIdempotentResponse) {
-		// This is fine.
-		// XXX: we should totally attach a warning to the function run here.
-		return nil
+		e.log.Warn("step output already persisted; keeping existing output", "error", err, "run_id", runCtx.Metadata().ID.RunID, "step_id", gen.ID)
+		err = nil
 	}
 	if err != nil {
 		return err
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -3717,9 +3717,11 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
 	}
 
 	hasPendingSteps, err := e.smv2.SaveStep(ctx, runCtx.Metadata().ID, gen.ID, []byte(output))
-	// A duplicate or idempotent save means the step was already persisted
-	// (typically via the checkpoint path). Keep the original output and
-	// continue so the run still advances.
+	// The step output was already persisted, typically by an earlier
+	// attempt that timed out or errored after succeeding. Clear the error so the
+	// discovery step still enqueues; otherwise the run stalls.
+	// XXX: hasPendingSteps can be stale on duplicate; worst case is
+	// redundant SDK invocations per in-flight parallel siblng.
 	if errors.Is(err, state.ErrDuplicateResponse) || errors.Is(err, state.ErrIdempotentResponse) {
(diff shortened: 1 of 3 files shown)


pkg/execution/executor/executor.go
```diff
@@ -3523,10 +3523,11 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
...
 
 	hasPendingSteps, err := e.smv2.SaveStep(ctx, runCtx.Metadata().ID, gen.ID, []byte(output))
+	// A duplicate or idempotent save means the step was already persisted
+	// (typically via the checkpoint path). Keep the original output and
+	// continue so the run still advances.
 	if errors.Is(err, state.ErrDuplicateResponse) || errors.Is(err, state.ErrIdempotentResponse) {
-		// This is fine.
-		// XXX: we should totally attach a warning to the function run here.
-		return nil
+		err = nil
```
Linell, line 3727: Could this lead to two parallel requests headed to the SDK? How did we actually get here "correctly" in the first place? 🤔

Linell: If we hit idempotency in the checkpointing endpoint handler, are we enqueueing a discovery step? Because we shouldn't be.

* Should the checkpointing API _never_ enqueue a discovery step? 
* Should it ever enqueue _anything_? Could that be an invariant that we document and test?

Linell: We _do_ need to add some sort of warning log here to let us know we've hit this case.

rhino1998 (approved):

pkg/execution/executor/executor.go
```diff
@@ -3646,10 +3646,12 @@
...
 
 	hasPendingSteps, err := e.smv2.SaveStep(ctx, runCtx.Metadata().ID, gen.ID, []byte(output))
+	// A duplicate or idempotent save means the step was already persisted
+	// (typically via the checkpoint path). Keep the original output and
+	// continue so the run still advances.
 	if errors.Is(err, state.ErrDuplicateResponse) || errors.Is(err, state.ErrIdempotentResponse) {
-		// This is fine.
-		// XXX: we should totally attach a warning to the function run here.
-		return nil
+		e.log.Warn("step output already persisted; keeping existing output", "error", err, "run_id", runCtx.Metadata().ID.RunID, "step_id", gen.ID)
```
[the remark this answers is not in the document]
rhino1998: Technically true I guess

pkg/execution/executor/executor.go
```diff
@@ -3658,10 +3646,12 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
...
+		err = nil
```
amh4r, line 3727: Can you add a little more context in a comment? We want readers to understand why we're unsetting the error. IIUC, we're unsetting it because we want to handle the case where the SDK's outgoing checkpoint request was process but the SDK still got a failed response (or no response at all, e.g. timeout). In that situation, the SDK returned the step output back in the HTTP response and the Executor processes it again. We hit idempotency because the step was already processed, but we need to still enqueue a discovery step. If we don't, then the run is stuck.

Also highlight some risks with this. For example, what happens if this was a parallel step?

Linell: I added a Linear ticket to improve the parallel step performance [here](https://linear.app/inngest/issue/EXE-1822/return-accurate-haspendingsteps-from-savestep-duplicate-path), and updated the comment in the code.

amh4r (approved):

PR #4140, opened 2026-05-07
fix: validate dispatch generation on async checkpoints

Linell:
When an HTTP timeout caused a dispatch to be requeued while the SDK was still running, the original dispatch could submit checkpoints against the new run and produce duplicate steps. Drivers now echo the queue item's GenerationID into the SDK request payload, and the async-checkpoint API loads the current queue item to compare; mismatches return HTTP 409 (ErrStaleDispatch).

Fails open on:
  - missing/zero GenerationID from the SDK (older clients)
  - zero GenerationID on the loaded queue item (pre-rollout items)
  - transient queue-load errors (Redis blips)

Stacked on top of the queue plumbing change.

---

This PR is a split from https://github.com/inngest/inngest/pull/4108

---

> [!NOTE]
> Adds a `RequestID`-based fence to async checkpoints to prevent duplicate step execution when an HTTP timeout causes a dispatch to be requeued while the original SDK is still running. Drivers stamp a deterministic ULID derived from `(RunID, JobID, GenerationID)` on outbound SDK requests; the async-checkpoint handler compares it against the current queue item's generation and rejects stale dispatches with HTTP 409 (`ErrStaleDispatch`). A `RequestStartedAt` short-circuit skips the queue load for dispatches younger than 10 seconds. The feature is gated per-account via `AllowAsyncDispatchValidation` and fails open for older SDK clients, pre-rollout queue items, and transient Redis errors. The `tests/golang/checkpoint_generation_test.go` that was blocking CI in all prior rounds was replaced entirely by the new unit/integration tests in `dispatch_validation_test.go`.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 69b981df7232f22d3852e0bc86ed9314924a17da.</sup>


pkg/execution/checkpoint/checkpoint.go
```diff
@@ -546,6 +556,52 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 	return nil
 }
 
+func (c checkpointer) validateAsyncDispatch(ctx context.Context, input AsyncCheckpoint) error {
```
BrunoScheufler: Flagging that this operation will run for every Checkpoint and will increase latency. We also have to be mindful that this is a queue request: Do we have rate limits for Checkpointing?

BrunoScheufler (dismissed):
We should check if there are any throughput protections on the Checkpointing API. On top of that, if we can find ways to check for staleness that doesn't involve hitting the queue, that would be preferable!

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -546,6 +562,64 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
...
+	if input.GenerationID == 0 {
+		return nil
+	}
+
+	// Skip the queue-item load when the dispatch is younger than the minimum
+	// requeue window. A Requeue is the only path that bumps GenerationID, and
+	// it can't fire until the queue lease expires, so a fresh dispatch is
+	// provably uncontested. Negative elapsed (future-dated stamp from clock
+	// skew or a buggy SDK) falls through to the existing validation.
+	if input.StepStartedAt != 0 {
```
Linell: Added to the SDK [here](https://github.com/inngest/inngest-js/pull/1511/commits/b493dd6f5c103df2098f1bbadab59c4295ba4392)

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -546,6 +562,64 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
...
+		elapsed := time.Since(time.UnixMilli(input.StepStartedAt))
+		if elapsed >= 0 && elapsed < freshDispatchWindow {
+			return nil
+		}
+	}
+
+	ref := queueref.Decode(input.QueueItemRef)
+	if ref[0] == "" {
+		return fmt.Errorf("%w: missing queue item reference", ErrStaleDispatch)
```
Linell: We need to add a metric around how many stale dispatches we're returning? Also need to tag how many times we validated and it _wasn't_ stale. We'll want a duration and a count. I should use a histogram from the metrics package.

Can we put this. behind a feature flag in general, so that we're able to enable/disable (per account) entirely. This lets us do a percentage rollout, too. Fail closed with LaunchDarkly issues in mind.

pkg/execution/executor/executor.go
```diff
@@ -1521,7 +1521,8 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		return nil, fmt.Errorf("no function loader specified running step")
 	}
 
-	requestID := ulid.MustNew(ulid.Timestamp(e.now()), rand.Reader).String()
+	seed := fmt.Appendf(nil, "%s:%d", id.RunID, queue.GenerationIDFromContext(ctx))
```
Linell: This needs to be the JobID, not the RunID.

pkg/execution/executor/executor.go
```diff
@@ -1592,7 +1521,7 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		return nil, fmt.Errorf("no function loader specified running step")
 	}
 
-	requestID := ulid.MustNew(ulid.Timestamp(e.now()), rand.Reader).String()
+	requestID := driver.DispatchRequestID(e.now(), id.RunID, queue.GenerationIDFromContext(ctx)).String()
```
amh4r, line 1613: Why are we using the run ID instead of the job ID? If we send 2 parallel requests at the same time then wouldn't they have the same request ID?

amh4r (approved):

pkg/flags/flags.go
```diff
@@ -0,0 +1,24 @@
...
+// access feature flags safely
+package flags
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+)
+
+func NewBoolFlag(fn func(ctx context.Context, acctID uuid.UUID) bool) BoolFlag {
```
BrunoScheufler: I don't think we need to attack this in this PR. We can probably do a follow-up to rethink passing flags down, but we'll have to consider more things for that anyway

Linell: Would you be open with shipping as-is and refactoring the flag later? Aaron wasn't a huge fan of the previous code for the flag - it looked like the metadata equivalent - but this _is_ a little cleaner. It'd just be nice to get this out for folks!

pkg/execution/queue/item.go
```diff
@@ -88,6 +103,9 @@ type QueueItem struct {
 	WorkspaceID uuid.UUID `json:"wsID"`
 	// LeaseID is a ULID which embeds a timestamp denoting when the lease expires.
 	LeaseID *ulid.ULID `json:"leaseID,omitempty"`
+	// GenerationID is a monotonic counter bumped by Requeue; mismatch fences
+	// a stale dispatch.
+	GenerationID int `json:"genID,omitempty"`
```
BrunoScheufler, line 108: Let's split out the generation ID into a separate queue-only PR for readability, thank you!

Linell: Opened [here](https://github.com/inngest/inngest/pull/4219)!

PR #4142, opened 2026-05-07
Add new "doctor" and "doctor healthcheck" commands

djfarrelly:
## Description

This command will be an umbrella for system and configuration checks that folks can run in their environment against their server. The first check is a healthcheck command which allows us to drop the curl requirement in the base image (see #4134).

```
inngest alpha doctor healthcheck
```

## Motivation
We can drop curl from the base image (see: #4134). This enables us to reduce the packages footprint for security purposes. This then allows folks running Inngest server with Docker Compose or other systems to run a healthcheck with zero other dependencies.

- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds `inngest alpha doctor` as an umbrella diagnostic command with `healthcheck` as the first subcommand, probing `/health` and `/ready` in parallel to replace `curl` in the base Docker image. Follow-up commits fix env-var Sources resolution through the parent command path, add a regression test with a `help`-skip to prevent a production panic, and strengthen the `TestRun_SchemeOverride` assertion by capturing probe errors via `ErrWriter` and checking for `"tls"` in the output.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit af3491d554f4539a1de717b698ce1a317175aec2.</sup>

### DIFF
--- a/cmd/devserver/cmd.go
+++ b/cmd/devserver/cmd.go
@@ -1,6 +1,7 @@
 package devserver
 
 import (
+	"github.com/inngest/inngest/pkg/api"
 	"github.com/inngest/inngest/pkg/devserver"
 	"github.com/urfave/cli/v3"
 )
@@ -36,10 +37,10 @@ func Command() *cli.Command {
 				Name:  "host",
 				Usage: "Inngest server host",
 			},
-			&cli.StringFlag{
+			&cli.IntFlag{
 				Name:    "port",
 				Aliases: []string{"p"},
-				Value:   "8288",
+				Value:   api.DefaultAPIPort,
 				Usage:   "Inngest server port",
 			},
 
--- a/cmd/devserver/devserver.go
+++ b/cmd/devserver/devserver.go
@@ -5,11 +5,11 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"strconv"
 	"syscall"
 	"time"
 
 	localconfig "github.com/inngest/inngest/cmd/internal/config"
+	"github.com/inngest/inngest/pkg/api"
 	"github.com/inngest/inngest/pkg/config"
 	connectConfig "github.com/inngest/inngest/pkg/config/connect"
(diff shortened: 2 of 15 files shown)


darwin67 (approved):

cmd/devserver/cmd.go
```diff
@@ -36,10 +37,10 @@ func Command() *cli.Command {
 				Name:  "host",
 				Usage: "Inngest server host",
 			},
-			&cli.StringFlag{
+			&cli.IntFlag{
```
djfarrelly, line 40: This should always be a int flag, we were handling the conversion manually and bubbling errors - this is cleaner.

cmd/devserver/devserver.go
```diff
@@ -45,12 +45,7 @@ func action(ctx context.Context, cmd *cli.Command) error {
 		os.Exit(1)
 	}
 
-	portStr := localconfig.GetValue(cmd, "port", "8288")
-	port, err := strconv.Atoi(portStr)
-	if err != nil {
-		fmt.Println(err.Error())
-		os.Exit(1)
-	}
+	port := localconfig.GetIntValue(cmd, "port", api.DefaultAPIPort)
```
djfarrelly, line 48: Now using an `IntFlag`, we can remove simplify this. This is tested to work with flags, config files and env vars that may list this as a string - conversion is automatic.

cmd/doctor/healthcheck/cmd.go
```diff
@@ -0,0 +1,142 @@
...
+	defaultScheme    = "http"
+	defaultTimeout   = 5 * time.Second
+	verboseBodyLimit = 200
+)
+
+func Command() *cli.Command {
+	return &cli.Command{
+		Name:  "healthcheck",
+		Usage: "Probe local inngest HTTP endpoints; exit non-zero on failure",
+		Flags: []cli.Flag{
```
djfarrelly, line 27: This aims to mimic the same flags as `start` as much as possible

darwin67 (approved):

PR #4136, opened 2026-05-07
feat: add in-band syncs to self-hosted

amh4r:
## Description

Add in-band syncs to Self-Hosted. Also 404 the `/fn/register` endpoint (used for out-of-band syncs).

Does not add in-band syncs to Dev Server.

See `pkg/appsync/ARCHITECTURE.md` for detailed info. The PR description is intentionally sparse because I want to ensure the docs are good 😄

---

> [!NOTE]
> Implements in-band (server-initiated) app syncs for `inngest start`. The server makes a signed PUT to the SDK URL, validates the signed response, and persists the result via a new `AppSyncer` interface. Adds `--allow-insecure-http` CLI flag, rate limiting on the sync endpoint, and comprehensive unit tests covering error sanitization, syscode propagation, and the full success path against a real `inngestgo` handler.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 7bd8b6e5d558f1de95b464558e29a901350a2e97.</sup>

PR #4143, opened 2026-05-07
fix(syncs): fix function orphaning on app resync after <v1.15 upgrade

djfarrelly:
## Description

Customer DBs from <v1.15 hold `apps` rows keyed by `sha1(URL)`; v1.15 (#3361) flipped the seed to `sha1(AppName)`. On resync the new id didn't match the legacy row, so register() inserted a parallel name-keyed row and every function attached to the legacy app_id was left orphaned under the now-archived URL-keyed app.

This PR mirrors the `functions(app_id, slug) WHERE archived_at IS NULL` partial-index pattern on `apps` so adoption happens at the SQL layer:

- New `apps_name_unique_key` on `(name) WHERE name <> ''` (empty names stay unconstrained for placeholder paths: `-u`, autodiscovery, UI add-by-URL).
- New `UpsertAppByName` uses the partial index as the conflict arbiter and preserves the existing row's id on conflict, so resyncs adopt legacy URL-keyed rows in place — id, function rows, and history all survive.
- Tightened `UpsertApp` (placeholder path) to skip the name update when `excluded.name = ''`, so placeholder upserts no longer blank a real name — the upstream defect that produced the legacy `(name='', URL-derived id, functions attached)` rows in the first place.
- `register()` hoists `UpsertAppByName` to the top of its tx so the function-insert loop uses the canonical app_id. The placeholder cleanup loop now distinguishes true placeholders (soft-delete) from legacy SDK rows whose name was blanked by the v1.15 ON CONFLICT(id) clobber (adopt by id so their attached functions don't orphan).
- CI: pass `--no-poll` to the dev server in e2e workflows. The pollSDKs loop can race with explicit URL changes by re-syncing a stale URL; the tests sync explicitly, so polling adds nothing but flake.

EXE-1738

## Release note

Inngest server: fix functions being orphaned (and runs/history pointing at a now-archived app) when re-syncing after upgrading from pre-v1.15. Legacy URL-keyed app rows are adopted in place by name on first re-sync.

## Migration note

Goose database migration `000004_apps_unique_active_name.sql` (sqlite) / `000006_apps_unique_active_name.sql` (postgres) adds the partial unique index and pre-archives any pre-existing same-name duplicates (active row or row with most functions wins, losers get their name suffixed with their id so they remain reachable for manual recovery). No action required for clean installs. This cleanup should also remove classes of bugs that were possible before.


pkg/devserver/api.go
```diff
@@ -253,24 +253,20 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*sync.Repl
...
 		for _, app := range allApps {
 			if app.Name == "" && util.NormalizeAppURL(app.Url, false) == normalizedURL {
-				// Since there's an app with the same URL but no name, we can
-				// assume it was a failed sync. We should delete it since we're
-				// in the process of syncing a replacement app.
-				//
-				// This situation happens when a user enters an unreachable URL
-				// in the UI. It'll still create an app, but in a placeholder
-				// state.
 				if err := a.devserver.Data.DeleteApp(ctx, app.ID); err != nil {
```
AndyInternet: Tested this branch live against Postgres with a row in the shape Cohere reported on 2026-05-05: `name=''`, URL-derived id, two functions attached. Sync against the SDK fixture, and the legacy row gets soft-deleted here before `UpsertAppByName` runs. The partial unique index excludes `name=''`, so `UpsertAppByName` sees no conflict and inserts a fresh row at `sha1("atlas")`. `GetActiveFunctionByAppAndSlug` then joins on `apps.name AND apps.archived_at IS NULL`, finds nothing, and the SDK-minted function uuids replace the legacy ones. Legacy function rows stay attached to the archived app, invisible to the dashboard.

This is the exact post-upgrade state on the T-7175 re-open. The `CASE WHEN excluded.name = ''` guard prevents *new* blanking but doesn't recover rows already in that state, and Cohere's `psql` dumps show they're sitting on at least two of them in prod.

Smallest fix that addresses the failure mode: before calling `DeleteApp` in this loop, check whether the placeholder has active functions attached. If yes, it's not really a placeholder, it's a legacy URL-keyed row whose name was blanked by the upstream defect. Adopt it via id-keyed `UpsertApp` with `name = r.AppName` instead of deleting it.

```go
for _, app := range allApps {
    if app.Name == "" && util.NormalizeAppURL(app.Url, false) == normalizedURL {
        fns, _ := a.devserver.Data.GetFunctionsByAppInternalID(ctx, app.ID)
        if len(fns) > 0 {
            adopt := cqrs.UpsertAppParams{ /* copy fields from r */ }
            adopt.ID = app.ID
            adopt.Name = r.AppName
            _, _ = a.devserver.Data.UpsertApp(ctx, adopt)
            continue
        }
        _ = a.devserver.Data.DeleteApp(ctx, app.ID)
    }
}
```

Pair with a regression test seeding the blanked-name + attached-functions state and asserting both `(name='atlas', archived_at IS NULL)` and that the original function uuids survive. Full repro evidence and SQL dumps in the test results doc.

pkg/db/postgres/migrations/000006_apps_unique_active_name.sql
```diff
@@ -0,0 +1,29 @@
...
+-- and the loser row may have runs attached. Keep the highest ctid (most
+-- recent insert) per name. Mirrors the precedent in 000005 for functions.
+UPDATE apps
+SET archived_at = NOW()
+WHERE id IN (
+    SELECT a.id
+    FROM apps a
+    JOIN apps b
+      ON a.name = b.name
+     AND a.ctid < b.ctid
```
AndyInternet: Repro with two active duplicates shaped like the Cohere DB on 2026-05-05 (older row created 2026-02-11 with two function attachments, newer empty duplicate created 2026-05-05):

```
  id          | created_at                 | fn_count | ctid
 211bf1db-... | 2026-02-11 23:45:02.353971 |        2 | (0,1)
 8657d7e6-... | 2026-05-05 18:15:00.718443 |        0 | (0,2)
```

Migration runs, archives the older row because `a.ctid < b.ctid`. Survivor has zero active functions. Next SDK resync routes through `UpsertAppByName`, hits the active row by name, function-insert loop finds no matching slugs, mints fresh uuids. Legacy `aaaa…`/`bbbb…` function rows stay attached to the archived app. Historical `function_runs` and `history` rows pointing at the legacy ids resolve to function rows on an archived app and drop out of every active view.

Confirmed live on Postgres 17.9. The SQLite migration at `pkg/db/sqlite/migrations/000004_apps_unique_active_name.sql:11` uses `MAX(rowid)`, which is the same pattern, but I didn't explicitly test the SQLite code path. Same fix should land in both.

Rank by attached active function count instead:

```sql
WITH ranked AS (
  SELECT a.id, a.name,
         (SELECT count(*) FROM functions f WHERE f.app_id = a.id AND f.archived_at IS NULL) AS fn_count,
         a.created_at
  FROM apps a WHERE a.archived_at IS NULL AND a.name <> ''
), losers AS (
  SELECT id FROM (
    SELECT id, row_number() OVER (PARTITION BY name ORDER BY fn_count DESC, created_at ASC) AS rn
    FROM ranked
  ) t WHERE rn > 1
)
UPDATE apps SET archived_at = NOW() WHERE id IN (SELECT id FROM losers);
```

While in there: `ctid` is a physical pointer that isn't stable across `VACUUM FULL` or table rewrites. Fine for a one-shot migration today, but `created_at DESC, id DESC` is the more idiomatic tiebreaker once function count is the primary signal.

Also needs a regression test. `TestAppUniqueActiveName` verifies the index rejects duplicates, but nothing tests the migration picks the right winner when the DB starts with duplicates. That's the bug in CI form.

pkg/devserver/api.go
```diff
@@ -350,15 +353,30 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*sync.Repl
...
 	}
 
+	// Resolve the canonical app row first so the function-insert loop below
+	// uses the correct app_id. UpsertAppByName conflicts on the partial unique
+	// index apps_name_active_key, so an existing active row with this name is
+	// adopted and the legacy id is returned in `app.ID` regardless of how it
+	// was originally minted.
+	app, err := tx.UpsertAppByName(ctx, appParams)
+	if err != nil {
+		_ = tx.Commit(ctx)
```
AndyInternet:
```go
app, err := tx.UpsertAppByName(ctx, appParams)
if err != nil {
    _ = tx.Commit(ctx)   // intent: abort, behavior: a Postgres tx already in error state no-ops Commit
    return nil, publicerr.Wrap(err, 500, "Error upserting app")
}
```

Should be `tx.Rollback(ctx)`. Behaviorally benign in Postgres because the query failure already aborted the tx, but it reads like "save partial state on error" and would pollute logs if the adapter ever surfaces Commit errors. Trivial, but worth flipping while the other two are in flight.

pkg/db/postgres/migrations/000006_apps_unique_active_name.sql
```diff
@@ -0,0 +1,29 @@
...
+    WHERE a.archived_at IS NULL
+      AND b.archived_at IS NULL
+      AND a.name <> ''
+);
```
djfarrelly: Should be safe - not expecting folks to have 1000s of apps.

pkg/devserver/api.go
```diff
@@ -350,15 +353,30 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*sync.Repl
...
-		// We want to save an app at the end, after handling each error.
+		// On error, persist the message to the row by re-upserting. On
+		// success the earlier UpsertAppByName call has already written the
+		// canonical state, so the deferred upsert is a no-op idempotent update.
 		if err != nil {
 			appParams.Error = sql.NullString{
 				String: err.Error(),
 				Valid:  true,
 			}
+			_, _ = tx.UpsertAppByName(ctx, appParams)
```
djfarrelly, line 409: Looking into this case with 2 uses within the transaction

pkg/db/postgres/migrations/000006_apps_unique_active_name.sql
djfarrelly: @AndyInternet - some solid ideas here - I'll adjust on the ctid and also have broader changes coming. 

In what situation do you know how the app would have:
* have the same app name (slug) creating a conflict
* a newer created at date
* not be archived
* not have any functions associated with it

I can't think of a case where this could happen and if it did, I would expect a lot of other bugs to be present if this could happen.

pkg/devserver/api.go
```diff
@@ -253,39 +253,21 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*sync.Repl
...
-					l.Error("error deleting app", "error", err)
-				}
-			}
-		}
-	}
-
 	method := enums.AppMethodServe
 	if r.IsConnect() {
 		method = enums.AppMethodConnect
 	}
```
AndyInternet, line 260: Should we check for an empty `appName` before we call the new codepath? We enforce the name in SDKs but a backend check is a good idea too because we'll get very unexpected behavour if one gets through.

Something like:

if r.AppName == "" {
	return nil, publicerr.Errorf(http.StatusBadRequest, "appName is required")
}

pkg/devserver/api.go
```diff
@@ -300,13 +282,57 @@ func (a devapi) register(ctx context.Context, r sdk.RegisterRequest) (*sync.Repl
...
+	//      filters archived apps, so the SDK-resolved slugs miss them and
+	//      fresh SDK-minted uuids replace the originals). Adopt it instead:
+	//      UpsertApp by id with name = r.AppName so the row re-enters the
+	//      active universe under the correct name, and UpsertAppByName below
+	//      then finds it by name and refreshes its mutable fields.
+	//
+	// Iterate all apps rather than using GetAppByURL because that query
+	// returns LIMIT 1 and may find the real (named) app instead when both
+	// share the same URL.
+	if allApps, err := a.devserver.Data.GetAllApps(ctx, consts.DevServerEnvID); err == nil {
```
AndyInternet: Should we log this error so it's not being silently swallowed?

AndyInternet (approved):

pkg/devserver/api.go
djfarrelly: We could do more validation of the payload, but I think it's OK for now - we'll likely plan a phased deprecation of this endpoint in favor of https://github.com/inngest/inngest/pull/4136 soon

pkg/devserver/api.go
djfarrelly: Added this

pkg/devserver/api.go
djfarrelly: Added this check

PR #4149, opened 2026-05-08
Fix dropped sleep trace attempts

jacobheric:
## Description

Agentic api testing revealed that we shipped extra internal spans to the ui that the ui was hiding. Properly hiding them here as the agentic api is sharing this mapping code so we have the same traces. 

## Motivation
Cleaner traces for the agentic api. 

- [ ] Docs

## Checklist

---

> [!NOTE]
> Fixes two bugs: (1) dropped execution spans (`MarkedAsDropped=true`) were still processed due to missing `continue`, causing spurious sleep attempt children in trace output; (2) `LoadMany` was indexing `errs[1]` instead of iterating for the first non-nil error, silently dropping the first error and panicking on single-result batches. Both fixes include regression tests.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit cf5c9d3c7f7a9a34d014737b5b580203e1fd1d4e.</sup>

### DIFF
--- a/pkg/coreapi/graph/loaders/loader.go
+++ b/pkg/coreapi/graph/loaders/loader.go
@@ -79,8 +79,10 @@ func LoadMany[T interface{}](
 ) ([]T, error) {
 	thunkMany := loader.LoadMany(ctx, keys)
 	results, errs := thunkMany()
-	if len(errs) > 0 {
-		return []T{}, errs[1]
+	for _, err := range errs {
+		if err != nil {
+			return []T{}, err
+		}
 	}
 
 	output := make([]T, len(keys))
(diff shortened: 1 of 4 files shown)


djfarrelly (approved):

pkg/coreapi/graph/loaders/loader.go
```diff
@@ -79,8 +79,10 @@ func LoadMany[T interface{}](
 ) ([]T, error) {
 	thunkMany := loader.LoadMany(ctx, keys)
 	results, errs := thunkMany()
-	if len(errs) > 0 {
-		return []T{}, errs[1]
+	for _, err := range errs {
```
jacobheric, line 82: This is technically unrelated but fixed an existing potential panic that mendral found.

PR #4159, opened 2026-05-11
fix(grpc): handle unimplemented error for `LoadDefersMeta` for backwards compatibility

admirsaheta:
## Description

When LoadDefersMeta is called on a gRPC state-service deployment that was compiled before the RPC was added to the proto (state.v2.RunService), the server returns codes.Unimplemented and the entire execution request fails with:

error loading defers in driver marshaller: non-retriable error: rpc error: code = Unimplemented desc = unknown method LoadDefersMeta for service state.v2.RunService
The fix detects the codes.Unimplemented gRPC status in MarshalV1 and falls back to an empty defers map, so executions continue normally on older state-service deployments instead of failing.

## Motivation

LoadDefersMeta was added to the StateLoader interface and called in driver.MarshalV1 in the same release that added the DeferAdd/DeferCancel opcodes. The corresponding gRPC RPC was added to the proto in a separate commit (22559d61f). Any state-service gRPC server deployed before that proto update doesn't know about LoadDefersMeta, causing all function executions routed through it to fail with a non-retriable error.

The fallback is safe because: prior to this feature, defers were never sent to the SDK at all — an empty map is exactly what older deployments would have sent.
- [ ] Docs

## Checklist

---

> [!NOTE]
> Adds backwards-compatible fallback in `MarshalV1` when the gRPC state-service returns `codes.Unimplemented` for `LoadDefersMeta`, treating it as an empty defers map to prevent execution failures during rolling upgrades.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 233336b6cc986deaab40f12d517587c05ed2ee0e.</sup>

### DIFF
--- a/pkg/execution/driver/driver.go
+++ b/pkg/execution/driver/driver.go
@@ -14,6 +14,8 @@ import (
 	sv2 "github.com/inngest/inngest/pkg/execution/state/v2"
 	"github.com/inngest/inngest/pkg/inngest"
 	"github.com/inngest/inngest/pkg/util/errs"
+	"google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
 )
 
 // DriverV2 represents a driver that makes requests to SDKs to re-enter and execute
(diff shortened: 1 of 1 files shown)


darwin67:
@admirsaheta thanks for catching this. could you rebase against the latest `main`? there's a new CI check added for conventional commits for PR titles. Your title is mostly correct already (probably need a space after colon), so there should be nothing needed except for the rebase.

admirsaheta:
@darwin67 done, should be good now 💯

PR #4170, opened 2026-05-12
feat: Phase 1 scoring

jakobevangelista:
## Description

Added support for the metadata tab in runs detail page

Also protects the `inngest.*` namespace for metadata in `pkg/api/apiv1/metadata.go`

SDK PR: https://github.com/inngest/inngest-js/pull/1521

Screenshot 2026-05-11 at 6 36 36 PM

---

> [!NOTE]
> Adds `inngest.score` as a first-class metadata kind with backend validation (allowlist, finite numeric values, regex-constrained names, step/run scope enforcement). Improves error handling to distinguish transient vs. not-found `LoadMetadata` errors. Adds a "Scores" tab in the frontend that aggregates scores recursively from child spans, with score-aware filtering in `StepInfo` and `TopInfo`. New commits (`961cb08`–`63ed066`) share score metadata collection via `collectScoreMetadata`, clarify reserved metadata validation, and update comments around metadata target handling.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 63ed0662d2ecb81f5cccb69d564dd890922d9c5b.</sup>


ui/packages/components/src/RunDetails/ScoresAttrs.tsx
```diff
@@ -0,0 +1,59 @@
+import { useMemo } from 'react';
```
amh4r: Is this file exactly the same as `RunDetailsV4/ScoresAttrs.tsx`? If so, should we share instead of duplicating?

pkg/tracing/metadata/metadata.go
```diff
@@ -144,3 +151,53 @@ func (m Update) Validate() error {
...
+	}
+
+	if m.Kind() == "inngest.score" {
+		return validateScoreValues(m.RawUpdate.Values)
+	}
+
+	return nil
+}
+
+func (m Update) ValidateAllowedForScope(scope Scope) error {
```
amh4r: 3 public `Validate*` methods is confusing. As a consumer, which should I call?

pkg/tracing/metadata/metadata.go
```diff
@@ -144,3 +151,53 @@ func (m Update) Validate() error {
...
+	if m.Kind() == "inngest.score" && scope != enums.MetadataScopeStep {
+		return fmt.Errorf("invalid score scope %q: %w", scope, ErrScoreScopeInvalid)
+	}
+
+	return nil
+}
+
+func (m ScopedUpdate) ValidateAllowed() error {
+	return m.Update.ValidateAllowedForScope(m.Scope)
+}
```
amh4r: Nit (non-blocking): moving the struct methods next to the structs make it more readable. I was scanning through `Update`'s methods and briefly got confused when I got here, then I saw it was for a different struct

pkg/tracing/metadata/metadata.go
```diff
@@ -144,3 +151,53 @@ func (m Update) Validate() error {
...
+func (m Update) ValidateAllowed() error {
+	if err := m.Validate(); err != nil {
+		return err
+	}
+
+	if err := m.Kind().ValidateAllowed(); err != nil {
+		return err
+	}
+
+	if m.Kind() == "inngest.score" {
```
amh4r: This string is used in a variety of places. Can we make it a constant/enum?

pkg/tracing/metadata/metadata_test.go
```diff
@@ -78,3 +81,84 @@ func TestValuesSizeNilMap(t *testing.T) {
...
+			wantErr: ErrScoreNameInvalid,
+		},
+		{
+			name: "non-score metadata keeps generic shape",
+			update: Update{RawUpdate: RawUpdate{
+				Kind:   "userland.score",
+				Op:     enums.MetadataOpcodeMerge,
+				Values: Values{"score": json.RawMessage(`{"value":1}`)},
+			}},
+		},
```
amh4r, line 170:
```suggestion
		},
		{
			name: "null is invalid",
			update: Update{RawUpdate: RawUpdate{
				Kind:   "userland.score",
				Op:     enums.MetadataOpcodeMerge,
				Values: Values{"score": json.RawMessage(`null`)},
			}},
			wantErr: ErrScoreValueInvalid,
		},
```

Do we want to disallow null? The above test fails

pkg/api/apiv1/metadata.go
```diff
@@ -81,13 +81,6 @@ func (a router) addRunMetadata(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	for _, md := range data.Metadata {
-		if err := md.Validate(); err != nil {
-			_ = publicerr.WriteHTTP(w, publicerr.Wrap(err, 400, "Invalid metadata"))
-			return
-		}
-	}
-
```
amh4r, line 90: Maybe doesn't matter, but removing this means validation happens after loading metadata (i.e. we don't avoid the StateStore operation when we get invalid metadata)

ui/packages/components/src/RunDetailsV3/TopInfo.tsx
```diff
@@ -258,7 +270,9 @@ export const TopInfo = ({
       {result?.error && <ErrorInfo error={result.error.message || 'Unknown error'} />}
       <div className="flex-1">
         <Tabs
-          defaultActive={result?.error ? 'error' : prettyPayload ? 'input' : 'output'}
+          defaultActive={
+            result?.error ? 'error' : prettyPayload ? 'input' : prettyOutput ? 'output' : ''
+          }
```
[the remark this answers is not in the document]
jakobevangelista: for both these comments, tried repro'ing with 

```
inngest.createFunction(
  {
    id: "score-only-step",
    name: "score-only step",
    triggers: { event: eventTypes.testExperimentPhase1ScoreOnlyStep },
    retries: 0,
  },
  async ({ step }) => {
    await step.score("score-only", {
      name: "answer_quality",
      value: 1,
    });
  },
);
```

and the behavior was as expected, so probably false positive

`Tabs` already falls back to the first tab when `defaultActive === ''`, so this seems okay unless there’s another route where tabs are ordered differently / no first tab is present

pkg/tracing/metadata/metadata.go
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/791e1d4dcddcacacb08e14446391e30350b0e4c9

pkg/tracing/metadata/metadata.go
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/791e1d4dcddcacacb08e14446391e30350b0e4c9

pkg/tracing/metadata/metadata.go
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/0547fb21ee7af9ed10752f5aa17119a12b89a2c8

pkg/tracing/metadata/metadata_test.go
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/1523327ee5b1ccb0b2cfdc0190f7bdfcdbd270e1

pkg/api/apiv1/metadata.go
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/cae9c383685bd74b66db6909750edcb4402b9f2c

ui/packages/components/src/RunDetails/ScoresAttrs.tsx
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/049a4e1e9849b7131d2aba8809bdddf955da9872

darwin67:
could you rebase against `main`? the conventional commit check should pass after.

pkg/tracing/metadata/metadata.go
```diff
@@ -121,6 +127,11 @@ type ScopedUpdate struct {
 	Update
 }
 
+// ValidateAllowed applies reserved-kind rules.
+func (m ScopedUpdate) ValidateAllowed() error {
```
AndyInternet: ScopedUpdate.ValidateAllowed is just a pass-through to Update.ValidateAllowed and the embedded method is already promoted onto ScopedUpdate. Is the wrapper necessary?

ui/packages/components/src/RunDetailsV4/types.ts
```diff
@@ -32,7 +32,7 @@ export type ResponseInfo = {
   headers: Record<string, string | string[]>;
 };
 
-export type SpanMetadataKind = GeneratedSpanMetadataKind;
+export type SpanMetadataKind = GeneratedSpanMetadataKind | 'inngest.score';
```
AndyInternet: KindInngestScore isn't flowing through tygo, so the union has to be manually augmented here. Looks like it's because the const lives in kind.go (grouped with the prefixes) without a //tygo:generate annotation, unlike warning.go / ai.go / etc. which each have their own file with the annotation.

Could we move KindInngestScore into its own score.go with //tygo:generate, add | typeof KindInngestScore to the union in tygo.yaml, and regen? That drops this manual override and keeps the next person adding a kind on the well-trodden path.

AndyInternet (approved):

pkg/tracing/metadata/metadata.go
jakobevangelista: mb, i cleaned this up over the weekend and didn't push https://github.com/inngest/inngest/pull/4170/commits/5de8055ac1325cab460318a4ffaf89888bee6557

ui/packages/components/src/RunDetailsV4/types.ts
jakobevangelista: https://github.com/inngest/inngest/pull/4170/commits/15bbdab30a3b13dc294cd0b3f091c8ce3f882b59

PR #4173, opened 2026-05-12
internal: some potential fixes for our vercel api docs deploy

jacobheric:
## Description

API docs deploy ci config tweaks.

## Release note

None.

## Migration note

None.

---

> [!NOTE]
> Fixes the Vercel API docs deploy by adding a `vercel pull` step, regenerates v2 proto/OpenAPI artifacts with new Runs endpoints, and adds a post-processing pass to strip proto package-level enum prefixes (e.g. `FUNCTION_RUN_STATUS_COMPLETED` → `COMPLETED`) from the public OpenAPI spec.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit e7cc87fa69e79bca923a2196f1abbef2d6f6faa4.</sup>


.github/workflows/api_docs_deploy.yml
```diff
@@ -50,30 +50,38 @@ jobs:
       - name: Install Vercel CLI
         run: npm install -g vercel
 
+      - name: Pull Vercel project settings
+        working-directory: docs/api-docs
+        run: vercel pull --yes --environment="${{ inputs.environment }}" --token="$VERCEL_TOKEN"
```
djfarrelly: They recommend not using the `--token` command as it leaves the token in the logs - the env var should work. 
```suggestion
        run: vercel pull --yes --environment="${{ inputs.environment }}"
```

djfarrelly (changes requested):
Can we try without the `--token` flag here? I think the env var should be fine. The other changes all see aligned with the failure we saw.

https://vercel.com/docs/cli#using-in-a-ci/cd-environment

jacobheric:
> Can we try without the `--token` flag here? I think the env var should be fine. The other changes all see aligned with the failure we saw.
> 
> https://vercel.com/docs/cli#using-in-a-ci/cd-environment

done

docs/api-docs/public/api-specs/v2.json
```diff
@@ -559,6 +685,108 @@
...
+        "default": "TRACE_STEP_OP_UNSPECIFIED",
+        "enum": [
+          "TRACE_STEP_OP_UNSPECIFIED",
+          "TRACE_STEP_OP_RUN",
+          "TRACE_STEP_OP_SLEEP",
+          "TRACE_STEP_OP_WAIT_FOR_EVENT",
+          "TRACE_STEP_OP_INVOKE",
+          "TRACE_STEP_OP_SEND_EVENT",
+          "TRACE_STEP_OP_AI_GATEWAY",
+          "TRACE_STEP_OP_WAIT_FOR_SIGNAL"
```
djfarrelly: Ooo, since we modify this in the response. Are the docs correct or not for this? 

I think this is the bit we use internally, but externally is drops the prefix.

djfarrelly (approved):
Approving, with an open question- could be addressed here or another pr if any change is needed for accurate docs.

docs/api-docs/public/api-specs/v2.json
jacobheric: good catch! will think on a fix.

jacobheric: I added a big of code detect these enums with package/schema prefixes and strip them from the docs.

PR #4177, opened 2026-05-12
feat: surface deferred runs in run details

Linell:
## Description

Builds the persistence, GraphQL, and UI plumbing for surfacing deferred runs in the dev-server UI.

To accomplish this, we added a `run_defers` table that records parent/child defer linkages durably, keyed on `parent_run_id` and `defer_id`. 

`Finalize` writes a `run_defers` row for each `AfterRun/Aborted` defer alongside emitting `inngest/deferred.schedule`. After a child schedules, `backfillDeferChildRunID` joins the child run ID back onto the parent's defer row using the _inngest-prefixed metadata on the trigger event.

`FunctionRunV2` gains `defers: [RunDefer!]!` and `deferredFrom: RunDeferredFrom`.

### Screenshots

Screenshot 2026-05-12 at 11 08 11 AM
Screenshot 2026-05-12 at 11 08 05 AM

## Release note

None.

## Migration note

None.

---

> [!NOTE]
> Adds persistence, GraphQL API, and UI for surfacing deferred runs in the dev-server. A new `run_defers` table records parent/child defer linkages. `FunctionRunV2` gains `defers: [RunDefer!]!`, `deferredFrom: RunDeferredFrom`, and `runType: RunType!` fields backed by dataloader-batched resolvers. The UI surfaces this in the run details panel with a "Linked Runs" section.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 98a10be5e1406dd5f1edfbac532ae782179eaf2c.</sup>


pkg/db/postgres/migrations/000006_run_defers.sql
```diff
@@ -0,0 +1,18 @@
+-- +goose Up
+
+CREATE TABLE IF NOT EXISTS run_defers (
+    parent_run_id BYTEA NOT NULL,
+    defer_id VARCHAR NOT NULL,
+    user_defer_id VARCHAR NOT NULL DEFAULT '',
```
amh4r:
```suggestion
    user_defer_id VARCHAR NOT NULL,
```

pkg/cqrs/defer.go
```diff
@@ -0,0 +1,62 @@
...
+	FnSlug      string
+	Status      RunDeferStatus
+}
+
+type RunDeferStatus string
+
+const (
+	RunDeferStatusScheduled RunDeferStatus = "SCHEDULED"
+	RunDeferStatusAborted   RunDeferStatus = "ABORTED"
+)
```
amh4r: Can we use the existing status enum?

pkg/cqrs/defer.go
```diff
@@ -0,0 +1,62 @@
...
+// UI can show the id the SDK caller typed.
+type DeferStore interface {
+	InsertRunDefer(ctx context.Context, parentRunID ulid.ULID, deferID, userDeferID, fnSlug string, status RunDeferStatus) error
+	InsertRunDefers(ctx context.Context, defers []RunDeferInsert) error
+	UpdateRunDeferChildRunID(ctx context.Context, parentRunID ulid.ULID, deferID string, childRunID ulid.ULID) error
+}
+
+type RunDeferInsert struct {
+	ParentRunID ulid.ULID
+	DeferID     string
```
amh4r:
```suggestion
	HashedDeferID     string
```

Renaming the field avoids the need to clarify that it isn't the userland ID in various places

pkg/cqrs/defer.go
```diff
@@ -0,0 +1,62 @@
+package cqrs
+
+import (
+	"context"
+
+	"github.com/oklog/ulid/v2"
+)
+
+// DeferStore writes parent-run → child-run defer linkages so the dev-server
+// GraphQL can expose deferred runs after the parent's state has been deleted.
```
amh4r: Don't mention Dev Server here since this is an interface that will be implemented by Cloud too.

Also no need to mention GraphQL. There could be other consumers too (e.g. REST API), and that isn't a concern of this package

pkg/cqrs/defer.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// deferID is the SHA1 hash of the user-supplied defer id (the join key with
+// UpdateRunDeferChildRunID). userDeferID preserves the original string so the
+// UI can show the id the SDK caller typed.
```
amh4r:
```suggestion
```

pkg/cqrs/defer.go
```diff
@@ -0,0 +1,62 @@
...
+// RunDefer is a single defer attached to a parent function run. Parsed from
+// the parent's history opcodes (DeferAdd/DeferAbort) and joined to the
+// resulting child run when one exists. Run is nil if the defer was aborted,
+// the schedule event has not yet been processed, or the child trace was
+// pruned.
+//
+// ID is the SHA1-hashed defer id used as the join key. UserDeferID is the
+// original SDK-caller-supplied id (the first arg to `defer("foo", ...)`) and
+// is what the dev-server UI shows. The SDK always emits userland.id, so
+// UserDeferID is expected to be non-empty for any defer this server records.
```
amh4r: The field clarifications are good, but move them onto their respective fields. This makes it easier for readers to discover this info

pkg/db/sqlite/adapter.go
```diff
@@ -26,7 +26,7 @@ type Adapter struct {
 func New(conn *sql.DB) *Adapter {
 	return &Adapter{
 		conn: conn,
-		q:    &sqliteQuerier{q: sqlc.New(conn)},
+		q:    &sqliteQuerier{conn: conn, q: sqlc.New(conn)},
```
amh4r: Necessary?

pkg/cqrs/traces.go
```diff
@@ -482,6 +485,17 @@ func GetRunOptFromContext(ctx context.Context) GetRunOpt {
 	return opt
 }
 
+// RunTypeFilter narrows the run list to primary runs, deferred (child) runs,
```
rhino1998: Maybe this should be in `enums`?

pkg/cqrs/defer.go
rhino1998: ~~QUEUED and CANCELED instead of SCHEDULED and ABORTED?~~

pkg/history_reader/reader.go
```diff
@@ -118,7 +119,7 @@ type RunHistoryInvokeFunctionResult struct {
 	Timeout bool       `json:"timeout"`
 }
 
-func NewRunHistoryResultFromHistoryResult(hr *cqrs.Result) *RunHistoryResult {
+func NewRunHistoryResultFromHistoryResult(hr *history.Result) *RunHistoryResult {
```
rhino1998: NIT: This is still (re)exported by cqrs, so this file's modifications can be reset

rhino1998 (approved):
I think the backend of this looks reasonable once the minor build issues are fixed.

pkg/coreapi/graph/resolvers/function_run_v2.resolver.go
```diff
@@ -25,6 +25,89 @@ func (r *functionRunV2Resolver) Function(ctx context.Context, fn *models.Functio
 	return models.MakeFunction(fun)
 }
 
+func (r *functionRunV2Resolver) Defers(ctx context.Context, fn *models.FunctionRunV2) ([]*models.RunDefer, error) {
```
amh4r, line 32: This isn't returning defers until the parent run ends

pkg/execution/executor/executor.go
```diff
@@ -831,10 +822,26 @@ func (e *executor) Schedule(ctx context.Context, req execution.ScheduleRequest)
 		"fn_id", req.Function.ID,
 		"fn_v", req.Function.FunctionVersion,
 		"evt_id", req.Events[0].GetInternalID(),
-		"run_id", runID,
 		"schedule_req", req,
 	)
 
+	deferred := findDeferredChild(l, req.Events)
```
amh4r: Why do we need to change the scheduling logic? I'm not following

pkg/event/defer.go
```diff
@@ -1,11 +1,20 @@
...
+	"github.com/inngest/inngest/pkg/consts"
+)
 
 type DeferredScheduleMetadata struct {
 	FnSlug       string `json:"fn_slug"`
 	ParentFnSlug string `json:"parent_fn_slug"`
 	ParentRunID  string `json:"parent_run_id"`
+	// HashedDeferID is the hashed per-run defer ID; the SDK-supplied ID is
+	// hashed inside SaveDefer before reaching this field.
+	HashedDeferID string `json:"defer_id"`
```
amh4r:
```suggestion
	HashedDeferID string `json:"hashed_defer_id"`
```

Let's be explicit that's it's a hash. Otherwise, it'll be messy if we every want to include the userland defer ID in the future.

IIUC, this is only necessary because we need the hashed defer ID to update the parent run's span while scheduling the child run. Is that correct?

pkg/execution/executor/finalize.go
```diff
@@ -222,28 +222,17 @@ func (e *executor) buildDeferEvents(
 			continue
 		}
 
+		if d.ScheduleStatus == enums.DeferStatusAfterRun || d.ScheduleStatus == enums.DeferStatusAborted {
+			e.emitDeferSpan(ctx, opts.Metadata, now, d)
```
amh4r: This appears to be creating the span for the first time. That's problematic because we'll only see defers in the UI after the parent run ends.

We shouldn't need to do any updating during finalization:
- If `DeferStatusAfterRun`, then we'll set its run ID elsewhere when scheduling.
- If `DeferStatusAborted`, then its span will already have been updated during the abort process. That said, we don't need to handle that in this PR because abort is out of scope for now.

pkg/execution/state/redis_state/lua/saveDefer.lua
```diff
@@ -12,7 +12,7 @@ KEYS[1] - defers meta hash key
 KEYS[2] - defers input hash key
 KEYS[3] - run metadata hash key
 ARGV[1] - hashedID
-ARGV[2] - meta JSON ({FnSlug, HashedID, ScheduleStatus} only)
+ARGV[2] - meta JSON ({FnSlug, HashedID, UserlandID, ScheduleStatus} only)
```
amh4r: Why do we need the userland ID in Redis?

pkg/util/ulid.go
```diff
@@ -25,3 +25,33 @@ func DeterministicULID(ts time.Time, seed []byte) (ulid.ULID, error) {
...
+	id, _ := DeterministicULID(ulid.Time(parent.Time()), []byte(parent.String()+hashedID))
+	return id
+}
+
+// DeterministicDeferSpanSeed returns the seed for the executor.defer span.
+// Determinism keeps the span ID stable across finalize retries so resolvers
+// don't see duplicates.
+func DeterministicDeferSpanSeed(parent ulid.ULID, hashedID string) []byte {
+	return []byte(parent.String() + hashedID + "s")
+}
```
amh4r: For `DeterministicChildRunID`, why do we need a deterministic child run ID?

Why do we need `DeterministicDeferEventID` and `DeterministicDeferSpanSeed`? Why can't we just used `DeterministicULID `directly in the business logic?

Also, this is a general ULID file so really specific business logic things like this don't belong in here

pkg/cqrs/manager/cqrs.go
```diff
@@ -2105,6 +2112,38 @@ type runsQueryBuilder struct {
...
+
+	key := meta.Attrs.DeferParentRunID.Key()
+	var jsonPred sq.Expression
+	switch dialect {
+	case "postgres":
+		jsonPred = sq.L("jsonb_exists(?, ?)", sq.I("dt.attributes"), key)
+	case "sqlite3":
+		jsonPred = sq.L(`json_extract(?, '$."`+key+`"') IS NOT NULL`, sq.I("dt.attributes"))
+	default:
+		panic(fmt.Sprintf("runTypeFilterExpr: unsupported dialect %q", dialect))
```
amh4r: Can we return an error instead of panicking? Intentionally panicking should be rare in Go

pkg/cqrs/manager/run_linkage.go
```diff
@@ -0,0 +1,303 @@
...
+  )) AS span_fragments
+FROM spans
+WHERE dynamic_span_id IN (
+  SELECT dynamic_span_id
+  FROM spans
+  WHERE name = ` + nameArg + ` AND ` + jsonPath + ` IN (` + inList + `)
+)
+GROUP BY run_id, trace_id, dynamic_span_id, parent_span_id
+ORDER BY run_id, start_time
+`
```
amh4r: Is it note possible to put this query in a SQL file and codegen? Also, I'm worried about SQL injection risk

pkg/coreapi/graph/loaders/defer.go
```diff
@@ -0,0 +1,65 @@
...
+			if r != nil {
+				continue
+			}
+			results[i] = &dataloader.Result{Error: err}
+		}
+		return results
+	}
+
+	for _, p := range parsedKeys {
+		results[p.index] = &dataloader.Result{Data: byRunID[p.runID]}
```
amh4r, line 58: `byRunID[p.runID]` could return a zero value. We should check each key's existence, and set an error in `results` if it doesn't exist

pkg/coreapi/graph/resolvers/function_run_v2.resolver.go
```diff
@@ -25,6 +25,89 @@ func (r *functionRunV2Resolver) Function(ctx context.Context, fn *models.Functio
...
+		return nil, fmt.Errorf("error converting deferred-from parent run: %w", err)
+	}
+
+	return &models.RunDeferredFrom{
+		ParentRunID: df.ParentRunID,
+		ParentRun:   parentV2,
+	}, nil
+}
+
+func (r *functionRunV2Resolver) InvokedFrom(ctx context.Context, fn *models.FunctionRunV2) (*models.RunInvokedFrom, error) {
```
amh4r: Why do we need `InvokedFrom`? We're only implementing deferred runs right now

pkg/coreapi/graph/resolvers/function_run_v2.resolver.go
```diff
@@ -25,6 +25,89 @@ func (r *functionRunV2Resolver) Function(ctx context.Context, fn *models.Functio
...
+	}
+
+	return &models.RunInvokedFrom{
+		ParentRunID: rif.ParentRunID,
+		ParentRun:   parentV2,
+		StepName:    rif.StepName,
+	}, nil
+}
+
+func (r *functionRunV2Resolver) RunType(ctx context.Context, fn *models.FunctionRunV2) (models.RunType, error) {
```
amh4r: Does this need to be a resolver? We're loading just to check the type, which seems excessive.

We're using a dataloader, which should usually dedupe calls. But if there's a way to easily avoid the call then we should prob do it

pkg/tracing/meta/attributes.go
```diff
@@ -40,6 +40,13 @@ var Attrs = struct {
 	IsDurableEndpointRun         attr[*bool]
 	DurableEndpointModeChangedAt attr[*time.Time]
 
+	// Defer attributes
+	DeferFnSlug      attr[*string]
+	DeferHashedID    attr[*string]
+	DeferParentRunID attr[*ulid.ULID]
```
amh4r: How will this work when we eventually add batching?

pkg/coreapi/graph/resolvers/function_run_v2.resolver.go
Linell: Resolved here: https://github.com/inngest/inngest/pull/4177/changes/4b3849abe1ffc78b135c7af417bdb3744924f59f

pkg/coreapi/graph/loaders/defer.go
Linell: A missing key is the "no linkages" case. It doesn't seem right to add an error since it'll _usually_ be empty, right?

pkg/cqrs/manager/run_linkage.go
Linell: I was having a bunch of trouble with the engine mishandling `sqlc.slice` when the `IN` operand's left side is a function call like `json_extract`. The run IDs we filter on live in the JSON attributes column. Theoretically the only interpolated token is `attrKey` and the run IDs are bound params, but I agree this is a big code stank.

Linell:
Closing in favor of https://github.com/inngest/inngest/pull/4290

PR #4182, opened 2026-05-13
fix: rerun from step

jacobheric:
## Description

* updates reconstruct to use new traces and gets them working for both oss and monorepo
* updates rerun from step gql op to pass env
Screenshot 2026-05-15 at 11 35 58 AM
Screenshot 2026-05-15 at 11 35 33 AM

## Release note

Gets rerun from step, including with the ability to update step inputs, working in cloud. This already works in the oss dev server.

## Migration note

None.

---

> [!NOTE]
> Rewrites `reconstruct` to use the new trace-v2 span tree API (`GetSpansByRunID`) instead of the legacy flat-span approach, enabling rerun-from-step to work with the new monorepo trace format. Also adds `envID` to the `Rerun` GraphQL mutation.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 2275bbecf2b291864a352c57fb463c231b994aa2.</sup>


pkg/execution/executor/reconstruct.go
```diff
@@ -4,106 +4,71 @@ import (
...
+			return fmt.Errorf("step found in stack but output not found in original run")
+		}
+
+		var outputIdentifier cqrs.SpanIdentifier
+		if err := outputIdentifier.Decode(*outputID); err != nil {
+			return fmt.Errorf("error decoding span output ID: %w", err)
+		}
+		if outputIdentifier.Preview == nil || !*outputIdentifier.Preview {
+			return fmt.Errorf("span output is not trace-v2 output")
+		}
```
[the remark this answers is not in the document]
jacobheric: legacy traces are dead. we are ripping the last vestiges out now.

jacobheric:
> **Needs attention** — 1 issue in 1 file
> 
> The migration logic is sound and well-tested. One hard block: runs whose output spans were recorded before trace-v2 will now fail with an opaque error instead of being handled — the new `Preview` guard is stricter than the old `LegacyGetSpanOutput` path it replaces. Confirm this is intentional (i.e. rerun-from-step is only supported for trace-v2 runs) or add a fallback.
> 
> Prompt for AI agents (all issues)
> ```
> Check if these issues are valid — if so, understand the root cause of each and fix them.
> 
> <assessment>
> The migration logic is sound and well-tested. One hard block: runs whose output spans were recorded before trace-v2 will now fail with an opaque error instead of being handled — the new `Preview` guard is stricter than the old `LegacyGetSpanOutput` path it replaces. Confirm this is intentional (i.e. rerun-from-step is only supported for trace-v2 runs) or add a fallback.
> </assessment>
> 
> <file name="pkg/execution/executor/reconstruct.go">
> <issue location="pkg/execution/executor/reconstruct.go:67">
> Silently breaks rerun-from-step for any run recorded before trace-v2. The old code used `LegacyGetSpanOutput` which had no such guard; this replacement returns an error for all pre-trace-v2 output IDs. If rerun-from-step is intentionally trace-v2-only, document that constraint; otherwise fall back to the legacy path.
> </issue>
> </file>
> ```
> 
> Tag @mendral-app with feedback or questions. [View session](https://app.mendral.com/sessions/github:inngest:inngest:pr:4182:01KRNRZNDW1G1QRKCW288DQVKP)

we're not supporting v1 traces for rerun!

pkg/execution/executor/reconstruct.go
```diff
@@ -135,3 +91,93 @@ func reconstruct(ctx context.Context, tr cqrs.TraceReader, req execution.Schedul
...
+
+		next := reconstructStep{
+			id:       stepID,
+			at:       span.StartTime,
+			stepSpan: span,
+		}
+
+		//
+		// Retries can produce more than one span for a step ID; use the newest one.
```
rhino1998: Can this use the one with the greatest attempt number instead of the one we iterate through last? (In theory it should be the same, but attempt number is a bit safer)

pkg/execution/executor/reconstruct.go
```diff
@@ -135,3 +91,93 @@ func reconstruct(ctx context.Context, tr cqrs.TraceReader, req execution.Schedul
...
+	}
+
+	sort.SliceStable(steps, func(i, j int) bool {
+		return reconstructStepLess(steps[i], steps[j])
+	})
+
+	return steps, foundStepToRunFrom
+}
+
+func reconstructStepLess(a, b reconstructStep) bool {
```
rhino1998, line 158: Same as above, can this compare by attempt number?

pkg/execution/executor/reconstruct.go
jacobheric: done

pkg/execution/executor/reconstruct.go
jacobheric: done

rhino1998 (approved):

PR #4197, opened 2026-05-14
feat: agentic api cli

jacobheric:
## Description

```
npx inngest-cli@latest alpha api --help
NAME:
   inngest alpha api - Call Inngest REST API v2 endpoints

USAGE:
   inngest alpha api [target/auth flags] <endpoint> [endpoint flags]

DESCRIPTION:
   By default, the command targets the local dev server.
   Set --prod to target Inngest Cloud Production, or --api-host/--api-port to target a custom API server.

COMMANDS:
   health                    GET /health
   create-env                POST /envs
   get-account               GET /account
   get-account-envs          GET /envs
   get-account-event-keys    GET /keys/events
   get-account-signing-keys  GET /keys/signing
   create-webhook            POST /env/webhooks
   get-webhooks              GET /env/webhooks
   patch-env                 PATCH /envs/{id}
   get-function-run          GET /runs/{run_id}
   sync-app                  POST /apps/{app_id}/syncs
   get-function-trace        GET /runs/{run_id}/trace
   invoke-function           POST /apps/{app_id}/functions/{function_id}/invoke

OPTIONS:
   --help, -h  show help

   Auth

   --api-key string      API key sent as a Bearer token [$INNGEST_API_KEY]
   --env string          Environment name sent as X-Inngest-Env [$INNGEST_ENV]
   --signing-key string  Signing key sent as a Bearer token [$INNGEST_SIGNING_KEY]

   Output

   --raw  Print the response body without JSON formatting (default: false)

   Target

   --api-host string   Custom API host or origin                                                              --api-port int      Custom API port (default: default local dev server port)
   --config string     Path to an Inngest configuration file
   --prod              Target Inngest Cloud Production API unless --api-host or --api-port is set (default: false)
   --timeout duration  HTTP request timeout (default: 30s)
```

```
npx inngest-cli@latest alpha api get-function-run --run-id 01KRKKJB7A1HJCQAW3XNJZ6927
{
  "data": {
    "app": {
      "id": "example-id"
    },
    "durationMs": "1279",
    "endedAt": "2026-05-14T16:02:11.125Z",
    "function": {
      "id": "example-id,
      "name": "retry-failed-ci"
    },
    "id": "fake-id",
    "queuedAt": "2026-05-14T16:01:59.018Z",
    "startedAt": "2026-05-14T16:02:09.846Z",
    "status": "COMPLETED",
    "trigger": {
      "eventIds": [
        "fake-id"
      ]
    }
  },
  "metadata": {
    "fetchedAt": "2026-05-14T18:30:59.687934221Z"
  }
}
```

## Release note

Adds alpha support for an agentic cli that can be used to call any of the Inngest V2 REST APIs: https://api-docs.inngest.com. See `npx inngest-cli@latest alpha api --help`.

## Migration note

None.

---

> [!NOTE]
> Adds an `inngest alpha api` CLI command that auto-discovers REST API v2 endpoints from the protobuf service descriptor and generates subcommands with typed flags. Supports targeting local dev server, self-hosted, or Inngest Cloud, with auth via API key or signing key.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a44aee4a47ecc728e3590578122bb227e47b0092.</sup>


darwin67:
why not just use the grpc endpoint directly instead of having to construct the REST payload?
you already have the grpc client structs available, so seems kinda redundant?

cmd/internal/config/config.go
```diff
@@ -23,6 +23,9 @@ type Config struct {
 	NoDiscovery *bool    `koanf:"no-discovery"`
 	NoPoll      *bool    `koanf:"no-poll"`
 	Host        string   `koanf:"host"`
+	APIHost     string   `koanf:"api-host"`
+	APIPort     int      `koanf:"api-port"`
```
djfarrelly, line 27: Is there a benefit to having "API host" and "API Port" but also "host" and "port"?

cmd/internal/config/config.go
```diff
@@ -23,6 +23,9 @@ type Config struct {
...
+	APICloud    bool     `koanf:"api-cloud"`
```
djfarrelly: I wonder if we should make this a `--target prod`/`--target dev` (default, doesn't assume it requires an auth key type) or `--prod` or `--cloud` flag if it follows similar other conventions for CLIs. Then host/port are used if connecting to a non-dev and also custom.

One question for this is if someone is using self-hosted and maybe they're connecting to a local server or kube port-forward, should the prod/cloud or custom host method always assume HTTPS? should that be true but a user needs to opt out to non https if not dev sever/localhost?

djfarrelly (commented):
This is awesome - it looks super clean. I have some open questions, but I think these are more about design, not totally implementation:

High level immediate thoughts:
* It would be nice to have consistency across verbs used (e.g. fetch-env vs. get-function-run ) - I'm assuming this is pulled right from the gRPC proto, so it might just be a lower level thing.
* A mechanism to hide certain endpoints from the CLI for now (like we could hide some of the partner apis like create-partner-account and create-env (this should require an account-level token and our api keys are currently just env-level. Could keep it tight for most agentic use cases for now - we can always un-hide them later. This could be a x- flag in the protofile (similar to the x-beta tag we use for docs).

cmd/internal/config/config.go
jacobheric: I was thinking of supporting self hosters who might use INNGEST_HOST & INNGEST_PORT already while also giving people who don't a clear set of env vars to set. I can just do api host and api port though for simplicity.

cmd/internal/config/config.go
jacobheric: I do enforce https for non-local calls. I have api-cloud to force production saas without them having to know our address. I avoided `prod` because of the ambiguity between self-hosters and saas users. 

Basically:
* if no env vars/env, use local dev server if it's there (we probe it).
* api-cloud always means saas production
* api-host/api-port overrides everything
* host/port is used only if the api-* ones aren't set

Happy to make any changes to this interface setup here. My goal was to make it work by default for the typical setups already out there (local dev, self-hoster) and then throw in the minimal amount of self-explanatory args/vars for the other cases.

jacobheric:
> This is awesome - it looks super clean. I have some open questions, but I think these are more about design, not totally implementation:
> 
> High level immediate thoughts:
> 
>     * It would be nice to have consistency across verbs used (e.g. fetch-env vs. get-function-run ) - I'm assuming this is pulled right from the gRPC proto, so it might just be a lower level thing.
> 
>     * A mechanism to hide certain endpoints from the CLI for now (like we could hide some of the partner apis like create-partner-account and create-env (this should require an account-level token and our api keys are currently just env-level. Could keep it tight for most agentic use cases for now - we can always un-hide them later. This could be a x- flag in the protofile (similar to the x-beta tag we use for docs).

Yeah names are generated from the protos, but I can see if we can normalize thing in the gen phase a bit. I can also add some blacklist mechanism to hide partner apis, etc. 

Ok, I added some very basic blacklisting and verb normalization, see:

```
   health                    GET /health
   create-env                POST /envs
   get-account               GET /account
   get-account-envs          GET /envs
   get-account-event-keys    GET /keys/events
   get-account-signing-keys  GET /keys/signing
   create-webhook            POST /env/webhooks
   get-webhooks              GET /env/webhooks
   patch-env                 PATCH /envs/{id}
   get-function-run          GET /runs/{run_id}
   sync-app                  POST /apps/{app_id}/syncs
   get-function-trace        GET /runs/{run_id}/trace
   invoke-function           POST /apps/{app_id}/functions/{function_id}/invoke
```

jacobheric:
> why not just use the grpc endpoint directly instead of having to construct the REST payload? you already have the grpc client structs available, so seems kinda redundant?

good point, I'll try it!

Looks like we don't have have grpc listeners set up. I had a go at setting them up on both sides and testing them out. Not bad but also not a super tiny lift. Will also require infra changes for cloud. You still think we should go this route?

https://github.com/inngest/monorepo/pull/7193
https://github.com/inngest/inngest/pull/4207

cmd/internal/config/config.go
jacobheric: Ok, I cleaned this up and simplified. It connects to dev by default. You can connect to prod with --prod. And self hosters can use --api-host/port. That's it. No more host/port since those are for dev server. No more magical fallback connecting behavior.

cmd/internal/config/config.go
jacobheric: Cleaned up and simplified. See: https://github.com/inngest/inngest/pull/4197#discussion_r3250264884

djfarrelly (approved):

PR #4198, opened 2026-05-14
refactor(queue): scope debounce and singleton operations

BrunoScheufler:
## Description

Adds an explicit queue `Scope` for debounce and singleton storage operations. The scope carries system state plus account, environment, and function IDs so future queue storage implementations can namespace debounce/singleton content under the tenant and function it belongs to.

This updates the debounce and singleton operation interfaces, Redis-backed implementations, callers, CLI debug commands, and tests to pass scope through without changing the current Redis key layout. It also updates the debug API protobufs so clients provide account/environment/function scope instead of the server assuming devserver IDs.

## Release note

Debug API debounce and singleton requests now include account and environment IDs so scoped queue storage can target the correct tenant/function namespace.

## Migration note

Debug API clients for debounce and singleton operations must send `account_id`, `env_id`, and `function_id` on scoped requests. The debug CLI now requires account/environment flags for these commands. Existing Redis storage behavior is unchanged.

---

> [!NOTE]
> Adds an explicit `queue.Scope` (account, env, function IDs) to debounce and singleton debug operations. The new `debugflags/scope.go` centralises flag definitions and UUID validation for the debug CLI, and `pkg/debugapi/scope.go` provides a `debugScope()` helper used by all debug API handlers. Proto requests now require account/environment/function IDs instead of the server assuming dev-server constants.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit fefee22dfbf78686a740c94c42b333f2cd544706.</sup>

### DIFF
--- a/pkg/debugapi/debounce.go
+++ b/pkg/debugapi/debounce.go
@@ -8,8 +8,9 @@ import (
 	"github.com/google/uuid"
 	"github.com/inngest/inngest/pkg/consts"
 	"github.com/inngest/inngest/pkg/execution/debounce"
-	"github.com/oklog/ulid/v2"
+	"github.com/inngest/inngest/pkg/execution/queue"
 	pb "github.com/inngest/inngest/proto/gen/debug/v1"
+	"github.com/oklog/ulid/v2"
 )
 
 // GetDebounceInfo retrieves the currently debounced event for a function and debounce key.
@@ -23,8 +24,14 @@ func (d *debugAPI) GetDebounceInfo(ctx context.Context, req *pb.DebounceInfoRequ
 		return nil, fmt.Errorf("invalid function_id: %w", err)
 	}
 
+	scope := queue.Scope{
+		AccountID:  consts.DevServerAccountID,
+		EnvID:      consts.DevServerEnvID,
+		FunctionID: fnID,
+	}
+
 	// Use the debouncer to get debounce info
-	info, err := d.debouncer.GetDebounceInfo(ctx, fnID, req.GetDebounceKey())
+	info, err := d.debouncer.GetDebounceInfo(ctx, scope, req.GetDebounceKey())
 	if err != nil {
 		return nil, fmt.Errorf("failed to get debounce info: %w", err)
 	}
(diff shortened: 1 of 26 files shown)


pkg/execution/queue/interface.go
```diff
@@ -201,116 +209,82 @@ const (
...
 
 	// DebounceGetItem retrieves the serialized debounce item from the
 	// hash. Returns ErrDebounceNotFound when absent.
-	DebounceGetItem(ctx context.Context, debounceID ulid.ULID) ([]byte, error)
+	DebounceGetItem(ctx context.Context, scope Scope, debounceID ulid.ULID) ([]byte, error)
 
 	// DebounceDeleteItems removes one or more debounce items from the
 	// hash. A no-op on an empty list.
-	DebounceDeleteItems(ctx context.Context, debounceIDs ...ulid.ULID) error
+	DebounceDeleteItems(ctx context.Context, scope Scope, debounceIDs ...ulid.ULID) error
```
lkasinathan, line 240: methods that directly take debounceID don't seem to use scope (and likely won't in fdb as well?) - if that's the case, consider removing it from the interface to keep it clean.

I noticed because debugapi calls these methods with functionID missing from scope.

lkasinathan (approved):
looks good! left a comment about some of the DebounceOperations methods potentially not needing scope.

pkg/execution/queue/interface.go
BrunoScheufler: Right now the implementations don't use the scope but we want all operations to get one for FDB since that allows us to properly namespace data without a global lookup index. I'm updating the remaining callers to pass in all IDs all the time. Redis won't namespace for now, but FDB can do that from day one.

pkg/debugapi/debounce.go
```diff
@@ -101,7 +112,10 @@ func (d *debugAPI) DeleteDebounceByID(ctx context.Context, req *pb.DeleteDebounc
 		parsed[i] = u
 	}
 
-	err := d.debouncer.DeleteDebounceByID(ctx, parsed...)
+	err := d.debouncer.DeleteDebounceByID(ctx, queue.Scope{
+		AccountID: consts.DevServerAccountID,
+		EnvID:     consts.DevServerEnvID,
```
lkasinathan: functionID missing, this is fine now because DeleteDebounceByID doesnt use scope. but good to keep consistent?

PR #4201, opened 2026-05-14
feat: EXE-1547 - Accept StepPlanned upon beginning to run a step

scottnuma:
## Description
* Users running checkpointing sometimes have long running steps. Currently in the dashboard, it is difficult to determine that the step is in progress rather than hanging.
* We do have the Execution > Attempt 0 row in the execution viewer, but this is separated from where on-going step will eventually live, especially in longer and more complex workflows. Additionally, this eventually disappears once the run finishes
* We would like to show users that we are working on their step
* On the SDK side, we https://github.com/inngest/inngest-js/pull/1530
* In the executor in this PR,  upon receiving this `StepPlanned` opcode, we persist a `Running` span with the step name and start time. Critically, we will set the dynamic span ID to match the incoming `StepRun` opcode that will conclude the step.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/enums/opcode.go
+++ b/pkg/enums/opcode.go
@@ -35,6 +35,7 @@ const (
 var opcodeSyncMap = map[Opcode]struct{}{
 	OpcodeStep:            {},
 	OpcodeStepRun:         {},
+	OpcodeStepPlanned:     {},
 	OpcodeRunComplete:     {},
 	OpcodeSyncRunComplete: {},
 	OpcodeStepFailed:      {},
--- a/pkg/execution/checkpoint/checkpoint.go
+++ b/pkg/execution/checkpoint/checkpoint.go
@@ -216,7 +216,7 @@ func (c checkpointer) CheckpointSyncSteps(ctx context.Context, input SyncCheckpo
 				meta.SpanNameStep,
 				&tracing.CreateSpanOptions{
 					Debug:      &tracing.SpanDebugData{Location: "checkpoint.SyncStep"},
-					Seed:       []byte(fmt.Sprintf("%s:%d", op.ID, runCtx.AttemptCount())),
+					Seed:       stepDynamicSeed(op, runCtx.AttemptCount()),
 					Parent:     tracing.RunSpanRefFromMetadata(input.Metadata),
 					StartTime:  op.Timing.Start(),
 					EndTime:    op.Timing.End(),
@@ -254,7 +254,7 @@ func (c checkpointer) CheckpointSyncSteps(ctx context.Context, input SyncCheckpo
 				meta.SpanNameStep,
 				&tracing.CreateSpanOptions{
 					Debug:      &tracing.SpanDebugData{Location: "checkpoint.SyncErr"},
-					Seed:       []byte(fmt.Sprintf("%s:%d", op.ID, runCtx.AttemptCount())),
+					Seed:       stepDynamicSeed(op, runCtx.AttemptCount()),
 					Parent:     tracing.RunSpanRefFromMetadata(input.Metadata),
 					StartTime:  op.Timing.Start(),
 					EndTime:    op.Timing.End(),
@@ -494,7 +494,7 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 				meta.SpanNameStep,
 				&tracing.CreateSpanOptions{
 					Debug:      &tracing.SpanDebugData{Location: "checkpoint.AsyncStep"},
-					Seed:       []byte(fmt.Sprintf("%s:%d", op.ID, 0)),
+					Seed:       stepDynamicSeed(op, 0),
 					Parent:     tracing.RunSpanRefFromMetadata(&md),
 					StartTime:  op.Timing.Start(),
 					EndTime:    op.Timing.End(),
--- a/pkg/execution/checkpoint/checkpoint.go
+++ b/pkg/execution/checkpoint/checkpoint.go
@@ -508,6 +508,33 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 
 			c.processMetadata(ctx, l, input.AccountID, &md, stepSpanRef, op, "checkpoint.AsyncStep.metadata")
 
+		case enums.OpcodeStepPlanned:
+			// When the SDK announces a step is about to run, we open a Running
+			// executor.step span to show progress to the user.
+			_, err := c.TracerProvider.CreateSpan(
+				tracing.WithExecutionContext(ctx, tracing.ExecutionContext{
+					Identifier: md.ID,
+					Attempt:    0,
+				}),
+				meta.SpanNameStep,
+				&tracing.CreateSpanOptions{
+					Debug: &tracing.SpanDebugData{Location: "checkpoint.AsyncStepPlanned"},
+
+					// Set the same dynamic span ID as the eventual completion arm.
+					// We use DynamicSpanIDOverride instead of Seed to avoid setting the same
+					// span ID.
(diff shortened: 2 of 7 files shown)


pkg/enums/opcode.go
```diff
@@ -35,6 +35,7 @@ const (
 var opcodeSyncMap = map[Opcode]struct{}{
 	OpcodeStep:            {},
 	OpcodeStepRun:         {},
+	OpcodeStepPlanned:     {},
```
scottnuma, line 38: Feels a little strange, but I guess this ticket's change moves StepPlanned from async to both sync and async.

An alternative is to create a new opcode, but planned felt pretty aligned with this use case.

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -508,6 +508,33 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
...
+					Identifier: md.ID,
+					Attempt:    0,
+				}),
+				meta.SpanNameStep,
+				&tracing.CreateSpanOptions{
+					Debug: &tracing.SpanDebugData{Location: "checkpoint.AsyncStepPlanned"},
+
+					// Set the same dynamic span ID as the eventual completion arm.
+					// We use DynamicSpanIDOverride instead of Seed to avoid setting identical
+					// the same span ID.
```
rhino1998:
```suggestion
					// We use DynamicSpanIDOverride instead of Seed to avoid setting
					// the same span ID.
```

pkg/execution/checkpoint/util.go
```diff
@@ -22,6 +30,18 @@ func stepRunAttrs(attrs *meta.SerializableAttrs, op state.GeneratorOpcode, runID
...
 }
 
+func stepPlannedAttrs(attrs *meta.SerializableAttrs, op state.GeneratorOpcode, runID ulid.ULID) *meta.SerializableAttrs {
+	return attrs.Merge(
+		meta.NewAttrSet(
+			meta.Attr(meta.Attrs.StepName, inngestgo.Ptr(op.UserDefinedName())),
+			meta.Attr(meta.Attrs.RunID, &runID),
+			meta.Attr(meta.Attrs.QueuedAt, inngestgo.Ptr(op.Timing.Start())),
+			meta.Attr(meta.Attrs.StartedAt, inngestgo.Ptr(op.Timing.Start())),
+			meta.Attr(meta.Attrs.DynamicStatus, inngestgo.Ptr(enums.StepStatusRunning)),
```
rhino1998, line 38: If there is a reasonable way to plumb the `_inngest.step.type` attribute here for the "planned" step, that could be nice as then we'd get partially run step states in Insights. I think a span like this would just work™️with the way the Insights MVs are structured currently if we have the step ID and step type.

pkg/execution/checkpoint/util.go
```diff
@@ -22,6 +30,18 @@ func stepRunAttrs(attrs *meta.SerializableAttrs, op state.GeneratorOpcode, runID
 	)
 }
 
+func stepPlannedAttrs(attrs *meta.SerializableAttrs, op state.GeneratorOpcode, runID ulid.ULID) *meta.SerializableAttrs {
+	return attrs.Merge(
+		meta.NewAttrSet(
+			meta.Attr(meta.Attrs.StepName, inngestgo.Ptr(op.UserDefinedName())),
```
rhino1998: ~~Can we also plumb the userland step ID (the step name slug the user provides) and the hashed step ID too for completeness/parity with completed step attributes?~~

rhino1998: Actually, it looks like this is merged with the output of GeneratorAttrs in which case the StepName attribute should already be set. Also I think the RunID attribute is already set in `execution_processor.go`

scottnuma: Removed both and confirmed that the span still retains those attributes

pkg/execution/checkpoint/util.go
scottnuma: I've adjusted the SDK over in https://github.com/inngest/inngest-js/pull/1530 to always send over the step.type with the last two commits. We can rely on this for the StepPlanned opcode since we're introducing it at the same time.

The latest commit here lets us pass it through all the way to the span.

scottnuma:
With those changes, we get a incoming arm span with these attributes:

```
  ┌──────────────────────────────┬──────────────────────────────────────────┐
  │             Key              │                  Value                   │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.account.id          │ 00000000-0000-4000-a000-000000000000     │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.app.id              │ b238544c-989a-56de-b9df-617ef377bd36     │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.dynamic.span.id     │ b66e1452a2e4a15b                         │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.dynamic.status      │ Running                                  │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.env.id              │ 00000000-0000-4000-b000-000000000000     │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.function.id         │ e7038ef7-f6a1-5a90-9a92-52937aef9a1f     │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.internal.location   │ checkpoint.AsyncStepPlanned              │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.queued_at           │ 1778886315136                            │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.run.id              │ 01KRPY632X22ZB3NESG7R64THN               │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.started_at          │ 1778886315136                            │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.attempt        │ 0                                        │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.id             │ 117ec720ea4b96702ba69fad917ef776ef5efe4e │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.name           │ uppercase                                │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.op             │ StepPlanned                              │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.type           │ run                                      │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.userland.id    │ uppercase                                │
  ├──────────────────────────────┼──────────────────────────────────────────┤
  │ _inngest.step.userland.index │ 0                                        │
  └──────────────────────────────┴──────────────────────────────────────────┘
```

rhino1998 (approved):

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -508,6 +508,33 @@
...
+					DynamicSpanIDOverride: tracing.DeterministicSpanConfig(stepDynamicSeed(op, 0)).SpanID.String(),
+					Parent:                tracing.RunSpanRefFromMetadata(&md),
+					StartTime:             op.Timing.Start(),
+					Attributes:            stepPlannedAttrs(attrs, op, input.RunID),
+				},
+			)
+			if err != nil {
+				// Processing the leading-edge is best effort both in the executor and SDK.
+				// We'll eventually get the completion arm even if this fails.
+				l.Warn("error saving leading-edge span for StepPlanned", "error", err, "step_id", op.ID)
```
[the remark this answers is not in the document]
scottnuma: Addressed this by sanitizing the user input.

PR #4207, opened 2026-05-15
internal: agentic cli grpc

jacobheric:
## Description

Add grpc listener to oss and use that directly instead of rest. Per: https://github.com/inngest/inngest/pull/4197#issuecomment-4453761112

```
NAME:
   inngest alpha api - Call Inngest API v2 endpoints over gRPC

USAGE:
   inngest alpha api [target/auth flags] <endpoint> [endpoint flags]

DESCRIPTION:
   By default, the command targets the local dev server's gRPC endpoint.
   Set --prod to target Inngest Cloud Production, or --api-host/--api-port to target a custom server.

COMMANDS:
   health                    GET /health
   create-env                POST /envs
   get-account               GET /account
   get-account-envs          GET /envs
   get-account-event-keys    GET /keys/events
   get-account-signing-keys  GET /keys/signing
   create-webhook            POST /env/webhooks
   get-webhooks              GET /env/webhooks
   patch-env                 PATCH /envs/{id}
   get-function-run          GET /runs/{run_id}
   sync-app                  POST /apps/{app_id}/syncs
   get-function-trace        GET /runs/{run_id}/trace
   invoke-function           POST /apps/{app_id}/functions/{function_id}/invoke

OPTIONS:
   --help, -h  show help

   Auth

   --api-key string      API key sent as a Bearer token [$INNGEST_API_KEY]
   --env string          Environment name sent as X-Inngest-Env [$INNGEST_ENV]
   --signing-key string  Signing key sent as a Bearer token [$INNGEST_SIGNING_KEY]

   Output

   --raw  Print the response as compact JSON (no pretty-printing) (default: false)

   Target

   --api-host string   Custom host (or host:port) for the gRPC endpoint
   --api-port int      Custom API v2 gRPC port (default: default API v2 gRPC port)
   --config string     Path to an Inngest configuration file
   --prod              Target Inngest Cloud Production unless --api-host or --api-port is set (default: false)
   --timeout duration  RPC call timeout (default: 30s)
```

## Release note

None.

## Migration note

None.


djfarrelly (commented):
One thing to note is that since this is an new port, we'll need to also document how self-hosters must configure the grpc port to make use of this CLI

PR #4225, opened 2026-05-19
refactor: score name moves into the kind suffix

AndyInternet:
## Description

Score metadata used to be one kind (`inngest.score`) with the user-supplied name as a key inside the values map. The name had to match `^[a-zA-Z_][a-zA-Z0-9_]{0,63}$`, which rejected things like `click-through rate` or `accuracy-rate`.

Now scores follow the `userland.<name>` pattern: kind is `inngest.score.<name>`, values is `{value: <number|bool>}`. ClickHouse already strips the `inngest.` prefix at ingest, so the Map storage shape is unchanged.

Score-name regex is gone. Bare `inngest.score` is no longer accepted (feature not released yet, so no back-compat).

## Release note

None.

## Migration note

None.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ gql: ## Generate GraphQL code
 .PHONY: tygo
 tygo: ## Generate TypeScript types from Go structs
 	go run github.com/gzuidhof/tygo@latest generate
-	cd ui && pnpx prettier@2.8.8 --write --no-config --single-quote "packages/components/src/generated/**/*.ts"
+	cd ui && pnpx prettier@2.8.8 --write --no-config --single-quote --print-width 100 "packages/components/src/generated/**/*.ts"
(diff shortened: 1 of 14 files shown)


jakobevangelista (approved):
lgtm

one thing to point out is since we're changing to `inngest.score.<name>`, `CreateMetadataSpanFromValues` still uses `kind.String()` as the `metadata_spans_total` `kind` tag, and only collapses userland kinds via `kind.IsUser()` to `userland.*` 

in `pkg/tracing/metadata.go`, since the score suffix is now defined by the user, should we also collapse scoped score kinds there to something like `inngest.score.*` / `inngest.score`

it's a nit for metric tracking as now the cardinality shoots up for tracking all `inngest.score.*` things, just being wary about it

PR #4228, opened 2026-05-19
fix(pauses): Retry enqueueing edge after consuming pause

lkasinathan:
## Description

Improves pause resume reliability by adding retry logic around enqueueing the next queue item after a pause is consumed.

In this [ticket](https://app.plain.com/workspace/w_01H5R1F69XQ35G4P7THFQGN5KB/thread/th_01KQSD1WR8FNAADPVEZD2QV8CR),  we see that a pause was consumed successfully, but the subsequent Enqueue of next step failed. 

When the pause job was retried, since the pause was already consumed, we early exit and declare success. Since the next step was not enqueued the user fn run continued to hang indefinitely.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -3225,9 +3225,13 @@ func (e *executor) Resume(ctx context.Context, pause state.Pause, r execution.Re
 				e.log.Debug("error creating span for next step after resume", "error", err)
 			}
 
-			err = e.queue.Enqueue(ctx, nextItem, e.now(), queue.EnqueueOpts{})
+			_, err = util.WithRetry(ctx, "executor.resume.enqueue", func(ctx context.Context) (any, error) {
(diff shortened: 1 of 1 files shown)


darwin67 (approved):
would be nice to have a regression test.

KiKoS0 (approved):

pkg/execution/executor/executor.go
lkasinathan: The fix below is a bandaid that tries to enqueue inline multiple times.

Another fix is to continue this handler even if DidConsume is false, and rely on queue item idempotency to avoid duplicate enqueues.

lkasinathan: from @KiKoS0: 
> maybe we can just drop the spans when DidConsume is false so that we keep tracing consistent with the state store though?

KiKoS0 (approved):

lkasinathan:
> would be nice to have a regression test.

unfortunately, this is not easily testable because of transient delays and races between pause timeout and wait-for-event firing.

PR #4235, opened 2026-05-20
internal: wrap getParentSpan in a retry

amh4r:
## Description
Wrap `getParentSpan()` call with a retry in the metadata update API endpoint. This is a short term fix to handle eventual consistency in ClickHouse.

Also include a histogram metric and sampled error logging

### DIFF
--- a/pkg/api/apiv1/metadata.go
+++ b/pkg/api/apiv1/metadata.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/google/uuid"
@@ -17,6 +18,7 @@ import (
 	statev2 "github.com/inngest/inngest/pkg/execution/state/v2"
 	"github.com/inngest/inngest/pkg/logger"
 	"github.com/inngest/inngest/pkg/publicerr"
+	"github.com/inngest/inngest/pkg/telemetry/metrics"
 	"github.com/inngest/inngest/pkg/tracing"
 	"github.com/inngest/inngest/pkg/tracing/meta"
 	"github.com/inngest/inngest/pkg/tracing/metadata"
@@ -118,10 +120,43 @@ type AddRunMetadataRequest struct {
 }
 
 func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID ulid.ULID, req *AddRunMetadataRequest) error {
-	parentSpan, scope, err := a.getParentSpan(ctx, auth, runID, &req.Target)
+	var parentSpan *cqrs.OtelSpan
+	var scope metadata.Scope
+	var err error
+	var attempts int
+	start := time.Now()
+
+	// This retry only exists because of eventual consistency in ClickHouse
+	// data. There's a race condition where the parent span may not be queryable
+	// when the metadata update arrives.
+	//
+	// There's also a related race where a successful retry span isn't queryable
+	// yet, which causes us to mistakenly update metadata on a prior failed
+	// attempt.
+	//
+	// TODO: We should replace this hack with a proper fix. But a proper fix
+	// likely requires changes in our ClickHouse trace schema.
+	_, err = util.WithRetry(
+		ctx,
+		"apiv1.AddRunMetadata.getParentSpan",
+		func(ctx context.Context) (any, error) {
+			attempts++
+			parentSpan, scope, err = a.getParentSpan(ctx, auth, runID, &req.Target)
+			return nil, err
+		},
+		util.NewRetryConf(),
+	)
 	if err != nil {
 		return err
 	}
+	metrics.HistogramMetadataGetParentSpanDuration(
+		ctx,
+		time.Since(start),
+		attempts,
+		metrics.HistogramOpt{
+			PkgName: pkgName,
+		},
+	)
 
 	// Load run metadata to enforce the per-run cumulative size limit against
 	// metadata that already exists in the run, not just this request.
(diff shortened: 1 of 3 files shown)


pkg/api/apiv1/metadata.go
```diff
@@ -118,10 +120,51 @@ type AddRunMetadataRequest struct {
...
+	// data. There's a race condition where the parent span may not be queryable
+	// when the metadata update arrives.
+	//
+	// There's also a related race where a successful retry span isn't queryable
+	// yet, which causes us to mistakenly update metadata on a prior failed
+	// attempt.
+	//
+	// TODO: We should replace this hack with a proper fix. But a proper fix
+	// likely requires changes in our ClickHouse trace schema.
+	_, err = util.WithRetry(
```
albertchae, line 143: oh cool, didn't know we had WithRetry

pkg/api/apiv1/metadata.go
```diff
@@ -118,10 +120,51 @@ type AddRunMetadataRequest struct {
...
+		util.NewRetryConf(),
+	)
 	if err != nil {
+		logger.StdlibLogger(ctx).ErrorSample(
+			1,
+			"failed to get parent span for metadata",
+			"error", err,
+			"attempts", attempts,
+			"run_id", runID,
+			"target", req.Target,
```
albertchae: hm actually this error sample might be overkill, since WithRetry already logs each attempt and then an Error at the end https://github.com/inngest/inngest/blob/d76dcaecb8fd54d92478df667c990af8ccd2932c/pkg/util/retry.go#L76-L142

pkg/telemetry/metrics/histogram.go
```diff
@@ -655,3 +655,29 @@ func HistogramDefersPerRun(ctx context.Context, count int64, opts HistogramOpt)
...
+	attempts int,
+	opts HistogramOpt,
+) {
+	if opts.Tags == nil {
+		opts.Tags = map[string]any{}
+	}
+
+	// Always be true, but we'll add this check to protect cardinality just in
+	// case
+	if attempts < 10 {
```
albertchae, line 671: looks like it'll always be less than 5, but this is a good safety measure https://github.com/inngest/inngest/blob/d76dcaecb8fd54d92478df667c990af8ccd2932c/pkg/util/retry.go#L60

pkg/api/apiv1/metadata.go
```diff
@@ -118,10 +120,51 @@ type AddRunMetadataRequest struct {
...
+	// likely requires changes in our ClickHouse trace schema.
+	_, err = util.WithRetry(
+		ctx,
+		"apiv1.AddRunMetadata.getParentSpan",
+		func(ctx context.Context) (any, error) {
+			attempts++
+			parentSpan, scope, err = a.getParentSpan(ctx, auth, runID, &req.Target)
+			return nil, err
+		},
+		util.NewRetryConf(),
```
albertchae: it looks like the backoff for the default config will give us

  | Attempt | Sleep before attempt | Sleep after failed attempt |
  |---:|---:|---:|
  | 1 | 0ms | 100ms |
  | 2 | 100ms | 200ms 
  | 3 | 300ms | 400ms |
  | 4 | 700ms | 800ms |
  | 5 | 1500ms | none |

we have a 3s timeout from kafka batches before inserting into ClickHouse. I think we need to configure this so the retry window covers at least 5-6s otherwise we might end up with not finding the span a lot of the time

pkg/api/apiv1/metadata.go
```diff
@@ -118,10 +120,51 @@ type AddRunMetadataRequest struct {
...
+	// attempt.
+	//
+	// TODO: We should replace this hack with a proper fix. But a proper fix
+	// likely requires changes in our ClickHouse trace schema.
+	_, err = util.WithRetry(
+		ctx,
+		"apiv1.AddRunMetadata.getParentSpan",
+		func(ctx context.Context) (any, error) {
+			attempts++
+			parentSpan, scope, err = a.getParentSpan(ctx, auth, runID, &req.Target)
```
AndyInternet, line 148: Heads-up: this retry doesn't fire on the actual race in the cloud path. Monorepo's `traceReader.GetRunSpanByRunID` returns `(nil, nil)` (not an error) when the ClickHouse rollup hasn't caught up, so the closure here returns `err == nil` on the first attempt, `WithRetry` exits, and `AddRunMetadata` then nil-derefs `parentSpan.RunID` at the `stateID` line below.

Simplest fix: surface a nil span as a 404 in `getParentSpan` so `WithRetry` sees an error and retries. OSS sqlite/postgres already returns `sql.ErrNoRows` on a miss, so the new case is unreachable there, no regression.

In the `switch` block at line ~293 below, add:

```go
	case span == nil:
		// monorepo's ClickHouse-backed TraceReader returns (nil, nil) when the
		// rollup hasn't caught up yet. Surface this as an error so WithRetry
		// fires on the eventual-consistency window.
		return nil, 0, publicerr.Errorf(404, "metadata target not found")
```

pkg/telemetry/metrics/histogram.go
amh4r: Yea, I just wanted something reasonable

pkg/api/apiv1/metadata.go
amh4r: I removed the sample

pkg/api/apiv1/metadata.go
amh4r: I changed the retry schedule

pkg/api/apiv1/metadata.go
amh4r: Done, thanks

AndyInternet (approved):

PR #4241, opened 2026-05-21
refactor(connect): gate gateway routing by connection phase

darwin67:
## Description

Implements phases 1 and 2 of the Connect gateway lifecycle controller.

- Adds an internal `gatewayConnPhase` model for one websocket handler: New, Handshaking, Ready, Draining, Disconnecting, and Closed.
- Records phase transitions in logs while keeping Redis connection status, lifecycle callbacks, and websocket transport cleanup separate.
- Adds `canForward` and `canWrite` helpers so new router work and `GATEWAY_EXECUTOR_REQUEST` writes require `Ready`.
- Releases pending request ACK waiters when a connection leaves `Ready`, instead of relying only on timeout cleanup.
- Keeps post-persistence reply ACK behavior non-fatal and preserves drain-time lease extension handling for in-flight work.
- Makes phase transitions monotonic so stale drain paths cannot move a handler backward from `Disconnecting` to `Draining`.
- Adds regression coverage for phase transitions, routing/write eligibility, pending ACK release, heartbeat/drain behavior, and non-fatal reply handling.
- Updates the plan checklist and agent guidance for intent-focused comments.

## Release note

None

## Migration note

None

### DIFF
--- a/pkg/connect/gateway_phase.go
+++ b/pkg/connect/gateway_phase.go
@@ -75,23 +75,31 @@ func (c *connectionHandler) canWrite(kind connectpb.GatewayMessageType) bool {
 // transition records phase movement. Most external side effects still live at
 // existing call sites so phase checks can be introduced without collapsing the
 // Redis, lifecycle listener, and transport-close responsibilities together.
-func (c *connectionHandler) transition(to gatewayConnPhase, reason string, attrs ...any) gatewayConnPhase {
-	from := gatewayConnPhase(c.connPhase.Swap(int32(to)))
+func (c *connectionHandler) transition(to gatewayConnPhase, reason string, attrs ...any) (gatewayConnPhase, bool) {
+	for {
+		from := c.phase()
+		logAttrs := []any{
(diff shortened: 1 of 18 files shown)


pkg/connect/gateway_phase.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (c *connectionHandler) canWrite(kind connectpb.GatewayMessageType) bool {
+	phase := c.phase()
+
+	switch kind {
+	case connectpb.GatewayMessageType_GATEWAY_EXECUTOR_REQUEST:
+		return phase == gatewayConnPhaseReady
+	case connectpb.GatewayMessageType_GATEWAY_HEARTBEAT:
+		return phase == gatewayConnPhaseReady
```
KiKoS0: we still want heartbeats when draining though right

darwin67: yup, good catch.

KiKoS0 (approved):

PR #4242, opened 2026-05-21
fix: prevent duplicate function.finished on overlapping finalize

dlambrig:
## Summary

This PR stops one run from sending `inngest/function.finished` more than once when two workers try to finish the same run at the same time.

The executor logic is: one caller wins, later callers clean up but do not send another finish event.

## What Changed

- added a small state-store adapter for "who is allowed to send the finish event?"
- changed `Finalize()` so it only uses that adapter and does not know whether the store is Redis, Cassandra, or something else
- kept the Redis implementation inside the Redis state adapter
- changed Redis release to use a per-claim token and compare-and-delete, so a stale finalizer cannot delete a newer claim
- kept the retry behavior: if sending the finish event fails, the claim is released so a later retry can send it
- kept the delete-failure behavior: if the finish event was sent but state deletion failed, a later retry can delete state without sending the finish event again

## Why This Shape

The dedupe logic is useful, but the executor should not depend on Redis-specific behavior. Different stores need different ways to make the "first caller wins" decision.

Redis currently does this with `SET NX` plus a guarded release. A Cassandra/Scylla-backed store should implement the same adapter with storage-safe behavior. Until that exists, this PR should be treated as the Redis implementation plus the shared executor shape, not a complete Cloud/Cassandra fix.

## Behavior

- first finalizer sends `inngest/function.finished`
- duplicate finalizers skip the finish event
- duplicate finalizers still try to clean up state
- if publishing the finish event fails, a later retry can try again
- if deleting state fails after publishing, a later retry cleans up state but does not publish again

## Tradeoff

This prevents duplicate finish effects, but it does not make finish-event publishing durable.

If a worker claims finalization and then dies before publishing, another overlapping finalizer will see the claim and skip publishing until the claim TTL expires. A stricter fix would need a larger design, such as a durable outbox, a published marker, or an idempotent publish path.

## Validation

- `go test ./pkg/execution/state/redis_state -run '^TestV2AdapterClaimFinalization$|^TestV2AdapterFinalizationReleaseDoesNotDeleteNewerClaim$|^TestV2AdapterWithDisabledRetries$' -count=1`
- `go test ./tests/execution/executor -run '^TestFinalizePublishesFnFinishedOnce$|^TestFinalizeRetriesFnFinishedAfterPublishFailure$|^TestFinalizeDeleteFailureDoesNotLoseFinishEffects$' -count=1`

I also started a broader direct-package run. It completed `pkg/execution/executor` and `pkg/execution/state/v2`, but the broader Redis/executor integration portion did not finish promptly, so the focused tests above are the reliable validation for this update.

## Scope

This does not fix the deeper problem where two workers can both think they own the same run. It only prevents the clearest bad side effect from that overlap: sending duplicate finish events.

For Cloud/Cassandra/Scylla, the next step is to add the same finalization-claim adapter to that state store, ideally after confirming the exact assumptions with Riadh.


darwin67 (approved):
nice, but before merging, you should figure how this should be done in the state store proxy in cloud

dlambrig:
> nice, but before merging, you should figure how this should be done in the state store proxy in cloud

I think I see what you mean... the finalize dedupe hook is in the OSS state backend interface, but cloud appears to access state through the state-proxy, and that does not currently expose ClaimFinalization. 

So I'll add the cloud-side support in monorepo and link it here before merging.

darwin67:
yup, and this change assumes the state store is redis, which won't work with the new state store backend that @riadh is working on right now. so best to consult him as well.

dlambrig:
> yup, and this change assumes the state store is redis, which won't work with the new state store backend that @riadh is working on right now. so best to consult him as well.

ok, won't merge it until this is sorted

dlambrig:
Updated Redis release to use a per-claim token and compare-and-delete, so a stale finalizer cannot delete a newer claim after TTL expiry. Added `TestV2AdapterFinalizationReleaseDoesNotDeleteNewerClaim` for that case.

dlambrig:
> > yup, and this change assumes the state store is redis, which won't work with the new state store backend that @riadh is working on right now. so best to consult him as well.
> 
> ok, won't merge it until this is sorted

I drafted a logical solution in PR https://github.com/inngest/monorepo/pull/7349 for ScyllaDB after discussing. We can evaluate if its too expensive.

PR #4243, opened 2026-05-21
fix: preserve debounce on migration failure

dlambrig:
## Summary

This fixes a debounce migration bug where we could delete the old debounce on the secondary cluster before we had actually proven that the new debounce existed on the primary cluster.

In plain English: a migration attempt could cut over too early. If primary debounce creation failed in that window, the debounce was gone instead of still being runnable on the old cluster.

Put another way, before this PR, migration looked roughly like:

1. Lock the existing debounce on the secondary cluster.
2. Delete the old debounce state and timeout queue item from the secondary cluster.
3. Create the debounce on the primary cluster.

The bug was if step 2 permanently removed the old debounce before step 3 had successfully created the replacement on the primary cluster.

If primary creation failed:

- old debounce already deleted
- new debounce never created
- work disappeared

The debounce was effectively lost.

## Root Cause

The old flow did this:

1. `prepareMigration()` swapped the secondary pointer to a fake ID and set the migrating lock.
2. `debounce()` immediately deleted the old secondary debounce state and old timeout queue item.
3. Only after that did it try to create the new primary debounce.

If step 3 failed, there was nothing left to run.

## What Changed

The migration flow is now treated as a prepare / commit / rollback sequence:

- `prepareMigration.lua` preserves the old secondary pointer TTL when it swaps in the fake pointer.
- `prepareMigration()` now returns the existing debounce ID, carried timeout, and pointer TTL.
- Moved Redis debounce operations behind the queue shard adapter (`DebounceOperations`) so migration code can operate on primary and secondary shards uniformly instead of reaching directly into Redis clients/Lua.
- `debounce()` no longer deletes the old secondary debounce during prepare.
- After the primary create/update path runs, we verify whether the primary debounce is actually ready:
  - if yes, we clean up the old secondary debounce and timeout item
  - if no, we roll back the migration lock, restore the original secondary pointer, and remove any partial primary artifacts
- `primaryDebounceReady()` now checks the hashed debounce timeout job ID, matching how debounce timeout jobs are stored.

This keeps the old debounce runnable unless the new one is actually in place.

## Repro

The bug is reproducible:

1. Create a debounce on the old / secondary cluster.
2. Turn on migration.
3. Trigger another debounce event so migration starts.
4. Make primary debounce creation fail after `prepareMigration()` has already run.

Before this change, the old debounce disappeared.
After this change, the old debounce still exists on the secondary cluster with its pointer and timeout queue item intact.

## Tests

The main regression test is:

- `TestDebounceMigrationFailurePreservesExistingDebounce`

That test forces primary creation to fail after `prepareMigration()` has already locked the old debounce, then asserts:

- the original secondary debounce state still exists
- the secondary pointer still points at the original debounce
- the original secondary timeout queue item still exists
- no primary debounce was created

Additional migration coverage added / kept:

- `TestFinalizePreparedMigrationCommitsWhenPrimaryReadyAfterError`: fills the opposite finalize gap. If primary debounce state and its timeout job are already ready, but a later migration operation reports an error, finalization must commit secondary cleanup instead of rolling back into duplicate runnable debounce state. This also caught the hashed queue-item ID mismatch in `primaryDebounceReady()`.
- `TestRollbackPreparedMigrationKeepsMigratingFlagWhenPointerRestoreFails`: covers rollback partial failure. If pointer restore fails, the migrating guard must remain so the secondary debounce is not stranded behind the fake pointer with no guard.
- `TestDebounceExecutionDuringMigrationWorks`: covers timeout execution winning before migration prepares.
- `TestDebounceExecutionShouldNotRaceMigration`: covers migration prepare winning before timeout execution.
- `TestJITDebounceMigration`: covers the normal JIT migration path.

## Validation

- `go test ./pkg/execution/debounce -run '^(TestFinalizePreparedMigrationCommitsWhenPrimaryReadyAfterError|TestDebounceMigrationFailurePreservesExistingDebounce|TestRollbackPreparedMigrationKeepsMigratingFlagWhenPointerRestoreFails|TestJITDebounceMigration|TestDebounceExecutionShouldNotRaceMigration)$' -count=1`
- `go test ./pkg/execution/debounce -count=1`

## Scope

This PR is intentionally narrow. It fixes the Redis debounce migration work-loss bug and the primary-readiness check used by that migration finalizer.

Known gaps left outside this PR:

- Crash-after-prepare recovery / scavenger behavior if the process dies after the pointer is swapped to the fake ID.
- Full rollout sequencing with old and new executors plus the future-timestamp feature flag.
- FDB-specific debounce semantics, which are covered separately by the FDB debounce PR.
- The broader atomic debounce upsert prototype work.

### DIFF
--- a/pkg/execution/debounce/debounce.go
+++ b/pkg/execution/debounce/debounce.go
@@ -259,6 +259,12 @@ type debouncer struct {
 	shouldMigrate func(ctx context.Context, accountID uuid.UUID) bool
 }
 
+type preparedMigration struct {
+	debounceID       ulid.ULID
+	timeoutUnixMilli int64
+	pointerTTL       time.Duration
+}
+
 func (d debouncer) hasSecondary() bool {
 	return d.secondaryShardName != ""
 }
@@ -513,6 +519,8 @@ func (d debouncer) Debounce(ctx context.Context, di DebounceItem, fn inngest.Fun
 func (d debouncer) debounce(ctx context.Context, di DebounceItem, fn inngest.Function, ttl time.Duration, n int, shouldMigrate bool) error {
 	newDebounceID := ulid.MustNew(ulid.Timestamp(d.c.Now()), rand.Reader)
 	var foundDebounce bool
+	var migration *preparedMigration
+	var err error
 
 	l := logger.StdlibLogger(ctx).With(
 		"fn_id", di.FunctionID.String(),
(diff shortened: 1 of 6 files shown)


BrunoScheufler (commented):
This sounds correct but it's a little hard to follow all the codepaths, I'll try to map this out

BrunoScheufler (approved):

PR #4245, opened 2026-05-21
chore: improve agentic api cli help 

SterlingChin:
## Summary

Improves the generated `inngest alpha api <endpoint> --help` output for coding agents and humans by:

- using OpenAPI operation summaries/descriptions for endpoint help
- adding an explicit inherited target/auth/output flag note to every generated endpoint command
- using existing proto field descriptions for generated endpoint flags instead of raw JSON field names

## Why

The alpha API CLI already has useful metadata in `proto/api/v2/service.proto`, but the generated CLI help was falling back to terse JSON field names like `data`, `url`, and `transform`. Subcommand help also showed `[target/auth flags]` in the usage line without naming the inherited flags, which makes it easy for agents to miss `--prod`, `--api-key`, and `--env` when they start from a specific endpoint.

This keeps the command behavior unchanged while making the generated help more self-contained and actionable.

## Testing

- `/tmp/go/bin/go test ./cmd/apiv2cli`

## Notes

This is intentionally scoped to help generation only. Follow-up PRs can tackle behavior changes such as auth preflight, explicit env targeting for writes, and discovery endpoints for apps/functions/runs.


ui/packages/components/src/RunDetailsV4/StepInfo.tsx
```diff
@@ -284,7 +284,7 @@ export const StepInfo = ({
 
   return (
     <div className="flex h-full flex-col justify-start gap-2">
-      <div className="min-h-11 flex w-full flex-row items-center justify-between border-none px-4">
```
jacobheric, line 287: Unrelated but someone skipped precommit hook linting/formatting earlier...so leaving here.

jacobheric (approved):

PR #4253, opened 2026-05-22
fix(executor): treat function-not-found as orphaned run during cancel

lasse-it:
## Description

`executor.Cancel()` loads the function definition to obtain its slug for finalization. When the function row no longer exists (e.g. dev server restarted with the function removed while a run was still active in Redis), the raw `sql.ErrNoRows` from `LoadFunction` was wrapped as `"unable to load function: sql: no rows in result set"` and returned as a fatal error.

The stale-run-recovery loop then logged the following on every scavenge tick, indefinitely, since the function will never come back:

```
ERROR error cancelling stale run, will retry on next tick
  component: stale-run-recovery
  run_id: 01KQTB9J08NQ4ZY2Z2GWK43P2M
  error: unable to load function: sql: no rows in result set
```

This change treats `sql.ErrNoRows` from `LoadFunction` the same way `Cancel()` already treats `ErrMetadataNotFound` and `ErrEventNotFound`: log a warning and return `nil`. That lets `stale-run-recovery`'s existing cleanup path call `RemoveActiveRun` and stop looping on a run whose function is gone for good. No new error sentinel is introduced — the codebase already uses `errors.Is(err, sql.ErrNoRows)` directly elsewhere (e.g. `pkg/execution/executor/service.go:1031`).

## Release note

Stale-run recovery no longer spams an error every tick when a deleted function still has orphaned runs in state — the runs are cleaned up from the ActiveRuns index instead.

## Migration note

None.


lasse-it:
Closing — opened against the wrong fork. Re-opening on Avido-AI/inngest.

PR #4266, opened 2026-05-22
fix(connect): encode worker request ack policy

darwin67:
## Description

Implements plan 003 phase 3 for Connect gateway lifecycle handling.

- Encodes Option A for WORKER_REQUEST_ACK: a parsed worker ACK completes local Forward delivery, while executor gRPC Ack notification is best-effort.
- Keeps executor client creation and Ack RPC failures non-fatal for the worker websocket after ACK acceptance.
- Adds regressions for gRPC client creation failure, executor Ack RPC failure, and Forward success semantics.
- Updates the phase 3 plan checklist and records the Option A decision.

Validation:

- go test ./pkg/connect -run 'Test.*Ack|Test.*Forward|Test.*Drain' -count=1
- go test ./pkg/connect -count=1
- scripts/check-plan-status.sh docs/plans/003-connect-gateway-lifecycle-controller.org

## Release note

None.

## Migration note

None.

### DIFF
--- a/docs/plans/003-connect-gateway-lifecycle-controller.org
+++ b/docs/plans/003-connect-gateway-lifecycle-controller.org
@@ -570,10 +570,13 @@ Either way, a transient =getOrCreateGRPCClient= or =Ack= failure should not
 automatically close the worker websocket after the worker ACK was parsed and
 accepted.
 
+Decision: use Option A. A parsed =WORKER_REQUEST_ACK= completes local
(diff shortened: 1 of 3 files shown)


dlambrig (commented):
one comment, after this change, executor notification failures are only logs. Should we also increment a metric/counter for gRPC client creation failure, Ack RPC failure, and reply.Success=false? Metrics would make it easier to detect degraded executor notification behavior in production.

PR #4269, opened 2026-05-25
fix(executor): durably notify parent runs of invoked-child completion

lasse-it:
## Description

Fixes parent runs that hang waiting for invoked children that have already finished — a race that surfaces during executor pod rotation.

**Root cause.** With the default `inmemory` event-stream backend ([`pkg/cuedefs/config/config.cue`](pkg/cuedefs/config/config.cue) sets `eventStream.service.backend: "inmemory"`), the `inngest/function.finished` event published in [`finalize.go`](pkg/execution/executor/finalize.go) only reaches subscribers inside the same Go process. Across a multi-replica deployment, the parent-resume notification therefore lives entirely on the pod that finalized the child:

1. Child finishes on Pod A → `e.smv2.Delete()` ([finalize.go:163](pkg/execution/executor/finalize.go)) → UI marks child completed.
2. SIGTERM hits Pod A → runner Subscribe loop exits → mem:// subscriber dies.
3. Retry-wrapped `service.Go(HandleInvokeFinish)` fast-path is also tied to Pod A's lifecycle.
4. No other pod sees the completion event. Parent's invoke pause sits in Redis with no consumer.

**Fix.** Add a durable cross-pod notification path that rides the existing Redis-backed queue:

- New `KindInvokeComplete` queue kind + `PayloadInvokeComplete` payload carrying the FnFinished `event.BaseTrackedEvent`.
- `Finalize()` enqueues one item per FnFinished event with a correlation ID **before** state deletion. Items go to a dedicated `invoke-complete` system queue (not the child run's per-function partition) so `finalizeRemoveJobs` won't tear them down.
- Executor queue worker dispatches `KindInvokeComplete` to the existing `HandleInvokeFinish`, which is idempotent — `ErrPauseNotFound` is swallowed when the fast path already consumed the pause.
- Existing retry-wrapped in-process fast-path goroutine retained as the low-latency happy path; the queue item is the durable backstop.

Layer 2 (orphan-pause scavenger + short-TTL result cache) from the diagnosis plan is intentionally deferred to a follow-up.

### Files changed

- `pkg/execution/queue/kind.go` — new `KindInvokeComplete = "ic"` constant.
- `pkg/execution/queue/item.go` — `PayloadInvokeComplete` type + `UnmarshalJSON` case.
- `pkg/execution/executor/finalize.go` — reorder `finalizeEvents` before `Delete`; new `enqueueInvokeCompletes` helper; new `invokeCompleteQueueName` system queue.
- `pkg/execution/executor/service.go` — new `handleInvokeComplete` queue-worker dispatch.
- `pkg/execution/deletion/delete.go` — `KindInvokeComplete` added to the no-cleanup-needed system kinds list.

### Test plan

- [x] `go build ./pkg/execution/...` and `go build ./cmd/` succeed.
- [x] `go test ./pkg/execution/executor/... ./pkg/execution/queue/... ./pkg/execution/deletion/...` pass.
- [ ] Cluster test: trigger an invoking function, `kubectl delete pod -l app.kubernetes.io/component=server` while the child is running. Confirm parent resumes within a few seconds and that the resume occasionally fires on a different pod than the one that finalized the child.
- [ ] Confirm via logs that for normal completions the fast path resumes and the queue handler observes `ErrPauseNotFound` (i.e., is a no-op).

## Release note

Fix self-hosted multi-replica deployments where parent runs could hang waiting for invoked children that had already completed when an executor pod was rotated mid-finalization. The parent-resume notification now rides the durable Redis-backed queue and can be processed by any pod.

## Migration note

None. The new `invoke-complete` system queue is created on-demand by the executor; no config or schema changes are required.


lasse-it:
Wrong target repo — reopening on Avido-AI/inngest fork.

PR #4270, opened 2026-05-25
fix(security): bump js-cookie v2 transitive to 3.0.7 (CVE-2026-46625)

lasse-it:
## Description

Aikido continues to flag `js-cookie@2.2.1` after the initial fix in [#593](https://github.com/Avido-AI/inngest/pull/593), which intentionally restricted the override to the 3.x range to avoid crossing a major-version boundary on react-use's transitive dep. The rescan keeps surfacing CVE-2026-46625 against the v2 copy, so widen the override:

```yaml
# ui/pnpm-workspace.yaml
- 'js-cookie@>=3.0.0 <3.0.7': 3.0.7
+ 'js-cookie@<3.0.7': 3.0.7
```

This is safe for our workspace: the only `react-use` entry points imported anywhere in `ui/apps/dashboard` and `ui/packages/components` are `useLocalStorage` (all via direct `react-use/esm/useLocalStorage` imports), which never touches `js-cookie`. `useCookie` — the only consumer of `js-cookie` inside `react-use` — is unreachable from our bundles, so bumping it across the v2→v3 major boundary has no runtime impact.

After regenerating the lockfile, `js-cookie@2.2.1` is gone and `react-use@17.6.0` resolves to `js-cookie@3.0.7`.

## Release note
None.

## Migration note
None.


lasse-it:
Wrong target repo — gh defaulted to upstream. Reopening against Avido-AI/inngest.

PR #4271, opened 2026-05-25
refactor(queue): Add Scope to more queue methods

BrunoScheufler:
## Description

Threads `queue.Scope` through the queue manager/reader APIs added around `c938782421a206cfe07910b91b47f30616d6769b` and updates callers/tests to pass account, environment, and function IDs where available.

Notable updates:

- Adds `ScopeFromQueueItem` in `pkg/execution/queue/item.go` for deriving queue scope from persisted queue items.
- Updates Redis queue method signatures to accept scope while preserving the existing Redis shard behavior.
- Updates cron `HealthCheck` to accept account ID, environment ID, and function ID, then uses that scope when checking for the scheduled queue item.
- Updates cron health-check callers, including the debug API and executor health-check flow, to supply account/env/function IDs.
- Keeps the debug API on dev-server account/env constants for now; cloud-specific ID resolution can be handled by the cloud implementation.

Validation:

- `go test ./pkg/execution/cron`
- `go test ./pkg/execution/queue`
- `go test -run '^$' ./pkg/execution/state/redis_state ./tests/execution/executor ./tests/execution/queue`
- `go test -run '^$' ./...` compiles the changed packages; the top-level `tests` package still requires `SDK_URL` during init.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/debugapi/key_queue.go
+++ b/pkg/debugapi/key_queue.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/inngest/inngest/pkg/consts"
 	"github.com/inngest/inngest/pkg/execution/queue"
 	pb "github.com/inngest/inngest/proto/gen/debug/v1"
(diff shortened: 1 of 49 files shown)


pkg/debugapi/key_queue.go
```diff
@@ -19,7 +20,14 @@ func (d *debugAPI) GetShadowPartition(ctx context.Context, req *pb.ShadowPartiti
 		return nil, fmt.Errorf("error finding shard: %w", err)
 	}
 
-	pt, err := d.queue.PartitionByID(ctx, shard, req.GetPartitionId())
+	scope := queue.Scope{
```
BrunoScheufler, line 23: These will be updated in cloud

KiKoS0 (approved):

PR #4274, opened 2026-05-26
feat: Deterministic finalized step span IDs

rhino1998:
## Description

This changes trace outputs pretty majorly:
* Always emit `executor.step` span after execution
* Emit a separate discovery span for planned steps
* Still attach request metadata to the request span
* Make step spans deterministic with the following conditions:
  * If the step is finalized (IE successful or the final failure), emit a span derived from the step ID
  * Otherwise (for non-final failed attempts), emit a span derived from step ID and attempt number

inngest/inngest#4305 and inngest/monorepo#7413 are needed before this is merged.

## Release note

None.

## Migration note

None.


pkg/execution/checkpoint/run_context.go
```diff
@@ -128,10 +130,27 @@ func (c *checkpointRunContext) ExecutionSpan() *meta.SpanReference {
...
 
+func (c *checkpointRunContext) IsCheckpointing() bool {
+	return true
+}
+
 func (c *checkpointRunContext) ReleaseCapacityLease() error {
 	// TODO: Implement this once capacity leases are supported in Checkpointing
 	return nil
 }
+
```
rhino1998: We'll want to plumb req ID in CheckpointSyncSteps in the future for better step attribution to the actual driver request.

pkg/execution/executor/executor.go
```diff
@@ -3412,7 +3418,7 @@ func (e *executor) HandleGeneratorResponse(ctx context.Context, i *runInstance,
 	// When this happens, we ALWAYS need to create a trace for each step.
 	//
 	// We pass this down in context, unfortunately.
-	if nonLazy > 1 {
+	if nonLazy > 0 {
```
rhino1998: This causes us to emit "checkpointed" step spans whenever we have any non-lazy opcodes

pkg/execution/executor/executor.go
```diff
@@ -1963,39 +1970,18 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 		}
 
 		if e.allowStepMetadata.Enabled(ctx, instance.Metadata().ID.Tenant.AccountID) {
-			l := l.With("step_metadata", true)
```
rhino1998, line 1966: This gets moved onto the checkpoint-like step span

pkg/execution/executor/executor.go
```diff
@@ -1963,39 +1970,18 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
-				}
-			}
-
 			// Extract HTTP timing metadata from httpstat if available.
 			// This captures the detailed connection timing breakdown (DNS, TCP, TLS, TTFB, transfer)
 			// from the HTTP request to the user's SDK function.
 			if resp.HTTPStat != nil {
 				httpTimingMd := extractors.ExtractHTTPTimingMetadata(resp.HTTPStat)
-				_, err := e.createMetadataSpan(
+				_, err := e.createMetadataSpanOnParent(
```
rhino1998, line 1978: This stays on the request spans it is inconsistent for checkpointed steps currently anyways

pkg/execution/executor/executor.go
```diff
@@ -2004,20 +1990,34 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
...
 				)
 				if err != nil {
 					l.Warn("error creating timing metadata span", "error", err)
 				}
 			}
 		}
 
-		if handleErr := e.HandleResponse(ctx, &instance); handleErr != nil {
+		handleErr := e.HandleResponse(ctx, &instance)
+		if resp.IsGatewayRequest() {
```
rhino1998, line 2008: Emit a new execution update span after gateway requests to mark the request completed

pkg/execution/executor/executor.go
```diff
@@ -4138,23 +4075,38 @@ func (e *executor) handleGeneratorStepPlanned(ctx context.Context, runCtx execut
...
 	lifecycleItem := runCtx.LifecycleItem()
-	span, err := e.tracerProvider.CreateDroppableSpan(
+
+	_, err := e.tracerProvider.CreateSpan(
 		ctx,
-		meta.SpanNameStep,
+		meta.SpanNameStepDiscovery,
 		&tracing.CreateSpanOptions{
+			Debug:       &tracing.SpanDebugData{Location: "executor.handleGeneratorStepPlanned"},
 			Carriers:    []map[string]any{nextItem.Metadata},
```
rhino1998, line 4073: Previously the StepPlanned span acted as a discovery span kinda. We emit a separate span now for that

pkg/execution/checkpoint/run_context.go
```diff
@@ -127,11 +129,24 @@ func (c *checkpointRunContext) ExecutionSpan() *meta.SpanReference {
 	return nil
 }
 
-func (c *checkpointRunContext) ParentSpan() *meta.SpanReference {
+func (c *checkpointRunContext) RootSpan() *meta.SpanReference {
 	return tracing.RunSpanRefFromMetadata(&c.md)
 }
 
+func (c *checkpointRunContext) ParentSpan() *meta.SpanReference {
```
rhino1998, line 137: Currently we don't differentiate between parent/root spans for checkpointing because checkpointing doesn't have discovery spans

pkg/enums/metadata_scope.go
```diff
@@ -11,4 +11,5 @@ const (
 	// We keep it here for backward compatibility with old metadata that was written with this scope.
 	MetadataScopeStepAttempt
 	MetadataScopeExtendedTrace
+	MetadataScopeRequest
```
rhino1998, line 14: This is used for http timing metadata on the driver requests

pkg/tracing/util.go
```diff
@@ -100,58 +142,24 @@ func DriverResponseAttrs(
 		// to collect as much data as it can.
 		dropSpan := true
 		meta.AddAttr(rawAttrs, meta.Attrs.DropSpan, &dropSpan)
+		return rawAttrs
```
rhino1998: Afaict, we don't handle dropping EXTEND spans anyways, so this doesn't actually work?

(we explicitly exclude EXTEND spans when processing the DropSpan attribute)

pkg/execution/executor/executor.go
```diff
@@ -3405,26 +3418,6 @@ func (e *executor) HandleGeneratorResponse(ctx context.Context, i *runInstance,
 		}
 	}
```
rhino1998, line 3420: We always want to emit `executor.step` spans now, so this check is no longer necessary

pkg/tracing/util.go
```diff
@@ -73,6 +73,42 @@ func ResumeAttrs(p *state.Pause, r *execution.ResumeRequest) *meta.SerializableA
...
+	rawAttrs := meta.NewAttrSet()
+
+	if resp == nil {
+		return rawAttrs
+	}
+
+	fnOutput, err := resp.GetTraceFunctionOutput()
+	rawAttrs := meta.NewAttrSet()
+
+	fnOutput, err := resp.GetTraceFunctionOutput()
```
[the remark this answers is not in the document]
rhino1998: [3ba7724](https://github.com/inngest/inngest/pull/4274/commits/3ba7724d8177e2105f9267e76bf7c987d0e0a109)

Your suggested changes need some better context handling

pkg/tracing/util.go
```diff
@@ -413,6 +448,58 @@ func RunSpanRefFromMetadata(md *statev2.Metadata) *meta.SpanReference {
...
+func FinalizedStepSpanRefFromMetadataAndStepID(md *statev2.Metadata, stepID string) *meta.SpanReference {
+	if md == nil {
+		return nil
+	}
+
+	cfg := DeterministicSpanConfig(md.ID.RunID[:])
+	stepSpanID := DeterministicSpanConfig(FinalizedStepDynamicSeed(stepID)).SpanID
+	return &meta.SpanReference{
+		DynamicSpanID: stepSpanID.String(),
+		// TODO: validate that -00 is ok
```
scottnuma: Do we still need to validate this and https://github.com/inngest/inngest/pull/4274/changes#diff-7b1054eb04e87a1d972caf4a170eaaa7bb010c987c3b7342999336fe0d4a0df7R487?

rhino1998: After reading [this](https://www.w3.org/TR/trace-context/), that is the flags field, and that only has one flag defined where `01` represents that the trace was sampled. 

I think I finally understand why we need http://github.com/inngest/inngest/blob/3ba7724d8177e2105f9267e76bf7c987d0e0a109/pkg/tracing/tracer.go#L103

`00` is definitely fine, but we could change it to `01` everywhere in a followup

pkg/execution/executor/executor.go
```diff
@@ -5763,25 +5664,34 @@ func (e *executor) emitExperimentMetadataFromOpts(ctx context.Context, runCtx ex
...
-		parent = tracing.RunSpanRefFromMetadata(runCtx.Metadata())
+		parent = tracing.RunSpanRefFromMetadata(runMD)
 	case enums.MetadataScopeStep, enums.MetadataScopeStepAttempt:
-		parent = runCtx.ExecutionSpan()
+		if op.Op == enums.OpcodeStepError && IsStepRetryable(op, runCtx) {
+			parent = tracing.RetryStepSpanRefFromMetadataAndStepID(runMD, op.ID, runCtx.AttemptCount())
+		} else {
+			parent = tracing.FinalizedStepSpanRefFromMetadataAndStepID(runMD, op.ID)
+		}
 	default:
```
scottnuma, line 5689: We added a new MetadataScope here: https://github.com/inngest/inngest/pull/4274/changes#diff-9239dab6c06975c6f62ab076b613a949bb21f0b6dcdb67cec8d78455048963f5R14,

with MetadataScopeRequest

Do we need to add handling for it here?

pkg/execution/executor/executor.go
```diff
@@ -5815,6 +5725,156 @@ func (e *executor) createMetadataSpan(ctx context.Context, runCtx execution.RunC
...
+	if result != nil {
+		attrs.Merge(apiAttributes(*result))
+	}
+
+	item := runCtx.LifecycleItem()
+
+	seed := tracing.RetryNonStepDynamicSeed(item)
+	if status == enums.StepStatusFailed {
+		seed = tracing.FinalizedNonStepDynamicSeed(item)
+	}
```
scottnuma: These seem to be our only uses of RetryNonStepDynamicSeed and FinalizedNonStepDynamicSeed. They have identical implementations.

They seem to follow the convention of having different methods for retry vs terminal, but I think we could consolidate them to simplify this branch + their defs and testing code.

Or might they change around in the future?

pkg/execution/checkpoint/run_context.go
```diff
@@ -127,10 +128,19 @@ func (c *checkpointRunContext) ExecutionSpan() *meta.SpanReference {
...
+	return c.RootSpan()
+}
+
+func (c *checkpointRunContext) StartTime() time.Time {
+	// Plumbed here for interface compatibility, but not currently used in checkpointing
+	return time.Time{}
+}
```
scottnuma, line 143: While looking through this PR, Claude pointed out this. I can't quite follow Claude all the way through. Returning an empty time.Time{} felt a little like a tripping hazard, so I figured worth forwarding this comment with a grain of salt.

> Question on the zero `StartTime()` here — this isn't quite dead "interface compatibility" plumbing; it's a live fallback on the checkpoint path.
> 
> The only caller is `opcodeTiming`'s fallback ([executor.go](https://github.com/inngest/inngest/pull/4274/changes#diff-ccd0679a9f24c84fef4f5d4d5fe2c32651d23886d85af8448bb968921c7b8a95R5754-R5756)), which `emitStepSpan`/`emitNonStepSpan` reach for every opcode checkpoint forwards through `HandleGenerator` (step errors/failures, gateway/AI-gateway, sync `RunComplete`). Two observations:
> 
> 1. On the checkpoint path the `startedAt.Before(queuedAt)` half of the fallback condition is vacuous: the synthetic `LifecycleItem` never sets `EnqueuedAt` ([run_context.go#L87-L107](https://github.com/inngest/inngest/blob/3ba7724d8177e2105f9267e76bf7c987d0e0a109/pkg/execution/checkpoint/run_context.go#L87-L107)), so `queuedAt` is also zero. Net effect: this returns zero time exactly when the SDK omitted `op.Timing`.
> 2. When that happens the span isn't corrupted — `AddTimingAttrs` skips zero values ([execution_processor.go](https://github.com/inngest/inngest/pull/4274/changes#diff-d7ebacb2f19a0fb1cc4663921b46f425fe5b576b0143283564283e44857d2486R87-R89)) and the tracer start-time guard falls back to `now` — but the resulting step/non-step span has no `startedAt` and `endedAt` = checkpoint-processing time, so its duration is junk.
> 
> What I'd like to confirm is the asymmetry with the async path: `runInstance.StartTime()` returns the request start ([run_context.go](https://github.com/inngest/inngest/pull/4274/changes#diff-35abf153fa7ab234148a7d7b40006c4b8691b7792f8d808b21d8dc04df4d4107R163-R165)), so the same SDK timing omission gets a sensible fallback on the async path but missing data on the checkpoint path.
> 
> Is `op.Timing` guaranteed to be populated on all checkpointed opcodes (in which case this is fine — maybe worth stating that in the comment instead of "not currently used")? Or should this return an analog of the request start, e.g. the checkpoint API's request-received time?

pkg/execution/executor/executor.go
```diff
@@ -4621,17 +4541,13 @@ func (e *executor) handleGeneratorAIGateway(ctx context.Context, runCtx executio
...
-			TargetSpan: runCtx.ExecutionSpan(),
-		}); spanErr != nil {
-			e.log.Debug("error updating span for successful gateway request during handleGeneratorAIGateway", "error", spanErr)
-		}
+		e.emitStepSpan(
+			ctx,
+			runCtx,
+			&gen,
+			md,
+			tracing.GatewayResponseAttrs(resp, nil, gen, nil))
```
scottnuma: In the previous call to this, we included rawBody, but now we pass nil. I see that we still have rawBody used in the logic of GatewayResponseAttrs, so I think we should probably still keep it around?

If not, that seems to be the last usage of the arg, so we could drop the arg - out of scope for this PR.

scottnuma (approved):
This looks good to me. 

I've reviewed mostly at close up level, so I think it could be useful for additional review particularly at the higher level, like changes to our span parent graph before and after.

For other reviewers, I also clauded my way this branch with some more piece wise commits to step through that match hashes: https://github.com/inngest/inngest/commits/scott/pr-4274-review/

pkg/execution/checkpoint/run_context.go
rhino1998: I think that all is true, but I don't think the timing field is ever missing from checkpointing?

We can set a real StartTime in the API handler for safety I guess?

pkg/execution/executor/executor.go
rhino1998: Combined in [443b5ad](https://github.com/inngest/inngest/pull/4274/commits/443b5adce23e3a7f02ed1300d2186f1fd8676104)

pkg/execution/executor/executor.go
rhino1998: [68f045f](https://github.com/inngest/inngest/pull/4274/commits/68f045ff8a04468f84b93c015e4bf09a4d639572)

pkg/execution/checkpoint/run_context.go
scottnuma: ehh, feels like it's not worth then. thanks!

rhino1998: [4970f9c](https://github.com/inngest/inngest/pull/4274/commits/4970f9cefcae3be26a7c58e4a5b9a71032c68ee6)

Adds a fallback timestamp so we have a roughly reasonable timestamp in the case that the opcode is missing a timestamp

pkg/execution/executor/executor.go
rhino1998: Technically no because we never call createMetadataSpan with MetadataScopeRequest (we use createMetadataSpanOnParent with an explicit parent), but we have enough info on the runCtx to handle it easily so [119b649](https://github.com/inngest/inngest/pull/4274/commits/119b649a9f2d43c714197f514b8968c26d9de049)

PR #4279, opened 2026-05-28
feat(api): add bulk function loaders

jacobheric:
## Description

Add bulk function providers to support our new rest v2 endpoints. Nice to have as it allows our oss API impls to match our cloud impls more closely with regard to pagination, etc and so that our self-hosters don't run into performance issues. 

## Release note

None.

## Migration note

None.


jacobheric:
This is too painful. Going to make a better contract  that allows us to fetch this in one pass.

PR #4280, opened 2026-05-28
fix: Properly calculate discover step

jakobevangelista:
## Description

Previously, we would calculate discovery time from the start of the run to queued at

This makes discovery timing calculation based on the previous completed step

### DIFF
--- a/ui/packages/components/src/RunDetailsV4/utils/traceConversion.ts
+++ b/ui/packages/components/src/RunDetailsV4/utils/traceConversion.ts
@@ -124,13 +124,14 @@ function getTimingFromMetadata(
  */
 function getInngestBreakdown(
   trace: Trace,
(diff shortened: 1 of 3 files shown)


ui/packages/components/src/RunDetailsV4/utils/traceConversion.ts
```diff
@@ -150,6 +151,40 @@ function getInngestBreakdown(
   return { discoveryMs, queueDelayMs, systemLatencyMs, totalMs };
 }
 
+function tracesToBarData(
```
jakobevangelista, line 154: `traceToBarData` used to be fully per-span, but discovery is now relative to the previous completed sibling

this is just the sibling aware wrapper that carries that previous end timestamp while recursively converting children

AndyInternet (approved):

PR #4284, opened 2026-05-29
refactor(queue): Remove queue migration locks

BrunoScheufler:
## Description

Since the Constraint API has been rolled out, there's no need to pause queues on the source or destination. Capacity will be enforced on either side.

## Release note

None.

## Migration note

None.


dlambrig:
with the migration lock removed, what prevents the source shard from continuing to refill/promote work for a function while Migrate() is running? 

1. function F has some ready items in its partition and more items still sitting in shadow/backlog
2. migration starts and Migrate() begins moving the ready items from shard A to shard B
3. while that is happening, shard A’s normal shadow processing refills more work for F out of shadow/backlog into the partition
4. Migrate() finishes the set of items it saw earlier and returns
5. now shard B has the migrated items, but shard A still has newly refilled items for the same function

What guarantees that this work is either migrated as well or otherwise picked up correctly? Previously it looked like the migration lock prevented shadow processing from racing with migration, so I'm trying to understand what replaces that guarantee.

BrunoScheufler:
> with the migration lock removed, what prevents the source shard from continuing to refill/promote work for a function while Migrate() is running?
> 
>     1. function F has some ready items in its partition and more items still sitting in shadow/backlog
> 
>     2. migration starts and Migrate() begins moving the ready items from shard A to shard B
> 
>     3. while that is happening, shard A’s normal shadow processing refills more work for F out of shadow/backlog into the partition
> 
>     4. Migrate() finishes the set of items it saw earlier and returns
> 
>     5. now shard B has the migrated items, but shard A still has newly refilled items for the same function
> 
> 
> What guarantees that this work is either migrated as well or otherwise picked up correctly? Previously it looked like the migration lock prevented shadow processing from racing with migration, so I'm trying to understand what replaces that guarantee.

This would also apply in the general case where execution could start and race with the migration itself.

We probably want to keep the source lock for the duration of the migration, but we _can_ loosen the destination lock and start processing right away.

BrunoScheufler:
Closing this PR in favor of https://github.com/inngest/monorepo/pull/7343

PR #4286, opened 2026-05-29
Fix: Update Events List and docs in v1/events endpoint to limit to 51 and return 400 if limit is greater than 51

campbell88:
## Description

Our v1 Rest API only returns 51 records although our api spec docs have maximum set to 100. Change is updating the docs to show 51 and returning 400 if a request send limits over 51

## Release note

  The /v1/events endpoint now returns a 400 Bad Request when the limit query parameter is 
  less than 1 or greater than 51. Previously, out-of-range values were silently clamped to
   the nearest valid value, which could cause callers to receive fewer events than        
  expected without any indication. The API documentation has also been updated to reflect 
  the correct maximum of 51.

## Migration note

None.

### DIFF
--- a/docs/api-docs/public/api-specs/v1.json
+++ b/docs/api-docs/public/api-specs/v1.json
@@ -55,7 +55,7 @@
             "schema": {
               "type": "number",
               "minimum": 1,
-              "maximum": 100
+              "maximum": 51
             },
             "in": "query",
             "name": "limit",
--- a/docs/api-docs/public/api-specs/v1.yaml
+++ b/docs/api-docs/public/api-specs/v1.yaml
@@ -57,7 +57,7 @@ paths:
         - schema:
             type: number
             minimum: 1
-            maximum: 100
+            maximum: 51
           in: query
           name: limit
           description: The number of events to list
--- a/docs/openapi/v3/api/v1/spec.yaml
+++ b/docs/openapi/v3/api/v1/spec.yaml
@@ -54,7 +54,7 @@ paths:
         - schema:
             type: number
             minimum: 1
-            maximum: 100
+            maximum: 51
           in: query
           name: limit
           description: The number of events to list
--- a/pkg/api/apiv1/events.go
+++ b/pkg/api/apiv1/events.go
@@ -13,7 +13,6 @@ import (
 	"github.com/inngest/inngest/pkg/enums"
 	"github.com/inngest/inngest/pkg/logger"
 	"github.com/inngest/inngest/pkg/publicerr"
-	"github.com/inngest/inngest/pkg/util"
 	"github.com/oklog/ulid/v2"
 )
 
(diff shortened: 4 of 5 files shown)


kang8:
**Thanks for jumping on this so quickly! Before this lands, I'd like to push back on the chosen direction — I think raising the cap to 100 is the better fix, and the current approach has an undeclared breaking change.**

**1. The `51` isn't a `50+1` "has-more" sentinel — it's just an arbitrary value.**

The original report speculated that `51` might be `50 + 1` for has-more detection. The code doesn't bear that out: the storage layer simply does `Limit(uint(opts.Limit))` and returns exactly those rows (`pkg/cqrs/manager/cqrs.go:1469`, with `out` built directly from the result set at `:1481-1503`), and the handler just does `WriteResponse(w, events)` with no pagination metadata (`pkg/api/apiv1/events.go:99`). So there's no has-more logic anywhere, and nothing depends on the value being 51. As the reporter noted, the constant hasn't changed since the original implementation.

Since `51` isn't a sentinel, isn't a performance boundary (it's a single indexed `LIMIT` query), and isn't otherwise load-bearing, is there a reason we're lowering the documented max to match it rather than raising the cap to match the spec?

**2. This direction makes the public contract weaker, not corrected.**

The OpenAPI spec already advertises `maximum: 100`, and "raise the cap to 100" was the reporter's first suggested resolution. This PR instead downgrades the documented capability to fit an internal arbitrary constant. When both directions are equally easy, lowering what the public API offers seems like the less desirable one — and the PR description doesn't explain why 100 was ruled out.

**3. Returning a 400 is itself an undeclared breaking change.**

Today:
- `limit=100` → `util.Bound(limit, 1, 51)` → **200 OK** with 51 events.

After this PR:
- `limit=100` → **400 Bad Request**.

Any caller that built against the documented `max=100` is currently succeeding (just under-fetching). After this change, those same requests start returning 400. Turning a previously-200 public REST request into a 400 is a behavioral breaking change — and the Migration note is set to `None`, which I don't think is accurate here.

By contrast, **raising the cap to 100 breaks no one**: every existing caller (whether they request ≤51 or ≤100) keeps working, and callers following the docs finally get the count they asked for.

**4. Minor: `cqrs.MaxEvents` is shared with the GraphQL resolvers.**

Worth noting that raising it would also widen the GraphQL page cap (`pkg/coreapi/graph/resolvers/events.go:51`, `events_v2.go:23`, `events_connection.go:20`). That's backward-compatible (raising a max only permits more), so low risk — just flagging it so the change is intentional. (Unrelated: `pkg/api/api.go:214` / `eventstream.go` use the separate `consts.MaxEvents = 5000`, so they're unaffected.)

**Suggestion:** raise `MaxEvents` to 100 to match the spec, keep the spec at 100, and keep the new validation as a `400` only for genuinely out-of-range values (`< 1` or `> 100`). That honors the published contract, breaks no existing caller, and still fixes the silent-clamp problem. If there's a reason to keep 51 that I'm missing, then at minimum the 400 behavior should be called out as a breaking change in the Migration note.

PR #4305, opened 2026-06-02
feat: UI: Group top-level step spans using step/group IDs

rhino1998:
## Description

This adds another stage of UI rollup to group top-level step attempt spans under a virtual parent span.

Needs #4304 and inngest/monorepo#7413

## Release note

None.

## Migration note

None.


ui/apps/dev-server-ui/src/coreapi.ts
```diff
@@ -382,6 +382,7 @@ export const TRACE_DETAILS_FRAGMENT = gql`
       updatedAt
     }
     outputID
+    groupID
```
rhino1998, line 386: Will need to do this for cloud too once #4304 and a followup vendor PR are in

Linell (approved):

PR #4308, opened 2026-06-02
feat(pauses): Move pause timeouts to system queue

BrunoScheufler:
## Description

> [!CAUTION]
> This breaks singletons. Some users want to maintain a singleton lock when the run is waiting for an event or signal, or invoking another function. Moving pause timeouts to the system queue means we won't use the same user level singleton lock on the account shard.

Upstream in https://github.com/inngest/monorepo/pull/7382

Going through all queue items, pause timeouts are enqueued to user queues and don't skip constraint checks, so use the Constraint API. This also means that in case of missing concurrency capacity, we will not run pause timeouts.

Pause timeouts only schedule the resume job as an `edge` item subject to constraint checks, so there's no reason why pause timeouts should not be a system queue like all other timeouts.

Here's a comprehensive table of all queue item types:

| Item Kind | Queue name | Is System? | Description |
| --- | --- | --- | --- |
| `internal-recovery` | `internal-recovery` | Yes | Replays |
| `schedule-batch` | `schedule-batch` | Yes | Batch timeouts |
| `pause-event` | `pause-event` | Yes | Pause events (not timeouts) |
| `debounce` | `debounce` | Yes | Debounce timeouts |
| `queue-migrate` | `queue-migrate` | Yes | Queue migrations (is this still used?) |
| `cancel` | `cancel` | Yes | Cancellation jobs |
| `fp` | `fp` | Yes | Function pausing jobs |
| `fup` | `fup` | Yes | Function unpausing jobs |
| `cron` | `cron` | Yes | Cron item |
| `cron-sync` | `cron-sync` | Yes | Cron sync |
| `jps` | `jps` | Yes | Job promotion |
| `pbf` | `pbf` | Yes | Pause block flush |
| `pause` | `pause` | No ⚠️ | Pause timeout job |

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -524,6 +524,8 @@ type executor struct {
 	clock             clockwork.Clock
 
 	conditionalTracer itrace.ConditionalTracer
+
+	pauseTimeoutSystemQueue func(ctx context.Context, accountID, envID, functionID uuid.UUID) bool
 }
 
 func (e *executor) SetFinalizer(f execution.FinalizePublisher) {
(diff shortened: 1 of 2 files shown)


pkg/execution/executor/executor.go
```diff
@@ -3290,9 +3299,16 @@ func (e *executor) Resume(ctx context.Context, pause state.Pause, r execution.Re
 
 		// And dequeue the timeout job to remove unneeded work from the queue, etc.
 		if q, ok := e.queue.(queue.QueueManager); ok {
+			// Previously we enqueued pause timeouts to the user queue, never a system queue.
+			// We are changing this below, creating per-function timeout partitions.
+			var queueName *string
+			if e.pauseTimeoutSystemQueue != nil && e.pauseTimeoutSystemQueue(ctx, md.ID.Tenant.AccountID, md.ID.Tenant.EnvID, md.ID.FunctionID) {
```
dlambrig, line 3305: Do all pause timeout producers now use the same queue-name logic?

Resume switches its timeout cleanup path based on this flag, but this PR only updates `handleGeneratorWaitForSignal`. If `handleGeneratorWaitForEvent` or `handleGeneratorInvokeFunction` still enqueue timeout jobs to the account queue, Resume could look in the wrong queue and leave stale timeout jobs behind.

dlambrig (commented):
left one remark

PR #4310, opened 2026-06-02
internal: EXE-1837: convert attribute data to .otlp.json

scottnuma:
## Description

- Previously, we parsed otel data into a Go format for test fixture data to validate our parsing
- We're going to be making more of this data. Let's make it more robust by using the Otel JSON directly and skip parsing in Go

## Release note

None.

## Migration note

None.


pkg/tracing/metadata/extractors/aimetadata_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+// TestAIMetadataExtractor_CapturedFixtures asserts AIMetadata fields against captured OTLP spans
+func TestAIMetadataExtractor_CapturedFixtures(t *testing.T) {
+	cases := []struct {
+		fixture  string
+		expected extractors.AIMetadata
+	}{
+		// official @opentelemetry/instrumentation-openai
+		{
+			fixture: "openai_otel_official/params_chat.otlp.json",
```
rhino1998, line 21: Maybe using goldie like the Insights tests would be cleaner as the number of test cases grows?

rhino1998 (approved):
Nothing blocking

pkg/tracing/metadata/extractors/aimetadata_test.go
scottnuma: I think this makes sense, esp. as it grows. I'll work this in with a future PR

scottnuma: https://linear.app/inngest/issue/EXE-1923/convert-tests-to-goldie-for-extraction

PR #4314, opened 2026-06-03
fix: add app name to run spans

jacobheric:
## Description

Add app name to run spans for v2 api fetches.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/debounce/debounce.go
+++ b/pkg/execution/debounce/debounce.go
@@ -54,6 +54,8 @@ type DebounceItem struct {
 	WorkspaceID uuid.UUID `json:"wsID"`
 	// AppID represents the app for the debounce item
 	AppID uuid.UUID `json:"appID"`
+	// AppName represents the app ID defined in user code.
+	AppName string `json:"appName,omitempty"`
 	// FunctionID represents the function ID that this debounce is for.
 	FunctionID uuid.UUID `json:"fnID"`
 	// FunctionVersion represents the version of the function that was debounced.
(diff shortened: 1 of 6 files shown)


pkg/execution/debounce/debounce.go
```diff
@@ -54,6 +54,8 @@ type DebounceItem struct {
 	WorkspaceID uuid.UUID `json:"wsID"`
 	// AppID represents the app for the debounce item
 	AppID uuid.UUID `json:"appID"`
+	// AppName represents the app ID defined in user code.
+	AppName string `json:"appName,omitempty"`
```
jacobheric, line 58: app name was lost on debounce so I opted to add it to the contract and pass it along rather than hit db cache in debounce like we do for functions. shout if hitting db cache is preferable.

pkg/execution/execution.go
```diff
@@ -239,6 +239,8 @@ type ScheduleRequest struct {
 	WorkspaceID uuid.UUID
 	// AppID is the app that this request belongs to.
 	AppID uuid.UUID
+	// AppName is the app ID defined in user code.
```
albertchae, line 242: ok unlike functions, we only have an id concept for apps and not a separate name as well? https://www.inngest.com/docs/reference/typescript/v4/client/create#configuration

albertchae (approved):

pkg/execution/execution.go
jacobheric: technically we have separate id and name in the table, but we don't expose via the config api I believe.

PR #4323, opened 2026-06-04
chore(tracing): add more conditional traces around scheduling

KiKoS0:
## Description
Even if these specific calls aren’t the main source of latency
this should give us a clearer picture of the timings between
these intermediate calls.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -1122,6 +1122,9 @@ func (e *executor) schedule(
 	// Some scenarios are happy path (e.g.  queue idempotency) and some are sad
 	// path (e.g. Executor borked)
 	sendSpans := func() {
+		_, span := e.conditionalTracer.NewSpan(ctx, "executor.schedule.send_spans", req.AccountID, req.WorkspaceID, req.Function.ID)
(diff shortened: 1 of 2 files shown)


albertchae (approved):

pkg/execution/executor/executor.go
```diff
@@ -1260,7 +1263,9 @@ func (e *executor) schedule(
 	// Check if the function should be skipped (paused, draining, backlog limit)
 	// Only check if not already marked as skipped (e.g., by singleton)
 	if skipReason == enums.SkipReasonNone {
+		_, span := e.conditionalTracer.NewSpan(ctx, "executor.schedule.check_backlog_limit", req.AccountID, req.WorkspaceID, req.Function.ID)
```
lkasinathan: did not look too closely at surrounding code, but is this name right?

lkasinathan (approved):

pkg/execution/executor/executor.go
KiKoS0: not really fixing it

mar-inngest (approved):

PR #4324, opened 2026-06-04
fix(cqrs): correct run EndedAt when a CEL event filter is active

### LINKED ISSUE #4268: Bug in dashboard (when using CEL query)
**Describe the bug**
When using CEL query to filter runs, the function `end at` time is wrong. Take a look at the below screenshots:
- Without CEL query, end time of the function with runID `01KSCWH6BN21RWPQ49YNPBNAQ3` is **3:09:43 PM** (which is correct)
Image

- But when CEL query applied, the function's end time becomes **3:09:39 PM**
Image

**To Reproduce**
1. Go to "runs"
2. Check "Ended at" column for a function
3. Write a CEL query so the resulting list includes that function
4. Now check the end time of that function

**System info**
- inngest: 1.17.9 (self-host using docker image)
- OS: Ubuntu 24.04

admirsaheta:
## Description

Fixes #4268 — the `Ended At` column in the runs list showing an earlier (wrong) timestamp whenever a CEL query filter was active. The non-filtered view was correct; apply literally any CEL expression and the end time would quietly slip back by a few seconds. Classic.

### Root cause

`GetSpanRuns` builds an aggregated query over the `spans` table, grouping all spans for a run to compute `MAX(spans.end_time)` as the run's end time. When a CEL `event.*` filter was present, the query joined the `events` table inline via an `INNER JOIN` on `spans.event_ids`:

```sql
FROM spans
INNER JOIN json_each(spans.event_ids) …
INNER JOIN events ON … WHERE events.event_name = 'user/created'
GROUP BY spans.run_id, spans.dynamic_span_id, …
SELECT MAX(spans.end_time) -- 🙂 only root span survived
```

The problem: only the root `executor.run` span carries `event_ids`. Every subsequent `EXTEND` span has `NULL`/empty `event_ids` and is silently dropped by the `INNER JOIN` before the `GROUP BY` even runs. So `MAX(end_time)` was computed from the root span alone — producing the earlier timestamp the user observed. The actual last span's end time never made it into the result.

### Fix

Moved event-field CEL predicates out of the main aggregation query and into a correlated `run_id` subquery:

```sql
FROM spans
WHERE spans.run_id IN (
    SELECT DISTINCT spans.run_id
    FROM spans
    INNER JOIN events ON …      -- join lives here now, safely isolated
    WHERE events.event_name = 'user/created'
)
GROUP BY spans.run_id, spans.dynamic_span_id, …
SELECT MAX(spans.end_time)      -- ✅ all spans participate
```

All spans for a matched run now participate in the `GROUP BY`, so `MAX(end_time)` correctly reflects the last `EXTEND` span's end time. The same subquery pattern is applied to `output.*`/`error.*` CEL predicates for the same reason.

**`ExpressionHandler` changes (`pkg/run/cel.go`):**
- Extracted `ToEventSQLFilters(ctx)` — SQL predicates for `event.*` CEL expressions
- Extracted `ToOutputSQLFilters(ctx)` — SQL predicates for `output.*`/`error.*` CEL expressions
- Refactored `ToSQLFilters(ctx)` to be their union — zero behaviour change for existing callers

**`GetSpanRuns` changes (`pkg/cqrs/manager/cqrs.go`):**
- Removed `BuildEventJoin` from the main span aggregation query
- Event filters → correlated `run_id IN (SELECT … INNER JOIN events …)` subquery
- Output/error filters → same subquery pattern anchored on `spans.output`

**Tests:**
- `TestExpressionHandlerSeparateSQLFilters` (6 subtests) — unit coverage for the new filter accessors
- `TestGetSpanRunsCELEventFilterPreservesEndedAt` (3 subtests) — integration regression that directly reproduces the bug: inserts a root + EXTEND span pair, verifies `EndedAt` equals the EXTEND span's end time both with and without a CEL event filter

## Release note

Fixed a bug where the `Ended At` time shown for a function run was incorrect when a CEL query filter was applied. The displayed time was taken from the root span rather than the actual last step, causing it to appear earlier than the real end time.

## Migration note

None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ pkg/execution/state/redis_state/testdata
 .specify/
 specs/**
 .retrospeced/**
+.codegraph/
--- a/pkg/cqrs/manager/cqrs.go
+++ b/pkg/cqrs/manager/cqrs.go
@@ -2994,9 +2994,12 @@ func (w wrapper) GetSpanRuns(ctx context.Context, opt cqrs.GetTraceRunOpt) ([]*c
 
 	builder := newSpanRunsQueryBuilder(ctx, opt)
 
-	// Parse CEL expressions using adapter's converter
-	var celFilters []sq.Expression
-	var useJoin bool
+	// Parse CEL expressions using adapter's converter.
+	// Event filters and output filters must be kept separate: event filters reference the
+	// events table (via a join) while output filters reference spans.output directly.
+	// Mixing them in the main span aggregation join corrupts MAX(end_time) — see fix below.
+	var eventCELFilters []sq.Expression
+	var outputCELFilters []sq.Expression
 	if opt.Filter.CEL != "" {
 		expHandler, err := run.NewExpressionHandler(ctx,
 			run.WithExpressionHandlerBlob(opt.Filter.CEL, "\n"),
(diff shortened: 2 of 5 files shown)


admirsaheta:
@jpwilliams could this be merged in next release?

PR #4329, opened 2026-06-04
internal: Convert AI Metadata Tests to use Goldie EXE-1923

scottnuma:
## Description

- Per a previous review convo, we convert from hand writing the expected output in Go to recording the expected output with Goldie
- We move context on the parsing into pkg/tracing/metadata/extractors/testdata/README.md

## Release note

None.

## Migration note

None.


rhino1998 (approved):
nice

PR #4331, opened 2026-06-04
feat: A/B test insights query-writer model and score tokens/cost

AndyInternet:
## Description

Split the insights SQL query-writer 50/50 between claude-sonnet-4-5 and
claude-opus-4-8 using the SDK's group.experiment, and record per-call
output-token and cost scores via step.score. Cost is derived from hardcoded
per-model pricing and rounded to micro-dollars.

Also trim floating-point noise from non-integer score values in the Scores
tab (ScoresAttrs), so values render as e.g. 0.02773 instead of
0.027729999999999998.

Bumps the dashboard's inngest dependency to the 4.4.1-pr-1521.15 prerelease,
which provides the scoring primitive (step.score / scoreMiddleware).

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/apps/dashboard/src/lib/inngest/functions/run-insights.ts
+++ b/ui/apps/dashboard/src/lib/inngest/functions/run-insights.ts
@@ -1,4 +1,4 @@
-import { anthropic } from 'inngest';
+import { anthropic, experiment } from 'inngest';
 import { v4 as uuidv4 } from 'uuid';
 
 import { inngest } from '../client';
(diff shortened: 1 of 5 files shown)


ui/apps/dashboard/src/lib/inngest/functions/run-insights.ts
```diff
@@ -155,17 +155,71 @@ export const runInsightsAgent = inngest.createFunction(
...
+      ];
+    const rawCostUsd =
+      (inputTokens / 1_000_000) * pricing.inputPerMTok +
+      (outputTokens / 1_000_000) * pricing.outputPerMTok;
+    // Round to micro-dollars to avoid floating-point noise in the score.
+    const costUsd = Math.round(rawCostUsd * 1_000_000) / 1_000_000;
+    await step.score('score-query-writer-cost', {
+      name: 'query_writer_cost_usd',
+      value: costUsd,
     });
```
djfarrelly, line 223: side note - this makes me think that batch scoring calls would be nice in the future as an array.

djfarrelly (approved):

PR #4344, opened 2026-06-05
feat: promote api cli, add redirect message and beta tag

jacobheric:
## Description

Take it out alpha, redirect any existing users who might try to use it on alpha:

```
NAME:
   inngest alpha api - Moved to inngest api

USAGE:
   inngest alpha api

DESCRIPTION:
   The alpha api command has moved. Use `inngest api` instead.
```

and mark it as beta:

```
NAME:
   inngest api - Call Inngest REST API v2 endpoints (beta)

USAGE:
   inngest api [target/auth flags] <endpoint> [endpoint flags]

DESCRIPTION:
   Beta: this command is under active development and may change.
   By default, the command targets the local dev server.
   Set --prod to target Inngest Cloud Production, or --api-host/--api-port to target a custom API server.
```

## Release note

The API CLI has moved out of alpha and is now available as a top-level beta command:

`npx inngest-cli@latest api --help`

## Migration note

None.

### DIFF
--- a/cmd/apiv2cli/cmd.go
+++ b/cmd/apiv2cli/cmd.go
@@ -58,14 +58,28 @@ type endpoint struct {
 func Command() *cli.Command {
 	return &cli.Command{
 		Name:      "api",
-		Usage:     "Call Inngest REST API v2 endpoints",
-		UsageText: "inngest alpha api [target/auth flags] <endpoint> [endpoint flags]",
+		Usage:     "Call Inngest REST API v2 endpoints (beta)",
+		UsageText: "inngest api [target/auth flags] <endpoint> [endpoint flags]",
 		Description: strings.Join([]string{
+			"Beta: this command is under active development and may change.",
 			"By default, the command targets the local dev server.",
 			"Set --prod to target Inngest Cloud Production, or --api-host/--api-port to target a custom API server.",
 		}, "\n"),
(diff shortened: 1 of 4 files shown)


cmd/apiv2cli/cmd.go
```diff
@@ -69,6 +70,19 @@ func Command() *cli.Command {
...
+		Usage:       "Moved to inngest api",
+		UsageText:   "inngest alpha api",
+		Description: "The alpha api command has moved. Use `inngest api` instead.",
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			_, err := fmt.Fprintln(cmd.Root().Writer, "The alpha api command has moved. Use `inngest api` instead.")
+			return err
+		},
+	}
+}
+
```
djfarrelly, line 85: I think given it was an alpha we can skip the old command as a placeholder and clean that up - nothing should be relying on it.
```suggestion
```

djfarrelly (approved):
Suggestion left to fully remove the `alpha api` command - your call if we cleanup now or later.

cmd/apiv2cli/cmd.go
jacobheric: I do have users using it from there already so I wanted to give them an in place heads up. Ok if I leave it there and remove it after a week or so?

PR #4352, opened 2026-06-08
refactor: Refactor tel package to remove noise

djfarrelly:
## Description

Simplify the use of telemetry to avoid sending empty metadata events. This ensures there are no empty "cmd" events which lead to lots of noise.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/coreapi/coreapi.go
+++ b/pkg/coreapi/coreapi.go
@@ -200,8 +200,6 @@ func (a CoreAPI) TrackEvent(w http.ResponseWriter, r *http.Request) {
 	eventName, ok := requestBody["eventName"].(string)
 	if ok {
 		tel.SendEvent(ctx, eventName, metadata)
-	} else {
-		tel.SendMetadata(ctx, metadata)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/coreapi/coreapi.go
```diff
@@ -200,8 +200,6 @@ func (a CoreAPI) TrackEvent(w http.ResponseWriter, r *http.Request) {
 	eventName, ok := requestBody["eventName"].(string)
 	if ok {
 		tel.SendEvent(ctx, eventName, metadata)
-	} else {
-		tel.SendMetadata(ctx, metadata)
```
djfarrelly, line 204: We should not send events without a specific event name. This may be the culprit of all of the empty noise.

jakobevangelista (approved):

jakobevangelista (commented):
non blocking:

this fixes empty cmd on`cli/command.executed` but non command events like `cli/dev_ui.loaded` can still include `cmd: ""`, since `Metadata.Event` always emits the field

seems okay for this pr to stop the command event noise, but maybe something to look into later

PR #4353, opened 2026-06-08
feat: implements v2 api/cli get function

jacobheric:
## Description

`curl http://localhost:8288/v2/apps/inngest-ai-dev/functions/inngest-ai-dev-hello-world` 
and
`npx inngest-cli@latest api get-function inngest-ai-dev hello-world`

```json
{
  "data": {
    "app": {
      "id": "inngest-ai-dev"
    },
    "configuration": {
      "concurrency": [
        {
          "limit": {
            "isPlanLimit": true,
            "value": -1
          },
          "scope": "ACCOUNT"
        }
      ],
      "retries": {
        "isDefault": true,
        "value": 4
      }
    },
    "id": "hello-world",
    "name": "hello-world",
    "slug": "hello-world",
    "triggers": [
      {
        "type": "EVENT",
        "value": "test/hello.world"
      }
    ]
  },
  "metadata": {
    "fetchedAt": "2026-06-08T22:42:52.322777Z"
  }
}
```

## Release note

Adds `GET /v2/functions/{id}` to the V2 API and CLI:

`npx inngest-cli@latest api get-function my-app-hello-world`

## Migration note

None.


djfarrelly:
One question about this @jacobheric is if we could somehow offer an option for passing app id and function id - in the user's codebase we only have them separately and joining them together is only something that happens after the app is synced to the server.

jacobheric:
> One question about this @jacobheric is if we could somehow offer an option for passing app id and function id - in the user's codebase we only have them separately and joining them together is only something that happens after the app is synced to the server.

great point. and I see that is how we did function invoke. I'll do the same.

proto/api/v2/service.proto
```diff
@@ -1619,6 +1637,117 @@ message AppRef {
...
+message FunctionApp {
+  string id = 1;
+  reserved 2, 3, 4;
+  reserved "external_id", "name", "latest_sync";
+}
+
+message FunctionTrigger {
+  FunctionTriggerType type = 1;
+  string value = 2;
+  optional string condition = 3;
```
djfarrelly: **question**: What is the context behind "condition" vs. "if" matching the SDK? I don't mind it, just trying to understand any potential new term or point of confusion between code, docs and APIs.

jacobheric: I borrowed "condition" from our graphql implementation! I've been trying to have the rest api match the ui as closely as possible. But it turns out "condition" is internal to graphql and we don't use that in the ui. I've switched to "if" here.

djfarrelly (approved):

PR #4361, opened 2026-06-09
fix: v2 api event/runs statuses

jacobheric:
## Description

Addresses some incorrect status issues reported by our early adopters. This hoists the main status check out of the sql (shared by multiple queries). It also better handles intermediate run span states so we accurately reflect state as we move from qeueud to running and running to completed. 

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/db/postgres/sqlc/queries.sql
+++ b/pkg/db/postgres/sqlc/queries.sql
@@ -223,7 +223,12 @@ LEFT JOIN LATERAL (
 ) run_start ON true
 LEFT JOIN LATERAL (
   SELECT
-    COALESCE((status_spans.attributes#>>'{}')::json->>'sys.function.status.code', status_spans.status, '')::TEXT AS status,
+    COALESCE(
+      (status_spans.attributes#>>'{}')::json->>'sys.function.status.code',
+      (status_spans.attributes#>>'{}')::json->>'_inngest.dynamic.status',
+      status_spans.status,
+      ''
+    )::TEXT AS status,
     end_time
   FROM spans status_spans
   WHERE status_spans.run_id = spans.run_id
@@ -233,7 +238,15 @@ LEFT JOIN LATERAL (
     -- run status.
     AND (
       COALESCE((status_spans.attributes#>>'{}')::json->>'sys.function.status.code', '') <> ''
-      OR COALESCE((status_spans.attributes#>>'{}')::json->>'_inngest.is.function.output', '') = 'true'
+      OR (
+        COALESCE((status_spans.attributes#>>'{}')::json->>'_inngest.is.function.output', '') IN ('true', '1')
+        AND COALESCE((status_spans.attributes#>>'{}')::json->>'_inngest.retryable', '') NOT IN ('true', '1')
+      )
+      -- Terminal runs may only update the root dynamic status.
+      OR (
+        COALESCE((status_spans.attributes#>>'{}')::json->>'_inngest.dynamic.span.id', '') = spans.span_id
+        AND COALESCE((status_spans.attributes#>>'{}')::json->>'_inngest.dynamic.status', status_spans.status, '') IN ('Completed', 'Failed', 'Errored', 'Cancelled', 'TimedOut', 'Skipped')
+      )
     )
   ORDER BY status_spans.end_time DESC
   LIMIT 1
@@ -245,7 +258,7 @@ LEFT JOIN LATERAL (
     AND output_spans.debug_run_id IS NULL
     AND output_spans.output IS NOT NULL
   ORDER BY
-    COALESCE((output_spans.attributes#>>'{}')::json->>'_inngest.is.function.output', '') != 'true',
+    COALESCE((output_spans.attributes#>>'{}')::json->>'_inngest.is.function.output', '') NOT IN ('true', '1'),
     output_spans.end_time DESC
   LIMIT 1
 ) run_output ON true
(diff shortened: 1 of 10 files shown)


pkg/db/postgres/sqlc/queries.sql
```diff
@@ -200,7 +200,8 @@ SELECT
   spans.app_id,
   COALESCE(run_start.start_time, spans.start_time) AS start_time,
   COALESCE(run_status.end_time, spans.end_time) AS end_time,
```
rhino1998: Is this correct? Isn't `spans` the first span on the run? (I think `run_status.end_time IS NOT NULL` if `spans.end_time IS NOT NULL` though, so maybe the COALESCE doesn't matter?)

pkg/db/postgres/sqlc/queries.sql
```diff
@@ -200,7 +200,8 @@ SELECT
...
-  COALESCE(run_status.status, spans.status, '') AS status,
+  -- Child spans mean user code has started even if the root run span still says queued.
+  COALESCE(run_status.status, CASE WHEN run_start.start_time IS NOT NULL THEN 'Running' END, spans.status, '') AS status,
```
rhino1998: I think this could be wrong if the run still has status Queued but the child span we extract has status Completed. Maybe if `spans.status IS NOT NULL, return RUNNING?`

pkg/db/postgres/sqlc/queries.sql
```diff
@@ -221,11 +222,19 @@ LEFT JOIN LATERAL (
   LIMIT 1
 ) run_start ON true
 LEFT JOIN LATERAL (
-  SELECT status, end_time
+  SELECT
+    COALESCE((status_spans.attributes#>>'{}')::json->>'sys.function.status.code', status_spans.status, '')::TEXT AS status,
```
rhino1998: I think this is the wrong status attribute name? That looks like the old traces one. I'd expect `_inngest.dynamic.status` Looking locally at least, I don't see that attribute on my spans

pkg/db/postgres/sqlc/queries.sql
```diff
@@ -221,11 +222,19 @@ LEFT JOIN LATERAL (
...
+    end_time
   FROM spans status_spans
   WHERE status_spans.run_id = spans.run_id
     AND status_spans.debug_run_id IS NULL
-    AND status_spans.status IN ('Completed', 'Failed', 'Errored', 'Cancelled', 'TimedOut')
+    -- Step spans can finish while the run is still active; only function
+    -- status code spans or final function-output spans should decide terminal
+    -- run status.
+    AND (
+      COALESCE((status_spans.attributes#>>'{}')::json->>'sys.function.status.code', '') <> ''
```
rhino1998: Same thing the attribute name as above

pkg/db/sqlite/sqlc/queries.sql
```diff
@@ -212,13 +222,19 @@ run_status AS (
   FROM (
     SELECT
       status_spans.run_id,
-      status_spans.status,
+      COALESCE(CAST(status_spans.attributes->>'$."sys.function.status.code"' AS TEXT), status_spans.status, '') AS status,
```
rhino1998: Same thing again about the attribute name

pkg/db/sqlite/sqlc/queries.sql
```diff
@@ -212,13 +222,19 @@ run_status AS (
...
       ROW_NUMBER() OVER (PARTITION BY status_spans.run_id ORDER BY status_spans.end_time DESC) AS rn
     FROM spans status_spans
     JOIN run_roots ON run_roots.run_id = status_spans.run_id
     WHERE status_spans.debug_run_id IS NULL
-      AND status_spans.status IN ('Completed', 'Failed', 'Errored', 'Cancelled', 'TimedOut')
+      -- Step spans can finish while the run is still active; only function
+      -- status code spans or final function-output spans should decide terminal
+      -- run status.
+      AND (
+        COALESCE(CAST(status_spans.attributes->>'$."sys.function.status.code"' AS TEXT), '') <> ''
```
rhino1998: Again

rhino1998 (commented):
Some notes about attributes & the coalesce logic

pkg/db/postgres/sqlc/queries.sql
jacobheric: you're right. I reworked the query to make it better mirror the logic in GetSpanRuns from gql. the run is now `executor.run` root plus the EXTEND rows sharing the root's dynamic_span_id and `end_time` is now `MAX(end_time)` and `start_time` is MIN(start_time) over the same group. and then we only use the ended at above when the run status says it's over (same as MakeFunctionRunV2).

pkg/db/postgres/sqlc/queries.sql
jacobheric: I removed it. status now comes only from the run's root dynamic-span group. we get the latest by end time. again I think this is the same as `GetSpanRuns`.

pkg/db/postgres/sqlc/queries.sql
jacobheric: Yes, that's v1, nice catch. There is some weird promotion of statuses by dynamic span id going on that I sidestepped by reading status directly from the root's dynamic span group.

jacobheric:
@rhino1998 thanks for the careful review!

rhino1998 (approved):

PR #4364, opened 2026-06-09
fix: Fudge planned step span times slightly to prevent rollup race condition

rhino1998:
## Description

Currently if the StepPlanned and actual step spans for parallel execution are emitted the same millisecond timestamp, they roll up incorrectly. This fudges the step span start_time (not the attribute timestamps) to prevent this from happening. This really only happens with no-op steps in the integration tests as otherwise there's almost always at least 1ms of latency.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -4051,7 +4051,9 @@ func (e *executor) handleGeneratorStepPlanned(ctx context.Context, runCtx execut
 
 	// Re-enqueue the exact same edge to run now.
 	jobID := fmt.Sprintf("%s-%s", runCtx.Metadata().IdempotencyKey(), gen.ID+"-plan")
-	now := e.now()
+	// NOTE: we fudge the time to be slightly in the past so that ultra-low-latency step executions don't return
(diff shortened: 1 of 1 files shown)


pkg/execution/executor/executor.go
```diff
@@ -4051,7 +4051,9 @@ func (e *executor) handleGeneratorStepPlanned(ctx context.Context, runCtx execut
 
 	// Re-enqueue the exact same edge to run now.
 	jobID := fmt.Sprintf("%s-%s", runCtx.Metadata().IdempotencyKey(), gen.ID+"-plan")
-	now := e.now()
+	// NOTE: we fudge the time to be slightly in the past so that ultra-low-latency step executions don't return
+	// with the same timestamp as the discovery step, which can cause issues with span ordering for rollup.
+	now := e.now().Add(-1 * time.Millisecond)
```
scottnuma: It might be nice to rename `now` to reflect that we've messed with it, maybe `adjustedSpanStartTime` or `fudgedNow`

scottnuma (approved):

PR #4371, opened 2026-06-09
feat(scoring): Scoring Dashboard UI

Linell:
## Description

Adds a Scoring page to the cloud dashboard, based on the design [in Figma](https://www.figma.com/design/E23uQifkuokz5e7VhrLeQS/Scoring?node-id=8902-614&p=f&t=2OlcwYxVFz38xrB3-0).

Screenshot 2026-06-11 at 11 22 15 AM

## Release note

None.

## Migration note

None.


ui/apps/dashboard/src/components/Scores/ScoreCard.tsx
```diff
@@ -0,0 +1,185 @@
...
+  </TabRow>
+);
+
+const TabRow = ({ children }: React.PropsWithChildren) => (
+  <div className="border-subtle mb-2 flex flex-row gap-4 border-b">
+    {children}
+  </div>
+);
+
+const TabButton = ({
```
anafilipadealmeida, line 164: Are these Tabs very different from the Tabs shared component? We shouldn't need to keep defining these types of components every time.

ui/apps/dashboard/src/components/Scores/Dashboard.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const timeRange = useMemo(
+    () => ({ from: range.from.toISOString(), to: range.to.toISOString() }),
+    [range],
+  );
+
+  const [{ data: lookupData, fetching: lookupFetching, error: lookupError }] =
+    useQuery({
+      query: ScoresLookupDocument,
+      variables: { envSlug, page: 1, pageSize: 1000 },
```
anafilipadealmeida, line 127: This seems like a big page number. Do we have any performance concerns?

ui/apps/dashboard/src/components/Scores/Dashboard.tsx
```diff
@@ -0,0 +1,264 @@
...
+                key={s.name}
+                name={s.name}
+                series={seriesByName.get(s.name)}
+                isLoading={seriesFetching}
+                error={seriesError}
+              />
+            ))
+          )}
+        </div>
+        <Legend
```
anafilipadealmeida: The Legend should probably not be displayed if there's an error

anafilipadealmeida (approved):
Approved so I don't block the progress but left some comments!

ui/apps/dashboard/src/components/Scores/Dashboard.tsx
Linell: That's vaguely copy/pated from `Metrics/Dashboard.tsx:121` and it's just this lightweight query:

```ts
const ScoresLookupDocument = graphql(`
  query ScoresLookup($envSlug: String!, $page: Int, $pageSize: Int) {
    envBySlug(slug: $envSlug) {
      workflows @paginated(perPage: $pageSize, page: $page) {
        data {
          name
          id
          slug
        }
      }
    }
  }
`);
```

ui/apps/dashboard/src/components/Scores/ScoreCard.tsx
Linell: The shared Tabs _could_ do it, but I think that they're different enough that it'd be more trouble than it's worth. It _does_ seem like we've got a few different places that use this style so a good followup may be something like a shared `UnderlineTabs` component? I can add that here if you'd rather!

PR #4386, opened 2026-06-11
feat(metadata): Use deterministic step span IDs

rhino1998:
## Description

This changes the metadata API to use deterministic span IDs for span attribution for new runs.

We still use CH for older runs

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/api/apiv1/metadata.go
+++ b/pkg/api/apiv1/metadata.go
@@ -195,18 +195,15 @@ func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID
 
 	default:
 		scope = enums.MetadataScopeExtendedTrace
-		// TODO: require that this is an extended trace span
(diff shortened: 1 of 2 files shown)


pkg/api/apiv1/metadata.go
```diff
@@ -113,6 +118,142 @@ type AddRunMetadataRequest struct {
...
+			stateMetadata = &md
+			loadedFromState = true
+		}
+	}
+
+	// Missing state uses a request-local fallback so this write still enforces
+	// the cumulative size limit within the request.
+	if stateMetadata == nil {
+		stateMetadata = &statev2.Metadata{ID: partialID}
+	}
```
scottnuma, line 163: nit: setting this state and the comment could live together in the errors.Is branch

scottnuma (approved):

pkg/api/apiv1/metadata.go
rhino1998: if a.opts.State == nil, I don't think that is valid (although I don't know if a.opts.State is ever nil in practice)

PR #4389, opened 2026-06-11
feat: v2 rest api get app

jacobheric:
## Description

Add v2 rest api get app, see usage below. 

response looks like:
```
{
  "data": {
    "createdAt": "2026-03-05T16:09:19Z",
    "functionCount": 28,
    "id": "inngest-ai-dev",
    "latestSync": {
      "framework": "deno/fresh",
      "sdkLanguage": "js",
      "sdkVersion": "v4.4.0",
      "url": "http://localhost:8000/api/inngest"
    },
    "method": "SERVE",
    "name": "inngest-ai-dev"
  },
  "metadata": {
    "fetchedAt": "2026-06-11T13:09:38.869421Z"
  }
}
```

## Release note

Added REST API v2 support for fetching a single app by it id.

`GET /v2/apps/{id}`

`npx inngest-cli@latest api get-app my-app --prod`

## Migration note

None.

### DIFF
--- a/proto/api/v2/service.proto
+++ b/proto/api/v2/service.proto
@@ -1946,22 +1946,17 @@ message App {
 }
 
 message AppSync {
-  reserved 10;
-  reserved "function_count";
+  reserved 2, 3, 10;
+  reserved "started_at", "completed_at", "function_count";
 
   optional string status = 1 [
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
       description: "Status of the latest sync"
     }
   ];
-  optional google.protobuf.Timestamp started_at = 2 [
+  optional google.protobuf.Timestamp synced_at = 11 [
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      description: "Timestamp when the latest sync started"
-    }
-  ];
-  optional google.protobuf.Timestamp completed_at = 3 [
-    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      description: "Timestamp when the latest sync completed, if recorded"
+      description: "Timestamp for the latest sync, if available"
     }
   ];
   optional string sdk_language = 4 [
--- a/proto/api/v2/service.proto
+++ b/proto/api/v2/service.proto
@@ -975,6 +975,24 @@ service V2 {
     };
   }
 
+  rpc GetApp(GetAppRequest) returns (GetAppResponse) {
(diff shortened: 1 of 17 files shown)


proto/api/v2/service.proto
```diff
@@ -1849,6 +1867,137 @@ message GetFunctionResponse {
...
+  optional google.protobuf.Timestamp created_at = 2 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Timestamp when the latest sync was created"
+    }
+  ];
+  optional google.protobuf.Timestamp last_synced_at = 3 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Timestamp when the latest sync completed"
+    }
+  ];
```
djfarrelly: question: whats the difference between these two? is this the de-duping in action when config isn't changed and a resync happens?

proto/api/v2/service.proto
```diff
@@ -1849,6 +1867,137 @@ message GetFunctionResponse {
...
+      description: "URL of the Inngest endpoint in the app, if synced via URL"
+      example: "\"https://example.com/api/inngest\""
+    }
+  ];
+  optional string error = 8 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Most recent sync or connectivity error for the app, if any"
+    }
+  ];
+  optional string app_version = 9 [
```
djfarrelly, line 1991: thought (non-blocking): this is smart to add here! makes me think that we should respond with this at some point in the app sync endpoint so the user can confirm the synced version.

djfarrelly (approved):

proto/api/v2/service.proto
jacobheric: I checked the codepaths and we do track when we start AND when we complete sync. so I changed these to `started_at` and `completed_at`

proto/api/v2/service.proto
jacobheric: I ticketed this.

proto/api/v2/service.proto
jacobheric: actually, scratch that. completedAt is not reliably set on monorepo and neither are readily available in dev server right now. so I'll just use a single syncedAt that will show up in monorepo. that will be useful for debugging. I'll ticket figuring out why we don't have it dev server so that self hosters can have it to at some point.

PR #4391, opened 2026-06-11
feat(queue): Track earliest peek time before Constraint API

BrunoScheufler:
### Description

This PR introduces a new way to track earliest lease time, correcting latency metrics.

image

#### Why?

- We need a good latency indicator for queue performance
- Latency can have different reasons, so it should be split into system vs user latency
- System latency measures the time to job discovery, which is the earliest peek time - desired time to run the job
- User latency measures the time between job discovery and leasing the item, after acquiring the required capacity
- Ideally, system latency is close to 0, this means the item is picked up as close to the desired run time as possible
- Ideally, user latency is close to 0, this means capacity is acquired as close to job discovery as possible

#### How?
- Currently, earliest peek time is set during Lease. When there's no capacity (i.e. Acquire does not return a valid capacity lease), we will skip Lease, so earliest peek time is never set
- Whenever we peek an item, we need to set the earliest peek time before running constraint checks.
- Of course, if there's no capacity, the iterator will potentially stop and requeue the partition before hitting all items, so earliest peek time will be in the future as well, even if the item was returned in the initial peek set.
- This is fixed by iterating over all remaining items and setting earliest peek time, in case of the early exit
- In case of large peek sizes, iterating over every item might get expensive, which is why we cap the maximum number using an option

This PR introduces a new lightweight earliest peek time key that is set during processor iteration if doesn't already exist. We then use that value to set the earliest peek time on the item which is used for metrics.

<details><summary>What does partition.Last do?</summary>
<p>

 partition.Last is the millisecond timestamp of the last
  successful partition lease.

  The authoritative update happens in Redis during PartitionLease:

  - partitionLease.lua reads the existing value into existingTime
  - then sets existing.last = currentTime
  - stores the updated partition JSON back into the partition hash

  See pkg/execution/state/redis_state/lua/queue/
  partitionLease.lua:37.

  Subtle part: the Lua script returns the previous existing.last,
  not the new currentTime. Go then only copies that returned
  previous value onto the in-memory p.Last:

  pkg/execution/state/redis_state/queue.go:1158

  So in the current ProcessPartition call, p.Last is effectively
  “the previous time this partition was leased before this lease.”
  The current lease time is persisted to Redis and will be visible
  when the partition is peeked again later.

  Other paths mostly preserve it:

  - Enqueue creates the partition if missing, with Last defaulting
    to 0.

  - Requeue updates lease/score/force metadata but does not change
    last.

  - Partition peek just reads the persisted partition JSON, so it
    sees whatever the last successful lease wrote.

</p>
</details> 

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/queue/processor_iterator.go
+++ b/pkg/execution/queue/processor_iterator.go
@@ -146,6 +146,21 @@ func (p *ProcessorIterator) Process(ctx context.Context, item *QueueItem) error 
 	item.Data.At = time.UnixMilli(item.AtMS)
 	item.Data.EnqueuedAt = time.UnixMilli(item.EnqueuedAt)
 
+	if item.EarliestPeekTime == 0 && p.Queue.Options().ItemEarliestPeekTimeConfig(ctx, p.Queue.Shard().Name(), *item).Enabled {
+		peekTime := p.StaticTime
+		if peekTime.IsZero() {
+			peekTime = p.Queue.Clock().Now()
+		}
+
+		earliestPeekTime, err := p.Queue.Shard().SetEarliestPeekTime(ctx, *item, peekTime)
+		if err != nil {
+			span.RecordError(err)
+			l.Warn("could not set earliest peek time", "error", err)
+		} else {
+			item.EarliestPeekTime = earliestPeekTime.UnixMilli()
+		}
+	}
+
 	if item.IsLeased(p.Queue.Clock().Now()) {
 		span.SetAttributes(attribute.String("skip_reason", "already_leased"))
 		metrics.IncrQueueItemProcessedCounter(ctx, metrics.CounterOpt{
--- a/pkg/execution/queue/processor_iterator.go
+++ b/pkg/execution/queue/processor_iterator.go
@@ -98,9 +98,15 @@ func (p *ProcessorIterator) Iterate(ctx context.Context) error {
 			if errors.Is(err, ErrQueueItemNotFound) {
 				continue
 			}
+
+			// If item processing was terminated early due to user constraints,
+			// set the earliest peek time for the remaining items.
+			// This ensures accurate tracking of peeked items waiting for
+			// user constraint capacity to become available (user latency).
 			if errors.Is(err, ErrProcessNoUserConstraintCapacity) {
 				config := p.Queue.Options().ItemEarliestPeekTimeConfig(ctx, p.Queue.Shard().Name(), *i)
 				if config.Enabled {
+					// Stamp remaining items from next item on (idx+1)
 					p.stampRemainingEarliestPeekTimes(ctx, idx+1, config.BulkStampLimit)
 				}
 			}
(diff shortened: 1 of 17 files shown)


pkg/execution/queue/processor_iterator.go
dlambrig: This fallback appears to overwrite a valid earliest-peek timestamp populated earlier by SetEarliestPeekTime. Since the comment says this path is for when the peek time wasn't set, should this be guarded by item.EarliestPeekTime == 0?

pkg/execution/queue/processor_iterator.go
```diff
@@ -98,6 +98,18 @@ func (p *ProcessorIterator) Iterate(ctx context.Context) error {
...
+
+			// If item processing was terminated early due to user constraints,
+			// set the earliest peek time for the remaining items.
+			// This ensures accurate tracking of peeked items waiting for
+			// user constraint capacity to become available (user latency).
+			if errors.Is(err, ErrProcessNoUserConstraintCapacity) {
+				config := p.Queue.Options().ItemEarliestPeekTimeConfig(ctx, p.Queue.Shard().Name(), *i)
+				if config.Enabled {
+					// Stamp remaining items from next item on (idx+1)
+					p.stampRemainingEarliestPeekTimes(ctx, idx+1, config.BulkStampLimit)
```
dlambrig, line 110: can we document or sanity-check the expected BulkStampLimit values? On a hot partition that is repeatedly function/account-concurrency blocked, this can add up to BulkStampLimit extra Redis writes per iterator pass. The writes are NX so correctness is fine, but this could still become noticeable Redis traffic if enabled broadly.

dlambrig (commented):
two comments

dlambrig:
The code emits `queue_partition_process_no_capacity_total` when local worker capacity is unavailable, but I don’t see that metric referenced in the inngest/infra dashboards/alerts.

Since this path intentionally avoids stamping earliest-peek time, should we add or confirm dashboard/alert coverage for this metric? i.e to distinguish “item was ready, but no queue worker was available” from other system-side queue delay.

Also, do you think it would it be useful to add a separate duration metric for local queue worker wait time?

pkg/execution/queue/processor_iterator.go
BrunoScheufler: Those would likely be set in feature flags but you're right: We want to monitor how much (compute) time we're spending on stamping remaining items.

BrunoScheufler:
> The code emits `queue_partition_process_no_capacity_total` when local worker capacity is unavailable, but I don’t see that metric referenced in the inngest/infra dashboards/alerts.
> 
> Since this path intentionally avoids stamping earliest-peek time, should we add or confirm dashboard/alert coverage for this metric? i.e to distinguish “item was ready, but no queue worker was available” from other system-side queue delay.
> 
> Also, do you think it would it be useful to add a separate duration metric for local queue worker wait time?

We do have a graph in DD but it needs to be more prominent and we should have alerts on growing rates of this error, as this means we should be scaling up. This will probably be handled by our autoscaling efforts.

dlambrig (approved):

Issue #4123, opened 2026-05-05
Self-hosted Postgres pool settings are not applied to shared sql.DB


feliche93:
**Describe the bug**
Self-hosted Inngest exposes Postgres pool settings such as:

- `INNGEST_POSTGRES_MAX_IDLE_CONNS`
- `INNGEST_POSTGRES_MAX_OPEN_CONNS`
- `INNGEST_POSTGRES_CONN_MAX_IDLE_TIME`
- `INNGEST_POSTGRES_CONN_MAX_LIFETIME`

In `v1.19.1`, these settings do not appear to reach the shared `database/sql` `*sql.DB` created for Postgres-backed CQRS/runtime storage. The runtime therefore behaves like a default Go `database/sql` pool (`MaxOpenConns == 0`, effectively unlimited, with very low idle retention), which can create high PgBouncer client login churn under queue worker load.

**Observed impact**
We noticed this in a self-hosted deployment of Inngest `v1.19.1` using Postgres behind PgBouncer with 100+ queue workers. During bursty function execution, PgBouncer showed tens of thousands of short-lived logins from `user=inngest`. While this churn was happening, Inngest/API function calls intermittently failed with TLS/request errors.

A local patch in our app repo that applies these pool settings directly to the shared `*sql.DB` before the first `Ping()` dramatically reduced PgBouncer login churn.

**Regression notes**
This appears to have been introduced by `64f35c775` / #3894 (`Wire db.Adapter into base_cqrs (Phase 3)`).

The original Postgres pool settings were added in `23f9c9149` / #2782 (`Add Postgres connection params for self-hosting`). That implementation passed `sqlc_postgres.NewNormalizedOpts` through `base_cqrs.NewCQRS`, `NewHistoryDriver`, and `NewQueries`, and `sqlc/postgres.NewNormalized(...)` called:

- `SetMaxIdleConns(...)`
- `SetMaxOpenConns(...)`
- `SetConnMaxIdleTime(...)`
- `SetConnMaxLifetime(...)`

#3894 then moved `base_cqrs` to `db.Adapter` and removed the `NewNormalizedOpts` constructor path from the runtime wiring. The shared `*sql.DB` is still opened centrally in `pkg/cqrs/base_cqrs/base_cqrs.go`, but the Postgres pool options are no longer propagated there. `v1.19.1` contains #3894.

[...]

PR #4392, opened 2026-06-11
fix(devserver): Reapply postgres connection pool settings

scottnuma:
## Description

- Mendral identified that one of our top CI flakes is [this one](https://app.mendral.com/insights/01KMP8EJV264SNJFKJ9KEGE22Q) where we're hitting test timeouts. In particular, "Postgres variants are worst".
- Investigated and identified that we're not properly passing Postgres configurations, in addition to other contributors.
- We regressed in #3894 during some rewiring
- Rewire and add a test for regressions
- Also wire up ENV VAR support for these pieces of config for self-hosted Inngest

## Release note

Fixed a bug where we were improperly configuring Postgres connection pool settings for the dev server and self-hosted Inngest.

Fixed a bug ignoring non-flag Postgres connection pool configuration methods (such as env vars and config files) for self-hosted Inngest.

## Migration note

For users with Postgres, the dev server and self-hosted Inngest will move its connection pooling settings from from 

    2 max idle connections
    unlimited max open connections
    unlimited  connection max idle time
    unlimited connection max lifetime 

to the configured settings. 

These settings can be configured via flags, which have always been supported, or environment variables and config files, which were not previously supported before for self-hosted Inngest.

When a setting is not specified, it will use the default. The default settings are 

      10 max idle connections
      100 max open connections
      5m connection max idle time
      30m connection max lifetime

### DIFF
--- a/pkg/db/postgres/migrations.go
+++ b/pkg/db/postgres/migrations.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/inngest/inngest/pkg/logger"
 	_ "github.com/jackc/pgx/v5/stdlib"
@@ -30,6 +31,26 @@ type Options struct {
 	// purposes. By default database handlers are singletons, but when this flag
 	// is enabled, each call creates a new connection.
 	ForTest bool
+
+	// MaxIdleConns sets the number of idle connections retained in the pool.
+	//
+	// Zero leaves the database/sql default.
+	MaxIdleConns int
+
+	// MaxOpenConns bounds the total connections the pool may open.
+	//
+	// Zero leaves the database/sql default.
+	MaxOpenConns int
+
+	// ConnMaxIdleTime sets how long an idle connection is retained.
+	//
+	// Zero leaves the database/sql default.
+	ConnMaxIdleTime time.Duration
+
+	// ConnMaxLifetime sets the maximum lifetime of a connection.
+	//
+	// Zero leaves the database/sql default.
+	ConnMaxLifetime time.Duration
 }
 
 func Open(ctx context.Context, opts Options) (*sql.DB, error) {
@@ -61,6 +82,8 @@ func Open(ctx context.Context, opts Options) (*sql.DB, error) {
 		return nil, err
 	}
 
+	configurePool(conn, opts)
+
 	if err := conn.Ping(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/db/postgres/migrations.go
```diff
@@ -73,6 +96,23 @@ func Open(ctx context.Context, opts Options) (*sql.DB, error) {
 	return conn, nil
 }
 
+// configurePool applies any non-zero pool options, leaving database/sql
+// defaults otherwise.
```
scottnuma, line 100: This introduces slightly new behavior than before - previous direct callers (mostly tests rather than users) would pass an empty Options{} and thus all zeros. This gave them no idle pooling rather than a default of 2. Idle pooling is nice to start new connections more quickly.

pkg/devserver/devserver.go
```diff
@@ -195,7 +195,14 @@ func start(ctx context.Context, opts StartOpts) error {
...
 	if opts.PostgresURI != "" {
-		db, err := dbpostgres.Open(ctx, dbpostgres.Options{URI: opts.PostgresURI})
+		// Pool settings are in minutes, mirroring the postgres-conn-max-* flags.
+		db, err := dbpostgres.Open(ctx, dbpostgres.Options{
+			URI:             opts.PostgresURI,
+			MaxIdleConns:    opts.PostgresMaxIdleConns,
+			MaxOpenConns:    opts.PostgresMaxOpenConns,
+			ConnMaxIdleTime: time.Duration(opts.PostgresConnMaxIdleTime) * time.Minute,
+			ConnMaxLifetime: time.Duration(opts.PostgresConnMaxLifetime) * time.Minute,
+		})
```
scottnuma, line 205: Users upgrading will see the devserver utilize Postgres differently, likely slightly better given we're giving an actual max connections bound. It is possible that performance could regress though. Are my call outs in the PR description enough?

pkg/db/postgres/migrations_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	t.Run("applies explicit options", func(t *testing.T) {
+		conn := open(t)
+		configurePool(conn, Options{
+			MaxIdleConns:    3,
+			MaxOpenConns:    7,
+			ConnMaxIdleTime: time.Minute,
+			ConnMaxLifetime: time.Hour,
+		})
+		require.Equal(t, 7, conn.Stats().MaxOpenConnections)
```
albertchae, line 35: any reason we're not checking the other 3 values? both here and in the default test?

pkg/devserver/devserver.go
albertchae: I think it'd be good to explicitly mention which users would be affected. Would it only be users using postgres in their devserver and setting any of `INNGEST_POSTGRES_MAX_IDLE_CONNS`, `INNGEST_POSTGRES_MAX_OPEN_CONNS`, `INNGEST_POSTGRES_CONN_MAX_IDLE_TIME`, `INNGEST_POSTGRES_CONN_MAX_LIFETIME`?

albertchae (approved):
This PR seems good but https://github.com/inngest/inngest/issues/4123 is definitely related and has some additional points not covered by this PR as is. It looks like configuring these settings by environment variables never worked for self hosted but works for dev server so we might want to fix that (maybe a separate PR if you think that's easier).

pkg/db/postgres/migrations_test.go
scottnuma: Checking more seems good, will add.

scottnuma: With a closer look, MaxOpenConnections is the only field of those that conn.Stats() exposes.

I think unit testing the single and visual inspection of the other setters is probably the right effort:value here

scottnuma:
I'll fix the env vars here too, it's a small change and will be nice to make sure this all rolls out together

pkg/devserver/devserver.go
scottnuma: Updated the migration note to dev server vs. self-hosted and that we will now support env vars and config files uniformly now.

PR #4403, opened 2026-06-13
feat: Add sessions dashboard UI

jakobevangelista:
## Description

sessions pages in the dashboard, behind the `sessions-ui` flag:
- list: enter a session key → its session ids with run count, failure rate, last active, functions
- detail: runs for a key + id (read-only runs list)

reads the cloud `sessions`/`sessionRuns`, depends on cloud for the api, no dependency on the oss go changes (#4404)

Screenshot 2026-06-01 at 12 13 49 AM
Screenshot 2026-06-01 at 12 13 58 AM
Screenshot 2026-06-16 at 3 59 14 PM

## Release note

None.

## Migration note

None.


ui/packages/components/src/Sessions/SessionResults.tsx
```diff
@@ -0,0 +1,165 @@
...
+    queryFn: () => getSessions({ sessionKey, startTime, endTime }),
+    refetchOnWindowFocus: false,
+  });
+
+  const sessions = useMemo(() => data ?? [], [data]);
+
+  return (
+    <div className="bg-canvasBase text-basis flex flex-1 flex-col overflow-hidden focus-visible:outline-none">
+      <div className="flex flex-col gap-4 px-3 pb-3 pt-6">
+        <h1 className="text-basis text-xl font-medium">{sessionKey} Results</h1>
```
anafilipadealmeida: Given that the `sessionKey` is an attribute, this should be styled as mono

ui/packages/components/src/Sessions/SessionsEmptyState.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { RiExternalLinkLine, RiSearchLine } from '@remixicon/react';
+
+type SessionsEmptyStateProps = {
+  onSubmit: (sessionKey: string) => void;
+  docsUrl?: string;
+};
+
+export function SessionsEmptyState({
+  onSubmit,
+  docsUrl = 'https://www.inngest.com/docs',
```
anafilipadealmeida: I am assuming we'll wire this to the docs once we have them, we should have a TODO comment so we don't forget it. At that point, the URL should have a ref so we can track where users came from

anafilipadealmeida (approved):

PR #4404, opened 2026-06-13
feat: session server parse + emit

jakobevangelista:
## Description

server-side sessions: parse `sessions` off event payloads and emit the run's sessions on the root span attr so related runs can be grouped

the read path is in the cloud pr, the send surface is the sdk pr (inngest/inngest-js#1547), and the dashboard ui is #4403. doesn't add anything to the ingest path, it's a span attr insights already captures

- a run's sessions = the union of its triggering events' sessions (batched runs union all of them). kept as `(key, id)` pairs, not a map, so a run can be in multiple sessions under the same key
- `normalizeRunSessions` sorts (deterministic across retries) + dedupes + caps. over the cap we drop + warn, never error
- two caps: 5 per event (at ingest), 25 per run (the post-batch union)
- booleans rejected, numbers coerced to strings

convo [here](https://inngest.slack.com/archives/C05QSAYPEG0/p1781721674384799?thread_ts=1781717694.974469&cid=C05QSAYPEG0) about top level `meta` changes
## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -123,6 +123,18 @@ const (
 	// MaxEvents is the maximum number of events we can parse in a single batch.
 	MaxEvents = 5_000
 
+	// MaxEventSessions is the maximum number of sessions on a single event.
+	MaxEventSessions = 5
+	// MaxRunSessions is the maximum number of sessions on a single run. A run
+	// aggregates the sessions of every event that triggered it (eg. when
+	// batching), so its bound is larger than MaxEventSessions. It keeps the
+	// run's session label a small fraction of the span metadata budget.
+	MaxRunSessions = 25
+	// MaxEventSessionKeyLength is the maximum number of bytes in a session key.
+	MaxEventSessionKeyLength = 128
+	// MaxEventSessionIDLength is the maximum number of bytes in a session ID.
+	MaxEventSessionIDLength = 512
+
 	InngestEventDataPrefix = "_inngest"
 	// InvokeSlugKey is the data key used to store the fn name when invoking a function
 	// via an RPC-like call, abstracting event-driven fanout.
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
@@ -64,8 +62,8 @@ type Event struct {
 	Timestamp int64  `json:"ts,omitempty"`
 	Version   string `json:"v,omitempty"`
 
-	// Sessions groups runs triggered by this event.
-	Sessions Sessions `json:"sessions,omitempty"`
+	// Meta carries event meta shared across runs triggered by this event.
+	Meta EventMeta `json:"meta,omitempty,omitzero"`
 
 	// User represents user-specific information for the event.
 	//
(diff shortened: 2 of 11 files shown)


jakobevangelista (commented):
some notes for reviewers

pkg/execution/executor/executor.go
```diff
@@ -1023,13 +1024,18 @@ func (e *executor) schedule(
...
 	evts := make([]json.RawMessage, len(req.Events))
+	sessions := meta.EventSessions{}
 	for n, item := range req.Events {
 		evt := item.GetEvent()
 		if eventName == nil {
 			name := evt.Name
 			eventName = &name
 		}
 
+		for name, id := range evt.Sessions {
```
jakobevangelista: a run's sessions = the union of its triggering events' sessions

for a batched run that's every event in the batch. kept as `(key, id)` pairs, not a map, so a run can be in multiple sessions under the same key

pkg/execution/executor/executor.go
```diff
@@ -1038,6 +1044,17 @@ func (e *executor) schedule(
 		evts[n] = byt
 	}
 
+	var droppedSessions int
+	sessions, droppedSessions = normalizeRunSessions(sessions)
+	if droppedSessions > 0 {
```
jakobevangelista, line 1049: drop + warn, not error

the events already passed validation at ingest (pun lol) and the run's about to do real work, so failing it over a metadata cap would throw away the whole job

pkg/execution/executor/executor.go
```diff
@@ -6078,3 +6098,29 @@ func updateDeferSpans(
...
+// normalizeRunSessions dedupes, sorts, and caps the session pairs collected
+// from a run's triggering events. Sorting makes the run's session label
+// deterministic across retries, since map and batch iteration order are not.
+// Returns the number of pairs dropped by the cap.
+func normalizeRunSessions(pairs meta.EventSessions) (meta.EventSessions, int) {
+	if len(pairs) == 0 {
+		return nil, 0
+	}
+
+	slices.SortFunc(pairs, func(a, b meta.EventSession) int {
```
jakobevangelista, line 6111: sort makes the run's session label deterministic across retries (map/batch iteration order isn't) and the cap's cut just below is an alphabetical/deterministic slice, not semantic

fine, because hitting the cap means misuse and the warn flags it

pkg/consts/consts.go
```diff
@@ -123,6 +123,18 @@ const (
 	// MaxEvents is the maximum number of events we can parse in a single batch.
 	MaxEvents = 5_000
 
+	// MaxEventSessions is the maximum number of sessions on a single event.
+	MaxEventSessions = 5
```
jakobevangelista, line 127: two caps: 

5 bounds a single event at ingest
25 bounds the post-batch union (batching is the only way to exceed 5 on one run) 

per-field length limits exist despite the event-size cap because sessions get copied onto hot paths 

the span attr budget, a column per run, ui labels, urls

pkg/event/event.go
```diff
@@ -62,6 +64,9 @@ type Event struct {
 	Timestamp int64  `json:"ts,omitempty"`
 	Version   string `json:"v,omitempty"`
 
+	// Sessions groups runs triggered by this event.
+	Sessions Sessions `json:"sessions,omitempty"`
```
tonyhb: tldr from convos:

```
Ctx EventCtx `json:"ctx,omitempty,omitzero"`
```

then we define EventCtx as a struct that contains _only session details right now_

pkg/execution/executor/executor.go
tonyhb: yea, and there needs to be _some_ way of deterministically dropping things for our docs -- alphabetical is a safe bet, and easy to review

pkg/event/event.go
```diff
@@ -145,6 +207,10 @@ func (e Event) Validate(ctx context.Context) error {
 		}
 	}
 
+	if err := e.Sessions.Validate(); err != nil {
+		return err
```
tonyhb, line 157: what happens if one event in an array of 5000 is invalid?  just to check here... i know that's not your code.

jakobevangelista: same as normal event batch ingestion where the api processes the array sequentially and stops on the first invalid item

if item `300` fails validation, the request returns `400`

items before it may already have been accepted/published, and items after it are not processed

pkg/event/event.go
jakobevangelista: https://github.com/inngest/inngest/pull/4404/commits/6e8d0f8f3dbe6662410378f5708737bc360a68d0

tonyhb (approved):
el gee tee em

PR #4410, opened 2026-06-15
feat: v2 api rerun

jacobheric:
## Description

Implements V2 API rerun endpoint: `POST /v2/runs/{run_id}/rerun`.

PR includes some cleanup, e.g. `RunsReader` -> `RunsProvider` to contain contract and provider sprawl. So instead of having a runs reader, writer, etc. we have a single `RunsProvider` like:

```
type RunProvider interface {
	GetRun
	GetRuns
	Rerun
}
```
## Release note

Adds Agentic/V2 REST API endpoint and CLI for `POST /v2/runs/{run_id}/rerun`:

`npx inngest-cli@latest api rerun 01KV5P4M13DQ5V9R3S1DT4JDVQ`

## Migration note

None.


pkg/devserver/run_provider.go
```diff
@@ -0,0 +1,289 @@
+package devserver
```
jacobheric, line 1: FYI, because we collapsed the `runs_reader` and `function_run _reader` into this provider it's not easy for me to get a good diff here. All non rerun implementation is the same as before with some name cleanup e.g. `GetFunctionRun` -> `GetRun` for consistency sake.

pkg/api/v2/endpoints_runs.go
```diff
@@ -112,6 +114,76 @@ func (s *Service) GetEventRuns(ctx context.Context, req *apiv2.GetEventRunsReque
...
+		return nil, s.base.NewError(http.StatusNotImplemented, apiv2base.ErrorNotImplemented, "Rerun is not yet implemented")
+	}
+
+	runID, err := ulid.Parse(req.RunId)
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Run ID must be a valid ULID")
+	}
+
+	if req.FromStep != nil {
+		// TODO: Re-enable once rerun from step reconstruction is fully implemented.
```
jacobheric, line 136: Adding this to the contract now but throwing a 501 until we can merge our open PRs to fix this up.

proto/api/v2/service.proto
```diff
@@ -1837,6 +1856,40 @@ message GetEventRunsResponse {
...
+  ];
+}
+
+message RerunResponse {
+  RerunData data = 1;
+  ResponseMetadata metadata = 2;
+}
+
+message RerunData {
+  string run_id = 1 [
```
djfarrelly, line 1885: Can't see on mobile easily, but confirming that this is the same response shape as Invoke?

djfarrelly (approved):
I like the refactoring - looks good, one note

proto/api/v2/service.proto
jacobheric: yup, exactly the same!

PR #4415, opened 2026-06-15
fix(executor): allow self-hosted crons to pass capacity-lease check (#4387)

### LINKED ISSUE #4387: Self-hosted: cron functions fail to schedule in v1.27.0 - CheckConstraints rejects zero envID ("missing envID")
**Describe the bug**
After upgrading our self-hosted Inngest from v1.26.0 to v1.27.0, cron functions silently stopped running. The scheduler fails before a run is ever created: every cron tick, the executor calls `CheckConstraints` to acquire a capacity lease, and that request is rejected with `missing envID`. Event-triggered functions still work fine.

The reason seems to be that in self-hosted mode, cron items are stamped with `consts.DevServerEnvID` — the zero UUID — as their workspace ID (visible in the logged request: `"WorkspaceID":"00000000-0000-0000-0000-000000000000"`). The constraint API validation added/used in the v1.27.0 cron rework ("Re-enable parallel processing for crons", "Split cron partitions by workspace") requires a non-zero `EnvID`, so the two are incompatible: self-hosted crons can never pass the check, and there's no config to supply a real env ID. The cron health check then notices the cron never ran and re-syncs it — with the same zero UUID — so it loops forever ("cron health check failed, re-syncing" every check).

**To Reproduce**
1. Run `inngest/inngest:v1.27.0` self-hosted (`inngest start`, Postgres + Redis, `INNGEST_DEV=0`).
2. Sync an app with any cron function (e.g. `cron: "TZ=UTC */15 * * * *"`).
3. Every tick logs the errors below; no run is created.

**Expected behavior**
Crons schedule and run as they did on v1.26.0 — rolling back the image restores them with identical config.

```
ERROR msg="acquiring capacity lease failed" err="invalid request: 1 error occurred:\n\t* missing envID\n\n" method="CheckConstraints"
ERROR msg="error scheduling cron function execution" handler="cron" cronExpr="TZ=UTC */15 * * * *" err="could not enforce constraints: invalid request: 1 error occurred:\n\t* missing envID\n\n"
ERROR msg="error handling queue item" error="error scheduling run for cron: ..." item_kind="cron"
WARN  msg="cron health check failed, re-syncing"
```
Constraint request (abridged): `"AccountID":"00000000-0000-4000-a000-000000000000","WorkspaceID":"00000000-0000-0000-0000-000000000000","Events":[{"event":{"name":"inngest/scheduled.timer"}}]`

**System info**
- inngest-cli version: v1.27.0 (Docker image, regression from v1.26.0)
- OS: Linux (Docker)
- SDK: inngest-js 4.5.1 (Next.js)

crossi-dev:
## Problem

After upgrading self-hosted Inngest to v1.27.0, cron functions silently stop scheduling (event-triggered functions are unaffected). Every cron tick the executor calls `CheckConstraints` to acquire a capacity lease, and the request is rejected with `missing envID`.

Root cause is in the cron health-check path. Self-hosted functions are stored without an env ID, so `cqrs.Function.EnvID` is the zero UUID (`uuid.Nil`). `handleCronHealthCheck` used that value directly:

```go
envID := cqrsFn.EnvID // uuid.Nil for self-hosted functions
```

That zero env ID is then stamped as the cron item's `WorkspaceID`, propagated through `ScheduleNext` → `handleCron` → `Executor().Schedule`, and finally used as the capacity-lease `EnvID`. `CapacityCheckRequest.Valid()` rejects `EnvID == uuid.Nil` with `missing envID`, so the cron can never schedule. The same zero env ID also makes the `HealthCheck` lookup miss the item that `InitializeCrons` enqueued under `consts.DevServerEnvID`, so the health check reports the cron as unscheduled and re-syncs it forever under the same zero env ID — matching the `WorkspaceID:"00000000-0000-0000-0000-000000000000"` seen in the logged request in #4387.

`InitializeCrons` already stamps the non-zero `consts.DevServerEnvID`; the health-check re-sync was the one place using the raw zero value, so the two paths were inconsistent.

## Fix

Resolve the env ID to `consts.DevServerEnvID` when a function has no env ID, matching `InitializeCrons`. Real env IDs (cloud / multi-tenant) are preserved unchanged. This fixes both the capacity-lease rejection and the false-negative health-check lookup, so self-hosted crons schedule as they did on v1.26.0.

## Verified

- New focused test `TestCronHealthCheckEnvID` proves the resolved env ID is non-nil and passes the real `constraintapi` `CapacityCheckRequest.Valid()` validation, while the old zero env ID fails it with `missing envID`. Confirmed red→green: with the pre-fix behavior the test fails with the exact `missing envID` error from the issue; with the fix it passes.
- `go test ./pkg/execution/executor/` passes (one unrelated pre-existing failure, `TestScheduleConstraintCacheDoesNotDropRetriesOnExhaustion`, fails identically on a clean `main` — an embedded-Redis Lua script issue unrelated to this change). `pkg/execution/cron` passes; `constraintapi` validation tests pass.

Fixes #4387

— Charles ([Choreless](https://choreless.dev))

### DIFF
--- a/pkg/execution/executor/cron_health_check_env_test.go
+++ b/pkg/execution/executor/cron_health_check_env_test.go
@@ -1,0 +1,66 @@
+package executor
+
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/inngest/inngest/pkg/constraintapi"
+	"github.com/inngest/inngest/pkg/consts"
+	"github.com/inngest/inngest/pkg/cqrs"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCronHealthCheckEnvID covers the self-hosted cron scheduling regression in
+// issue #4387: self-hosted functions are stored without an env ID, so cqrs.Function.EnvID
+// is uuid.Nil. The cron health-check re-sync used that zero env ID directly as the cron
+// item's workspace ID, which the capacity-lease CheckConstraints validation rejects with
+// "missing envID", so self-hosted crons could never schedule and the health check looped
+// forever re-syncing them.
+func TestCronHealthCheckEnvID(t *testing.T) {
+	t.Run("falls back to DevServerEnvID for self-hosted functions with no env ID", func(t *testing.T) {
+		// Self-hosted function as returned by GetFunctions: EnvID is the zero UUID.
+		fn := &cqrs.Function{EnvID: uuid.Nil}
+
+		got := cronHealthCheckEnvID(fn)
+
+		require.Equal(t, consts.DevServerEnvID, got)
+		require.NotEqual(t, uuid.Nil, got, "resolved env ID must be non-nil so constraint validation passes")
+	})
+
+	t.Run("nil function falls back to DevServerEnvID", func(t *testing.T) {
+		require.Equal(t, consts.DevServerEnvID, cronHealthCheckEnvID(nil))
+	})
+
+	t.Run("preserves a real env ID when present (cloud/multi-tenant)", func(t *testing.T) {
+		real := uuid.New()
+		require.Equal(t, real, cronHealthCheckEnvID(&cqrs.Function{EnvID: real}))
+	})
+
+	// Regression guard: prove the resolved env ID actually passes the same constraint
+	// validation that was rejecting the request, and that the old zero env ID did not.
+	t.Run("resolved env ID passes capacity-lease validation while uuid.Nil fails", func(t *testing.T) {
+		fn := &cqrs.Function{EnvID: uuid.Nil}
+		resolved := cronHealthCheckEnvID(fn)
+
+		base := func(envID uuid.UUID) *constraintapi.CapacityCheckRequest {
+			return &constraintapi.CapacityCheckRequest{
+				AccountID: consts.DevServerAccountID,
+				EnvID:     envID,
+				Constraints: []constraintapi.ConstraintItem{
+					{Kind: constraintapi.ConstraintKindConcurrency},
+				},
+			}
+		}
+
+		// Before the fix: the raw zero env ID is rejected with "missing envID".
+		err := base(fn.EnvID).Valid()
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "missing envID")
+
+		// After the fix: the resolved env ID is accepted (no "missing envID").
+		err = base(resolved).Valid()
+		if err != nil {
+			require.NotContains(t, err.Error(), "missing envID")
+		}
+	})
+}
--- a/pkg/execution/executor/service.go
+++ b/pkg/execution/executor/service.go
@@ -903,6 +903,19 @@ func (s *svc) handleEagerCancelBulkRun(ctx context.Context, c cqrs.Cancellation)
 	return nil
 }
 
+// cronHealthCheckEnvID resolves the env (workspace) ID used to health-check and
+// re-sync a self-hosted cron function. Self-hosted functions carry no env ID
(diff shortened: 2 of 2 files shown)


tsushanth (approved):
Surgical fix. The root-cause analysis in the PR body ties the symptom (capacity-lease rejection with `missing envID`) through the chain — `handleCronHealthCheck` → `cqrsFn.EnvID == uuid.Nil` → `WorkspaceID` stamping → `CapacityCheckRequest.Valid()` rejection — and shows why the two paths (`InitializeCrons` and the health-check re-sync) were silently inconsistent. Consistency-with-`InitializeCrons` is the right design argument; the file-local helper is the right abstraction for it.

### Worth verifying

**1. Sibling call sites that read `cqrsFn.EnvID` directly.**

This PR fixes one site in `pkg/execution/executor/service.go` (the re-sync at the new line 957). If any other code path reads `cqrsFn.EnvID` for a self-hosted function and propagates it as a workspace ID into a constraint check, that path will have the same bug — the helper just hasn't been applied there yet.

A quick `rg 'cqrsFn?\.EnvID\b' pkg/execution` (and the same against `pkg/cqrs`) to enumerate read sites would close that risk. If there are no others, the file-local helper is sized right; if there are, promoting it (or attaching it as a method on `*cqrs.Function`) into a shared location would be cleaner than fixing each site independently when the next regression surfaces.

**2. Test #4's soft assertion is silently vacuous on schema changes.**

```go
// After the fix: the resolved env ID is accepted (no "missing envID").
err = base(resolved).Valid()
if err != nil {
    require.NotContains(t, err.Error(), "missing envID")
}
```

The intent reads as "the resolved env ID passes validation," but the assertion only fires when `err != nil`, and even then only checks the error message doesn't include the magic string. If `Valid()` returns nil (success), the test makes no claim — which is the intended behavior here. But if a future `Valid()` change adds a *different* required field and starts returning an unrelated error, this test passes silently while the resolved env ID isn't actually being verified anymore.

Two tightening options:

- **(a)** If the request is supposed to be a fully valid `CapacityCheckRequest`, swap the `if err != nil` block for `require.NoError(t, err)`. Any future `Valid()` requirement surfaces here loudly instead of silently.
- **(b)** If you specifically want to assert "the *missing envID* gate doesn't fire," invert the check so it always evaluates: `require.False(t, err != nil && strings.Contains(err.Error(), "missing envID"), "resolved env ID must not be rejected for missing envID")`. Same targeted intent, but the assertion runs in both branches.

The first three test cases are tight; only #4 has the silent-vacuous failure mode.

### Worth a comment-pass, not a blocker

**A.** `consts.DevServerEnvID` reads misleading when used in production self-hosted deployments — the constant predates the multi-tenant split. The new helper docstring at `service.go:906` is accurate but doesn't flag this; a one-line note at the `consts.DevServerEnvID` declaration site itself ("also used as the canonical workspace ID for self-hosted production deployments, not just the dev server") would save the next reader the same archaeology. Out of scope for this PR; just flagging.

**B.** The unrelated `TestScheduleConstraintCacheDoesNotDropRetriesOnExhaustion` failure noted under "Verified" (embedded-Redis Lua script) is worth a separate issue if one isn't already open — easy to lose track of and easy to misread as a regression from a future PR.

### Decision

Approving. The fix is correct, the four test cases prove it red→green against the real `constraintapi.CapacityCheckRequest.Valid()` (not a mock — important detail for a validation-layer regression), and the helper is well-scoped. The two tightening suggestions above (sibling-site grep + test #4 assertion) would close out the residual risk; neither is a merge blocker.

Issue #4326, opened 2026-06-04
Dashboard /runs list creates excessive DB load from per-row app/function lookups


Liooo:
**Describe the bug**

The self-hosted dashboard `/runs` page appears to issue too many DB reads for a single list load.

For a broad range, `GetRuns` returns up to 100 run rows and asks for:

- `node.app`
- `node.function`

Those fields appear to be resolved separately for each run row. So one page load can become roughly:

- 1 runs query
- up to 100 app lookups
- up to 100 function lookups

This creates avoidable DB pressure.

**To Reproduce**

1. Run self-hosted Inngest with Postgres persistence.
2. Open the dashboard `/runs` page.
3. Use a broad range, for example last 1 day.
4. Compare with a narrow range, for example last 10 minutes.

Observed in our deployment:

- Last 10 minutes works.
- Last 1 day fails.
- `CountRuns` succeeds for last 1 day, so runs exist.
- `GetRuns` for last 1 day returns `data: null`.
- In our staging environment, the failure mode is DB connection/connector errors under load.
- Errors are under:
  - `runs.edges[*].node.app`
  - `runs.edges[*].node.function`

The UI then shows "No results found", which is misleading because the query failed.

**Expected behavior**

A single `/runs` page load should not perform separate app/function DB lookups for every run row.

Please batch or preload these fields, for example by:

- using dataloaders
- hydrating app/function data in the runs query
- adding a purpose-built list endpoint for the dashboard

Also, query failure should be shown as an error, not "No results found".

**Screenshots / Stack trace dump**

Relevant code shape:

```graphql
type FunctionRunV2 {
  app: App!
  function: Function!
}
```

`FunctionRunV2.app` calls `GetAppByID`.

`FunctionRunV2.function` calls `GetFunctionByInternalUUID`.

The dashboard list query requests both fields for every run row.

**System info**

- Inngest version observed: v1.19.1
- Source shape also checked in v1.25.0
- Deployment: self-hosted Inngest on Cloud Run
- Database: Postgres

**Additional context**

The main issue is DB load amplification from the dashboard list query. The DB/connector failure is the observed failure mode in our environment, but the avoidable per-row DB work is what makes the page fragile for broad ranges.

PR #4425, opened 2026-06-16
fix(coreapi): dataloader-cache FunctionRunV2.App/Function lookups (refs #4326)

tsushanth:
## Description

Refs #4326. The dashboard \`/runs\` list resolver fans out one \`GetAppByID\` and one \`GetFunctionByInternalUUID\` call per row. On a 100-row page that's up to **200 separate DB reads** for two fields that almost always have far smaller unique cardinality across the page. The reporter (@gbradley) hit connection-pool exhaustion as a consequence; \`GetRuns\` returned \`data: null\` and the UI rendered it as \"No results found,\" masking the real failure.

The fix routes both fields through per-request dataloaders. The dataloader's identity cache scopes to the GraphQL request lifetime: a page of 100 runs sharing 5 distinct apps now calls \`GetAppByID\` 5 times instead of 100.

## Scope of this PR

**In scope:** dataloader plumbing for App and FunctionByInternalUUID. This gives meaningful dedup for the common case where many runs share an app/function on a single page.

**Out of scope:** adding \`GetAppsByIDs\` / \`GetFunctionsByIDs\` batch SQL queries (\`WHERE id = ANY(\\$1)\`). That's the right complete fix for the worst case (100 runs each referencing a distinct app or function) but it touches the sqlc-generated layer across both Postgres and SQLite plus regen. Keeping that as a follow-up so this PR stays reviewable in isolation. With the loader infrastructure now in place, swapping the inner per-key loop in \`appReader.GetApps\` / \`functionReader.GetFunctions\` for a single batch SQL call is a one-file change in the reader.

Happy to roll the batch SQL into this PR if maintainers prefer — let me know.

## Code shape (mirrors existing \`Defers\` resolver)

\`\`\`go
// before
func (r *functionRunV2Resolver) App(ctx context.Context, run *models.FunctionRunV2) (*cqrs.App, error) {
    return r.Data.GetAppByID(ctx, run.AppID)  // 1 DB call per row
}

// after
func (r *functionRunV2Resolver) App(ctx context.Context, run *models.FunctionRunV2) (*cqrs.App, error) {
    return loader.LoadOneWithString[cqrs.App](ctx, loader.FromCtx(ctx).AppLoader, run.AppID.String())
}
\`\`\`

Same shape used by \`Defers\` (line 33 of the same file), which already passes through the dataloader infrastructure. The change is just bringing \`App\` and \`Function\` into the same pattern.

## Changes

- **\`pkg/coreapi/graph/loaders/app_function.go\`** — new. \`appReader\` / \`functionReader\` batch funcs plus a generic \`loadByUUID\` helper that mirrors \`loadByRunID\` (defer.go) but accepts a per-key fetcher rather than a bulk fetcher. Invalid UUID keys surface per-row errors without aborting the rest of the batch — a single malformed key shouldn't 500 a whole page.
- **\`pkg/coreapi/graph/loaders/loader.go\`** — register \`AppLoader\` and \`FunctionByInternalUUIDLoader\` on \`Loaders\`; wire them in \`NewLoaders\`.
- **\`pkg/coreapi/graph/resolvers/function_run_v2.resolver.go\`** — swap \`App\` and \`Function\` resolvers to call the new loaders.
- **\`pkg/coreapi/graph/loaders/app_function_test.go\`** — new. Three focused unit tests:
  - \`TestLoadByUUID_DedupesUniqueKeysAndPreservesOrder\` — pins per-key invocation count and output ordering.
  - \`TestLoadByUUID_InvalidKeyErrorsPerRowWithoutAborting\` — a single bad key produces a per-row error, fetch is NOT called for it, and valid neighbors still resolve.
  - \`TestLoadByUUID_PerKeyFetchErrorDoesNotAbortBatch\` — a fetch error on one key surfaces only on that row.

## Verification

\`\`\`
\$ go test ./pkg/coreapi/... -count=1
ok    github.com/inngest/inngest/pkg/coreapi/graph/loaders    0.475s
ok    github.com/inngest/inngest/pkg/coreapi/graph/models     0.898s
\`\`\`

\`go vet\` clean on the touched packages.

## Release note

Fix dashboard \`/runs\` page connection-pool pressure caused by per-row \`GetAppByID\` and \`GetFunctionByInternalUUID\` calls. The GraphQL resolver now uses per-request dataloaders so repeated app/function references on a single page collapse to one underlying lookup each.

## Migration note

None. The GraphQL schema and the public \`cqrs.Manager\` interface are unchanged. The behavior change is internal to the resolver layer.

### DIFF
--- a/pkg/coreapi/graph/loaders/app_function.go
+++ b/pkg/coreapi/graph/loaders/app_function.go
@@ -1,0 +1,71 @@
+package loader
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/graph-gophers/dataloader"
+	"github.com/inngest/inngest/pkg/cqrs"
+)
+
+// appReader is the dataloader-backed reader for App lookups.
+//
+// The current implementation calls GetAppByID once per unique key inside the
+// batch function — it does NOT yet issue a single batched SQL query. The
+// per-request win comes entirely from dataloader's identity cache: when the
+// dashboard /runs page (or any list view) renders 100 runs that share, say,
+// 5 distinct apps, dataloader collapses them to 5 GetAppByID calls instead
+// of 100. See https://github.com/inngest/inngest/issues/4326 for the original
+// report. A follow-up that adds GetAppsByIDs to the cqrs interface and a
+// `WHERE id = ANY($1)` SQL variant would replace the inner loop here with a
+// single call, fully closing the worst-case 100-unique-apps gap.
+type appReader struct {
+	reader cqrs.Manager
+}
+
+func (ar *appReader) GetApps(ctx context.Context, keys dataloader.Keys) []*dataloader.Result {
+	return loadByUUID(ctx, keys, func(ctx context.Context, id uuid.UUID) (*cqrs.App, error) {
+		return ar.reader.GetAppByID(ctx, id)
+	})
+}
+
+// functionReader is the dataloader-backed reader for Function lookups by
+// internal UUID. Shares the same architecture and follow-up trajectory as
+// appReader (see appReader docstring).
+type functionReader struct {
+	reader cqrs.Manager
+}
+
+func (fr *functionReader) GetFunctions(ctx context.Context, keys dataloader.Keys) []*dataloader.Result {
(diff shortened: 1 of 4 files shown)


Liooo:
@darwin67 

any chances this is merged? this one's blocking us from updating server version.

PR #4443, opened 2026-06-17
fix(extended traces): Reparent orphaned extended trace spans using span attributes

rhino1998:
## Description

This reparents orphaned extended trace spans using their step ID and attempt number attributes.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/cqrs/manager/cqrs.go
+++ b/pkg/cqrs/manager/cqrs.go
@@ -573,6 +573,9 @@ func mapRootSpansFromRows[T normalizedSpan](ctx context.Context, spans []T) (*cq
 	// executor.step is the preferred parent; executor.execution is the fallback.
 	stepSpanByKey := make(map[extendedTraceKey]string)
 	execSpanByKey := make(map[extendedTraceKey]string)
+	// Dynamic span IDs of all userland spans, used to detect interior nodes of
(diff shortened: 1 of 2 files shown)


pkg/cqrs/manager/cqrs.go
```diff
@@ -811,6 +857,17 @@ func walkMetadataSize(span *cqrs.OtelSpan, total *int) {
 	}
 }
 
+// extendedTraceKey is the map key used when indexing and reparenting orphaned
+// extended trace spans. A zero Attempt means no attempt was recorded.
+type extendedTraceKey struct {
+	StepID  string
+	Attempt int // 0 when unknown
```
scottnuma: With our step attempts, I thought that 0 was the value for the first attempt. Does it mean unknown here? Especially given that we're only populating this when .StepAttempt != nil?

scottnuma (approved):

pkg/cqrs/manager/cqrs.go
rhino1998: Ah, that comment is a leftover from claude's original draft

[6b32ff1](https://github.com/inngest/inngest/pull/4443/commits/6b32ff13fecc450319df996b5161817adbea3c3b)

PR #4445, opened 2026-06-17
chore: adding metrics for run duration

mar-inngest:
## Release note

None.

## Migration note

None.


pkg/telemetry/metrics/histogram.go
```diff
@@ -681,6 +698,38 @@ func HistogramDefersPerRun(ctx context.Context, count int64, opts HistogramOpt)
...
+	})
+}
+
+func HistogramRunStateStepCount(ctx context.Context, count int64, opts HistogramOpt) {
+	switch {
+	case count < 0:
+		count = 0
+	case count > 20:
+		count = 20
+	}
```
KiKoS0: why do we cap at 20?

pkg/telemetry/metrics/histogram_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+)
+
+func TestHistogramRunStateResidenceDuration(t *testing.T) {
```
KiKoS0: These histogram tests might not be very helpful since they're asserting the histogram's internal implementation. For example, simply adding a new bucket would cause them to fail even though the change is completely valid.

pkg/telemetry/metrics/histogram.go
mar-inngest: We're capping at 20 mostly for metric cardinality. No other reason. I'll remove it.

pkg/telemetry/metrics/histogram_test.go
mar-inngest: Good point!

KiKoS0 (approved):

PR #4453, opened 2026-06-18
feat(api): add v2 score endpoint

jacobheric:
## Description

Adds a v2 REST scoring endpoint for recording named scores against a function run or individual steps in a run.

See the mono side [PR](https://github.com/inngest/monorepo/pull/7625) for complete implementation details. Note there is a follow up to extract the metadata writer from the v1 endpoint [here](https://github.com/inngest/inngest/pull/4502) so we don't couple v2 and v1 endpoints. This can be merged without that refactor and it will work.

## Release note

Adds a v2 REST scoring endpoint for recording named scores against a function run or individual steps in a run.  This

`POST /v2/runs/{runId}/scores`

with body:
```json
[
  {
    "name": "accuracy",
    "value": 0.95,
   "stepId": "step 1"
  }
]
```

Scores are run scoped when `stepId` is omitted. Score values can be a number or boolean. Scores can be submitted in batches of up to 100. Experiment scores are supported at run level only (I copied this from the sdk, shout if we don't want to support this in the api) like so:

```json
[
  {
    "name": "accuracy",
    "value": 0.95,
    "experiment": {
      "id": "model-routing",
      "variant": "baseline"
    }
  }
]
```

## Migration note

None.

### DIFF
--- a/pkg/api/apiv1/metadata.go
+++ b/pkg/api/apiv1/metadata.go
@@ -286,6 +286,10 @@ func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID
 // deterministicSpanIDCutoff. It queries ClickHouse to locate the parent span
 // with a retry loop to absorb Kafka→ClickHouse propagation latency.
 func (a router) addRunMetadataLegacy(ctx context.Context, auth apiv1auth.V1Auth, runID ulid.ULID, req *AddRunMetadataRequest) error {
+	if a.opts.TraceReader == nil {
+		return publicerr.Errorf(http.StatusBadRequest, "trace reader not provided, aborting legacy metadata retrieval")
+	}
+
 	var parentSpan *cqrs.OtelSpan
 	var scope metadata.Scope
 	var err error
--- a/pkg/api/apiv1/metadata.go
+++ b/pkg/api/apiv1/metadata.go
@@ -117,6 +117,29 @@ type AddRunMetadataRequest struct {
 	Metadata []metadata.Update `json:"metadata"`
 }
 
+type AddRunMetadataOpts struct {
+	State          statev2.RunService
+	TracerProvider tracing.TracerProvider
+	TraceReader    cqrs.TraceReader
+}
+
+func AddRunMetadata(ctx context.Context, opts AddRunMetadataOpts, auth apiv1auth.V1Auth, runID ulid.ULID, req *AddRunMetadataRequest) error {
+	tracerProvider := opts.TracerProvider
+	if tracerProvider == nil {
+		tracerProvider = tracing.NewNoopTracerProvider()
+	}
+	r := router{
+		API: &API{
+			opts: Opts{
+				State:          opts.State,
+				TracerProvider: tracerProvider,
+				TraceReader:    opts.TraceReader,
+			},
+		},
+	}
+	return r.AddRunMetadata(ctx, auth, runID, req)
+}
+
 func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID ulid.ULID, req *AddRunMetadataRequest) error {
 	if err := metadata.ValidateUpdatesAllowed(req.Metadata); err != nil {
 		return publicerr.Wrap(err, 400, "Invalid metadata")
--- a/pkg/api/apiv1/metadata.go
+++ b/pkg/api/apiv1/metadata.go
@@ -254,6 +254,15 @@ func (a router) AddRunMetadata(ctx context.Context, auth apiv1auth.V1Auth, runID
 			return err
 		}
 	}
+	if parentSpanRef != nil && parentSpanRef.DynamicSpanID != "" {
+		if err := a.opts.TracerProvider.UpdateSpan(ctx, &tracing.UpdateSpanOptions{
+			Debug:      &tracing.SpanDebugData{Location: "router.AddRunMetadata.refreshTarget"},
+			Metadata:   stateMetadata,
+			TargetSpan: parentSpanRef,
+		}); err != nil {
+			return err
+		}
+	}
 
 	// Persist the cumulative metadata size delta back to the state store.
 	// Only persist when we successfully loaded from state; the fallback
@@ -391,6 +400,15 @@ func (a router) addRunMetadataLegacy(ctx context.Context, auth apiv1auth.V1Auth,
 			addTenantIDs,
 		)
 		if err != nil {
+			return err
+		}
+	}
+	if parentSpanRef != nil && parentSpanRef.DynamicSpanID != "" {
+		if err := a.opts.TracerProvider.UpdateSpan(ctx, &tracing.UpdateSpanOptions{
+			Debug:      &tracing.SpanDebugData{Location: "router.AddRunMetadata.refreshTarget"},
+			Metadata:   stateMetadata,
+			TargetSpan: parentSpanRef,
+		}); err != nil {
 			return err
 		}
 	}
--- a/pkg/api/v2/endpoints_scores.go
+++ b/pkg/api/v2/endpoints_scores.go
@@ -1,0 +1,175 @@
+package apiv2
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"net/http"
+	"strings"
+
+	"github.com/inngest/inngest/pkg/api/v2/apiv2base"
+	"github.com/inngest/inngest/pkg/tracing/metadata"
+	apiv2 "github.com/inngest/inngest/proto/gen/api/v2"
+	"github.com/oklog/ulid/v2"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const maxScoreBatchSize = 100
+
+func (s *Service) CreateScore(ctx context.Context, req *apiv2.CreateScoreRequest) (*apiv2.CreateScoreResponse, error) {
+	if req.RunId == "" {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorMissingField, "Run ID is required")
+	}
+
+	if result := s.rateLimiter.CheckRateLimit(ctx, apiv2.V2_CreateScore_FullMethodName); result.Limited {
+		return nil, s.base.NewError(http.StatusTooManyRequests, apiv2base.ErrorRateLimited, "API rate limit exceeded. The request was rejected and no score was recorded.")
+	}
+
+	if s.scores == nil {
+		return nil, s.base.NewError(http.StatusNotImplemented, apiv2base.ErrorNotImplemented, "Create score is not yet implemented")
+	}
+
+	runID, err := ulid.Parse(decodePathParam(req.RunId))
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Run ID must be a valid ULID")
+	}
+
+	scoreInputs, err := scoreInputsFromRequest(req)
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, err.Error())
+	}
+
+	err = s.scores.CreateScores(ctx, CreateScoresParams{
+		RunID:  runID,
+		Scores: scoreInputs,
+	})
+	switch {
+	case errors.Is(err, ErrScoresNotEnabled):
+		return nil, s.base.NewError(http.StatusForbidden, apiv2base.ErrorAccessDenied, "Scores are not enabled for this account")
+	case errors.Is(err, ErrScoreTargetNotFound):
+		return nil, s.base.NewError(http.StatusNotFound, apiv2base.ErrorNotFound, "Run or step not found")
+	case errors.Is(err, metadata.ErrMetadataSpanTooLarge), errors.Is(err, metadata.ErrRunMetadataSizeExceeded):
+		return nil, s.base.NewError(http.StatusRequestEntityTooLarge, apiv2base.ErrorValidationError, "Score exceeds the run metadata size limit")
+	case err != nil:
+		return nil, s.base.NewError(http.StatusInternalServerError, apiv2base.ErrorInternalError, "Unable to record score")
+	}
+
+	scores := make([]*apiv2.Score, 0, len(scoreInputs))
+	for _, score := range scoreInputs {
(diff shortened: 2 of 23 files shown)


pkg/api/apiv1/metadata.go
```diff
@@ -117,6 +117,29 @@ type AddRunMetadataRequest struct {
 	Metadata []metadata.Update `json:"metadata"`
 }
 
+type AddRunMetadataOpts struct {
+	State          statev2.RunService
+	TracerProvider tracing.TracerProvider
+	TraceReader    cqrs.TraceReader
+}
+
+func AddRunMetadata(ctx context.Context, opts AddRunMetadataOpts, auth apiv1auth.V1Auth, runID ulid.ULID, req *AddRunMetadataRequest) error {
```
jacobheric, line 126: I temporary shim here so we can use the v1 metadata code without muddying this PR with diffs to extract that. That extraction is done here on a stacked PR: https://github.com/inngest/inngest/pull/4502

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+		loaded, err := p.state.LoadMetadata(ctx, id)
+		if err == nil {
+			md = &loaded
+		} else if !errors.Is(err, statev2.ErrRunNotFound) && !errors.Is(err, statev2.ErrMetadataNotFound) {
+			return fmt.Errorf("%w: %w", ErrScoreTargetNotFound, err)
+		}
+	}
+
+	for stepID := range stepIDs {
+		if md != nil && scoreStepExistsInState(md, stepID) {
```
amh4r: Is this comparing the unhashed step ID with the hashed step ID?

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+		return nil
+	}
+
+	var md *statev2.Metadata
+	if p.state != nil {
+		loaded, err := p.state.LoadMetadata(ctx, id)
+		if err == nil {
+			md = &loaded
+		} else if !errors.Is(err, statev2.ErrRunNotFound) && !errors.Is(err, statev2.ErrMetadataNotFound) {
+			return fmt.Errorf("%w: %w", ErrScoreTargetNotFound, err)
```
amh4r: Is this conditional backwards? It looks like we return `ErrScoreTargetNotFound` when we get an error that _isn't_ a "not found"

proto/api/v2/service.proto
```diff
@@ -2408,6 +2427,94 @@ message InvokeFunctionData {
...
+  ];
+  optional ScoreExperiment experiment = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Optional experiment context. When provided, the score is recorded with the same experiment metadata emitted by the SDK's score.experiment API. Experiment scores should be run-scoped for experiment aggregate and detail APIs."
+    }
+  ];
+}
+
+message ScoreExperiment {
+  string experiment_name = 1 [
```
amh4r: Are we locked into using `name` for the experiment identifier? I'm wondering because our platform often uses `id` for "unchanging identifier for programmatic logic" and `name` for "human readable string". For example, we introduced function `id` in TS SDK v3 because users didn't understand that `name` needed to be durable in v2

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+		switch {
+		case errors.Is(err, statev2.ErrRunNotFound), errors.Is(err, statev2.ErrMetadataNotFound):
+			return fmt.Errorf("%w: %s", ErrScoreTargetNotFound, params.RunID)
+		case err != nil:
+			return err
+		}
+	}
+
+	return nil
+}
```
amh4r, line 86: Since we process scores sequentially and return on the first error, a partial failure (e.g. `score[0]` persisted, `score[1]` errors) leaves the run with some scores written but returns a single coarse error with no index, so the client can't tell what landed or where to retry from.

Three things worth deciding:
1. Should the write error include the failing index (`scores[%d]: ...`) like the validation path does, so a retry can be targeted?
2. Retrying is value-safe today only because scores use `MetadataOpcodeMerge` (re-merging the same name/value is a no-op). `MetadataOpcodeAdd` already exists (Values.Combine sums on each apply), and the moment scores can use it "retry the whole batch" double-counts the value. So I'd rather not lean on op idempotency for retry safety.
3. Minor: the metadata-size accounting isn't idempotent (each `AddRunMetadata` re-adds its delta), so retries inflate `metadata_size` even when no new data is stored. Low severity given how small score values are (you'd need thousands of resubmits to one run to approach the 1 MB cap), but worth noting since the counter is shared across all metadata kinds.

Not asking for full atomicity (hard given per-call `Target.StepID`), just want the partial-failure behavior to be intentional.

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+	if p.enabled != nil && !p.enabled(ctx, accountID) {
+		return ErrScoresNotEnabled
+	}
+
+	if p.state == nil {
+		return errors.New("score provider state is nil")
+	}
+
+	stateID := statev2.ID{
+		RunID: params.RunID,
```
amh4r: Critical: how are we authorizing that the run belongs to the tenant? Are we trusting that metadata read/write will automatically do it? If so, did we confirm that metadata read/write does proper authz?

That said, even if metadata read/write does authz right now, should we trust that that remains true in the future? If we explicitly authz the run ID in this handler then we don't need to worry

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+	state              statev2.RunService
+	tracerProvider     tracing.TracerProvider
+	traceReader        cqrs.TraceReader
+	auth               ScoreAuthResolver
+	enabled            ScoresEnabledFlag
+	missingStateLoader MissingScoreMetadataLoader
+	stepValidator      ScoreStepTargetValidator
+}
+
+func (p *stateScoreProvider) CreateScores(ctx context.Context, params CreateScoresParams) error {
```
amh4r, line 53: Should we add request validation here?
- How many scores per request?
- Experiment name length
- Cannot specify both step ID and experiment (IIUC)

proto/api/v2/service.proto
```diff
@@ -2408,6 +2427,94 @@ message InvokeFunctionData {
...
 }
 
+message CreateScoreRequest {
+  string run_id = 1 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Run ID (ULID) of the function run to score"
+      example: "\"01JLW1NNJ69KCD1KMJEXAMPLE0\""
+    }
+  ];
+  repeated CreateScoreInput scores = 2 [
```
amh4r, line 2539: What does it mean if I set `Experiment` in a score? Like what does it mean to score a run and an experiment at the same time? I thought they were mutually exclusive targets

pkg/api/v2/endpoints_scores_test.go
```diff
@@ -0,0 +1,527 @@
...
+	require.NoError(t, err)
+	require.Equal(t, metadata.KindInngestExperiment, update.Kind())
+	require.Equal(t, enums.MetadataOpcodeMerge, update.Op())
+	require.Equal(t, metadata.Values{
+		"name":    json.RawMessage(`"model-routing"`),
+		"variant": json.RawMessage(`"baseline"`),
+	}, update.RawUpdate.Values)
+}
+
+func TestScoreParentSpanRefHashesSDKStepID(t *testing.T) {
```
amh4r: Delete. It's testing an unused function (`scoreParentSpanRef`)

pkg/api/v2/endpoints_scores.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/inngest/inngest/pkg/tracing/metadata"
+	apiv2 "github.com/inngest/inngest/proto/gen/api/v2"
+	"github.com/oklog/ulid/v2"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const maxScoreBatchSize = 100
+
+func (s *Service) CreateScore(ctx context.Context, req *apiv2.CreateScoreRequest) (*apiv2.CreateScoreResponse, error) {
```
amh4r, line 24: Validation smell here: `ScoreMetadataUpdate`/`ScoreExperimentMetadataUpdate` both build and validate the `metadata.Update`, and we call them once in the handler purely to validate (discarding the result), then again in the provider to build it, then `AddRunMetadata` re-validates.

```
CreateScore
├─ scoreInputsFromRequest
│   └─ scoreInputFromFields                     [per score]
│       ├─ ScoreMetadataUpdate                  → builds Update, calls ValidateAllowed
│       └─ ScoreExperimentMetadataUpdate        → builds Update, calls ValidateAllowed
│
└─ s.scores.CreateScores
    └─ stateScoreProvider.CreateScores
        ├─ scoreMetadataUpdates                 [per score]
        │   ├─ ScoreExperimentMetadataUpdate    → builds Update, calls ValidateAllowed
        │   └─ ScoreMetadataUpdate              → builds Update, calls ValidateAllowed
        └─ apiv1.AddRunMetadata
            └─ metadata.ValidateUpdatesAllowed
                └─ update.ValidateAllowed       [per update]
```

I wonder if we should build the updates once and pass them through, rather than validate-then-discard then rebuild

pkg/api/v2/score_provider.go
```diff
@@ -0,0 +1,315 @@
...
+			stepIDs[*score.StepID] = struct{}{}
+		}
+	}
+	if len(stepIDs) == 0 {
+		return nil
+	}
+
+	var md *statev2.Metadata
+	if p.state != nil {
+		loaded, err := p.state.LoadMetadata(ctx, id)
```
amh4r: Does run metadata exist after the run ends? If not, then that means we can't score after the run ends, which would be a massive limitation

pkg/api/v2/endpoints_scores.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	runID, err := ulid.Parse(decodePathParam(req.RunId))
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, "Run ID must be a valid ULID")
+	}
+
+	scoreInputs, err := scoreInputsFromRequest(req)
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, err.Error())
```
amh4r, line 44: We shouldn't parrot errors back to the client because it leaks internal details. I see this in other spots too

pkg/api/v2/score_provider.go
rhino1998: Writing metadata for an arbitrary target w/o validating anything else is safe because we use the accountID/envID from the API auth. W/o additional validation (IE reading CH) it is very possible to write orphan metadata, but that is something we already accept in the metadata API.

rhino1998: I don't think we actually care if RunIDs are globally unique but rather that `(account_id, env_id, run_id)` is unique.

pkg/api/v2/endpoints_scores.go
jacobheric: `err.Error()` here are for request validation errors that we map to user facing messages in `scoreInputsFromRequest`. internal provider errors are handled below and mapped to user facing message. so I don't see any paths here that leak internal errors.

proto/api/v2/service.proto
jacobheric: I got this from sdk `score.experiment()`, which seems to be a run level experiment context for a score. Let me if I should not be exposing this. I added a better description to the proto as well that this is run level.

proto/api/v2/service.proto
jacobheric: I switched all user facing references to `id` from `name`

pkg/api/v2/score_provider.go
jacobheric: Added a check to handle this.

pkg/api/v2/endpoints_scores.go
jacobheric: Yup, done, see [here](https://github.com/inngest/inngest/blob/7775f25349fc95c72b4c46277d141c1f261530ae/pkg/api/v2/endpoints_scores.go#L127-L165) and [here](https://github.com/inngest/inngest/blob/7775f25349fc95c72b4c46277d141c1f261530ae/pkg/api/v2/score_provider.go#L102-L115).

pkg/api/v2/score_provider.go
jacobheric: * added 100 score limit
* added a length check that matches the sdk [here](https://github.com/inngest/inngest/blob/7775f25349fc95c72b4c46277d141c1f261530ae/pkg/api/v2/endpoints_scores.go#L138-L145)
* added step id + experiment check

pkg/api/v2/score_provider.go
jacobheric: I can do some follow up work to tighten this up and add the failing index. for now I just added a description to the protos notes about scoring not being atomic and about partial failure modes, see [here](https://github.com/inngest/inngest/blob/7775f25349fc95c72b4c46277d141c1f261530ae/proto/api/v2/service.proto#L2539-L2543).

pkg/api/v2/score_provider.go
jacobheric: we support ended runs via the `MissingStateLoader` fallback...in monorepo that's provided by the `cloudScoreMetadataLoader` [here](https://github.com/inngest/monorepo/blob/1e80c084ee1c724fb6aefe5ca953f5def948d03f/api/rest/v2/score_provider.go#L77-L125).

pkg/api/v2/endpoints_scores_test.go
jacobheric: moved to the follow on metadata writer refactor

pkg/api/v2/score_provider.go
jacobheric: fixed

pkg/api/v2/score_provider.go
amh4r: I'm confused about why we need to ensure the step exists in state at all. This check is skipped if the run has ended, so why do we need to do it when the run hasn't ended?

In other words, it's a check that we can't do all the time, which seems like it defeats its purpose

rhino1998 (approved):
Looks good metadata-wise

PR #4456, opened 2026-06-18
fix(trace view): Fix the ordering/timestamps of finalization spans

rhino1998:
## Description

Currently we emit finalization spans with timestamps derived from the execution as a whole. If there were steps checkpointed during the execution, these timestamps are too early which causes display issues. This just bumps the timestamps visually so that the view is correct until we do SDK-side plumbing to emit better timestamps after checkpointing.

Before:
image

After: (notice the sleep bar bumps the finalization bar)
image

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/packages/components/src/utils/date.ts
+++ b/ui/packages/components/src/utils/date.ts
@@ -172,6 +172,24 @@ export function toMaybeDate<T extends string | null | undefined>(value: T): Date
   return new Date(value);
 }
 
+export function maxDateString<T extends string | null | undefined>(
+  date1: T,
+  date2: T | null | undefined
+): T {
+  if (!date2) {
+    return date1;
+  }
+
+  const d1 = toMaybeDate(date1);
+  const d2 = toMaybeDate(date2);
+
+  if (!d1 || !d2) {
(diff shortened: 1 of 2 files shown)


ui/packages/components/src/utils/date.ts
```diff
@@ -172,6 +172,24 @@ export function toMaybeDate<T extends string | null | undefined>(value: T): Date
...
+  if (!date2) {
+    return date1;
+  }
+
+  const d1 = toMaybeDate(date1);
+  const d2 = toMaybeDate(date2);
+
+  if (!d1 || !d2) {
+    return date1;
+  }
```
scottnuma, line 188: Is there a case where get here, and `!d1 && d2` and we'd rather give back date2?

scottnuma (approved):

ui/packages/components/src/utils/date.ts
rhino1998: In the callsites where it is used, no, but this also implies that the date string was null or invalid for both anyways, so neither are actually dates

PR #4504, opened 2026-06-24
feat(api): experiments api

jacobheric:
## Description

Adds beta rest v2 APIs for reading experiment and session data. See monorepo PR for complete implementation details and examples: https://github.com/inngest/monorepo/pull/7678.

## Release note

Adds experiment endpoints to REST V2 API and CLI:
GET /v2/experiments
GET /v2/apps/{app_id}/functions/{function_id}/experiments
GET /v2/apps/{app_id}/functions/{function_id}/experiments/{experiment_id}

Adds sessions endpoints to REST V2 API and CLI:
GET /v2/sessions
GET /v2/sessions/{session_key}
GET /v2/sessions/{session_key}/{session_id}/runs

## Migration note

None.


proto/api/v2/service.proto
```diff
@@ -1620,6 +1622,78 @@ service V2 {
...
+          key: "BearerAuth"
+          value: {}
+        }
+      }
+    };
+  }
+
+  rpc GetExperiment(GetExperimentRequest) returns (GetExperimentResponse) {
+    option (google.api.http) = {
+      get: "/experiments/{function_id}/{experiment_name}"
```
amh4r: What happens if an experiment name contains a URL path unsafe character? For example, experiment `checkout/v2`

proto/api/v2/service.proto
```diff
@@ -1620,6 +1622,78 @@ service V2 {
...
+          key: "BearerAuth"
+          value: {}
+        }
+      }
+    };
+  }
+
+  rpc ListSessionRuns(ListSessionRunsRequest) returns (ListSessionRunsResponse) {
+    option (google.api.http) = {
+      get: "/sessions/{session_key}/{session_id}/runs"
```
amh4r: What happens if an session ID contains a URL path unsafe character? For example, session `thread/123`

pkg/api/v2/endpoints_experiments_sessions.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, s.base.NewError(http.StatusNotImplemented, apiv2base.ErrorNotImplemented, "Sessions not implemented in OSS")
+	}
+
+	cursor, limit, err := sessionsPageOpts(req.GetCursor(), req.GetLimit())
+	if err != nil {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorInvalidFieldFormat, err.Error())
+	}
+
+	result, err := s.sessions.ListSessions(ctx, SessionListOpts{
+		SessionKey: decodePathParam(req.SessionKey),
```
amh4r: `req.SessionKey` comes from a query param so we shouldn't decode it as a path param

pkg/api/v2/endpoints_experiments_sessions.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, s.base.NewError(http.StatusNotImplemented, apiv2base.ErrorNotImplemented, "Experiments not implemented in OSS")
+	}
+
+	detail, err := s.experiments.GetExperiment(ctx, ExperimentDetailOpts{
+		FunctionID:     decodePathParam(req.FunctionId),
+		ExperimentName: decodePathParam(req.ExperimentName),
+		Variant:        req.Variant,
+	})
+	if err != nil {
+		return nil, s.base.NewError(http.StatusInternalServerError, apiv2base.ErrorInternalError, "Unable to fetch experiment")
```
amh4r: Will "experiment not found" become a 500 instead of a 404?

pkg/api/v2/endpoints_experiments_sessions.go
```diff
@@ -0,0 +1,380 @@
...
+		Page:     experimentsPage(result.Experiments, limit, result.HasMore),
+	}, nil
+}
+
+func (s *Service) GetExperiment(ctx context.Context, req *apiv2.GetExperimentRequest) (*apiv2.GetExperimentResponse, error) {
+	if req.FunctionId == "" || req.ExperimentName == "" {
+		return nil, s.base.NewError(http.StatusBadRequest, apiv2base.ErrorMissingField, "Function ID and experiment name are required")
+	}
+
+	if result := s.rateLimiter.CheckRateLimit(ctx, apiv2.V2_GetExperiment_FullMethodName); result.Limited {
```
amh4r: Nit: rate limit before anything for consistency sake.

Also, should we be adding explicit rate limiting on methods like this? Is there not an implicit way we can do it with middleware? Out of scope for this PR, but I didn't know if that was already a pattern

pkg/api/v2/interfaces.go
```diff
@@ -119,3 +119,137 @@ type FunctionTraceReader interface {
 	GetSpansByRunID(ctx context.Context, runID ulid.ULID) (*cqrs.OtelSpan, error)
 	GetSpanOutput(ctx context.Context, id cqrs.SpanIdentifier) (*cqrs.SpanOutput, error)
 }
+
+type TimeRange struct {
```
amh4r: Unused

proto/api/v2/service.proto
```diff
@@ -2681,3 +2755,145 @@ message InsightsEventSchema {
...
+  int32 variant_count = 6;
+  int32 total_runs = 7;
+  google.protobuf.Timestamp first_seen = 8;
+  google.protobuf.Timestamp last_seen = 9;
+}
+
+message GetExperimentRequest {
+  string function_id = 1;
+  string experiment_name = 2;
+  optional string variant = 5 [
```
amh4r:
```suggestion
  optional string variant = 3 [
```

pkg/api/v2/endpoints_experiments_sessions.go
AndyInternet: leaving as-is. The existing endpoints (GetApp, GetFunction, GetEventRuns, etc.) all validate required fields first, then rate limit, then nil-check. Moving the rate limit above the field check would make these the only endpoints that don't follow that order, so the current code is the consistent with other endpoints.
On the middleware question: agree, but there's not a current pattern for this.

pkg/api/v2/endpoints_experiments_sessions.go
AndyInternet: Yes, good catch and fixed

pkg/api/v2/endpoints_experiments_sessions.go
AndyInternet: Good catch, fixed.

pkg/api/v2/interfaces.go
AndyInternet: Removing

pkg/api/v2/endpoints_experiments_sessions.go
jacobheric: I can ticket some follow up work to move rate limits to middleware. Not a tiny lift as these are entitlement overridable by account/endpoint.

proto/api/v2/service.proto
jacobheric: added support for this and a test that asserts it: https://github.com/inngest/inngest/blob/b5dd710d0756887c213d5a634904081bafbd4e87/pkg/api/v2/endpoints_experiments_sessions_test.go#L47

proto/api/v2/service.proto
jacobheric: added support for this and a test that asserts it here: https://github.com/inngest/inngest/blob/b5dd710d0756887c213d5a634904081bafbd4e87/pkg/api/v2/endpoints_experiments_sessions_test.go#L47

proto/api/v2/service.proto
```diff
@@ -1620,6 +1622,99 @@ service V2 {
...
+          key: "BearerAuth"
+          value: {}
+        }
+      }
+    };
+  }
+
+  rpc GetExperiment(GetExperimentRequest) returns (GetExperimentResponse) {
+    option (google.api.http) = {
+      get: "/apps/{app_id}/functions/{function_id}/experiments/{experiment_id}"
```
AndyInternet: {experiment_id} only matches a single path segment, and experiment names are free-form user strings. An experiment named A/B rollout (which is likely given A/B testing) shows up fine in ListExperiments, but its detail URL is unreachable: %2F doesn't match a single-segment variable, so GET .../experiments/A%2FB%20rollout 404s. Verified against the local stack: listable, not fetchable.

Since experiment_id is the last segment of this template, {experiment_id=**} is legal per the http-rule spec and would accept both raw and encoded slashes. That's exactly the trick ListSessionRuns already uses for {session_id=**}.

AndyInternet (approved):

proto/api/v2/service.proto
jacobheric: nice catch, I've fixed it!

PR #4516, opened 2026-06-25
fix: Update AI Metadata field names

scottnuma:
## Description

- Currently, we have the model and system fields
- model -> request model to remove ambiguity of what model this is (request vs response) and match OTel and OI standards
- system -> provider to match OTel and OI standards
- These bring OSS into alignment with the JS SDK as well
- This is a breaking change for Insights queries that utilize AI Metadata

## Release note

For AI Metadata extracted from extended traces, step.ai.wrap, and step.ai.infer, the .model field has been renamed to .request_model and .system has been renamed .provider to better match the OpenTelemetry GenAI standards.

## Migration note

Insights queries that use AI Metadata fields of .model and .system should be updated to use .request_model and .provider, respectively.

### DIFF
--- a/pkg/tracing/metadata/extractors/ai_test.go
+++ b/pkg/tracing/metadata/extractors/ai_test.go
@@ -94,9 +94,9 @@ func TestAIMetadataExtractor_OpenAISpan(t *testing.T) {
 	assert.Equal(t, 97.0, data["output_tokens"], "Should extract output tokens")
 
 	// Verify model and operation data
-	assert.Equal(t, "gpt-4", data["model"], "Should extract request model")
+	assert.Equal(t, "gpt-4", data["request_model"], "Should extract request model")
 	assert.Equal(t, "chat", data["operation_name"], "Should extract operation name")
-	assert.Equal(t, "openai", data["system"], "Should extract AI system")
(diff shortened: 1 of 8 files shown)


pkg/tracing/metadata/extractors/ai_test.go
```diff
@@ -251,11 +251,11 @@ func TestExtractAIOutputMetadata_VercelAISDK(t *testing.T) {
 	assert.Equal(t, 429.0, data["output_tokens"], "Should extract output tokens")
 	assert.Equal(t, 440.0, data["total_tokens"], "Should extract total tokens")
 
-	// Verify model
-	assert.Equal(t, "gpt-4-turbo-2024-04-09", data["model"], "Should extract model from response.modelId")
+	// Verify request model
+	assert.Equal(t, "gpt-4-turbo-2024-04-09", data["request_model"], "Should extract request model from response.modelId")
```
amh4r: Should this assert `response_model` instead?

scottnuma: Good catch, yes that should be response model, but this extractor doesn't yet differentiate between request and response model. Fixing that over in https://github.com/inngest/inngest/pull/4518

rhino1998 (approved):

PR #4517, opened 2026-06-25
feat(experiments): Visualize numeric metrics using box plots

rhino1998:
## Description

Needs https://github.com/inngest/monorepo/pull/7705

This makes numeric experiment metrics render using box plots
image

Also adds a snapping reference line on hover for easier visual comparison (especially relevant for non-adjacent variants)
image

## Release note

None.

## Migration note

None.


tonyhb (approved):

jpwilliams (approved):
@sanjanaladdha also mentioned that we can further iterate here, which we should, but even this is great at showing more information right now and doesn't look hellish. 😄 

Screenshot + link for posterity from @sanjanaladdha's suggestions: 

Image

https://inngest.slack.com/archives/C0AFJB68X99/p1782760796587839?thread_ts=1782395034.978389&cid=C0AFJB68X99

We can merge this to ship it whenever (the demo has many datapoints so it won't suffer from any issues that low data boxplots will have) and then add those changes from @sanjanaladdha later.

Solid work all!

PR #4538, opened 2026-06-30
fix: should not coalesce parallel steps for the V1 execution engine

KiKoS0:
## Description
V1 expects that each parallel step completion to emit its own discovery
step and I broke that by coalescing them to the same idempotency key in
the queue. It was fine for a while because https://github.com/inngest/inngest/pull/4021
forced the `SavePending` call on V1 thus making the synchronization
happen at the state store level (atomic updated on the pending set in
Redis) but as soon as I reverted it in https://github.com/inngest/inngest/pull/4534
it made it that only one discovery step is allowed to be enqueued in the
queue which executed too early (other pending steps are still ongoing)
and did not pick up the runs after the fan-in thus hanging the runs
forever.

The logs would have really helped for recovery so I added them now
I don't expect them to be too noisy and if they are, there is likely an issue
but I'll remove them if they are noisy.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/discovery_coalesce_test.go
+++ b/pkg/execution/executor/discovery_coalesce_test.go
@@ -177,6 +177,110 @@ func TestResumeCoalesceJobID(t *testing.T) {
 	}
 }
 
+// trackingRunSvc records SavePending invocations so tests can assert whether
+// pending-step tracking was populated for a given batch.
+type trackingRunSvc struct {
+	sv2.RunService
+	savePendingCalls [][]string
+}
+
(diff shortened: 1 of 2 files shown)


albertchae (approved):
did https://github.com/inngest/inngest/pull/4452 introduce the original issue?

Also what do you think about sharing more of the test setup https://github.com/inngest/inngest/pull/4539

KiKoS0:
Yes https://github.com/inngest/inngest/pull/4452 changes were based on top of the Mendral fix: https://github.com/inngest/inngest/pull/4021

mar-inngest (approved):

PR #4541, opened 2026-06-30
refactor(queue): clean up QueueManager interface

lkasinathan:
## Description

Clean up QueueManager interface by moving some of its methods to either `QueueProcessor` or `JobQueueReader`

- Currently, `QueueManager` has a lot of methods that are directly and only used by the executors to process the queue. The `QueueProcessor` interface is a better home for these
- The `QueueManager` also has a bunch of methods that help various callers read the state of the queue. The `JobQueueReader` is a better home for the latter.
- NOTE: No API definitions have changed with this PR, they are simply moved from one interface to another - both of which are already implemented by a single `queueProcessor` object.

**Why**
- Today, both the QueueManager along with its various embedded interfaces  and the QueueProcessor interface are implemented by one core mega object - the queueProcessor.
- All callers that need even the tiniest surface area of contact to the queue needs to instantiate this mega object for any access
- Cleaning up the interfaces and allowing for well scoped concrete implementations of smaller interfaces will help us avoid constructing a full blown queueProcessor in areas like debugAPI that simply don't need it.

**Next up**
This is the first PR in a stack of PRs that aim to get us closer to the following final state. In following PRs,
- we will break up `queueProcessor` into smaller objects that implement specific interfaces 
- `queueProcessor` will only be instantiated by executor services. All other services should instantiate the smallest interface they need.
- `QueueManager` will cease to exist in favor of simply `Queue` - other interfaces will subsume the QueueManager's remaining functionalities
- `Queue` interface APIs will be re-designed. Common access patterns will determine whether they will already take a resolved shard (executor dequeue path) or will resolve the shard itself based on item (enqueue path)
- A proxy layer will be added either directly on top of `Queue` or for a subset of `Queue` APIs to
   - (a) allow for streamlined access to queue from various services in Cloud and 
   - (b) serve as an optional centralized layer of optimization to batch non-critical updates to the Queue.

## Release note

None.

## Migration note

None.


pkg/execution/checkpoint/checkpoint.go
```diff
@@ -719,16 +715,7 @@ func (c checkpointer) validateAsyncDispatch(ctx context.Context, input AsyncChec
 		return nil
 	}
 
-	loader, ok := c.Queue.(queueItemLoader)
```
lkasinathan, line 722: @Linell I am moving `LoadQueueItem` directly to `Queue` so this check is no longer needed. Can you please confirm?

dlambrig (approved):
I noticed one contract change.  Before this PR, async dispatch validation would "fail open" if the configured queue did not implement `LoadQueueItem`, so the checkpoint could still update state. 

Now `LoadQueueItem` is part of `JobQueueReader`, and `queue.Queue` embeds `JobQueueReader`, so every `checkpointer.Queue` must support item loading and `validateAsyncDispatch` calls it unconditionally.

Probably its what you wanted but might be worth documenting.

pkg/execution/checkpoint/checkpoint.go
Linell: Yup, that should be fine! 👍

PR #4547, opened 2026-06-30
test: regression tests for parallel step in V3 SDK

KiKoS0:
## Description
Add regression tests for parallel steps in the V3 SDK which have a sligthly different behavior than V4 because `optimizeParallelism` default to `false` in that version and it doesn't end up coalescing discovery steps. Fix is: https://github.com/inngest/inngest/pull/4538

## Release note

None.

## Migration note

None.

### DIFF
--- a/tests/js/src/inngest/sdk_parallel_fan_in_test.ts
+++ b/tests/js/src/inngest/sdk_parallel_fan_in_test.ts
@@ -12,11 +12,11 @@ export const sleepRandom = inngest.createFunction(
   }
 );
 
-// Regression test for parallel fan-in runs wedging.  Drives 100 concurrent
-// ops (50 step.run, 50 step.invoke) through one Promise.all and asserts the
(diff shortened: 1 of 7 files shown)


tests/js/src/inngest/sdk_parallel_fan_in_test.ts
```diff
@@ -0,0 +1,55 @@
+import { inngest } from "@/inngest/client";
```
albertchae, line 1: this is V3 because we're importing from `@/inngest` instead of `@/inngest-v4`? Do we have equivalent tests for V4 SDK as well?

also, i might be reading this wrong but it seems these tests are on an old version of the V3 SDK too https://github.com/inngest/inngest/blob/8d0854283996ef549f2d39864a413faf26c1356e/tests/js/pnpm-lock.yaml#L30-L35

PR #4560, opened 2026-07-06
feat: Replace Insights AI with agent loop

AndyInternet:
## Description

Insights AI was a fixed three-stage pipeline: an event matcher picked event names up front, a query writer produced SQL in a single forced tool call, and a summarizer described the result. Every prompt got a query whether it wanted one or not, and nothing ever checked that the SQL actually ran.

This replaces the pipeline with a durable agent loop inside the same Inngest function. Each iteration is one LLM call in a `step.run`; the model either calls a tool (also a durable step) or responds with text, which ends the run. No forced `tool_choice` anywhere, so the model decides what the turn needs:

- **Write or modify a query.** Discovery is tool-driven: `find_events` and `get_event_schemas` fetch event names and `data.*` schemas only when the events table is actually in play. Static table schemas live in the unified system prompt (the SQL rules body carries over from the old query-writer prompt unchanged). `submit_query` records the final SQL, title, and reasoning.
- **Validate before returning.** A `validate_query` tool executes the candidate SQL for real and feeds back result columns and row count, or the exact transpiler diagnostics, so the model fixes broken queries instead of shipping them.
- **Ask instead of guess.** Ambiguous prompts ("show me failures") end in a clarifying question with no SQL.
- **Answer general questions** about Insights and Inngest in plain text.

`run.completed` gains an additive `kind: 'query' | 'answer'` field. The chat UI needed no other changes; the realtime contract, both entry points (`/api/chat` and `/v2/insights/query/prompt`), the `query-writer-model` experiment, and client-managed history are all preserved. The three stage modules are deleted.

### Validation without new credentials

The agent function has no workspace credentials, and it shouldn't. Validation is a round trip through the already-authenticated chat client: the loop publishes `validation.requested` on the existing realtime stream, the browser runs the SQL through the normal `insights` GraphQL query with the user's own session, posts the outcome to a new `/api/chat-validate` route, and the waiting `step.waitForEvent` picks it up (20s timeout). The agent can only execute what the user could run themselves, and no tokens ever ride in event payloads.

Completions are pinned to the run's initiating user: the route stamps the poster's Clerk `userId` into the event server-side (the request body can't supply it; the schema strips unknown keys), and the wait condition matches on both the validation id and that user id. The schema also hard-caps every field (diagnostic count and message length, column count and name length), so nothing oversized rides into an event or a prompt.

The headless API path doesn't get the tool at all (`canValidate` flag on the event), and the loop only honors `validate_query` when the tool was actually offered; a hallucinated call falls through to an unknown-tool observation instead of stalling 20s waiting for a browser that isn't there. There's a test for that too.

### Replay safety

Inngest persists a step's return value, not closure mutations, so tools return their effects (`observation`, optional `draftPatch`, optional realtime publish) and the loop applies them outside the memoized step boundary. Getting this wrong silently drops the final SQL on replay. There's a test for it.

### Scoring

Per-run scores for watching whether the loop earns its keep: `insights_agent_submitted`, `_iterations`, `_tool_calls`, `_validation_attempts`, `_validation_failures`, plus the existing latency/token/cost scores. Latency now measures the whole loop: the start time is memoized in its own step, so suspends (`waitForEvent`) and checkpoint resumes don't reset the clock. The deferred haiku judge now fires on every run and branches: `insights_judge_relevance` when SQL was produced, `insights_judge_no_query_appropriate` when it wasn't. The average of the latter is the inverse of the submit-miss rate. Both stay attributed to the experiment variant.

## Release note

None.

## Migration note

None.


ui/apps/dashboard/src/lib/inngest/functions/agent/tools.ts
```diff
@@ -0,0 +1,174 @@
...
+        ? `### ${name}\n${found.schema}`
+        : `### ${name}\n(no schema available — avoid referencing this event's data fields)`;
+    });
+    return { observation: parts.join('\n\n') };
+  },
+};
+
+// Execution is intercepted by the loop (it needs durable wait primitives);
+// this entry only contributes the tool schema the model sees.
+export const validateQueryTool: ToolDef = {
```
rhino1998, line 89: Do you want a new GQL handler to just validate a query without executing it? Adding that would be straightforward (although notably it would only catch errors that the transpiler can catch (which is most tbf))

ui/apps/dashboard/src/lib/inngest/functions/agent/tools.ts
```diff
@@ -0,0 +1,174 @@
+import { z } from 'zod';
+
+import { VALIDATE_QUERY, type ToolDef } from './loop';
+
+// Tools for the Insights agent. The static table schemas live in the system
```
rhino1998, line 5: Should the agent use the tools to get current table schemas so we don't need to keep a separate schema updated?

The validate query tool should work for this with queries like `SHOW TABLES` and `DESCRIBE <table>`

ui/apps/dashboard/src/lib/inngest/functions/agent/tools.ts
```diff
@@ -0,0 +1,174 @@
+import { z } from 'zod';
+
```
rhino1998, line 2: Maybe we should also allow the agent to access a user's saved queries? A properly named saved query could really help with deciphering what a user means by some customer-specific filter concept

rhino1998 (approved):
Looks great! I added some suggestions for extensions/additional tools

PR #4602, opened 2026-07-10
fix(constraintapi): fix Constraint API marshaling

BrunoScheufler:
## Description

This PR should fix marshaling for the Constraint API 

This broke in https://github.com/inngest/inngest/pull/4596

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/constraintapi/acquire.go
+++ b/pkg/constraintapi/acquire.go
@@ -162,14 +162,14 @@ type acquireScriptResponse struct {
 		LeaseID             ulid.ULID `json:"lid"`
 		LeaseIdempotencyKey string    `json:"lik"`
 	} `json:"l"`
-	LimitingConstraints     flexibleIntArray        `json:"lc"`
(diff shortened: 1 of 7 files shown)


dlambrig:
Diagnosis from codex LGTM

The bug: the Constraint API builds responses in Lua and encodes with cjson, which turns an empty list into {} instead of []. The new cu field is a list on the Go side, so any response with no usage entries (common — e.g. functions without concurrency/throttle/rate limits) comes back as "cu":{} and fails to decode (cannot unmarshal object into … cu). That error surfaces as a gRPC error from Acquire, so capacity can't be acquired and runs don't schedule.
  
We already handle this same cjson quirk for the sibling fields (lc/ec/d use the flexible-array types) — cu just missed it. This PR gives cu the same {}-tolerant handling across acquire/check/extend/release, with tests.

albertchae (approved):

KiKoS0 (approved):

PR #4604, opened 2026-07-10
fix(queue): skip constraint lease on missing account

dlambrig:
## Summary

- treats a confirmed missing account during Constraint API item lease as non-blocking
- lets the executor continue to its inactive-account cleanup path instead of retrying constraint lease forever
- keeps transient Constraint API/network errors fail-closed

## Testing

- `go test ./pkg/execution/queue -run '^$' -count=0`

### DIFF
--- a/pkg/execution/queue/constraints.go
+++ b/pkg/execution/queue/constraints.go
@@ -298,7 +298,9 @@ func (q *queueProcessor) BacklogRefillConstraintCheck(
 			metrics.IncrBacklogRefillConstraintCheckCounter(ctx, enums.BacklogRefillConstraintCheckReasonAccountMissing.String(), metrics.CounterOpt{
 				PkgName: pkgName,
 			})
-			return nil, nil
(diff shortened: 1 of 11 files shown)


pkg/execution/queue/constraints.go
```diff
@@ -488,6 +490,13 @@ func (q *queueProcessor) ItemLeaseConstraintCheck(
 		},
 	})
 	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
```
BrunoScheufler: I wonder if there's a way that we don't have to run a database-scoped check here. I generally agree that we have to handle errors in this path properly.

BrunoScheufler: Ideally we could return a known error from the upstream shard selector implementation

dlambrig: done

BrunoScheufler (approved):

PR #4607, opened 2026-07-13
chore: Add usage tracking for Experiments

AndyInternet:
## Description

Add tracking with Segment for Experiments usage.

## Release note

None.

## Migration note

None.


ui/apps/dashboard/src/components/Experiments/useExperiments.ts
```diff
@@ -75,6 +76,8 @@ export function useExperimentsList({
       firstSeen: new Date(exp.firstSeen),
       lastSeen: new Date(exp.lastSeen),
     }));
+
+    return items;
```
Linell, line 80: I don't think it matters at all, but these changes look like they may be from debugging? `const items / return items`

Linell (approved):

PR #4608, opened 2026-07-13
fix: fix the ordering and queue time of async opcodes

scottnuma:
## Description

- When checkpointing is on, the async opcodes set their queuedAt at times based on queuedAt of their discovery span. However, other steps may execute between the start of discovery and the async opcode. This leads to displaying too long of queue times and incorrect order

invoke-child should be between step-before and step-after, and its queue time should not precede step-before

Screenshot 2026-07-13 at 3 26 03 PM

- We adjust the queuedAt of invoke-child's queued at instead. This fixes queueing and ordering

Screenshot 2026-07-13 at 3 27 06 PM

- We make our change server-side in the executor. This applies to our async opcodes: invoke, sleep, waitForEvent, waitForSignal, and Promise.all
  - Promise.all doesn't suffer from the same bug before this change, due to the overwrites we perform. We adjust it anyway for consistency.
- We introduce testing at the opcode -> trace layer to validate our change and prevent regressions

## Release note

Fix the ordering and display of queueing delay for invoke, sleep, waitForEvent, waitForSignal, and Promise.all when checkpointing is on

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -3698,6 +3698,17 @@ func (e *executor) HandleGenerator(ctx context.Context, runCtx execution.RunCont
 	return e.handleGeneratorOp(ctx, runCtx, gen, OpcodeGroup{})
 }
 
+// opcodeHandledAt returns the time async-opcode handlers stamp as their step
+// span's queue time: the group's shared handling time when the opcode arrived
+// in an SDK response (keeping parallel opcodes' trace order deterministic),
+// else the handler's own clock (e.g. opcodes fed via the checkpoint API).
+func (e *executor) opcodeHandledAt(group OpcodeGroup) time.Time {
+	if !group.HandledAt.IsZero() {
+		return group.HandledAt
+	}
+	return e.now()
+}
+
 func (e *executor) handleGeneratorOp(ctx context.Context, runCtx execution.RunContext, gen state.GeneratorOpcode, group OpcodeGroup) error {
 	// Grab the edge that triggered this step execution.
 	lifecycleItem := runCtx.LifecycleItem()
@@ -3727,7 +3738,7 @@ func (e *executor) handleGeneratorOp(ctx context.Context, runCtx execution.RunCo
 	case enums.OpcodeStepPlanned:
 		return e.handleGeneratorStepPlanned(ctx, runCtx, gen, edge, group)
 	case enums.OpcodeSleep:
-		return e.handleGeneratorSleep(ctx, runCtx, gen, edge)
+		return e.handleGeneratorSleep(ctx, runCtx, gen, edge, group)
 	case enums.OpcodeWaitForEvent:
 		return e.handleGeneratorWaitForEvent(ctx, runCtx, gen, edge, group)
 	case enums.OpcodeInvokeFunction:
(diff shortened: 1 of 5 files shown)


pkg/execution/executor/executor.go
```diff
@@ -4362,7 +4380,15 @@ func (e *executor) handleGeneratorStepPlanned(ctx context.Context, runCtx execut
 	}
 
 	attrs := tracing.GeneratorAttrs(&gen)
-	tracing.AddQueueTimestampAttrs(attrs, runCtx.LifecycleItem())
+	// The planned step is queued the moment this opcode is handled (its own
+	// queue item's EnqueuedAt is only stamped at Enqueue, below). The queued
+	// segment still covers any wait for a concurrency slot: handling time →
+	// execution start. See OpcodeGroup.HandledAt, which avoids a stale
+	// queuedAt inherited from the request.
+	{
```
tonyhb, line 4388: the block scoping is nice 🙌

tonyhb (approved):

pkg/execution/executor/executor.go
```diff
@@ -4439,7 +4465,13 @@ func (e *executor) handleGeneratorSleep(ctx context.Context, runCtx execution.Ru
 	lifecycleItem := runCtx.LifecycleItem()
 	metadata := runCtx.Metadata()
 	attrs := tracing.GeneratorAttrs(&gen)
-	tracing.AddQueueTimestampAttrs(attrs, runCtx.LifecycleItem())
+	// The sleep begins the moment this opcode is handled — not when the
+	// reporting request was enqueued. See OpcodeGroup.HandledAt, which avoids
+	// a stale queuedAt inherited from the request.
+	{
+		handledAt := e.opcodeHandledAt(group)
+		tracing.AddTimingAttrs(attrs, handledAt, handledAt, time.Time{}, time.Time{})
```
rhino1998, line 4473: Do we want to also set started_at? (Or is that handled by the start time on the step span in the span creation logic?)

pkg/execution/executor/executor.go
```diff
@@ -4863,7 +4895,13 @@ func (e *executor) handleGeneratorWaitForSignal(ctx context.Context, runCtx exec
 
 	lifecycleItem := runCtx.LifecycleItem()
 	attrs := tracing.GeneratorAttrs(&gen)
-	tracing.AddQueueTimestampAttrs(attrs, lifecycleItem)
+	// The wait begins the moment this opcode is handled — not when the
+	// reporting request was enqueued. See OpcodeGroup.HandledAt, which avoids
+	// a stale queuedAt inherited from the request.
+	{
+		handledAt := e.opcodeHandledAt(group)
+		tracing.AddTimingAttrs(attrs, handledAt, handledAt, time.Time{}, time.Time{})
```
rhino1998, line 4903: Do we want to set started_at here too?

rhino1998 (approved):
Seems reasonable. I think there might still be an issue with step ordering in the case that checkpointing ended due to hitting the checkpoint timeout (Which I think would return a normal step.run opcode that would otherwise have been checkpointable) That can be a future PR fix tho as it is out of scope here probably

amh4r (approved):

pkg/execution/executor/executor.go
scottnuma: The previous `tracing.AddQueueTimestampAttrs(attrs, lifecycleItem)` doesn't add startedAt or endedAt, so this keep consistency.

pkg/execution/executor/executor.go
scottnuma: Similar to the other location

scottnuma:
Did some digging around, and there does seem to be an issue with step ordering upon checkpoint timeout. I'll put that out for a separate PR!

scottnuma:
Ticket for tracking similar issue that Riley called out: https://linear.app/inngest/issue/EXE-2026/checkpointed-steprun-inherits-stale-queuedat-on-checkpoint-api-failure

PR #4610, opened 2026-07-14
feat(queue): define grpc service interfaces for queue producer and consumer

lkasinathan:
## Description

Adds queue producer/consumer gRPC contracts and proto conversion coverage needed for queue-proxy work. 

Unused in OSS.

## Release note

None.

## Migration note

None.


pkg/execution/queue/convert.go
```diff
@@ -0,0 +1,651 @@
...
+
+	return Item{
+		JobID:                 msg.JobId,
+		GroupID:               msg.GetGroupId(),
+		WorkspaceID:           workspaceID,
+		Kind:                  msg.GetKind(),
+		Identifier:            identifier,
+		Attempt:               int(msg.GetAttempt()),
+		MaxAttempts:           int64PtrToIntPtr(msg.MaxAttempts),
+		Payload:               payloadFromProto(msg.GetPayload()),
```
dlambrig: This evaluates to raw JSON for every kind. But the existing queue decode path turns known kinds back into typed payload structs (like edge, pause timeout, and job promote).

I think this changes behavior. The executor relies on those concrete types, so after crossing this gRPC boundary those jobs look like they would fail type assertions at runtime.

Could we share or reuse the existing kind-aware payload decoding logic from the current queue item unmarshal path here, and add a test that verifies the payload type for a few built-in kinds?

pkg/execution/queue/convert.go
```diff
@@ -0,0 +1,651 @@
...
+	if len(semaphores) == 0 {
+		return nil
+	}
+	result := make([]*pb.Semaphore, len(semaphores))
+	for i, semaphore := range semaphores {
+		result[i] = &pb.Semaphore{
+			Id:               semaphore.ID,
+			EvaluatedKeyHash: semaphore.EvaluatedKeyHash,
+			Weight:           semaphore.Weight,
+			Release:          strconv.Itoa(int(semaphore.Release)),
```
dlambrig: This sends the release mode over the wire as its numeric value encoded as a string, for example `"0"` or `"1"`. That makes the contract harder to understand and couples it to the current Go enum numbering.

On decode, any numeric value is accepted and cast to `SemaphoreReleaseMode`, so an unsupported value such as `"17"` would still deserialize into an invalid enum state.

Since this is a new RPC contract, could we use a protobuf enum here, or at least send and validate a named string value?

dlambrig (changes requested):
lg! left two comments

BrunoScheufler (approved):

pkg/execution/queue/convert.go
lkasinathan: Fixed in 1f010a55e70498009657570d7a19b69f040efc2e

> so after crossing this gRPC boundary those jobs look like they would fail type assertions at runtime.

I don't think it's quite that catastrophic as eventually Enqueue via proxy simply persists the json.RawMessage to the underlying shard (same as what a direct producer would).  I added some [compatibility tests in Cloud](https://github.com/inngest/monorepo/blob/1d2a2a4f96408fcfb3b04cc117feb6a724d86433/pkg/queue_proxy/queue_proxy_compatibility_test.go) to verify that.

However, I'm adopting the suggestion to make the proxy-Enqueue and direct-Enqueue both receive the same typed payload.

pkg/execution/queue/convert.go
lkasinathan: Good catch! fixed in c7f2c00362fcf8ffb5a498b29c08398ca9080fa9

dlambrig (approved):
Took another pass — LGTM to merge. Payload-typing and semaphore fixes look solid, and sharing `decodePayloadForKind` between `UnmarshalJSON` and the proto path means storage and proxy decode identically (checked the refactor preserves the old kind coverage). The reflection-based field-coverage test is a great guard.

One optional follow-up: it guards struct fields but not payload kinds — a future in-package Kind+Payload added without a `decodePayloadForKind` case would silently fall back to raw JSON. A small kind→type test would close that. Not blocking. 👍

PR #4613, opened 2026-07-14
feat(queue): route RequeueByJobID through producer

lkasinathan:
### Summary

  - Export the queue producer constructor/options for reuse outside the queue processor.
  - Add `queue.Scope` to `Producer.RequeueByJobID`.
  - Route debounce `RequeueByJobID` calls through the producer instead of directly through the shard.

All three changes are no-op in OSS and Cloud

###  Why

  Queue proxy routing needs account/workspace/function scope to decide whether a producer operation should go through the proxy or stay local. `RequeueByJobID` previously only carried shard/job/time, so it could
  not be account feature-flagged.

  Keeping shard-level `RequeueByJobID` scope-free preserves the low-level backend primitive, while moving production callers through the producer gives routing layers enough context.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/queue/convert.go
+++ b/pkg/execution/queue/convert.go
@@ -90,6 +90,44 @@ func QueueItemFromProto(msg *pb.QueueItem) (QueueItem, error) {
 	return item, nil
 }
 
+func ScopeToProto(scope Scope) *pb.Scope {
+	return &pb.Scope{
+		IsSystem:   scope.IsSystem,
+		AccountId:  scope.AccountID.String(),
+		EnvId:      scope.EnvID.String(),
+		FunctionId: scope.FunctionID.String(),
+	}
+}
+
+func ScopeFromProto(msg *pb.Scope) (Scope, error) {
+	if msg == nil {
+		return Scope{}, fmt.Errorf("scope is required")
+	}
+
+	scope := Scope{IsSystem: msg.GetIsSystem()}
+	if scope.IsSystem {
+		return scope, nil
+	}
+
+	accountID, err := parseUUID(msg.GetAccountId(), "scope account_id")
+	if err != nil {
+		return Scope{}, err
+	}
+	envID, err := parseUUID(msg.GetEnvId(), "scope env_id")
+	if err != nil {
+		return Scope{}, err
+	}
+	functionID, err := parseUUID(msg.GetFunctionId(), "scope function_id")
+	if err != nil {
+		return Scope{}, err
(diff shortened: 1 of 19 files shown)


pkg/execution/queue/convert.go
```diff
@@ -91,6 +91,44 @@ func QueueItemFromProto(msg *pb.QueueItem) (QueueItem, error) {
...
+	}
+}
+
+func ScopeFromProto(msg *pb.Scope) (Scope, error) {
+	if msg == nil {
+		return Scope{}, fmt.Errorf("scope is required")
+	}
+
+	scope := Scope{IsSystem: msg.GetIsSystem()}
+	if scope.IsSystem {
```
dlambrig: For system scopes, `ScopeFromProto()` returns early and drops any populated account, env, or function IDs. This seems to make is_system the field that decides how the scope is interpreted, but it also means that if a payload includes both is_system and IDs, those IDs get dropped when the payload is decoded and encoded again.

Could we document that IDs are ignored for system scopes, or add a test that makes the intended behavior explicit?

dlambrig (commented):
this is cool. can you add a few more tests?

- ScopeToProto / ScopeFromProto do not have focused coverage yet, especially for system scope and malformed UUID input.
  - The debounce changes do not appear to verify that Producer.RequeueByJobID is called with the expected scope and shard name.

BrunoScheufler (approved):

pkg/execution/queue/convert.go
lkasinathan: Good catch! This was definitely not intentional. scope is used mainly in Cloud client implementation to feature flag the proxy requests based on tenant identifiers. To address this comment:
- added a comment in queue.go clarifying that RequeueByJobID requires non-empty IDs in scope.
- added unit tests for producer.RequeueByJobID
- Fixed ScopeFromProto to always return IDs, added round trip tests to ensure IDs are not dropped for system scopes.

lkasinathan:
> this is cool. can you add a few more tests?
> 
> * ScopeToProto / ScopeFromProto do not have focused coverage yet, especially for system scope and malformed UUID input.
>   
>   * The debounce changes do not appear to verify that Producer.RequeueByJobID is called with the expected scope and shard name.

Sorry, I really should have done this before sending this out.
- Added unit tests for ScopeToProto and ScopeFromProto
- Added unit tests for producer.RequeueByJobID to test invalid scope IDs.

PR #4614, opened 2026-07-15
Feat/add dpa form to support

campbell88:
## Description

Added a DPA form in the Support Portal
Screenshot 2026-07-15 at 6 55 26 AM

## Release note

Allow users to submit a DPA request form through the Support Portal

## Migration note

None.


ui/apps/support/src/data/plain.ts
```diff
@@ -94,11 +102,10 @@ export type TicketStatusFilter =
   | typeof TICKET_STATUS_ALL;
 
 export const getTicketsByEmail = createServerFn({ method: "GET" })
-  .middleware([authMiddleware])
```
scottnuma, line 97: It looks like we're removing the authMiddleware across ~9 functions and replacing with a manual `requireAuthenticatedEmail()`.

I think the authMiddleware already requires an authenticated email, so I think we can just keep the middleware

ui/apps/support/src/routes/_authed/new.tsx
```diff
@@ -184,7 +269,9 @@ function NewTicketPage() {
       setResult({
         ok: false,
         message:
-          "Failed to create support ticket. Please email hello@inngest.com if the problem persists.",
+          error instanceof Error
+            ? error.message
```
scottnuma, line 273: This can expose the direct error message to the user, which could contain sensitive information. It's safer to just always give them the generalized message below and ensure that we're logging the message.

Worth checking this in the several other locations that we're presenting errors to the user.

ui/apps/support/src/data/commonpaper.ts
```diff
@@ -0,0 +1,390 @@
...
+  );
+  const templateId = requireEnv(
+    "COMMONPAPER_DPA_TEMPLATE_ID",
+    process.env.COMMONPAPER_DPA_TEMPLATE_ID,
+  );
+
+  const companyAddress = toCompanyAddress(
+    request.companyAddress,
+    request.country,
+  );
```
scottnuma, line 237: toCompanyAddress looks likes it leaves city, state, and zip blank - might need to massage the data around to the proper fields.

scottnuma (commented):
looks like a useful feature - few pointers

PR #4620, opened 2026-07-15
feat: normalize and instrument AI feature empty state pages

AndyInternet:
## Description

Standardizes the Experiments, Scores, and Sessions empty-state pages on one new shared FeatureEmptyState component 
- title
- description (tracked docs link)
- value props
- un-collapsed coding-agent prompt (tracked copy event)
- get-started example (tracked copy event) 

Instruments the onboarding funnel with consistent per-feature Segment events that denote empty state page actions
_object_ **Empty State** _action_

## Release note

None.

## Migration note

None.


ui/apps/dashboard/src/components/Experiments/tracking.ts
```diff
@@ -188,3 +192,19 @@ export function trackExperimentTimeRangeChanged({
...
+  trackExperimentsEvent('Experiment Empty State Prompt Copied');
+}
+
+export function trackExperimentEmptyStateExampleCopied() {
+  trackExperimentsEvent('Experiment Empty State Example Copied');
+}
+
+export function trackExperimentEmptyStateDocsLinkOpened() {
+  trackExperimentsEvent('Experiment Empty State Docs Link Opened');
+}
```
djfarrelly: I think we should normalize these to remove the prefix and lean on the `features` property to aggregate. 

This also allows us to do aggregations across queries as well, so in that sense:

```
Empty State Viewed
Empty State Prompt Copied
Empty State Example Copied
Empty State Docs Link Opened
```

I think at some point (not required in this PR), we should make these shard event names constants in a tracking events file. We also can store related Types there as well like:

```ts
export const eventNameEmptyStateViewed = 'Empty State Viewed';
export type EventEmptyStateViewed = {
  feature: 'scores' | 'experiments' | 'sessions';
}
// maybe even helpers like 
export function trackEmptyStateViewed(attributes: EventEmptyStateViewed): {
   analytics.track(eventNameEmptyStateViewed, attributes);
}
```

Centralizing this can help this a ton for maintaining and avoiding devs or agents re-creating all sorts of events that create 1000s of events to sort though.

ui/packages/components/src/FeatureEmptyState/FeatureEmptyState.tsx
```diff
@@ -0,0 +1,144 @@
...
+export type FeatureEmptyStateProps = {
+  title: string;
+  description: ReactNode;
+  docsUrl: string;
+  onDocsLinkClick?: () => void;
+  valueProps: FeatureEmptyStateValueProp[];
+  prompt: {
+    description: ReactNode;
+    content: string;
+    onCopy?: () => void;
```
djfarrelly: If we take the approach of standard event name and use "features" for tagging each, then we can remove the need for each of these `onCopy` functions and instead just ask for the `feature` slug in these props and this component can just pass that down through to each analytics tracking.

djfarrelly (changes requested):
Left notes on naming and data governance

djfarrelly:
thought: should these be more generic for diff types of properties to be suited to other places in the app, like metrics or function dashboard? It may also be ok

ui/packages/components/src/FeatureEmptyState/FeatureEmptyState.tsx
djfarrelly: Let's still do this one!

ui/apps/dashboard/src/components/Experiments/useExperiments.ts
```diff
@@ -277,13 +277,18 @@ export function useExperimentDetail(
...
+        'feature' | 'experimentName' | 'functionSlug'
       >,
     ) => {
       if (!shouldTrack) return;
       trackedViewKeyRef.current = viewKey;
-      trackExperimentDetailViewed({ experimentName, functionSlug, ...props });
+      trackDetailViewed({
+        feature: 'experiments',
+        experimentName,
+        functionSlug,
```
djfarrelly: thought: should this be a generic "name" prop so it could work for function detail views too, so feature: functions, name = "My cron job" or maybe what you're tracking here is an experiment ID?

second thought: do we need to track the experiment name at all here? it's probably going to be high cardinality and low value maybe. We could just drop `experimentName` and `functionSlug`

djfarrelly (changes requested):

ui/apps/dashboard/src/utils/analyticsEvents.ts
```diff
@@ -0,0 +1,267 @@
...
+}: DetailViewedArgs) {
+  track('Detail Viewed', feature, {
+    experiment_name: experimentName,
+    function_slug: functionSlug,
+    duration_ms: durationMs,
+    error_type: errorType,
+    result,
+    run_count: runCount,
+    selection_strategy: selectionStrategy,
+    variant_count: variantCount,
```
djfarrelly: we probably could drop all of these for now

ui/apps/dashboard/src/utils/analyticsEvents.ts
```diff
@@ -0,0 +1,267 @@
...
+};
+
+export function trackWaitlistFormSubmitted({
+  feature,
+  canContact,
+  message,
+}: WaitlistFormSubmittedArgs) {
+  track('Waitlist Form Submitted', feature, {
+    can_contact: canContact,
+    message: message,
```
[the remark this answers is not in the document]
AndyInternet: This is intentional. No fix needed

ui/apps/dashboard/src/components/Experiments/useExperiments.ts
AndyInternet: Dropping these

ui/apps/dashboard/src/utils/analyticsEvents.ts
AndyInternet: Dropping these

ui/apps/dashboard/src/components/Experiments/ExperimentDetailPage.tsx
```diff
@@ -176,16 +172,6 @@ export function ExperimentDetailPage({
...
-      );
-      trackExperimentTimeRangeChanged({
-        experimentName,
-        functionSlug,
-        rangeDays,
-        hitMaxRange: rangeDays >= daysAgoMax,
-      });
       onTimeRangeChange(nextRange);
     },
     [onTimeRangeChange, experimentName, functionSlug, daysAgoMax],
```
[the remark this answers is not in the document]
djfarrelly: Makes sense

djfarrelly:
Resolved

djfarrelly (approved):

PR #4621, opened 2026-07-15
feat(api): add v2 runs list endpoints

jacobheric:
## Description

Adds beta v2 REST endpoints for listing function runs:

- `GET /v2/runs`
- `GET /v2/apps/{appId}/functions/{functionId}/runs`

Supports cursor pagination, time ranges and fields, status, public app/function IDs, deferred-run filters, ordering, and optional outputs.

Depends on [#4623](https://github.com/inngest/inngest/pull/4623).

### Examples

```bash
# Failed or cancelled runs
curl -G "$API/runs" \
  --data-urlencode status=FAILED \
  --data-urlencode status=CANCELLED

# Function runs started within a time range
curl -G "$API/apps/inngest-ai/functions/hello-world/runs" \
  --data-urlencode from=2026-07-01T00:00:00Z \
  --data-urlencode until=2026-07-15T00:00:00Z \
  --data-urlencode timeField=startedAt \
  --data-urlencode order=ASC
```
note you can also pass app id and function id as params to `/v2/runs` in addition to using `/v2/apps/{appId}/functions/{functionId}/runs`, see:

```
curl -sS "${AUTH[@]}" -G "$API/runs" \
  --data-urlencode appId="inngest-ai" \
  --data-urlencode functionId="hello-world" \
  --data-urlencode limit=20 
```

## Release note

Adds beta v2 REST endpoints for listing function runs:

- `GET /v2/runs`
- `GET /v2/apps/{appId}/functions/{functionId}/runs`

Supports cursor pagination, time ranges and fields, status, public app/function IDs, deferred-run filters, ordering, and optional outputs.

## Migration note

None.

### DIFF
--- a/pkg/api/v2/endpoints_runs.go
+++ b/pkg/api/v2/endpoints_runs.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/inngest/inngest/pkg/api/v2/apiv2base"
@@ -23,6 +24,8 @@ import (
 const (
 	defaultEventRunsLimit = 20
 	maxEventRunsLimit     = 40
+	defaultRunsLimit      = 20
+	maxRunsLimit          = 100
 )
 
 func (s *Service) GetFunctionRun(ctx context.Context, req *apiv2.GetFunctionRunRequest) (*apiv2.GetFunctionRunResponse, error) {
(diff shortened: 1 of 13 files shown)


proto/api/v2/service.proto
```diff
@@ -3063,3 +3099,121 @@ message SessionRun {
...
+  optional google.protobuf.Timestamp from = 4 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Start of the run time range"
+    }
+  ];
+  optional google.protobuf.Timestamp until = 5 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "End of the run time range"
+    }
+  ];
```
albertchae, line 3125: are these both inclusive?

proto/api/v2/service.proto
```diff
@@ -3063,3 +3099,121 @@ message SessionRun {
...
+  string time_field = 6 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Run timestamp field used for filtering and ordering. Accepts queuedAt, startedAt, or endedAt."
+      default: "queuedAt"
+    }
+  ];
+  repeated string status = 7 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Statuses to include, using response status values such as COMPLETED, FAILED, RUNNING, QUEUED, or CANCELLED"
```
albertchae: the casing inconsistency is a little weird here. Should we use something like  QUEUED_AT/STARTED_AT/ENDED_AT for time_field?

proto/api/v2/service.proto
```diff
@@ -3063,3 +3099,121 @@ message SessionRun {
...
+  optional google.protobuf.Timestamp until = 7 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "End of the run time range"
+    }
+  ];
+  string time_field = 8 [
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Run timestamp field used for filtering and ordering. Accepts queuedAt, startedAt, or endedAt."
+      default: "queuedAt"
+    }
```
albertchae, line 3189: same questions as other RPC endpoint

pkg/api/v2/endpoints_runs.go
```diff
@@ -189,6 +326,77 @@ func runsPage(runs []*RunListItem, limit int, hasMore bool) *apiv2.Page {
...
+}
+
+func runTimeFieldFromAPI(field string) (RunTimeField, error) {
+	switch normalizeRunFilterToken(field) {
+	case "", "QUEUEDAT", "QUEUED_AT":
+		return RunTimeFieldQueuedAt, nil
+	case "STARTEDAT", "STARTED_AT":
+		return RunTimeFieldStartedAt, nil
+	case "ENDEDAT", "ENDED_AT":
+		return RunTimeFieldEndedAt, nil
```
albertchae, line 372: oh it looks like we support the caps snake case already. Maybe that's what we should advertise in the generated openAPI

pkg/devserver/run_provider.go
```diff
@@ -91,6 +104,24 @@ func (p *runProvider) GetRuns(ctx context.Context, opts apiv2.GetRunsOpts) (*api
...
 }
 
+func cqrsRunTimeField(field apiv2.RunTimeField) enums.TraceRunTime {
+	switch field {
+	case apiv2.RunTimeFieldStartedAt:
+		return enums.TraceRunTimeStartedAt
+	case apiv2.RunTimeFieldEndedAt:
+		return enums.TraceRunTimeEndedAt
+	default:
+		return enums.TraceRunTimeQueuedAt
```
albertchae: should we add an explicit case for queued_at and then default should be some kind of error? in case we add another time_field in the future it'd be easier to catch

albertchae (approved):

pkg/api/v2/service_test.go
```diff
@@ -920,6 +920,221 @@ func TestService_GetFunctionRun(t *testing.T) {
...
+		Status:       enums.RunStatusCompleted,
+		EndedAt:      &endedAt,
+		FunctionID:   "test-fn",
+		FunctionName: "Test function",
+		AppID:        "my-app",
+	}
+
+	t.Run("returns mapped runs with filters", func(t *testing.T) {
+		reader := &mockRunProvider{}
+		reader.On("GetRuns", mock.Anything, GetRunsOpts{
```
albertchae, line 954: I don't love testing all these with mocks but it seems that's the pattern so far with api v2 so it's fine.

Do we need corresponding tests in `pkg/api/v2/http_test.go` as well?

proto/api/v2/service.proto
jacobheric: They weren't, good catch. I made them both inclusive, which matches the cloud behavior as well.

proto/api/v2/service.proto
jacobheric: generally the v2 open api docs reference camel case everywhere but accept that and snake case. I've updated the docs/desc here to reflect that.

proto/api/v2/service.proto
jacobheric: kept camel case for consistency, but called out that we also support snake case.

pkg/devserver/run_provider.go
jacobheric: done.

pkg/api/v2/service_test.go
jacobheric: added tests in `pkg/api/v2/http_test.go` as well. I can do a cleanup pass after this for these mock tests. what would you prefer to see instead, more full e2e testing or some other fakes pattern?

pkg/api/v2/endpoints_runs.go
jacobheric: updated the docs

pkg/api/v2/service_test.go
albertchae: I generally have more confidence with e2e testing but it can make tests costly... I guess if we already have confidence that internal CQRS stuff is tested well it's not necessary 🤔

proto/api/v2/service.proto
albertchae: oh ok, slightly weird to me if the fields are snake case but values can be camel, but as long as we're consistent across all our APIs

albertchae (approved):

PR #4622, opened 2026-07-15
refactor(runs): new shared run list contract

jacobheric:
## Description

A prerequisite for the new v2 runs api so we can more easily consolidate the event/runs, runs and gql paths and share the fetch, pagination, filtering etc. 

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/db/postgres/helpers.go
+++ b/pkg/db/postgres/helpers.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"encoding/json"
+	"strings"
 	"time"
 
 	sq "github.com/doug-martin/goqu/v9"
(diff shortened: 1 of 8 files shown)


pkg/db/postgres/helpers.go
```diff
@@ -28,6 +29,32 @@ func (h *helpers) RootEventIDsExpr() sqexp.Expression {
 	return sq.L("spans.event_ids::text").As("event_ids")
 }
 
+func (h *helpers) EventIDsContain(ids []string) sqexp.Expression {
```
jacobheric, line 32: helpers so our api event/runs an api runs with outputs can use the new consolidated list.

PR #4623, opened 2026-07-15
refactor(runs): consolidate oss runs queries

jacobheric:
## Description

creates a new `GetRun` cqrs contract based on the existing goqu query that uses the latest `spans` table so we can use this as the single source of runs truth going forward.  switches the existing v2 event/runs endpoint to use that and adds all the missing bits we need for that and the new runs api. Removes the sqlc version that did the equivalent. 

v2 runs api is built on top of this [here](https://github.com/inngest/inngest/pull/4621).

## Release note

None.

## Migration note

None.


pkg/api/v2/endpoints_runs.go
```diff
@@ -156,28 +156,26 @@ func (s *Service) Rerun(ctx context.Context, req *apiv2.RerunRequest) (*apiv2.Re
 	}, nil
 }
 
-func runsPageOpts(cursor string, requestedLimit int32) (ulid.ULID, int, error) {
+func runsPageOpts(cursor string, requestedLimit int32) (string, int, error) {
```
jacobheric, line 158: A minor change to the existing event runs cursor behavior here to use the better composite start time + run time cursor that the v2 runs api and gql use. it's in beta so I think it's ok to change now and will allow us to consolidate to just one contract eventually.

pkg/cqrs/cqrs.go
```diff
@@ -29,6 +29,7 @@ type Manager interface {
 	EventManager
 	HistoryReader
 	HistoryWriter
+	GetRuns(ctx context.Context, opt GetTraceRunOpt) ([]*TraceRun, error)
```
jacobheric, line 32: just GetRuns so our cqrs contracts are not tightly coupled to the db layer....and so everything including gql can eventually just use GetRuns and not stuff that is reflective of data structures like GetTraceRuns.

pkg/api/v2/endpoints_function.go
```diff
@@ -314,21 +314,11 @@ func decodePathParam(value string) string {
 }
 
 func publicFunctionID(appID string, requestedFunctionID string, fn inngest.DeployedFunction) string {
-	if fn.Function.Slug != "" && fn.Function.Slug != fn.Slug {
```
jacobheric, line 317: This logic was duped across 3 places, deduped.

pkg/devserver/run_provider.go
```diff
@@ -204,61 +220,42 @@ func runEventID(root *cqrs.OtelSpan) (ulid.ULID, error) {
...
 }
 
-func runListItemFromRow(row *db.RunListItemRow, includeOutput bool) *apiv2.RunListItem {
-	fn := inngest.Function{}
-	_ = json.Unmarshal([]byte(row.FunctionConfig), &fn)
-
-	functionName := fn.Name
-	if functionName == "" {
-		functionName = row.FunctionName
+func runListItemFromCQRS(row *cqrs.TraceRun, includeOutput bool) *apiv2.RunListItem {
```
jacobheric, line 223: we're no longer mapping sqlc rows, we mapping the cqrs rows returned by the consolidated GetRuns on runProviderDataReader.

pkg/cqrs/manager/cqrs.go
```diff
@@ -3671,10 +3692,91 @@ func (w wrapper) getSpanRunsPushdown(ctx context.Context, opt cqrs.GetTraceRunOp
...
+
+	q := sq.Dialect(w.dialect()).
+		From("spans").
+		Select("spans.run_id", w.helpers().RunOutputExpr()).
+		Where(
+			sq.C("name").Eq(meta.SpanNameRun),
+			sq.C("debug_run_id").IsNull(),
+			sq.C("run_id").In(runIDs),
+		).
+		Order(sq.C("start_time").Asc())
```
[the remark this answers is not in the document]
jacobheric: I don't think this issues is real. `loadSpanRunPageOutput` executes one batched query for all page run ids. it filters `name = executor.run` and extension rows are named `EXTEND` so they are excluded.
the correlated output expression therefore runs once per root/run not once per extension.

pkg/cqrs/manager/cqrs.go
```diff
@@ -3843,6 +3965,70 @@ func spanRunGroupByCols() []interface{} {
...
+	if len(opt.Filter.AppName) > 0 || len(opt.Filter.FunctionSlug) > 0 {
+		return spanRunsWithMetadata(q)
+	}
+	return q
+}
+
+func spanRunAggregatePredicates(opt cqrs.GetTraceRunOpt, cursor *cqrs.TracePageCursor) []sq.Expression {
+	fieldName := "start_time"
+	aggregate := sq.L("MIN(spans.start_time)")
+	preds := []sq.Expression{}
```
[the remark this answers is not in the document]
jacobheric: I also don't think this is a real problem. a dynamic run span consists of the initial `executor.run` root followed by later `EXTEND` rows so `MIN(start_time)` is the root start time. and this is documented [here](https://github.com/inngest/inngest/blob/d037eb066d91303c811cd77f197ad594333d3a24/pkg/cqrs/manager/cqrs.go#L3893-L3900).

pkg/api/v2/endpoints_function.go
```diff
@@ -314,21 +314,11 @@ func decodePathParam(value string) string {
...
-
-	functionID := fn.Function.Slug
-	if functionID == "" {
-		functionID = fn.Slug
-	}
-	if trimmed := strings.TrimPrefix(functionID, appID+"-"); trimmed != functionID {
-		return trimmed
-	}
-	return functionID
+	return PublicFunctionID(appID, fn.Slug, fn.Function.Slug)
```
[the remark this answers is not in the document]
jacobheric: I think the refactored helper is equivalent, we have a test covering it that asserts the trimming behavior you call out so I think we're ok, see [here](https://github.com/inngest/inngest/blob/d037eb066d91303c811cd77f197ad594333d3a24/pkg/api/v2/service_test.go#L532-L546).

pkg/api/v2/function_ids.go
```diff
@@ -0,0 +1,19 @@
+package apiv2
+
+import "strings"
+
+// PublicFunctionID keeps function lookup and response IDs on the same rule.
+func PublicFunctionID(appID, storedFunctionID, configuredFunctionID string) string {
```
albertchae, line 6: so storedFunctionID might be appID prefixed and configuredFunctionID never is, right?

pkg/db/postgres/helpers.go
```diff
@@ -39,13 +82,17 @@ func (h *helpers) BuildEventJoin(q *sq.SelectDataset) *sq.SelectDataset {
...
+		return ids
+	}
+	if err := json.Unmarshal([]byte(*raw), &ids); err == nil {
+		return ids
+	}
+
+	var encoded string
+	if err := json.Unmarshal([]byte(*raw), &encoded); err == nil {
+		_ = json.Unmarshal([]byte(encoded), &ids)
 	}
```
albertchae, line 96: apparently we started storing single encoded JSON as of https://github.com/inngest/inngest/pull/3894/files#diff-3c3f2823d8473bb526981a84669c1b8a8717f4c5d563fb62aaa26f87dbcc5927R255, so it is necessary to try both single and double encoded, but it'd be nice to remove double encoding sometime if we provide a corresponding data migration or something

pkg/db/postgres/helpers.go
```diff
@@ -28,6 +29,48 @@ func (h *helpers) RootEventIDsExpr() sqexp.Expression {
...
+	return sq.L(`CASE
+		WHEN COALESCE((run_functions.config::json)->>'slug', '') <> ''
+			AND (run_functions.config::json)->>'slug' <> run_functions.slug
+			THEN (run_functions.config::json)->>'slug'
+		WHEN COALESCE(run_apps.name, '') <> ''
+			AND substr(COALESCE(NULLIF((run_functions.config::json)->>'slug', ''), run_functions.slug), 1, length(run_apps.name) + 1) = run_apps.name || '-'
+			THEN substr(COALESCE(NULLIF((run_functions.config::json)->>'slug', ''), run_functions.slug), length(run_apps.name) + 2)
+		ELSE COALESCE(NULLIF((run_functions.config::json)->>'slug', ''), run_functions.slug)
+	END`)
+}
```
albertchae, line 68: This looks like a SQL version of PublicFunctionID and is required if we want to support filtering by function_id without supplying app_id. Do we want that? seems like extra complexity and potentially slower queries too

pkg/devserver/run_provider.go
```diff
@@ -204,61 +220,42 @@ func runEventID(root *cqrs.OtelSpan) (ulid.ULID, error) {
...
+	runID, _ := ulid.Parse(row.RunID)
+	run := &apiv2.RunListItem{
+		RunID:        runID,
+		Cursor:       row.Cursor,
+		RunStartedAt: row.StartedAt,
+		FunctionID:   row.FunctionID.String(),
+		AppID:        row.AppID.String(),
```
albertchae, line 230: these should always be overridden by FunctionSlug and AppName later below right?

pkg/cqrs/manager/cqrs_test.go
```diff
@@ -1753,6 +1773,53 @@ func TestCQRSGetTraceRunsPagination(t *testing.T) {
...
+		DynamicSpanID: "dyn-other-app",
+		Name:          "executor.run",
+		StartTime:     baseTime.Add(4 * time.Second),
+		AccountID:     accountID.String(),
+		AppID:         otherAppID.String(),
+		FunctionID:    otherFunctionID.String(),
+		EnvID:         workspaceID.String(),
+	})
+
+	t.Run("function-only public filter matches every app", func(t *testing.T) {
```
albertchae: this is related to my other comment about whether we should support this

pkg/cqrs/manager/cqrs_test.go
```diff
@@ -1753,6 +1773,53 @@ func TestCQRSGetTraceRunsPagination(t *testing.T) {
...
+			},
+			Order: []cqrs.GetTraceRunOrder{{Field: enums.TraceRunTimeStartedAt, Direction: enums.TraceRunOrderDesc}},
+		})
+		require.NoError(t, err)
+		require.Len(t, runs, 4)
+		for _, run := range runs {
+			assert.Equal(t, "hello-world", run.FunctionSlug)
+			assert.Equal(t, "Hello world", run.FunctionName)
+			assert.Contains(t, []string{appName, otherAppName}, run.AppName)
+		}
```
albertchae: if we do keep this behavior, maybe a slight improvement to this test is to track the app counts explicitly

```
  appCounts := map[string]int{}
  for _, run := range runs {
        require.Equal(t, "hello-world", run.FunctionSlug)
        require.Equal(t, "Hello world", run.FunctionName)
        appCounts[run.AppName]++
  }

  require.Equal(t, map[string]int{
        appName:      3,
        otherAppName: 1,
  }, appCounts)
  ```

File: pkg/cqrs/manager/cqrs.go [hunk 14]
```diff
@@ -3843,6 +3965,70 @@ func spanRunGroupByCols() []interface{} {
...
+	if opt.Filter.TimeField == enums.TraceRunTimeEndedAt {
+		fieldName = "end_time"
+		aggregate = sq.L("MAX(spans.end_time)")
+		preds = append(preds,
+			aggregate.Gte(opt.Filter.From.UTC()),
+			aggregate.Lt(opt.Filter.Until.UTC()),
+		)
```
albertchae, line 4013: maybe add a small comment saying the SQLite UTC thing applies here too?

albertchae (approved):

pkg/db/postgres/helpers.go
jacobheric: yeah, we specifically don't support function id only in cloud so I'll remove it here for consistency.

pkg/api/v2/function_ids.go
jacobheric: `configredFuncitonId` is user defined so they can put the app id in there if they want, like `someapp-test-fn`, but then currently we store that as `someapp-someapp-test-fn`. so the first condition here tries to handle that so we don't strip a prefix in that situation.

pkg/db/postgres/helpers.go
jacobheric: ok, I can ticket a migration cleanup.

pkg/devserver/run_provider.go
jacobheric: yeah maybe this is duplicative and overly defensive. maybe it's better to return empty ids if we have some dat integrity issues so we don't give the user the impression they can use these.

pkg/cqrs/manager/cqrs_test.go
jacobheric: removed

pkg/cqrs/manager/cqrs.go
jacobheric: will do.

pkg/cqrs/manager/cqrs_test.go
jacobheric: removed this and the support for function only fetches.

pkg/devserver/run_provider.go
jacobheric: I clean this up on follow up.

PR #4630, opened 2026-07-16
test: Split V3 and V4 tests into different dirs and CI jobs

scottnuma:
## Description

- I'd like to bump the V3 and V4 versions that we test against to be up to date.
- We need to split them into two separate CI jobs, as the latest versions conflict with access to local state.

## Release note

None.

## Migration note

None.


.github/workflows/e2e.yml
```diff
@@ -18,12 +18,26 @@ env:
...
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
         experimentalKeyQueues: [false, true]
+        # The v3 and v4 SDKs cannot share a process: they cache their
+        # AsyncLocalStorage under the same global symbol with incompatible
+        # shapes, so whichever runs second reads a corrupt store and crashes.
+        # Run the v3 (pages-router) tests and the v4 durable-endpoint tests in
+        # separate jobs so each Next.js process only ever loads one SDK major.
```
scottnuma:
```suggestion
        # Run the v3 tests and the v4 tests in
        # separate jobs so each Next.js process only ever loads one SDK major.
```

PR #4632, opened 2026-07-17
internal: support null tombstones with session propagation

scottnuma:
## Description

- Users should be able to clear propagated session keys with a manual `key: null`
- Users should also be able to clear all propagated session keys by setting the sessions object to `null`
- https://linear.app/inngest/issue/EXE-2032/users-can-use-null-to-cancel-out-propagated-sessions

## Release note

None.

## Migration note

None.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/event/event_test.go
+++ b/pkg/event/event_test.go
@@ -56,12 +56,14 @@ func TestUnmarshalJSON(t *testing.T) {
 		r.Error(err)
 	})
 
-	t.Run("invalid session value type returns error", func(t *testing.T) {
+	t.Run("null manual session value is a tombstone, not an error", func(t *testing.T) {
 		r := require.New(t)
 
 		var evt Event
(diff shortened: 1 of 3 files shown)


amh4r:
How does validation work with the propagated sessions? Do we validate the keys (no empty strings, max length)? Do we validate the values (e.g. no nulls)? Do we validate the session count?

scottnuma:
We validate server side after merging/resolving propagated and manual sessions together. Our validation rules are unchanged from as we launched, so:

Do we validate the keys (no empty strings, max length)? yes 
Do we validate the values (e.g. no nulls)? yes, this one's type enforced I believe
Do we validate the session count? yes, after resolution. This means someone could add a ton of key:null, which would get dropped during resolution.

Validating all the way at server side is quite far way, especially for something like a defer. I was thinking it could be nice to do some simple validation client side later to get devs earlier feedback why their event was rejected.

amh4r (approved):

PR #4638, opened 2026-07-17
feat(api): v2 api cancel run

jacobheric:
## Summary

Adds `POST /api/v2/runs/{run_id}/cancel` to the v2 rest api. 

cloud pr [here](https://github.com/inngest/monorepo/pull/7876).

# Release Notes

Adds cancel run to the  v2 rest api and cli.

`POST /api/v2/runs/{run_id}/cancel`

`inngest api POST /api/v2/runs/{run_id}/cancel`

## Migration note

None

### DIFF
--- a/pkg/devserver/run_provider.go
+++ b/pkg/devserver/run_provider.go
@@ -17,6 +17,7 @@ import (
 	"github.com/inngest/inngest/pkg/event"
 	"github.com/inngest/inngest/pkg/execution"
 	"github.com/inngest/inngest/pkg/execution/executor"
+	state "github.com/inngest/inngest/pkg/execution/state/v2"
 	"github.com/inngest/inngest/pkg/tracing/meta"
 	"github.com/inngest/inngest/pkg/util"
 	"github.com/oklog/ulid/v2"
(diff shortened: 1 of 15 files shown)


pkg/devserver/run_provider.go
```diff
@@ -42,10 +48,33 @@ type runSpanReader interface {
...
+	}
+
+	return p.scheduler.Cancel(ctx, state.ID{
+		RunID:      runID,
+		FunctionID: fnrun.FunctionID,
+		Tenant: state.Tenant{
+			EnvID:     consts.DevServerEnvID,
+			AccountID: consts.DevServerAccountID,
+		},
+	}, execution.CancelRequest{})
```
Linell: Do you think we need `ForceLifecycleHook: true` here? With it false, `executor.Cancel` returns `nil` without finalizing or firing any cancellation lifecycle when the run's state metadata is already gone (`executor.go:3220-3249`), so a run that still shows as running gets a `200` with nothing recorded.

Linell (approved):

pkg/devserver/run_provider.go
jacobheric: done

PR #4644, opened 2026-07-20
internal: Add persistence support for sessions with defer

scottnuma:
## Description

- We're adding sessions support for defer. Deferred functions should be able to use both manual and propagated sessions
- In this PR, we add piping so that meta.sessions can be travel from the SDK's opcode -> persistence in Redis -> executor attaches it on to the `deferred.schedule` event
- In later monorepo PRs, we'll add scylladb support
- EXE-2035

## Release note

None.

## Migration note

None.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/execution/executor/finalize.go
+++ b/pkg/execution/executor/finalize.go
@@ -312,12 +312,17 @@ func (e *executor) buildDeferEvents(
 	now := e.now()
 	var events []event.Event
 
+	l := logger.StdlibLogger(ctx).With(
+		"run_id", util.SanitizeLogField(opts.Metadata.ID.RunID.String()),
+	)
+
 	for _, d := range defers {
+		l := l.With("fn_slug", d.FnSlug, "hashed_id", d.HashedID)
+
 		if err := d.Validate(); err != nil {
-			logger.StdlibLogger(ctx).Error(
+			l.Error(
 				"invalid defer",
 				"error", err,
-				"run_id", opts.Metadata.ID.RunID,
 			)
 			metrics.IncrDefersFinalizedCounter(ctx, "invalid", metrics.CounterOpt{PkgName: pkgName})
 			continue
@@ -331,11 +336,9 @@ func (e *executor) buildDeferEvents(
 
 		eventID, err := event.DeferEventID(opts.Metadata.ID.RunID, d.HashedID)
 		if err != nil {
-			logger.StdlibLogger(ctx).Error(
+			l.Error(
 				"failed to create defer event ID",
 				"error", err,
-				"hashed_id", d.HashedID,
-				"run_id", opts.Metadata.ID.RunID,
 			)
 			metrics.IncrDefersFinalizedCounter(ctx, "invalid", metrics.CounterOpt{PkgName: pkgName})
 			continue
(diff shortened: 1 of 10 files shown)


KiKoS0 (approved):

pkg/execution/state/opcode.go
```diff
@@ -424,6 +424,12 @@ func (g GeneratorOpcode) DeferAddOpts() (*DeferAddOpts, error) {
...
 	Input  json.RawMessage `json:"input,omitempty"`
+
+	// Meta carries control metadata (SDK-stamped session layers) for the
+	// deferred run. Opaque here: persisted with the defer and resolved onto
+	// the inngest/deferred.schedule event at finalize. Distinct from Input,
+	// which is the user payload.
+	Meta json.RawMessage `json:"meta,omitempty"`
 }
 
 func (d *DeferAddOpts) Validate() error {
```
Linell, line 435: Any reason `Meta` doesn't get the same `IsJSONObject` and size checks as `Input`? It rides in ARGV[2], so `saveDefer.lua` doesn't count it toward `MaxDeferInputAggregateSize` either.

The shape check feels like the useful half. Valid-but-not-an-object Meta isn't caught until `buildDeferEvents`, and at that point all we can do is drop the deferred run. Catching it here turns it into a `rejectReason` via `SaveFromOp`, where the metric and span already exist.

pkg/execution/executor/finalize.go
```diff
@@ -378,11 +378,45 @@ func (e *executor) buildDeferEvents(
...
+		manualSessions, propagatedSessions := len(evtMeta.Sessions) > 0, len(evtMeta.PropagatedSessions) > 0
+		evtMeta.ResolveSessions()
+		metrics.IncrEventSessionsResolvedCounter(ctx, "defer", manualSessions, propagatedSessions, metrics.CounterOpt{PkgName: pkgName})
+
+		// Validate the sessions after the merge.
+		//
+		// Failure to validate will drop the event, rather than truncate sessions or
+		// fail the parent run.
+		//
+		// Failures are unexpected as we also validate sessions SDK side.
```
Linell, line 412: On "we also validate sessions SDK side" - we may want a check for this over there too? `normalizeEventSessions` covers the count of 5 but not the 128/512 byte limits, and `propagatedSessions` is user-settable on `step.invoke()`, so a long id can sail through.

Linell (approved):
Nice work, this holds up well under scrutiny. Capturing tombstones off-map in `UnmarshalJSON` so `Sessions` stays a plain `map[string]string` for downstream readers is the right call, and I like that the single-hop invariant is pinned by real tests rather than asserted in a comment. The cjson reasoning on the new `deferMeta.Meta` field checks out for every path that exists today.

A couple of non-blocking comments inline. Plus one nit I couldn't attach since the file isn't in the diff: `saveDefer.lua`'s ARGV[2] header still says `{FnSlug, HashedID, ScheduleStatus} only`, but `Meta` rides in there now.

pkg/execution/state/opcode.go
scottnuma: Useful check, I've added this into https://github.com/inngest/inngest/pull/4649, which is more validation focused than this PR

scottnuma:
> saveDefer.lua's ARGV[2] header still says {FnSlug, HashedID, ScheduleStatus} only, but Meta rides in there now.

added a commit for this

pkg/execution/executor/finalize.go
scottnuma: Added a ticket to track this one https://linear.app/inngest/issue/EXE-2134/sdk-validates-the-keyvalue-length-limits

scottnuma:
Pulled https://github.com/inngest/inngest/pull/4702's logging improvements to appease the user provided logging concern that a bot raised.

PR #4647, opened 2026-07-21
fix(api): clarify run list commands and docs

jacobheric:
## Description

Generate API docs had duplicated nav entries for our two ways of calling the runs api (`/runs` and `/apps/{id}/function/{id}/runs`) this dedupes those and a similar case for experiments. Also noticed the same two runs api paths bled out into two runs cli commands. I deduped those as well. 

## Release note

None.

## Migration note

None.

### DIFF
--- a/docs/api-docs/scripts/generate-docs.ts
+++ b/docs/api-docs/scripts/generate-docs.ts
@@ -8,7 +8,9 @@ const appRoot = resolve(import.meta.dirname, '..');
 const repoRoot = resolve(appRoot, '../..');
 
 type Operation = {
+  description?: string;
   requestBody?: { content?: Record<string, unknown> };
+  summary?: string;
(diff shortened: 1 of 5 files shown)


docs/api-docs/scripts/generate-docs.ts
```diff
@@ -22,6 +24,25 @@ type OpenAPIDocument = {
 
 const HTTP_METHODS = ['get', 'post', 'put', 'patch', 'delete', 'head', 'options', 'trace'];
 
+const OPERATION_DOC_OVERRIDES: Record<string, Pick<Operation, 'description' | 'summary'>> = {
```
jacobheric, line 27: This is a bit hacky...because the experiment actually has a single rpc with a second binding.

albertchae (approved):

PR #4653, opened 2026-07-22
fix: use snapshots to handle map race access

darwin67:
## Description

`ScheduleRequest` have read/write concurrent access to the embedded maps so creating snapshots to avoid that for now.

We can revisit this later to make it more elegant.

## Release note

None.

## Migration note

None.

### DIFF
--- a/pkg/execution/executor/executor.go
+++ b/pkg/execution/executor/executor.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"net/url"
 	"runtime/debug"
 	"slices"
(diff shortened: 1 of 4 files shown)


mar-inngest (approved):

BrunoScheufler (approved):

dlambrig:
pkg/execution/executor/executor.go:1021 snapshots req too early. A few lines later, schedule() reuses req.Context as config.Context and then populates it with derived fields like url, cron schedule, function slug, debounce flag, event ID mapping, and debug IDs (executor.go:1124-1144). Before this PR, async event lifecycles saw those values because they were handed the live req; after this PR they get reqSnapshot, which was cloned before those writes happened. That means lifecycle callbacks can now see stale or missing context. I’d move the reqSnapshot := cloneScheduleRequest(req) until after schedule() finishes enriching req.Context, or build the snapshot from the fully-populated context source.

PR #4657, opened 2026-07-23
feat(api): add v2 apps list endpoint

jacobheric:
## Summary

Adds apps list to the v2 rest api.

monorepo pr is [here](https://github.com/inngest/monorepo/pull/7952).

## Releaes notes

Adds apps list to the v2 rest api and cli.

`GET /api/v2/apps`

`inngest api GET /api/v2/apps`


pkg/cqrs/manager/cqrs.go
```diff
@@ -1195,23 +1195,36 @@ func (w wrapper) InsertQueueSnapshotChunk(ctx context.Context, params cqrs.Inser
 
 // GetApps returns apps that have not been deleted.
 func (w wrapper) GetApps(ctx context.Context, envID uuid.UUID, filter *cqrs.FilterAppParam) ([]*cqrs.App, error) {
```
jacobheric, line 1197: updated the existing app fetch to support proper cursor based pagination and optional support for archived. without the params it behaves as before (all unarchived apps).

pkg/db/postgres/querier.go
```diff
@@ -74,14 +74,34 @@ func (pq *pgQuerier) GetAppByURL(ctx context.Context, url string) (*db.App, erro
...
+		Archived:  arg.Archived,
+		Method:    arg.Method,
+	})
 	if err != nil {
 		return nil, err
 	}
 	return convertSlice(rows, appFromPG), nil
 }
 
+func (pq *pgQuerier) GetAppFunctionCounts(ctx context.Context, appIDs []uuid.UUID) ([]db.AppFunctionCount, error) {
```
jacobheric, line 92: rather than doing this n+1 like gql, added batched fetch for that.

djfarrelly: Great addition

djfarrelly (approved):

pkg/db/postgres/sqlc/queries.sql
```diff
@@ -51,7 +51,22 @@ RETURNING *;
 SELECT * FROM apps WHERE id = $1;
 
 -- name: GetApps :many
-SELECT * FROM apps WHERE archived_at IS NULL;
+SELECT * FROM apps
+WHERE (
+    (sqlc.arg('archived')::boolean AND archived_at IS NOT NULL)
```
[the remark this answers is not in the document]
jacobheric: I don' think this is a real issue with the oss data model. apps.archived_at is an immediate archive flag: DeleteApp sets it to the current database time, upserts clear it, and existing app readers consistently treat any non-null value as archived. oss has no write path for scheduled future app archives. the Before(time.Now()) mapping also predates this PR. A manually inserted future timestamp would produce the described behavior, but that is outside the scope of this work.

PR #4660, opened 2026-07-23
feat(metadata): Extract More AI Metdata

Linell:
## Description

Fills derivable gaps in `inngest.ai` metadata (`total_tokens`, `estimated_cost`)
at the metadata-span choke point (`CreateMetadataSpanFromValues`), covering every emission path: SDK generator opcodes, sync checkpoints, and the public
run-metadata API. Caller-supplied values are never overwritten; delete ops are
untouched. The three AI extractors now share the same `AIMetadata.Enrich` logic.

This fixes an issue where the following function's AI calls do *not* produce equivalent metadata:

```ts
import { inngest } from "../client";
import { Agent, run } from "@openai/agents";

export const helloWorld = inngest.createFunction(
	{ id: "hello-world", triggers: [{ event: "test/hello.world" }] },
	async ({ step }) => {
		const agent = new Agent({
			name: "Bubba",
			instructions: "You are a helpful assistant.",
		});

		const haikuTopic = await step.run("get-topic", async () => {
			const response = await run(
				agent,
				"Choose a single, specific programming related topic, such as Inngest, TypeScript, or Go.",
			);
			return response.finalOutput;
		});

		const result = await step.ai.wrap("get-agent-output", async () => {
			const response = await run(
				agent,
				`Write a haiku about this topic: ${haikuTopic}`,
			);
			return response.finalOutput;
		});

		const rewrittenResult = await step.ai.infer("revise-agent-output", {
			model: step.ai.models.openai({ model: "gpt-4o" }),
			body: {
				messages: [
					{
						role: "assistant",
						content: `Make any highly artistic, poetic, and high brow changes to this haiku while keeping the overall message: ${result}`,
					},
				],
			},
		});

		return {
			topic: haikuTopic,
			original: result,
			rewritten: rewrittenResult.choices[0].message.content,
		};
	},
);
```

Also extracts more from AI gateway calls — response ID, finish reasons, request params (`temperature`, `top_p`, `max_tokens`, `seed`), and a normalized `provider` (`openai`, `anthropic`, `gcp.gemini`, `aws.bedrock`) — and fixes the `Temprature` typo (Go identifier only).

## Release Note

AI step metadata now includes derived `total_tokens` and `estimated_cost` when
omitted by the SDK, plus richer AI gateway metadata: response ID, finish
reasons, request parameters, and a normalized provider.

## Migration note

AI gateway `provider` values changed from the raw request format (e.g.
`openai-chat`) to normalized identifiers (e.g. `openai`), and
`total_tokens`/`estimated_cost are now omitted instead of `0` when no tokens
are reported.

### DIFF
--- a/pkg/tracing/metadata/extractors/ai.go
+++ b/pkg/tracing/metadata/extractors/ai.go
@@ -347,6 +347,44 @@ func backfillEstimatedCost(md *AIMetadata) {
 	md.EstimatedCost = estimatedCostForTokens(md.ResponseModel, md.RequestModel, md.InputTokens, md.OutputTokens)
 }
 
+// backfillTotalTokens sets md.TotalTokens by summing input and output tokens
+// only when the emitter didn't supply a total and there is usage to sum.
+func backfillTotalTokens(md *AIMetadata) {
+	if md.TotalTokens != nil || (md.InputTokens == 0 && md.OutputTokens == 0) {
+		return
+	}
+	totalTokens := md.InputTokens + md.OutputTokens
+	md.TotalTokens = &totalTokens
+}
+
+// BackfillTotalTokensInValues fills a "total_tokens" entry into raw
+// "inngest.ai" metadata values when one isn't already present, mirroring
+// BackfillEstimatedCostInValues for token totals.
+func BackfillTotalTokensInValues(values metadata.Values) {
+	if values == nil {
+		return
+	}
+
+	if raw, ok := values["total_tokens"]; ok {
+		var existing *int64
+		if err := json.Unmarshal(raw, &existing); err == nil && existing != nil {
+			return
+		}
+	}
+
+	var inputTokens, outputTokens int64
+	_ = json.Unmarshal(values["input_tokens"], &inputTokens)
+	_ = json.Unmarshal(values["output_tokens"], &outputTokens)
+	if inputTokens == 0 && outputTokens == 0 {
+		return
+	}
+
+	totalTokens := inputTokens + outputTokens
+	if b, err := json.Marshal(totalTokens); err == nil {
+		values["total_tokens"] = b
+	}
+}
+
 // BackfillEstimatedCostInValues fills an "estimated_cost" entry into raw
 // "inngest.ai" metadata values when one isn't already present. Unlike
 // backfillEstimatedCost, this operates on untyped metadata.Values — the shape
--- a/pkg/tracing/metadata/extractors/ai.go
+++ b/pkg/tracing/metadata/extractors/ai.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"math"
-	"net"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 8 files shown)


scottnuma (commented):
I think we may have double implemented some of these features in 
https://github.com/inngest/inngest-js/pull/1647 and marginally https://github.com/inngest/inngest/pull/4661 😬 

can you rebase and re-request me for review?

pkg/tracing/metadata/extractors/ai.go
```diff
@@ -347,6 +347,44 @@ func backfillEstimatedCost(md *AIMetadata) {
 	md.EstimatedCost = estimatedCostForTokens(md.ResponseModel, md.RequestModel, md.InputTokens, md.OutputTokens)
 }
 
+// backfillTotalTokens sets md.TotalTokens by summing input and output tokens
+// only when the emitter didn't supply a total and there is usage to sum.
+func backfillTotalTokens(md *AIMetadata) {
+	if md.TotalTokens != nil || (md.InputTokens == 0 && md.OutputTokens == 0) {
+		return
+	}
+	totalTokens := md.InputTokens + md.OutputTokens
```
scottnuma, line 370: Noting that this doesn't quite hold true for Anthropic or Gemini. However, we are becoming more consistent with ourselves across the various AI metadata entrypoints, and that feels more important first.

pkg/tracing/metadata/extractors/ai.go
```diff
@@ -138,6 +153,31 @@ func ExtractAIGatewayMetadata(req aigateway.Request, respStatus int, resp []byte
 	}, nil
 }
 
+// aiProviderFromRequest resolves a provider identifier from the request
+// format. The OpenAI chat format is shared by many compatible providers
+// (Groq, DeepSeek, local endpoints), so only api.openai.com itself maps to
+// "openai"; other hosts keep the format verbatim, with the endpoint visible
+// in the adjacent HTTP metadata.
+func aiProviderFromRequest(host, format string) string {
```
scottnuma: I'm less familiar with step.ai.infer/wrap and the pieces of data it delivers. How does this transformation affect consistency with the AI metadata via extended traces or the JS SDK AI metadata processor?

scottnuma (commented):
Passing along more information looks good - one questions on our provider processing.

pkg/tracing/metadata/extractors/ai.go
Linell: So it turns out `provider` means different things depending on the path. The JS SDK processor and extended traces both pull it from the `gen_ai.provider.name` OTel attribute, so they get actual provider names like `"openai"`. But for `step.ai.infer` we make the HTTP call ourselves — no OTel involved — and all we've got is the SDK's request format (`"openai-chat"`) and the URL. So infer was the one path reporting a format string instead of a provider name.

This function was my attempt to map format + host onto real provider names, but it can't fully work: Groq, DeepSeek, and local endpoints all speak the OpenAI-compatible format, so they'd still end up as `"openai-chat"`. Half a fix wasn't worth it, so I've dropped it from this PR and left `provider` as the raw format — happy to tackle vocabulary alignment as a follow-up.

PR #4665, opened 2026-07-24
refactor(queue): add scanner runtime

BrunoScheufler:
## Description

Adds a `QueueScannerRuntime` boundary so scanner implementations receive the shared item leaser, dispatch function, and worker semaphore together. This lets backend-specific scanners discover work with their own storage model while still handing candidate items to the common queue lease and worker-dispatch path, and lets scanners avoid scanning when worker capacity is exhausted.

Decouples queue item leasing and processing from queue partitions:

- `LeaseItemRequest` now carries only the item plus scanner-resolved metadata: priority, continuation count, earliest-peek fallback, and static time.
- `LeaseItem` derives account, environment, function, and run identifiers from the item itself instead of requiring a `QueuePartition`.
- `ProcessItem` no longer depends on queue partition state; it uses item identifiers and scanner-provided priority/continuation metadata.
- Partition priority is resolved during scanning and passed through to workers.
- `LeaseItemResult` carries `RetryAfter` for constrained statuses so scanner implementations can defer their own scheduling pointers without parsing errors or rechecking constraints.

Adds a dispatch completion contract:

- `DispatchFunc` now returns a `DispatchedItem` future when an item is accepted for processing.
- The default dispatcher creates the completion handle, sends the item to the worker channel with context cancellation, and returns the handle to the scanner.
- Workers complete the handle after `ProcessItem` returns, including whether the run scheduled immediate follow-up work.
- Completion channels are buffered so workers do not block if no scanner is waiting.
- `NewCompletedDispatchedItem` provides a standard helper for synchronous scanner/test implementations.

Moves continuation ownership out of `ProcessItem` and into the partition scanner path. When continuations are enabled, the partition processor watches dispatched item completion and adds a continuation only after successful immediate follow-up work. The watcher exits on completion, scanner context cancellation, or a 10-minute timeout to avoid stale goroutines.

Organizes scanner and dispatch types into focused files:

- `pkg/execution/queue/scanner.go` owns scanner runtime/interface code.
- `pkg/execution/queue/dispatch.go` owns dispatch result/future types and helpers.

## Release note

None

## Migration note

Out-of-tree queue scanner implementations must update `Run(ctx, dispatch)` to `Run(ctx, runtime)`, use `runtime.Dispatch` to submit leased work, and may inspect `runtime.WorkerSemaphore` before scanning. Implementations that call `DispatchFunc` must handle the returned `DispatchedItem`; callers may ignore it, or wait on `Done()` with context cancellation or a timeout.

Callers of `LeaseItem` must stop passing queue partitions and instead provide item metadata on `LeaseItemRequest`. Callers of `ProcessItem` must handle the new `(ProcessItemResult, error)` return shape.

## Validation

- `go test ./pkg/execution/queue`
- `go test ./tests/execution/queue`
- `make lint`

### DIFF
--- a/pkg/execution/queue/consts.go
+++ b/pkg/execution/queue/consts.go
@@ -55,14 +55,15 @@ const (
 	DefaultQueuePeekMax  int64 = 750
 	AbsoluteQueuePeekMax int64 = 5000
 
-	QueuePeekCurrMultiplier      int64 = 4 // threshold 25%
-	QueuePeekEWMALen             int   = 10
-	QueueLeaseDuration                 = 30 * time.Second
-	QueueItemEarliestPeekTimeTTL       = 7 * 24 * time.Hour
-	RoleLeaseDuration                  = 10 * time.Second
-	RoleLeaseMax                       = 20 * time.Second
-	ShardLeaseDuration                 = 10 * time.Second
-	ShardLeaseMax                      = 20 * time.Second
+	QueuePeekCurrMultiplier        int64 = 4 // threshold 25%
+	QueuePeekEWMALen               int   = 10
+	QueueLeaseDuration                   = 30 * time.Second
+	QueueContinuationResultTimeout       = 10 * time.Minute
+	QueueItemEarliestPeekTimeTTL         = 7 * 24 * time.Hour
+	RoleLeaseDuration                    = 10 * time.Second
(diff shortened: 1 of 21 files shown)


dlambrig (approved):
LGTM

lkasinathan (approved):

PR #4671, opened 2026-07-27
feat(cli): use the stored device-login key for inngest api requests

AndyInternet:
Stacked on #4658 so that one needs to go in first. Backend half is in inngest/monorepo#8015.

## Description

`inngest api` authenticates with the credential stored by `inngest auth login` when no key flag is set, so `inngest api --prod` just works once you're logged in.

`inngest api` is the only command that makes authenticated cloud calls so it's the only thing that needed to change. Order is `--api-key`/`INNGEST_API_KEY`, then `--signing-key`/`INNGEST_SIGNING_KEY`, then the stored credential. Explicit flags win, so nothing about existing scripted usage changes.

The part worth a careful look is that the stored key only goes to the host that minted it. It gets attached when the resolved target host matches whatever `inngest auth login` authenticated against (`$INNGEST_API`, defaulting to `api.inngest.com`), and not otherwise. Without that check a real cloud key would get attached to whatever `--api-host` happens to name, and the default target is the local dev server which needs no auth at all. So `--prod` gets the key, the dev server never does, and an arbitrary `--api-host` doesn't either, you'd pass `--api-key` for that. Pointing `$INNGEST_API` at a local stack does get the key, which is deliberate, that's where you logged in and where the key came from, and it's what makes the flow testable locally.

The CLI sends the credential and nothing else, no `X-Inngest-Env`. The key's env is resolved server side in inngest/monorepo#8015, so **that needs to be deployed first**: until it is, a login bound to a non-production env fails on the four v2 endpoints that assume production instead of reading the key. Everything else works either way.

Two smaller things:

- `$INNGEST_AUTH_TOKEN` overrides the stored key, since the fallback reads through `clistate.GetState` rather than `StoredCredentials`. `logout` deliberately uses `StoredCredentials` so it never revokes a key the CLI didn't mint. That asymmetry is on purpose.
- Pulled `client.APIBaseURL()` out of `client.New()` so "which host did we log in to" has one definition instead of a duplicated `$INNGEST_API` lookup.

## Testing

Unit tests all run against a stub API, no network:

- `TestCommandUsesStoredCredential`, plus `TestCommandWithholdsStoredCredentialFromOtherHosts` for when you're logged into cloud but pointing somewhere else, which also covers the dev server default.
- `TestStoredCredentialSentToCloudTarget` closes the `--prod` path alongside the existing `TestResolveBaseURLProdUsesCloud`.
- `TestCommandPrefersFlagsOverStoredCredential` and `TestCommandStoredCredentialHonorsAuthTokenOverride` for precedence.
- Tests point `$HOME` at a temp dir and clear the key env vars so they never read your real login. Checked that they fail with the fallback removed.

Ran the whole thing end to end against a local stack driving the real approval page in a browser. Login prints the code, `/device` is reachable through sign in with the redirect preserved, and the env picker is honored (picked branch, and the minted key came back bound to the branch workspace with `env: "branch"` in local state). The key is a normal `CLI login` key, env-scoped with `created_by` set and visible in the API keys table, and the device token gets marked confirmed and consumed so the code is single use. `inngest auth logout` revokes it server side and the saved key then 401s.

Then ran all 29 endpoints through `inngest api` with no flags at all. No auth or env failures. With the login host pointed at the local stack, a request to the default dev server target sent no `Authorization` header, and a stub confirmed the CLI sends the credential and no `X-Inngest-Env`.

## Release note

`inngest api` now authenticates with the credential stored by `inngest auth login` when no `--api-key` or `--signing-key` is provided.

## Migration note

None.


AndyInternet:
Folding this into #4658 as part of restructuring the stack. It was a 3-file follow-up rather than something worth reviewing on its own, and having it as a separate PR meant #4708 sat two levels deep for no benefit. The commit is unchanged, it's just the tip of #4658 now.

New shape: #4658 (CLI OAuth, now including this) → #4708 (v2 grant declarations, server-side only) → #4717 (the dashboard permission screens, split out of #4708).

PR #4674, opened 2026-07-28
feat(mcp): generate mcp tools from rest api v2

jacobheric:
## Summary

makes mcp tools list use the generated protobuf based list as api and cli so they are aligned. adds an mcp blacklist for dev server so we don't expose tools that are not supported there (webhooks, scoring, etc). 

I've kept the old static list of tools available in dev server as well for backwards compatibility, but marked them as deprecated. where two of those (list_function and invoke_function) collide with the new generated tools I've moved the old static ones to different names (e.g. invoke_function_sync, etc) and put notes on the new ones about that. I did not use the tool change notification feature from the mcp spec as that seems to be more for live updates with connected clients and this is more of a release update. I did add note to the setup docs though.

I've also mcp setup page a shared/dynamic component that we can use in cloud dashboard as well. 

companion monorepo pr is [here](https://github.com/inngest/monorepo/pull/8018).

Screenshot 2026-07-29 at 3 36 56 PM

## Release note

Updates dev server mcp tools list to expose more tools that align more closely with the v2 rest api and cli


cmd/apiv2cli/cmd.go
```diff
@@ -19,14 +19,11 @@ import (
 	"time"
 	"unicode"
 
-	openapiv2 "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options"
 	localconfig "github.com/inngest/inngest/cmd/internal/config"
 	"github.com/inngest/inngest/pkg/api"
 	"github.com/inngest/inngest/pkg/api/tel"
-	apiv2 "github.com/inngest/inngest/proto/gen/api/v2"
+	"github.com/inngest/inngest/pkg/api/v2/apiv2endpoint"
```
jacobheric, line 25: moved most of the endpoint discovery stuff to `pkg/api/v2/apiv2endpoint` so we can more easily share it with cloud mcp.

pkg/api/v2/apiv2mcp/tools.go
```diff
@@ -0,0 +1,342 @@
+package apiv2mcp
```
jacobheric, line 1: generates our mcp tools from our rest api v2 endpoints.

djfarrelly (approved):

PR #4676, opened 2026-07-29
fix(epxressions):  add support for falsy terns

tonyhb:
this commit adds full support for dynamic typing of false/empty/null and missing event data in ternary expressions without needing the `has` macro


mar-inngest (approved):

pkg/expressions/unknown.go
```diff
@@ -40,6 +44,44 @@ func unknownDecorator() interpreter.InterpretableDecorator {
 	}
 }
 
+func wrapConditionalCondition(i interpreter.Interpretable) bool {
+	attr, ok := i.(interpreter.InterpretableAttribute)
+	if !ok {
```
mar-inngest, line 49: Is there a probability/benchmark of how often it will return from here?

PR #4681, opened 2026-07-29
internal(queue): warn and count dropped retries on missing queue item

scottnuma:
## Description

- I'm trying to investigate [this plain ticket](https://app.plain.com/workspace/w_01H5R1F69XQ35G4P7THFQGN5KB/thread/th_01KY0QP89210PRA65NPFGGDEP7/), where we hit requeueing problems
- It would be much easier some more logs, let's add them in

## Release note

None.

## Migration note

None.


darwin67 (approved):

scottnuma:
Removed the metric, which I'm less familiar with, but kept the log, which is the real value driver here.

Will merge upon passing tests

PR #4683, opened 2026-07-30
feat(ui): nudge metadata usage in run details

Linell:
## Description

Show the Metadata tab on completed runs, even when no metadata exists, rendering a nudge that links to AI metadata and `step.metadata()` docs. The nudge also appears beneath existing metadata when the run has no `inngest.ai` metadata anywhere in its trace.

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/packages/components/src/RunDetailsV4/AIMetadata.tsx
+++ b/ui/packages/components/src/RunDetailsV4/AIMetadata.tsx
@@ -1,0 +1,62 @@
+import { RiSparkling2Line } from '@remixicon/react';
+
+import { Button } from '../Button/Button';
+import { InlineCode } from '../Code';
+import { KindInngestAI } from '../generated';
+import { traceWalk } from './runDetailsUtils';
+import type { Trace } from './types';
+
+const AI_METADATA_DOCS_URL =
+  'https://www.inngest.com/docs/examples/open-telemetry?ref=app-run-metadata';
+
+const CUSTOM_METADATA_DOCS_URL =
+  'https://www.inngest.com/docs/reference/typescript/v4/functions/metadata?ref=app-run-metadata';
+
+const links = [
+  { label: 'Set up AI metadata', href: AI_METADATA_DOCS_URL },
+  { label: 'Add custom metadata', href: CUSTOM_METADATA_DOCS_URL },
+];
+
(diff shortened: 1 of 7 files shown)


ui/packages/components/src/RunDetailsV4/AIMetadata.tsx
```diff
@@ -0,0 +1,62 @@
...
+import type { Trace } from './types';
+
+const AI_METADATA_DOCS_URL =
+  'https://www.inngest.com/docs/examples/open-telemetry?ref=app-run-metadata';
+
+const CUSTOM_METADATA_DOCS_URL =
+  'https://www.inngest.com/docs/reference/typescript/v4/functions/metadata?ref=app-run-metadata';
+
+const links = [
+  { label: 'Set up AI metadata', href: AI_METADATA_DOCS_URL },
```
scottnuma, line 18: We may want to tweak this wording to align with whatever Aaron set from [this convo](https://inngest.slack.com/archives/C0AFJB68X99/p1785355800144609?thread_ts=1785354396.237569&cid=C0AFJB68X99) - perhaps for a different pr

scottnuma (approved):
nice!

scottnuma (approved):
Tweaked wording looks good. We can always touch up more later too!

For spans with custom metadata, but no AI metadata, our nudge does include instructions for custom metadata, which the user is already doing. It could be nice to:

- Conditionally show the custom metadata when there's no custom metadata
- Drop the custom metadata prompt altogether, since we're focusing on AI Metadata right now
- Leave as is, since it's pretty minor

I think all three are reasonable choices, just wanted to surface.


Image

Linell:
> Tweaked wording looks good. We can always touch up more later too!
> 
> For spans with custom metadata, but no AI metadata, our nudge does include instructions for custom metadata, which the user is already doing. It could be nice to:
> 
> * Conditionally show the custom metadata when there's no custom metadata
> * Drop the custom metadata prompt altogether, since we're focusing on AI Metadata right now
> * Leave as is, since it's pretty minor
> 
> I think all three are reasonable choices, just wanted to surface.
> 
> Image

Super super fair!! I'm going to roll it as is just because it is minor, and I think mentioning custom metadata makes it less of an annoying AI feature nag for people who *aren't* doing AI stuff. Super duper open to any changes though, for posterity 👍

PR #4700, opened 2026-08-03
feat: add SDK feature observations

amh4r:
## Summary

Add OSS backend and UI support for SDK feature observations so apps can surface SDK-reported feature readiness and configuration.

## Changes

- Added SDK feature observation proto/types and app sync handling.
- Persist feature observations on app sync.
- Expose feature observations through GraphQL.
- Add AI Metadata Extraction (a.k.a. "AI OTel") and Extended Traces to the UI.

Does not include Cloud UI changes. That will be in a future PR


proto/connect/v1/connect.proto
```diff
@@ -33,6 +34,7 @@ message AppConfiguration {
 	string app_name = 1;
 	optional string app_version = 2;
 	bytes functions = 4;
+	repeated sdk_feature_observations.v1.FeatureObservation feature_observations = 5;
```
scottnuma: Do we define this proto twice: once in oss and once in inngest-js? Any context on this pattern?

scottnuma (approved):

proto/connect/v1/connect.proto
amh4r: Yes, the proto file exists in both place. We should eventually add a way to share it, but copy-pasting it hasn't been onerous so far. And the nature of proto's backcompat lowers the risk

scottnuma (approved):

PR #4716, opened 2026-08-05
feat(api): add v2 sandbox contracts

jacobheric:
## Summary

adds rest v2 protobuf contracts for the sandbox apis so they can be used by cloud with the grpc gateway and exposed to the cli. excludes these from the dev server mcp catalog.

monorep side is [here](https://github.com/inngest/monorepo/pull/8150). <- see there for using the cli locally.


pkg/api/v2/enum_json.go
```diff
@@ -32,19 +35,34 @@ func NewResponseEnumMarshaler() runtime.Marshaler {
 	return responseEnumMarshaler{JSONPb: &runtime.JSONPb{}}
 }
 
+func (responseEnumMarshaler) StreamContentType(any) string {
```
jacobheric, line 38: cloud has it's own streaming handlers and side steps the generated one so it can handle chunked streaming, etc.

proto/third_party/google/api/httpbody.proto
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 Google LLC
```
jacobheric, line 1: using `google.api.HttpBody` gives us the file’s content type and raw bytes...instead of doing an Inngest specific wrapper.

cmd/apiv2cli/cmd.go
```diff
@@ -627,6 +666,35 @@ func queryParams(cmd *cli.Command, ep endpoint) (url.Values, error) {
...
+	}
+	path := cmd.String("body-file")
+	if path == "-" {
+		reader := cmd.Root().Reader
+		if reader == nil {
+			reader = os.Stdin
+		}
+		return reader, nil
+	}
+	return os.Open(path)
```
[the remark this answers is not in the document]
jacobheric: I think this might be a false positive....*os.File implements io.ReadCloser, http.NewRequestWithContext assigns it to req.Body, and http.Client.Do closes request bodies even on errors.

jakobevangelista (approved):

jakobevangelista (approved):

PR #4736, opened 2026-08-11
feat(scores): show recorded scores on timeline bars

Linell:
## Description

Spans that recorded scores now show a badge beside their bar in the run details timeline. Hovering it lists each score name and value.

The badge reuses the experiment badge treatment.

### Screenshots

Screenshot 2026-08-11 at 1 14 14 PM
Screenshot 2026-08-11 at 1 14 18 PM

## Release note

Scores recorded on a step now appear as a badge on that step's bar in the run
timeline, with a hover card listing each score and its value.

## Migration note

None.

## Related

[EXE-1832: Scoring with the trace view](https://linear.app/inngest/issue/EXE-1832/scoring-with-the-trace-view)

### DIFF
--- a/ui/packages/components/src/RunDetailsV4/TimelineBar.tsx
+++ b/ui/packages/components/src/RunDetailsV4/TimelineBar.tsx
@@ -613,7 +613,7 @@ function ScoreBadge({ scores }: { scores: ScoreBadgeData[] }) {
           <RiPercentLine className="h-3 w-3" />
         </span>
       </HoverCardTrigger>
-      <HoverCardContent side="top" align="start" className="border-muted max-w-none border">
+      <HoverCardContent side="top" align="start" className="border-muted max-w-xs border">
         <ScoreHoverCardContent scores={scores} />
       </HoverCardContent>
     </HoverCardRoot>
@@ -627,7 +627,6 @@ function ScoreBadge({ scores }: { scores: ScoreBadgeData[] }) {
 function ScoreHoverCardContent({ scores }: { scores: ScoreBadgeData[] }) {
   return (
     <div className="whitespace-nowrap px-1 py-0.5 text-xs">
-      <p className="text-light mb-0.5">Scores</p>
       <div className="border-subtle flex justify-between gap-6 border-b pb-1">
         <span className="text-light">Score</span>
         <span className="text-light">Value</span>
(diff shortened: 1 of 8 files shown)


sanjanaladdha:
image
This looks great. can we just remove the "scores" heading from the top in the tooltip and just keep the score and the value title? that will clean up the UI a bit.

Maybe for the long strings, we want to truncate after certain number of characters.

Linell:
Screenshot 2026-08-25 at 9 22 47 AM

Good calls, I agree that's better 👍

ui/packages/components/src/RunDetailsV4/TimelineBar.tsx
```diff
@@ -910,9 +962,11 @@ export function TimelineBar({
           </span>
         </div>
 
-        {/* Experiment badge - centered between left panel and bars */}
-        <span className="inline-flex w-7 shrink-0 items-center justify-center">
+        {/* Experiment and score badges - centered between left panel and bars.
+            Width is fixed so bars stay aligned across rows regardless of badges. */}
+        <span className="inline-flex w-12 shrink-0 items-center justify-center gap-1">
```
scottnuma, line 967: There's a pre-existing problem where our green bars for spans aren't aligned with the ticks of the timeline above it.

I think this PR makes the problem worse as we move from  `w-7` to `w-12`.

In this pic, the span should be 0 to 5.

timeline

A potential fix might b extracting a shared constant (e.g. `BADGE_COLUMN_WIDTH_PX` in `TIMELINE_CONSTANTS`) and adding a matching spacer to `TimelineHeader`, so the two layouts can't drift again.

This pre-exists, so feel free to not address in this PR.

ui/packages/components/src/RunDetailsV4/TimelineBar.tsx
```diff
@@ -597,6 +600,54 @@ function ExperimentHoverCardContent({
...
+function ScoreHoverCardContent({ scores }: { scores: ScoreBadgeData[] }) {
+  return (
+    <div className="whitespace-nowrap px-1 py-0.5 text-xs">
+      <div className="border-subtle flex justify-between gap-6 border-b pb-1">
+        <span className="text-light">Score</span>
+        <span className="text-light">Value</span>
+      </div>
+      {scores.map((score) => (
+        <div
+          key={score.name}
```
scottnuma, line 636: We might have metadata that looks like:

```
metadata: [{ relevance: 0.5, updatedAt: 00:00:05 },
           { relevance: 0.9, updatedAt: 00:00:09 }]
→ bar.scores = [{ name: "relevance", value: 0.5 },
                { name: "relevance", value: 0.9 }]
```

I think rather than showing multiple of these for `relevance`, we'd rather just show only the latest one? That also avoids duplicate key problems.

This is a Claude-generated case, so feel free to push back if we probably won't land here anyway.

ui/packages/components/src/RunDetailsV4/utils/traceConversion.ts
```diff
@@ -211,6 +213,27 @@ function getHTTPTimingFromMetadata(metadata?: SpanMetadata[]): HTTPTimingBreakdo
...
+          return typeof value === 'boolean' || (typeof value === 'number' && Number.isFinite(value))
+            ? [{ name, value }]
+            : [];
+        })
+      )
+      .sort((a, b) => a.name.localeCompare(b.name)) ?? [];
+
+  return scores.length > 0 ? scores : undefined;
+}
+
```
scottnuma, line 229: Looks like we might be able to reuse `scoreRows` from `RunDetails/ScoresAttrs.tsx` here and save some lines and tests

scottnuma (approved):

scottnuma (approved):

PR #4741, opened 2026-08-13
feat(dashboard): add sandbox accounting views

jakobevangelista:
- replace the sandbox waitlist state with a time-filtered sandbox accounting dashboard
- show persisted sandbox totals, active count, p95 start time, failure rate, launch-unknown count, and a paginated sandbox table
- add a sandbox detail route for persisted configuration, resources, lifecycle timestamps, and networking details
- use existing dashboard table, status, time filter, error, header, and routing conventions

Depends on inngest/monorepo#8218.


jakobevangelista:
superseded by #4784

PR #4755, opened 2026-08-19
fix(dashboard): allow vercel and do marketplace users to manage api keys

jacobheric:
## Summary

when we restricted access for api creation to admins only, we used clerks useOrg hook for this, which broke api creation for vercel and do marketplace users because they use plain jwt. since vercel and do marketplace users are by definition admins, we can simply enable for them.

## Release note

fixes api key generation for vercel marketplace users.

## Migration note

None.


jacobheric:
I think mendral might be done...a day early.

mitchellalderson (approved):

mar-inngest (approved):

albertchae (approved):

PR #4758, opened 2026-08-20
feat(ai): add run-level AI usage summary

Linell:
## Description

Adds a synthetic run-scoped `inngest.ai.summary` metadata kind aggregating `inngest.ai` usage across a run (tokens, cost, models, providers), computed at span-tree read time in the CQRS manager and never persisted. Step-scoped usage takes precedence over `extended_trace` duplicates.

Backend half of [EXE-2087](https://linear.app/inngest/issue/EXE-2087/ai-metdatatraces-aggregate-gen-aiusage-across-runs-and-child). Stacked under #4759 (UI); cloud counterpart: inngest/monorepo#8272.

## Release note

None (user-facing note on the UI PR).

## Migration note

None.

### DIFF
--- a/pkg/cqrs/manager/cqrs.go
+++ b/pkg/cqrs/manager/cqrs.go
@@ -34,6 +34,7 @@ import (
 	"github.com/inngest/inngest/pkg/run"
 	"github.com/inngest/inngest/pkg/tracing/meta"
 	"github.com/inngest/inngest/pkg/tracing/metadata"
+	"github.com/inngest/inngest/pkg/tracing/metadata/extractors"
 	"github.com/inngest/inngest/pkg/util"
 	connpb "github.com/inngest/inngest/proto/gen/connect/v1"
 	"github.com/oklog/ulid/v2"
@@ -280,6 +281,16 @@ type spanRollupInfo struct {
 	// This is needed because parent_span_id in the DB is an OTEL span ID,
 	// but the span tree is keyed by dynamic_span_id.
 	otelToDynamic map[string]string
+	// aiCandidates collects every raw inngest.ai fragment emission, before
+	// rollup: the AI merge opcode is last-write-wins per key, so summing from
+	// rolled-up entries would silently drop all but the last emission that
+	// shares a dynamic span ID.
+	aiCandidates []aiCandidate
+}
+
+type aiCandidate struct {
+	scope metadata.Scope
+	raw   []byte
 }
 
 // fragmentAttributesJSON returns the attributes field of a span fragment as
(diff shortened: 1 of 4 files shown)


pkg/tracing/metadata/extractors/ai_summary.go
```diff
@@ -0,0 +1,258 @@
...
+	return &AISummaryBuilder{
+		models:    map[string]struct{}{},
+		providers: map[string]struct{}{},
+	}
+}
+
+// aiUsageValues is the minimal projection of an inngest.ai entry needed to
+// aggregate usage. It deliberately does not reuse AIMetadata: that struct
+// types optional counts like cache_read_tokens as *int64, but producers can
+// emit them as floats, so unmarshalling the full struct fails and would drop
```
rhino1998: Why not just change AIMetadata to use float if those are observed in practice?

pkg/tracing/metadata/kind.go
```diff
@@ -48,6 +48,10 @@ func (k Kind) Validate() error {
 // to prevent spoofing of internal metadata. The score kind is the bare
 // constant inngest.score; the user-supplied score name is a key in the values
 // map, not a kind suffix.
+//
+// Synthetic read-time kinds (inngest.usage, inngest.ai.summary) are computed
+// when a span tree is read and must never be allowlisted here: a client-
+// written entry of either kind would be spoofable and could double-count.
```
rhino1998: Do we care about spoofing? The AI metadata that the summary is calculated from is spoofable anyways and all spoofing accomplishes is making that user's run's AI summary invalid

pkg/cqrs/manager/cqrs.go
```diff
@@ -280,6 +281,16 @@ type spanRollupInfo struct {
 	// This is needed because parent_span_id in the DB is an OTEL span ID,
 	// but the span tree is keyed by dynamic_span_id.
 	otelToDynamic map[string]string
+	// aiCandidates collects every raw inngest.ai fragment emission, before
+	// rollup: the AI merge opcode is last-write-wins per key, so summing from
+	// rolled-up entries would silently drop all but the last emission that
+	// shares a dynamic span ID.
```
rhino1998: Why have a separate aiCandidates slice here? The AI metadata after rollup/merging *is* the real state of the AI metadata. Multiple separate emissions at the step level (that are intended to represent separate calls) are a bug. 

If we later emit a new metadata span with just updated pricing (for instance). That doesn't represent a new call that should be summed separately.

Linell: Yeah, you're right, merge is last-write-wins per key so a later emission is an update, not a new call. Switched to summing from rolled-up entries (c2fd1d2) and mirrored it in monorepo#8272.

pkg/tracing/metadata/kind.go
Linell: Fair point. Dropped the strip pass and reworded the comment: keeping these off the allowlist just means no stored entry can exist, so the computed one is the only one.

pkg/tracing/metadata/extractors/ai_summary.go
Linell: AIMetadata is producer-side and never decoded from JSON, the floats come from external emitters, so the read path needs the leniency regardless. Kept the minimal projection since it also survives malformed unrelated fields that a full-struct decode wouldn't.

rhino1998 (approved):

PR #4761, opened 2026-08-20
cloud(fix): wiring up the view functions button

sanjanaladdha:
## Description

the view functions button on the app card now points to the filtered view of the function dashboard for the said app
image
Screenshot 2026-08-20 at 19 19 58

## Release note

None.

## Migration note

None.


albertchae (approved):
looks good although I realized we should probably use app names in the URL rather than UUIDs, i'll fix that in a follow up

albertchae:
> looks good although I realized we should probably use app names in the URL rather than UUIDs, i'll fix that in a follow up

never mind this is consistent with how we do app filter on other pages like runs

PR #4762, opened 2026-08-20
perf(expressions): avoid cloning event data per evaluation

KiKoS0:
## Description
Avoid recursively cloning event data for every lifted
expression evaluation. Each evaluation now copies only the
root map for its expression-specific `vars`, while sharing
the immutable normalized event data.

Benchmark results for a prod representative 3.3 KB event:

- 54.1% fewer bytes: 6,296 → 2,888 B/op
- 34.2% faster: 5,316.6 → 3,499.4 ns/op
- 19.3% fewer allocations: 109 → 88 allocs/op

## Release note

None.

## Migration note

None.


albertchae (approved):
nice

PR #4763, opened 2026-08-21
perf(batching): adding metrics for batching to capture redis memory usage and batch duration

mar-inngest:
## Release note

None.

## Migration note

None.


pkg/execution/batch/lua/append.lua
```diff
@@ -58,22 +58,24 @@ if not newEvent then newEvent = false end
...
+resp.committedBytes = string.len(event)
 
 if len == 1 then
     -- newly started batch
-    resp = { status = "new", batchID = batchID, batchPointerKey = batchPointerKey }
+    resp.status = "new"
 end
 
-local size = redis.call("MEMORY", "USAGE", batchKey)
+local size = redis.call("MEMORY", "USAGE", batchKey) or 0
```
KiKoS0: is this a defensive measure or does it happen that batchKey is an empty key here?

pkg/telemetry/metrics/batch.go
```diff
@@ -0,0 +1,85 @@
...
+		5 * 60_000,
+		10 * 60_000,
+		30 * 60_000,
+		60 * 60_000,
+		6 * 60 * 60_000,
+		12 * 60 * 60_000,
+		24 * 60 * 60_000,
+	}
+)
+
```
KiKoS0, line 24: I believe the default timeout for a batch is 60s, slightly less for the free plan but it maxes out at 10 or 20mins for specific ent plans. Also can't we calculate this with a query from the timeout function configs?

KiKoS0 (approved):

pkg/execution/batch/lua/append.lua
mar-inngest: This was mostly defensive

pkg/telemetry/metrics/batch.go
mar-inngest: updated

PR #4766, opened 2026-08-22
fix(queue): renew capacity leases before dispatch

BrunoScheufler:
## Description

Prevent key-queue items from starting with a capacity lease that can expire before the first renewal tick.

Near-expiry leases are now renewed synchronously before item dispatch, using a reuse threshold based on the configured renewal interval plus a safety margin. If the renewal fails or returns no replacement lease, the item is not dispatched. Processing also watches the active lease expiry and cancels and requeues the item if renewal is not confirmed in time.

Adds a Redis-backed integration regression test for overlapping steps at concurrency one, plus focused coverage for pre-dispatch renewal failure and runtime expiry.

## Release note

Fixed an issue where steps using key queues could overlap despite a concurrency limit of one when a pre-acquired capacity lease expired before its first renewal.

## Migration note

None.

### DIFF
--- a/pkg/execution/queue/constraints.go
+++ b/pkg/execution/queue/constraints.go
@@ -18,6 +18,8 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 )
 
+const capacityLeaseRenewalGracePeriod = 2 * time.Second
+
 type PartitionConstraintConfig struct {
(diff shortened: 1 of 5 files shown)


jakobevangelista (approved):
nice

PR #4780, opened 2026-08-25
feat(auth): add cli oauth login

jacobheric:
## Description

adds top-level login/logout commands that do device authorization. store credentials in the os keyring with an explicit file fallback for envs that don't support that. use oauth sessions (and handle refresh) for underlying v2 api commands. adds the shared dashboard oauth approval screen and permission picker. main driver here is the client oauth lib: `golang.org/x/oauth2`. 

vendoring is huge: `zalando/go-keyring` gives us cross-platform credential storage support, that pulls in `godbus/dbus` for linux and `danieljoos/wincred` for windows. mac uses the built-in security command. and we use `flock` to prevent multiple cli from stomping each others tokens.

companion monorepo prs are [here](https://github.com/inngest/monorepo/pull/8380) and [protobuf permissions #4833](https://github.com/inngest/inngest/pull/4833). 

[EXE-2180](https://linear.app/inngest/issue/EXE-2180/cli-oauth)

For local testing before it's merged you can do like: `INNGEST_API_HOST=http://127.0.0.1:8090 go run ./cmd login`

Screenshot 2026-09-14 at 1 18 52 PM

## Release note

Adds cloud login support to the Inngest CLI. OAuth credentials are stored in the operating system’s credential store and refreshed automatically.

```sh
# log in
inngest login

# check the current session
inngest auth status

# start a new login and replace the current session
inngest login --force

# revoke the session and remove local credentials
inngest logout
```


ui/apps/dashboard/src/components/APIKeys/CreateAPIKeyForm.tsx
```diff
@@ -570,63 +552,14 @@ export function CreateAPIKeyForm({
     );
   }
 
-  const permissionGroups = [...(catalogRes.data?.apiKeyPermissionCatalog ?? [])].sort((a, b) => {
-    return a.resource.localeCompare(b.resource)
-  })
+  const permissionGroups = catalogRes.data?.apiKeyPermissionCatalog ?? [];
   let permissionsContent = (
-    <div className="border-subtle rounded border">
```
jacobheric, line 577: pulled the picker out so I can re-use it for oauth.

cmd/apiv2cli/cmd.go
```diff
@@ -890,11 +892,23 @@ func parseTimestamp(value string) (string, error) {
...
-	return cmd.String("signing-key"), nil
+	// an explicit key overrides the stored login
+	if signingKey := cmd.String("signing-key"); signingKey != "" {
+		return signingKey, nil
+	}
+	manager, err := cliauth.NewManager()
+	if err != nil {
+		return "", err
+	}
+	token, _, err := manager.AccessToken(ctx, resource)
```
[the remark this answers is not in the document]
jacobheric: this behavior is intended. oauth tokens take precedence over signing keys and even if they are expired we do not want to silently fall back to legacy signing keys as that behavior could quietly confer unintended global write privileges to the cli.

cmd/internal/auth/client.go
```diff
@@ -0,0 +1,314 @@
...
+func (m *Manager) AccessToken(ctx context.Context, target string) (string, *Metadata, error) {
+	metadata, credential, err := m.loadToken(target)
+	if err != nil {
+		return "", metadata, err
+	}
+	if credential.Expiry.After(m.now().Add(refreshLeeway)) {
+		return credential.AccessToken, metadata, nil
+	}
+
+	unlock, err := m.store.Lock(ctx)
```
jacobheric, line 111: lock across cli processes so that multiple refreshes don't revoke their own session

pkg/api/v2/apiv2oauth/scopes.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/inngest/inngest/pkg/api/v2/apiv2endpoint"
+)
+
+// Keep credential-management and partner operations outside CLI delegation.
+var cliExcludedResources = map[string]struct{}{
+	"api_keys":     {},
+	"event_keys":   {},
+	"partners":     {},
+	"signing_keys": {},
+}
```
albertchae: do we ever expect this to diverge with `pkg/auth/authz/grants.go`'s `nonGrantableResources` from https://github.com/inngest/monorepo/pull/8655

pkg/api/v2/apiv2oauth/scopes.go
```diff
@@ -0,0 +1,52 @@
...
+	result := make([]string, 0, len(scopes))
+	for scope := range scopes {
+		result = append(result, scope)
+	}
+	slices.Sort(result)
+	return result
+}
+
+func scopeParts(permission string) (string, string, bool) {
+	parts := strings.Split(permission, ":")
```
albertchae: similar comment to https://github.com/inngest/monorepo/pull/8655#discussion_r4021065761 about naming the split parts. this case we could name them in the return type

pkg/api/v2/apiv2oauth/scopes.go
```diff
@@ -0,0 +1,52 @@
...
+	scopes := map[string]struct{}{}
+	for _, endpoint := range endpoints {
+		resource, access, ok := scopeParts(endpoint.AuthzPermission)
+		if !ok {
+			continue
+		}
+		if _, excluded := excludedResources[resource]; excluded {
+			continue
+		}
+		scopes[resource+":"+access+":*"] = struct{}{}
```
albertchae: non blocking but since a lot of this authz code is similar, I wonder if it's worth factoring out a go package that we can use as a library. It could own

- Permission{Resource, Access, Operation}
- Parsing and serialization
- Valid read/write access values
- Concrete-versus-wildcard validation
- Conversion to resource:access:*
- Possibly wildcard matching

jacobheric:
> Overall the OAuth CLI login implementation (device flow, token storage, refresh locking, URL validation) is careful and well-tested. One real regression: the reworked `authToken()` in `cmd/apiv2cli/cmd.go` only falls back to `INNGEST_SIGNING_KEY` when the OAuth manager error is exactly `ErrNotLoggedIn`; any other manager error (e.g. an expired stored session, a corrupt config dir, `NewManager` failure) now surfaces as a hard failure instead of using a valid signing key, whereas previously the signing-key env var always worked directly since it was bound via `cli.EnvVars`.
> 
> **Re-review (2026-08-31):** Only one commit landed since the last review (`8ae37216`, "fix(auth): simplify device code confirmation"). It removes the "The codes match" checkbox that previously gated the device-flow Approve button — the code is still displayed but the user is no longer required to affirmatively confirm it matches their CLI before approving, which weakens the standard OAuth device-flow phishing mitigation (RFC 8628 §5.4) with no compensating control added.
> 
> **Re-review (2026-08-31):** No material changes since the last review — this push is a rebase of the branch onto an updated `poc-authz` base (all commit hashes changed but messages/diffs are identical), which is why the file list shrank (files already merged into the new base are no longer part of this PR's diff). Verified the auth-related code (cmd/auth/cmd.go, cmd/internal/auth/client.go, the device-authorization page) is byte-for-byte the same as previously reviewed. Nothing new to flag.
> 
> **Re-review (2026-08-31):** Only one new commit since the last review (`b69ede5d`, "fix(auth): require explicit environment selection"), which drops the implicit "production" default for `--env`/`X-Inngest-Env` and defaults the OAuth device-approval UI's environment boundary to `single_env` instead of `null`. The change is consistent with existing server-side handling (endpoints that require the env header already return an explicit error when it's missing) and the test expectations were updated accordingly. No new defects found.
> 
> **Re-review (2026-08-31):** The only change since the last review is a commit amendment (`0ded1bf2` replacing `b69ede5d`) that just reworded the `--env` flag's usage/help text ("required for environment-specific requests with all-environment credentials"); the actual logic and tests are unchanged. Nothing new to flag.
> 
> **Re-review (2026-08-31):** The only change since the last review amends `fix(auth): require explicit environment selection` to also remove the now-unreachable `if (!boundary)` check in the device-approval page (dead since `boundary` was changed to a non-nullable default). This is a clean cleanup with no new issues introduced.
> 
> **Re-review (2026-09-15):** Reviewed the new CLI OAuth login/logout/status commands, token storage (keyring/file with locking), scope derivation, and the dashboard device-authorization UI. The Go implementation is careful about redirect handling, resource/issuer validation, atomic credential writes, and concurrent refresh locking, with solid test coverage across client, storage, and cmd packages. No new defects found beyond what's already tracked on this PR (the signing-key fallback behavior was previously raised and dismissed by a human, and the removed device-code confirmation checkbox in the dashboard UI is still open from an earlier pass).
> 
> **Re-review (2026-09-15):** Re-reviewed changes since the last pass. The only material change is the `cmd/auth/cmd.go` refactor (commit [ff3317a](https://github.com/inngest/inngest/commit/ff3317a9497458ee20662c8f6217ef5dfa58fd7e)) that splits `Store().Load()` into `Metadata()`/`Load()` calls for `login`/`logout`/`status` to avoid unnecessary credential reads and to clear local metadata when the credential file is missing; the logic checks out against its new tests and doesn't introduce a regression. The remaining diffs since then are unrelated upstream merges and a stale-test fix, so no new findings. The two previously flagged items (dismissed signing-key fallback behavior, and the still-open removal of the device-code confirmation gate) remain unchanged.

the two remaining issue mentioned here are intentional. we don't want to silently fall back to signing keys if they are attempting to oauth and it fails. and the checkbox on code verification is not a requirement of the spec...and most of the other popular oauth device code flows do not have this so I don't think it's worth the friction.

cmd/apiv2cli/cmd.go
```diff
@@ -481,7 +481,7 @@ func buildRequest(ctx context.Context, cmd *cli.Command, ep endpoint) (*http.Req
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("User-Agent", "inngest-cli/"+version.Version)
 
-	if token, err := authToken(cmd); err != nil {
+	if token, err := authToken(ctx, cmd, baseURL); err != nil {
```
scottnuma, line 484: non-blocking defense in depth: we could ban our http client from following redirects. Such redirects could move us from HTTPS (which we enforced at the start) to HTTP.

This requires the server to act incorrectly, which we're already trusting -> non-blocking

scottnuma (approved):

cmd/internal/auth/client.go
amh4r: I think the lock should be at the top of the method. Concurrent calls are unlikely, but locking before any side effects (i.e. the `loadToken` call) is at least a best practice

amh4r (approved):

cmd/apiv2cli/cmd.go
jacobheric: added a credential safety check to redirects here...will think about the redirect ban.

PR #4781, opened 2026-08-25
feat(auth): add mcp oauth support

jacobheric:
## Description

reuses the same approval screen as the cli. adds the mcp authorization route and callback redirect. 

builds on the cli equivalent [here](https://github.com/inngest/inngest/pull/4780). monorepo companion is [here](https://github.com/inngest/monorepo/pull/8379). [EXE-2179](https://linear.app/inngest/issue/EXE-2179/mcp-oauth)

Screenshot 2026-09-14 at 10 54 49 AM

## Release note

Inngest Cloud MCP clients can now authenticate via OAuth. Users can now select the account, permissions, environment access, etc. in the Inngest Dashboard.


jacobheric:
> **`ui/apps/dashboard/src/routes/_authed/oauth/device/index.tsx` (around the `approve()` function, ~line 219):**
> 
> Dead code introduced by the "require explicit environment selection" commit: `boundary` was changed from `useState<Boundary | null>(null)` to `useState<Boundary>('single_env')`, and every `setBoundary` call site only ever sets `'single_env'` or `'all_envs'`. Since `Boundary` no longer includes `null`, `boundary` can never be falsy, so the `if (!boundary) { setError('Select an environment boundary.'); return; }` branch is now unreachable. If the intent was still to force an explicit boundary choice, don't default the state to `'single_env'`; otherwise this dead branch should be removed.

this comment is obsolete after recent changes.

ui/apps/dashboard/src/routes/_authed/oauth/device/index.tsx
```diff
@@ -311,23 +360,37 @@ function DeviceAuthorizationForm() {
     <Page>
       <div className="flex flex-col gap-1">
         <h1 className="text-basis text-2xl">Connect {details.client_name}</h1>
+        {details.flow === 'authorization_code' && details.client_id && (
```
jacobheric, line 363: mcp clients specify their names....e.g. codex plugin or whatever....so show it.

scottnuma (approved):
Looks good to me after my read through and agent review

PR #4786, opened 2026-08-26
fix(metrics): handle absent values explicitly in backlog and steps-running charts

albertchae:
## Description

Changes backlog and steps-running charts to handle sparse metric responses correctly.

- Gaps shorter than three minutes are treated as missing, not as 0. Lines connect across them without plotting a zero-valued point, and the tooltip reports that no data was recorded.
- Displays gaps lasting at least three minutes as inferred zeroes. This handles some of our edge cases when we stop publishing metrics from the backend, which is most often when a backlog has been canceled so it actually should be 0 but has not explicitly recorded as one. Separately we should look into recording a 0 officially at this point so that we're reporting true data.
- Updates tooltips to distinguish missing observations and inferred zeroes from observed zeroes.

This should work with current backend that does `WITH FILL INTERPOLATE` and after backend is updated to return absent values (since `WITH FILL INTERPOLATE` is equivalent to there being no absent data)

### Real examples
We could have false 0s in the middle of a curve or at beginning or end
false zero at front
false zero at end2
intermittent absent value

### Before (synthetic data)
1 - no backend no frontend

### After (synthetic data)

We've connected over intervals of missing data that are shorter than 3 minutes, but longer ones we still infer as 0
4 - infer zeros for 3 min of no data

We show a tooltip for no data recorded. For charts with more than one line, we show the value for the one that is present along with absent for the other.
5a - tooltip for absent

5b - tooltip for absent

If there's a trailing absent value, we just don't graph beyond the last present value
Screenshot 2026-08-26 at 2 15 50 PM

## Test plan

- Unit tests cover timestamp alignment, short gaps, sustained internal and edge gaps, empty responses, coarse granularities, and Running metrics.

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
+++ b/ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
@@ -31,6 +31,7 @@ type SimpleLineChartProps = {
     values: {
       [key: string]: number | boolean | undefined;
     };
+    inferred?: string[];
   }[];
   legend: {
     dataKey: string;
@@ -101,7 +102,7 @@ export default function SimpleLineChart({
   const flattenData = useMemo(() => {
     return data.map((d) => {
       const values = omit(d.values, referenceAreaKeys);
-      return { ...values, name: d.name };
+      return { ...values, name: d.name, inferred: d.inferred };
     });
   }, [data, referenceAreaKeys]);
 
@@ -224,6 +225,9 @@ export default function SimpleLineChart({
                       ) : (
                         metricPayload.map((p, idx) => {
                           const l = legend.find((l) => l.dataKey == p.name);
+                          const inferred =
+                            Array.isArray(p.payload?.inferred) &&
+                            p.payload.inferred.includes(p.name);
                           return (
                             <div
                               key={idx}
@@ -233,7 +237,12 @@ export default function SimpleLineChart({
                                 className="mr-2 inline-flex h-3 w-3 rounded"
                                 style={{ backgroundColor: l?.color || p.color }}
                               ></span>
-                              {p.value == null ? (
+                              {inferred ? (
+                                <>
+                                  {l?.name || p.name}: No data recorded —
+                                  displayed as 0
+                                </>
+                              ) : p.value == null ? (
                                 <>{l?.name || p.name}: No data recorded</>
                               ) : (
                                 <>
--- a/ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
+++ b/ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
@@ -29,7 +29,7 @@ type SimpleLineChartProps = {
   data?: {
     name: string;
     values: {
-      [key: string]: number | boolean;
+      [key: string]: number | boolean | undefined;
     };
   }[];
   legend: {
@@ -39,6 +39,7 @@ type SimpleLineChartProps = {
     default?: boolean;
     referenceArea?: boolean;
   }[];
+  connectNulls?: boolean;
   isLoading: boolean;
   error?: Error;
 };
@@ -88,6 +89,7 @@ export default function SimpleLineChart({
   desc,
   data = [],
   legend = [],
+  connectNulls = false,
   isLoading,
   error,
 }: SimpleLineChartProps) {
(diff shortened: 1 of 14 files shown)


ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
```diff
@@ -236,6 +267,7 @@ export default function SimpleLineChart({
               />
               {legend.map((l) => (
                 <Line
+                  connectNulls={connectNulls}
```
albertchae, line 270: this setting connects over absent values

ui/apps/dashboard/src/components/Functions/StepBacklogChart.tsx
```diff
@@ -64,16 +65,25 @@ export default function StepBacklogChart({
...
-
-  const maxLength = Math.max(scheduled.length, sleeping.length);
-
-  const metrics = Array.from({ length: maxLength }).map((_, idx) => ({
-    name: scheduled[idx]?.bucket || sleeping[idx]?.bucket || '',
-    values: {
-      scheduled: scheduled[idx]?.value ?? 0,
-      sleeping: sleeping[idx]?.value ?? 0,
-    },
-  }));
```
albertchae, line 76: the previous implementation essentially did a zip on 2 arrays, which was fine because if the backend guaranteed they weren't sparse, they'd be aligned by timestamp. Since we are changing it to return sparse arrays, we should align by the actual timestamp and not just array index.

albertchae: 2 - sparse backend, misaligned, false zeros at end

This is what would happen without the change, we shift the arrays silently to the left and then pad the end with extra false 0s

ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
albertchae: we could show gaps instead but i feel it's worse
3b - aligned, gaps instead of connect

ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
```diff
@@ -0,0 +1,60 @@
...
+  series: MetricSeries[],
+  granularity: string,
+) {
+  const byBucket = new Map<number, MetricRow>();
+
+  for (const { key, data, mapValue } of series) {
+    for (const { bucket, value } of data) {
+      const timestamp = new Date(bucket).getTime();
+      const metric = byBucket.get(timestamp) ?? { name: bucket, values: {} };
+      metric.values[key] = mapValue ? mapValue(value) : value;
```
albertchae, line 34: sometimes we convert the backend value to something more suitable for the frontend. e.g. ConcurrencyLimit becomes a boolean

ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
```diff
@@ -0,0 +1,60 @@
...
+export function mergeMetricRows(
+  series: MetricSeries[],
+  granularity: string,
+) {
+  const byBucket = new Map<number, MetricRow>();
+
+  for (const { key, data, mapValue } of series) {
+    for (const { bucket, value } of data) {
+      const timestamp = new Date(bucket).getTime();
+      const metric = byBucket.get(timestamp) ?? { name: bucket, values: {} };
```
albertchae, line 33: this keeps values time aligned even if they are sparse

ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
```diff
@@ -29,7 +29,7 @@ type SimpleLineChartProps = {
   data?: {
     name: string;
     values: {
-      [key: string]: number | boolean;
+      [key: string]: number | boolean | undefined;
```
albertchae, line 32: `undefined` means absent value

ui/apps/dashboard/src/components/Charts/SimpleLineChart.tsx
```diff
@@ -198,36 +201,64 @@ export default function SimpleLineChart({
...
               <ChartTooltip
+                filterNull={false}
                 content={(props) => {
                   const { label, payload } = props;
+                  const metricPayload =
+                    payload?.filter((item) => {
+                      return !legend.find(
+                        (entry) => entry.dataKey === item.name,
+                      )?.referenceArea;
+                    }) ?? [];
```
jacobheric, line 212: nice! nullish coalesce operator is not something ai does normally. I see signs of human intervention!

jacobheric (approved):

PR #4794, opened 2026-08-31
fix(metrics): handle nullable metric values

albertchae:
Blocked on API change: https://github.com/inngest/monorepo/pull/8456.

## Description

- Regenerate dashboard GraphQL types for nullable metric values.
- Coalesce absent values to zero at numeric aggregation and comparison sites, preserving previous behavior and present zero values.
- Keep the chart value contract nullable so absent values remain null chart gaps instead of being converted to undefined.

Testing:
- `pnpm type-check`
- Dashboard production build

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
+++ b/ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
@@ -2,7 +2,7 @@ import { findSustainedMissingBuckets } from './findSustainedMissingBuckets';
 
 export type Metric = {
   bucket: string;
-  value: number;
+  value: number | null;
 };
 
 export type MetricSeries = {
@@ -14,7 +14,7 @@ export type MetricSeries = {
 
 type MetricRow = {
   name: string;
-  values: Record<string, number | boolean | undefined>;
+  values: Record<string, number | boolean | null>;
   inferred?: string[];
 };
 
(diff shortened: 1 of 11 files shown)


ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
```diff
@@ -14,7 +14,7 @@ export type MetricSeries = {
 
 type MetricRow = {
   name: string;
-  values: Record<string, number | boolean | undefined>;
+  values: Record<string, number | boolean | null>;
```
albertchae, line 17: previous type was kind of dumb since `null` works fine and this way we don't have to convert a null to undefined

ui/apps/dashboard/src/components/Functions/mergeMetricRows.ts
```diff
@@ -31,7 +31,10 @@ export function mergeMetricRows(
     for (const { bucket, value } of data) {
       const timestamp = new Date(bucket).getTime();
       const metric = byBucket.get(timestamp) ?? { name: bucket, values: {} };
-      metric.values[key] = mapValue ? mapValue(value) : value;
+
+      // Preserve null as a chart gap and don't pass it to mappers; for example,
+      // Boolean(null) would conflate absence with a reported false value.
+      metric.values[key] = value === null ? null : (mapValue?.(value) ?? value);
```
albertchae, line 37: we do still need this explicit `=== null` case though

jacobheric (approved):

PR #4795, opened 2026-08-31
feat(duckdb): Dual-write DuckDB POC

rhino1998:
## Description

This adds duckdb as a dual-write path for runs, events, and run_trace_spans

This also simplifies trace emission to a write-once flat model.

Adds a quack client for the duckdb subprocess

Resolves EXE-2182
Resolves EXE-2186
Resolves EXE-2192
Resolves EXE-2194

## Release note

None.

## Migration note

None.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #4796, opened 2026-08-31
feat(duckdb): Implement defers on the duckdb dual-write path

rhino1998:
## Description

This implements the defer tracing path as separate span emission rather than dynamic span Updates.

Currently multiple spans are still rolled up into the logical defer returns because we don't preserve enough state to emit fully flat spans for the defer child updates or aborts.

## Release note

None.

## Migration note

None.


rhino1998:
Absorbed into #4795

PR #4799, opened 2026-08-31
feat(traces): Remove preview=false legacy trace path

rhino1998:
## Description

This removes the (unused) codepaths used when `preview=false` for the traces GQL endpoints.

closes EXE-2190

## Release note

None.

## Migration note

None.


tests/client/function_run.go
```diff
@@ -478,6 +475,51 @@ type runTraceSpan struct {
...
+	}
+
+	require.Failf(t, "child span not found", "no child span named %q found (children: %v)", name, spanNames(s.ChildSpans))
+	return runTraceSpan{}, -1
+}
+
+// FindStepAfter is FindStep plus a relative-order check: it additionally
+// requires the found span's position to come after `after` (typically the
+// index returned by the previous FindStepAfter call in a sequence of
+// steps). Pass -1 for `after` to check only that the span exists.
```
scottnuma: nit: we can drive folks directly to FindStep instead of passing -1?

scottnuma (approved):

scottnuma:
Looks good!

tests/client/function_run.go
rhino1998: That is just a side-effect of FindStep's index returning -1 for non-existent steps and FindStepAfter checking that the found step's index is `> after`, but sure I can tweak the comment

PR #4800, opened 2026-08-31
fix: preserve newer debounce events during queue item recovery

dlambrig:
## Summary
- preserve the existing newer debounce event before handling the missing timeout queue-item recovery path
- avoid letting an older out-of-order event overwrite debounce state when the pointer/hash exists but the timeout queue item is not visible yet
- add a regression test for the missing queue-item window

## Context
The CI failure in TestDebounce_OutOfOrderTS was a failed flaky test that found a bug in the debounce code. The function ran with the older “now” event instead of the newer “future” event. After that assertion failed, the test waited until its outer timeout expired, which made the failure look like a timeout.

The race is possible because debounce creation writes the pointer/hash before enqueueing the timeout job. During that gap, updateDebounce.lua previously checked for the missing queue item before checking whether the stored event was newer, so an older event could overwrite the stored debounce item.

## Testing
- go test ./pkg/execution/debounce -run ^TestDebounceUpdateMissingQueueItemPreservesNewerEvent$ -count=1 -v
- go test ./pkg/execution/debounce -count=1
- SDK_URL=http://localhost:3000/api/inngest API_URL=http://localhost:8288 go test ./tests/golang -run ^TestDebounce_OutOfOrderTS$ -count=10 -v

Also attempted local repro before the fix:
- isolated TestDebounce_OutOfOrderTS passed 50/50
- the CI log for run 33378436977 showed expected future, actual now

### DIFF
--- a/pkg/execution/debounce/debounce_test.go
+++ b/pkg/execution/debounce/debounce_test.go
@@ -1287,6 +1287,101 @@ func TestDebounceExplicitMigration(t *testing.T) {
 	})
 }
 
+func TestDebounceUpdateMissingQueueItemPreservesNewerEvent(t *testing.T) {
+	unshardedCluster := miniredis.RunT(t)
+
+	unshardedRc, err := rueidis.NewClient(rueidis.ClientOption{
+		InitAddress:  []string{unshardedCluster.Addr()},
+		DisableCache: true,
+	})
+	require.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


dlambrig:
Good catch — I agree. Returning out-of-order before checking the timeout item skipped the existing missing-item recovery path.

I’ve changed the flow so a missing timeout item still returns the not-found status and is re-enqueued for the existing debounce ID, while an older incoming event does not overwrite the newer stored debounce data. I also updated the regression test to verify that the timeout item is restored, resolves to the newer stored event, and can start execution with that event.

BrunoScheufler (approved):

PR #4813, opened 2026-09-02
fix(api): normalize unknown REST v2 body field errors

albertchae:
## Description

grpc-gateway’s protobuf JSON decoder already rejects unknown fields in declared request bodies, but REST v2 previously returned the underlying protobuf diagnostic:

```text
proto: ... unknown field "scootyPuffJr"
```

This change translates that decoding failure into the REST v2 error response matching other errors:

```json
{
  "errors": [{
    "code": "invalid_request",
    "message": "Unexpected request body field \"scootyPuffJr\""
  }]
}
```

This does not make request body decoding stricter yet. It only changes how existing unknown-field errors are presented. However, we are explicitly setting the `DiscardUnknown` option so we aren't getting the zero value by chance.

## Release note

REST v2 now returns a clear `invalid_request` error when a JSON request body contains a field unsupported by the server.

## Migration note

None

### DIFF
--- a/pkg/api/v2/enum_json.go
+++ b/pkg/api/v2/enum_json.go
@@ -3,12 +3,17 @@ package apiv2
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"github.com/inngest/inngest/pkg/api/v2/apiv2base"
 	"google.golang.org/protobuf/proto"
 )
+
+var unknownProtoJSONFieldRegexp = regexp.MustCompile(`unknown field "([^"]+)"`)
 
 var responseEnumPrefixes = []string{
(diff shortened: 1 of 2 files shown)


pkg/api/v2/enum_json.go
```diff
@@ -32,7 +38,9 @@ func newResponseEnumMarshaler() runtime.Marshaler {
 }
 
 func NewResponseEnumMarshaler() runtime.Marshaler {
-	return responseEnumMarshaler{JSONPb: &runtime.JSONPb{}}
+	return responseEnumMarshaler{JSONPb: &runtime.JSONPb{
+		UnmarshalOptions: protojson.UnmarshalOptions{DiscardUnknown: false},
+	}}
```
albertchae, line 43: This is a no-op change because we were already getting this with the zero value, but making it explicit since we're gonna depend on this behavior going forward.

jacobheric (approved):

PR #4814, opened 2026-09-02
do not merge: document unknown v2 request fields

albertchae:
POC documenting test behavior


albertchae:
We've implemented the stricter behavior via https://github.com/inngest/inngest/pull/4816, so can close this POC

PR #4815, opened 2026-09-02
POC: error on unknown rest v2 query params or body fields

albertchae:
This shows how much code we have to duplicate out of grpc-gateway because things aren't exposed or exported


albertchae:
using the fork from https://github.com/inngest/grpc-gateway/pull/1, so closing this

PR #4816, opened 2026-09-02
feat(api): reject unmapped v2 request fields via grpc-gateway fork

albertchae:
## Description

Enforce strict HTTP request contracts for the V2 API through a fork of grpc-gateway

Because grpc-gateway doesn't give us clean ways to override this behavior (see POC in https://github.com/inngest/inngest/pull/4815), I added it to a fork instead. This PR uses that fork and sets the options added to enable.

| Case | Example | Before | After |
| --- | --- | --- | --- |
| Unknown query parameter alongside a supported parameter | `GET /api/v2/apps?limit=10&unknown=value` | Ignored `unknown`; request continued and returned HTTP 200 | Rejects the request with HTTP 400 |
| Query parameter on a queryless endpoint | `GET /api/v2/health?unknown=value` | Ignored `unknown`; request returned HTTP 200 | Rejects the request with HTTP 400 |
| Body on an endpoint without a body mapping | `POST /api/v2/sandboxes/{sandboxId}/processes/{processId}/wait` with `{"unknown":"value"}` | Ignored the body; request reached the unimplemented handler and returned HTTP 501 | Rejects the request with HTTP 400 |
| Unknown field on an endpoint with a body mapping | `POST /api/v2/events` with `{"name":"test/event","unknown":"value"}` | Rejected the request with HTTP 400 | Still rejects the request with HTTP 400 |

## Release note

API V2 now rejects unknown query parameters and request bodies on endpoints that do not define a body.

## Migration note

Clients using API V2 must remove unsupported query parameters and must not send request bodies to endpoints without body mappings. Such requests now return HTTP 400 instead of silently ignoring the unsupported input.

### DIFF
--- a/pkg/api/v2/endpoints_event_test.go
+++ b/pkg/api/v2/endpoints_event_test.go
@@ -216,6 +216,33 @@ func TestHTTPGateway_SendEvent(t *testing.T) {
 	require.Equal(t, map[string]any{"message": "hello"}, sender.event.Data)
 }
 
+func TestHTTPGateway_SendEventRejectsQueryForBodyMappedField(t *testing.T) {
+	sender := &testEventSender{}
+	handler, err := newTestHTTPHandler(context.Background(), ServiceOptions{EventSender: sender.Send}, HTTPHandlerOptions{})
+	require.NoError(t, err)
+
+	req := httptest.NewRequest(
+		http.MethodPost,
+		"/api/v2/events?name=query/event",
+		strings.NewReader(`{"name":"body/event"}`),
+	)
+	req.Header.Set("Content-Type", "application/json")
+	recorder := httptest.NewRecorder()
+
+	handler.ServeHTTP(recorder, req)
+
+	require.Equal(t, http.StatusBadRequest, recorder.Code, recorder.Body.String())
+	require.Nil(t, sender.event)
+	var response errorResponse
+	require.NoError(t, json.Unmarshal(recorder.Body.Bytes(), &response))
+	require.Len(t, response.Errors, 1)
+	require.Equal(
+		t,
+		`query parameter "name" is mapped to the request body or path and cannot be set in the query string`,
+		response.Errors[0].Message,
+	)
+}
+
 func TestHTTPGateway_SendEventNullFields(t *testing.T) {
 	tests := []struct {
 		name       string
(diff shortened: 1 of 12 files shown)


proto/gen/api/v2/service.pb.gw.go
```diff
@@ -127,6 +210,15 @@ func local_request_V2_CreateEnv_0(ctx context.Context, marshaler runtime.Marshal
 	if err := marshaler.NewDecoder(req.Body).Decode(&protoReq); err != nil && !errors.Is(err, io.EOF) {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
+
+	if req.URL.RawQuery != "" {
+		if err := req.ParseForm(); err != nil {
+			return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
+		}
+		if err := runtime.PopulateQueryParameters(&protoReq, req.Form, filter_V2_CreateEnv_0); err != nil {
```
jacobheric, line 218: I think this means query params will quietly overwrite body fields. Not terrible...but probably slightly better to explicitly reject query params if they would overwrite body fields. Not sure how easy that is to accomplish in the forked template.

jacobheric (approved):
lgtm. I think better param handling/validating is worth the cost of the fork maintenance...as a user I would not want those silently ignored.

proto/gen/api/v2/service.pb.gw.go
albertchae: hm yes I'll look into that since that'd be confusing. But just to check, we probably wouldn't define endpoints that take both query params and request bodies?

albertchae: handling this in https://github.com/inngest/grpc-gateway/pull/1#discussion_r3991373796

pkg/api/v2/endpoints_event_test.go
```diff
@@ -216,6 +216,33 @@ func TestHTTPGateway_SendEvent(t *testing.T) {
...
 
+func TestHTTPGateway_SendEventRejectsQueryForBodyMappedField(t *testing.T) {
+	sender := &testEventSender{}
+	handler, err := newTestHTTPHandler(context.Background(), ServiceOptions{EventSender: sender.Send}, HTTPHandlerOptions{})
+	require.NoError(t, err)
+
+	req := httptest.NewRequest(
+		http.MethodPost,
+		"/api/v2/events?name=query/event",
+		strings.NewReader(`{"name":"body/event"}`),
```
albertchae, line 227: @jacobheric this tests the param overwriting body case

PR #4820, opened 2026-09-02
refactor(metadata): Remove span metadata feature flag

rhino1998:
## Description

This removes the span metadata feature flag (it was always enabled) but keeps enough of the symbols referenced by monorepo to not break things

## Release note

None.

## Migration note

None.


scottnuma (approved):
✨

PR #4851, opened 2026-09-10
fix(dashboard): allow bulk cancellation when run count fails

albertchae:
## Description

Preserve GraphQL run-count errors instead of misreporting them as a missing function. Suggest a shorter time range for an accurate estimate and allow cancellation submission without a count.

image

## Release note

None.

## Migration note

None.

### DIFF
--- a/ui/apps/dashboard/src/components/Functions/CancelFunction/useRunCount.ts
+++ b/ui/apps/dashboard/src/components/Functions/CancelFunction/useRunCount.ts
@@ -38,19 +38,33 @@ export function useRunCount(input?: RunCountInput) {
     },
   });
 
+  if (res.error) {
+    return {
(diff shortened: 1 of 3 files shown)


ui/apps/dashboard/src/components/Functions/CancelFunction/useRunCount.ts
```diff
@@ -38,19 +38,33 @@ export function useRunCount(input?: RunCountInput) {
     },
   });
 
+  if (res.error) {
+    return {
+      ...baseInitialFetchFailed,
+      error: res.error,
+      kind: 'count-error' as const,
```
albertchae, line 45: maybe we codify this error in the backend later

jakobevangelista (approved):

PR #4857, opened 2026-09-11
fix(api): REST v1 GET requests should keep their error codes

scottnuma:
student passing 200 request to next student who sees it's actually an error

## Description

- We receive a customer report of 200 requests with error bodies. Part of this is due to our REST v1 caching semantics. Let's minimally tighten them to ensure proper status codes
- Previously, we could cache errors and then serve them with non-matching 200 statuses

Define and enforce the REST v1 response-cache contract:

- Forward the recorded status, headers, and body for uncached responses instead of allowing body writes to implicitly produce HTTP 200.
- Cache only HTTP 200 responses that explicitly set a positive `max-age`.
- Return cached response bodies explicitly as HTTP 200 JSON without extending client-side cache lifetime.
- Preserve the existing URL-and-authorization cache-key isolation and non-GET bypass behavior.

### Behavior changes

| Scenario | Before | After |
| --- | --- | --- |
| Uncached handler returns 400, 404, 429, or 500 | Client receives HTTP 200 with the original error body | Client receives the handler's original status, headers, and body |
| HTTP 200 response with positive `max-age` | Response body is cached | Response body remains cached |
| Cache hit for a successful response | Body is returned with an implicit HTTP 200 and may be detected as `text/plain` | Body is returned with an explicit HTTP 200 and `Content-Type: application/json` |
| Non-200 response with positive `max-age` | Body is cached and a later hit returns it as HTTP 200 | Response is forwarded with its original status and is not cached |
| Response without positive `max-age` | Not cached | Remains uncached |
| Non-GET request | Bypasses the cache | Continues to bypass the cache |

The commits follow TDD: the first commit captures both existing behavior and the intended changes, and fails against the old middleware; the second makes the suite green. Coverage includes 200, 201, 400, 404, 429, and 500 responses, cache hits, missing `max-age`, non-GET requests, key isolation, and both byte and string cache values.

## Release note

REST v1 endpoints now preserve non-200 HTTP statuses through the response cache, and only successful HTTP 200 responses are cached.

## Migration note

None.

### DIFF
--- a/pkg/api/apiv1/cache_test.go
+++ b/pkg/api/apiv1/cache_test.go
@@ -1,0 +1,194 @@
+package apiv1_test
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/coocood/freecache"
+	"github.com/eko/gocache/lib/v4/cache"
+	"github.com/eko/gocache/lib/v4/store"
+	freecachestore "github.com/eko/gocache/store/freecache/v4"
+	"github.com/go-chi/chi/v5"
+	"github.com/golang/mock/gomock"
+	"github.com/inngest/inngest/pkg/api/apiv1"
+	"github.com/stretchr/testify/require"
+)
+
+func newCacheTestHandler(t *testing.T, next http.Handler) http.Handler {
+	t.Helper()
+	c := cache.New[[]byte](freecachestore.NewFreecache(freecache.NewCache(1024 * 1024)))
+	return apiv1.NewCacheMiddleware(c).Middleware(next)
+}
+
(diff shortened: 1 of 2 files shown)


pkg/api/apiv1/cache_test.go
```diff
@@ -0,0 +1,193 @@
+package apiv1
```
scottnuma: Could our test be in apiv1_test? I'd prefer that

scottnuma: Done — moved the suite to `apiv1_test` so it exercises the cache through the exported `NewCacheMiddleware` boundary only. `go test`, race testing, and vet for `pkg/api/apiv1` all pass.

Linell (approved):

tonyhb (approved):

PR #4858, opened 2026-09-11
feat(api): support CEL queries in REST run lists

albertchae:
## Description

Add a `query` parameter to both REST v2 run-list endpoints that accepts CEL filter expressions.

Wire CEL through the OSS devserver’s existing full-range CQRS filtering and return `expression_invalid` for malformed or unsupported expressions. Queries are limited to 2048 bytes as a conservative URL compatibility bound.

Allow run providers to return an opaque continuation cursor alongside `HasMore`. Callers should use this cursor rather than infer pagination state from the last returned run. In OSS, the cursor still corresponds to the last matching run because it uses ordinary matching-result pagination. However for Cloud, this contract lets us expose a scan frontier (i.e. the position after the last candidate scanned) which may differ from the last match or accompany an empty page.

Also add function slug, triggering event name, deferred state, and AI usage to REST run-list rows so the dashboard retains its existing information when moving from GraphQL.

## Release note

REST v2 run-list endpoints now support CEL filtering through the optional `query` parameter and return additional run metadata.

## Migration note

Run-list clients should use `page.cursor` for continuation instead of deriving a cursor from returned runs.

## Testing

```sh
nix develop --command go test -race -count=1 ./pkg/api/v2 ./pkg/devserver
nix develop --command go test -count=1 ./pkg/cqrs/manager -run '^TestCQRSGetSpanRuns(CELJoinQualifiesTenantFilters|RejectsUnsupportedCELPredicates)$'
```


pkg/api/v2/endpoints_runs.go
```diff
@@ -589,7 +626,13 @@ func toFunctionRunStatus(status enums.RunStatus) apiv2.FunctionRunStatus {
 	case enums.RunStatusCancelled:
 		return apiv2.FunctionRunStatus_FUNCTION_RUN_STATUS_CANCELLED
 	case enums.RunStatusRunning:
+		// paused is currently not stored on the run directly, so we must infer it from the function state
```
albertchae, line 632: same behavior as cloud gql

https://github.com/inngest/monorepo/blob/54c799232aa202f2ceee5f80ca01dba65fd77795/api/gql/gqlmodels/augmented.go#L273-L274

https://github.com/inngest/monorepo/blob/54c799232aa202f2ceee5f80ca01dba65fd77795/api/gql/resolver/workspace.go#L981-L992

albertchae: This will currently only be used on cloud anyway because we don't store real function paused state in devserver

pkg/api/v2/endpoints_runs.go
```diff
@@ -500,7 +529,7 @@ func toFunctionRun(run *cqrs.FunctionRun, fn inngest.DeployedFunction) *apiv2.Fu
 		App: &apiv2.AppRef{
 			Id: appRefID(fn),
 		},
-		Status:    toFunctionRunStatus(run.Status),
+		Status:    toFunctionRunStatus(run.Status, false),
```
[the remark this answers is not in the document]
albertchae: hardcoding was wrong, but PausedAt still doesn't do what we want in devserver. changed it anyway for cloud

jacobheric (approved):

PR #3313, opened 2025-11-16
tanstack dashboard

jacobheric:
## Description

Convert cloud dashboard from next to tanstack.

FYI, the preview link below is broken as I created this PR when we had a dedicated domain for tanstack test which was causing us some issues. Instead use: https://tanstack-test-git-jacob-tanstack-dashboard-inngest.vercel.app/

## Type of change (choose one)
- [ ] Chore (refactors, upgrades, etc.)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [ ] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #2115, opened 2025-01-27
Make integration pages work with Vercel Marketplace

amh4r:
# Description
Make the integration pages work with Vercel Marketplace. This uses a new GraphQL resolver that moves much of the "merge Vercel API data with our backend data" logic to our backend.

PR #1236, opened 2024-03-15
Move Vercel integration page to client

djfarrelly:
## Description

Server components were causing issues with caching across organizations that a user owned (no leaked data), just browser/session caching. Moving this to the client simplifies this as a part of a long term move away from RSC.

## Motivation
Report from agency user in Slack that had many orgs, each with a different Vercel org authorization.

## Type of change (choose one)
- [ ] Chore (refactors, upgrades, etc.)
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Security fix (non-breaking change that fixes a potential vulnerability)
- [ ] Docs
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist
- [ ] I've linked any associated issues to this PR.
- [x] I've tested my own changes.

*[Check our Pull Request Guidelines](https://github.com/inngest/inngest/blob/main/docs/PULL_REQUEST_GUIDELINES.md)*

PR #4861, opened 2026-09-13
chore(dashboard): remove unused REST request hook

albertchae:
## Description

Remove the unused `useRestAPIRequest` dashboard helper. It confused an agent working on adding a REST call to the dashboard.

The helper was introduced with its sole consumer for the Vercel integration in https://github.com/inngest/inngest/pull/1236. That consumer was replaced with `useGraphQLQuery` in https://github.com/inngest/inngest/pull/2115, leaving the helper unused. The TanStack migration in https://github.com/inngest/inngest/pull/3313 adapted the orphaned file's Clerk import but added no consumer.

A current repository-wide exact search finds only the helper definition, and the dashboard has no glob, barrel, generated, or dynamic import path that loads it. Deleting the file is therefore safe and does not change runtime behavior.

Checks:
- `pnpm --filter @inngest/dashboard type-check`
- `pnpm --filter @inngest/dashboard lint`
- `git diff --check`

## Release note

None.

## Migration note

None.


albertchae:
also the POC in https://github.com/inngest/inngest/pull/4174 wasn't going in this direction either, so seems fine to delete

jacobheric (approved):

PR #4863, opened 2026-09-14
fix(debounce): finish source cleanup before execution

dlambrig:
## Description

When a process dies after copying a debounce to the primary shard but before source cleanup, both timeout jobs remain executable. Before starting a primary debounce, check whether the same item remains on the secondary and run the existing idempotent completion cleanup first. The durable timeout retries that cleanup after errors or another process death.

This also adds explicit coverage that a failed primary create rolls back to one executable source copy.

## Release note

Fixed duplicate debounce execution after a crash during queue-shard migration.

## Migration note

None

## Verification

- `go test ./pkg/execution/debounce -count=1`
- `golangci-lint run ./pkg/execution/debounce/...`

Refs SYS-1258.


pkg/execution/debounce/debounce.go
```diff
@@ -450,6 +450,23 @@ func (d debouncer) StartExecution(ctx context.Context, di DebounceItem, fn innge
 	if err != nil {
 		return fmt.Errorf("could not resolve shard: %w", err)
 	}
+	if shouldMigrate && !di.isSecondary && d.hasSecondary() {
```
lkasinathan, line 453: nit: This is a weird corner case that may not be obvious to future readers, especially since the event that causes this duplicate secondary timeout to exist happens in a different code path (in the debounce write path)

Might be worth (a) a quick comment explaining why this would happen in the first place and (b) an unconditional warning log inside this branch so we know when it has happened

lkasinathan (approved):
just a tiny nit, thanks for fixing this!

pkg/execution/debounce/debounce.go
dlambrig: will do, thanks!

dlambrig: Added a comment explaining the interrupted copy-before-cleanup sequence and an unconditional warning with the debounce ID and both shard names when this duplicate state is detected.

pkg/execution/debounce/debounce.go
```diff
@@ -450,6 +450,23 @@ func (d debouncer) StartExecution(ctx context.Context, di DebounceItem, fn innge
...
+		}
+		if _, err := secondary.DebounceGetItem(ctx, scopeForDebounceItem(di), debounceID); err == nil {
+			// A durable primary timeout can outlive the request that copied this
+			// debounce. Finish source cleanup before execution so a process crash
+			// cannot leave both timeout jobs runnable. The timeout retries this
+			// idempotent cleanup if it fails or the process exits midway through it.
+			if err := d.completePreparedMigration(ctx, di, fn, &preparedMigration{debounceID: debounceID}); err != nil {
+				return fmt.Errorf("could not complete debounce migration before execution: %w", err)
+			}
+		} else if !errors.Is(err, ErrDebounceNotFound) {
```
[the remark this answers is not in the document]
dlambrig: Added a regression test that injects a secondary DebounceGetItem failure and verifies StartExecution returns the error before primary execution begins.

PR #4879, opened 2026-09-17
feat(execution): Add SyncLifecycleListeners & supporting logic

rhino1998:
## Description

This adds `execution.SyncLifecycleListener`, a synchronous (inline, non-goroutine) hook interface for run/step/event lifecycle events, invoked directly from the executor, runner, checkpoint, and defer-scheduling paths.

Wires the interface into the checkpoint API and OTLP span ingestion (apiv1), so a registered listener sees userland (extended-trace) spans as they're committed.

Adds the `pkg/tracing` v3 tracer and generated metadata/attribute-extraction plumbing the interface's hooks need to carry span data through.

No concrete listener implementation ships in this PR — this is the base for downstream consumers (e.g. the DuckDB dual-write branch) to hang their own SyncLifecycleListener off of.

Resolves EXE-2234

## Release note

None.

## Migration note

None.


pkg/execution/state/opcode.go
```diff
@@ -505,6 +505,8 @@ func (g GeneratorOpcode) DeferAbortOpts() (*DeferAbortOpts, error) {
 
 type DeferAbortOpts struct {
 	TargetHashedID string `json:"target_hashed_id"`
+
+	FnSlug string `json:"fn_slug"`
```
rhino1998, line 509: This was already plumbed by the SDK but just not unmarshaled

pkg/api/apiv1/traces.go
```diff
@@ -403,6 +407,8 @@ func (a router) commitSpan(ctx context.Context, l logger.Logger, auth apiv1auth.
 		return fmt.Errorf("failed to create span: %w", err)
 	}
 
+	a.emitUserlandSpan(ctx, runID, fn, auth, parent, s, attrs)
```
scottnuma, line 410: Issue: `commitSpan` emits `OnExtendedTraceSpan`, but its extracted-metadata `CreateMetadataSpan` call does not pass `WithMetadataSyncListeners`, so that metadata never reaches `OnMetadataEntry`. `ExtendedTraceSpan.AppName` is also left empty. The later DuckDB PRs consume both callbacks, which makes the missing metadata and app name visible in the new read path but does not fix the producer.

Suggested test: Exercise `commitSpan` end to end with two registered listeners. Cover a successful span, canonical span creation failure, and a span containing extracted metadata. Assert complete identity, parent/span IDs, timestamps, filtered plus injected attributes, callback order (`OnExtendedTraceSpan` before `OnMetadataEntry`), context propagation, and no callbacks after failed primary span creation. The test should also assert the chosen `AppName` behavior.

pkg/api/apiv1/apiv1.go
```diff
@@ -96,6 +96,13 @@ type Opts struct {
 
 	// MetadataOpts represents the required opts for the metadadata API
 	MetadataOpts MetadataOpts
+
+	// SyncLifecycleListeners are notified synchronously (see
+	// execution.SyncLifecycleListener) when a userland (extended-trace) span
+	// is committed via OTLP ingestion -- see traces.go's commitSpan. Nil/empty
+	// is a no-op; only a caller wiring DuckDB dual-write
+	// (pkg/execution/dualwrite) sets this today.
+	SyncLifecycleListeners []execution.SyncLifecycleListener
```
scottnuma, line 105: Issue: Neither `AddRunMetadata` path passes sync listeners to `CreateMetadataSpan`, and standalone `AddRunMetadataOpts` cannot carry them. API-created metadata therefore does not produce `OnMetadataEntry`, despite the interface documentation describing API metadata as part of the contract. The later PRs do not change these call sites.

Suggested test: Add a table-driven API test for run, step-attempt, and explicit extended-trace metadata. Assert the exact scope, parent, tenant/run identity, SDK-facing step ID, index/attempt, and values. Verify the callback occurs only after successful span creation, with zero callbacks for validation, size-limit, or tracer failures.

pkg/api/apiv1/checkpoint.go
```diff
@@ -118,6 +118,7 @@ func NewCheckpointAPI(o Opts) CheckpointAPI {
 		BackoffFunc:                  o.CheckpointOpts.BackoffFunc,
 		EnforceStepSizeLimits:        o.CheckpointOpts.EnforceStepSizeLimits,
 		AllowAsyncDispatchValidation: o.CheckpointOpts.AllowAsyncDispatchValidation,
+		SyncLifecycleListeners:       o.SyncLifecycleListeners,
```
scottnuma, line 121: Robustness improvement: PR #4880 supplies one shared listener slice through the dev-server's executor, runner, and API composition, and this constructor currently forwards the API listeners into checkpoint options. However, no API-boundary test proves that forwarding. Both option structs could continue to compile while a constructor refactor silently drops this assignment.

Suggested test: Construct `NewCheckpointAPI` with a uniquely identifiable listener and submit one metadata-bearing step or accepted defer through the API boundary. Assert that the exact listener receives the expected callback. This should test behavior rather than inspect private fields.

pkg/execution/defers/defers.go
```diff
@@ -83,14 +88,16 @@ func SaveFromOp(
 		}
 	}
 
+	d := statev2.Defer{
```
scottnuma, line 91: Issue: When `DeferAddOpts()` returns `(nil, err)`, `SaveFromOp` handles the error but later dereferences `opts.FnSlug`, so malformed serialized options can panic. The defer listener timestamp is also captured separately from the span timestamp, allowing one logical event to receive different times.

Suggested test: Add a listener-aware table covering accepted defers, persisted input/metadata/aggregate rejection, non-persisted count rejection, rejected-record persistence failure, infrastructure failure, malformed serialized options, and duplicate delivery. Assert no panic, exact callback count and payload, persistence before span/listener dispatch, no callback for failed or non-persisted mutations, and one shared timestamp across the span and all listeners.

pkg/execution/defers/defers.go
```diff
@@ -157,12 +180,21 @@ func AbortFromOp(
...
+
 	if err := rs.SetDeferStatus(ctx, md.ID, opts.TargetHashedID, enums.DeferStatusAborted); err != nil {
 		log.Error("error aborting defer", "error", err)
 		return fmt.Errorf("error aborting defer: %w", err)
 	}
 
+	now := time.Now()
 	updateDeferSpanStatus(ctx, tp, log, md, opts.TargetHashedID, enums.DeferStatusAborted)
+	for _, sl := range syncListeners {
+		sl.OnDeferAbort(ctx, *md, opts.TargetHashedID, opts.FnSlug, userlandID, now)
```
scottnuma, line 196: Robustness improvement: Current defer-abort coverage checks only a missing target. It would not detect callbacks emitted before persistence, callbacks after a failed state mutation, or mixing the hashed target ID, function slug, and userland ID. Later DuckDB tests invoke `OnDeferAbort` directly, so they do not guard this helper's ordering or fanout.

Suggested test: Add a table for successful abort, malformed or missing target, and `SetDeferStatus` failure. On success, assert state mutation occurs before span update and listener dispatch, with two listeners called once in registration order and receiving the same timestamp and complete identifiers. On parse or state failure, assert no span update and no callback.

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -583,6 +604,13 @@ func (c checkpointer) checkpointAsyncSteps(ctx context.Context, input AsyncCheck
 
 			c.processMetadata(ctx, l, input.AccountID, &md, stepSpanRef, op, "checkpoint.AsyncStep.metadata")
 
+			// See checkpointSyncSteps's identical call for why this is
+			// needed. item is a placeholder zero value here (like
+			// inngest.SourceEdge elsewhere in this file) — reconstructing
+			// the real queue.Item behind input.QueueItemRef isn't needed
+			// for a sync listener to record this step.
+			c.Executor.RunStepRunFinishedLifecycle(ctx, md, queue.Item{}, inngest.SourceEdge, op, time.Now())
```
scottnuma, line 612: Issue: Existing checkpoint mocks silently accept `RunStepRunFinishedLifecycle`, so deleting the production dispatches would still pass. The async path also supplies `queue.Item{}`, meaning listeners receive none of the job, group, or attempt identity available on normal executor paths. The later DuckDB consumer makes these missing queue attributes observable but does not repair them.

Suggested test: Run a shared table through sync and async checkpoints covering `OpcodeStep`, `OpcodeStepRun` with metadata, duplicate save, save failure, span failure, and multiple successful operations. Assert persistence/span ordering, exact callback counts and payloads, metadata scope and parent, and no callbacks for duplicate or failed mutations. For async checkpoints, either assert a reconstructed queue item or explicitly lock down the documented zero-value limitation.

pkg/execution/checkpoint/checkpoint.go
```diff
@@ -287,6 +295,14 @@ func (c checkpointer) CheckpointSyncSteps(ctx context.Context, input SyncCheckpo
 
 			c.processMetadata(ctx, l, input.AccountID, input.Metadata, stepSpanRef, op, "checkpoint.SyncStep.metadata")
 
+			// Notify sync listeners (e.g. dualwrite) the same way
+			// executor.handleGeneratorStep does for the async/queue-backed
+			// path — checkpointed steps build their own span directly above
+			// rather than going through that path at all, so without this
+			// call sync listeners would never see step.run/step opcodes
+			// checkpointed from a sync (API-based) function.
+			c.Executor.RunStepRunFinishedLifecycle(ctx, *input.Metadata, runCtx.LifecycleItem(), inngest.SourceEdge, op, time.Now())
```
scottnuma, line 304: Robustness improvement: Sync checkpoint `OpcodeStepError` and `OpcodeStepFailed` already flow through `Executor.HandleGenerator`, whose real implementation dispatches `OnStepRunFailed`. The checkpoint tests stop at a mocked `HandleGenerator`, however, so they do not prove the resulting status, attempt, payload, or ordering. Async checkpoints reject these failure opcodes rather than following the same path.

Suggested test: Add an integration-focused table for a retryable sync `OpcodeStepError`, terminal sync `OpcodeStepFailed`, and the explicit async rejection behavior. Exercise the real executor dispatch where practical and assert `Errored` versus `Failed`, the documented 1-indexed attempt, exact generator payload, and ordering relative to retry enqueue or final persistence. Failure to enqueue or save should not produce a success-like secondary record.

pkg/execution/executor/executor.go
```diff
@@ -4194,6 +4317,7 @@ func (e *executor) handleGeneratorStep(ctx context.Context, runCtx execution.Run
 	))
 
 	e.emitStepSpan(ctx, runCtx, &gen, extraMetadata, nil)
+	e.RunStepRunFinishedLifecycle(ctx, *runCtx.Metadata(), runCtx.LifecycleItem(), edge.Edge, gen, e.now())
```
scottnuma, line 4320: Issue: Step-success, step-failure, and gateway callbacks can run before `SaveStep`. If persistence fails, the listener may record an outcome for a step that never committed to authoritative state. Later DuckDB-backed reads make these phantom rows user-visible but do not change the producer ordering.

Suggested test: Drive real executor handlers through representative outcomes: step success, retryable and terminal step failure, transport retry and permanent failure, cancellation, and one pause primitive. Include `SaveStep`, enqueue, and finalization failures. Assert exact callback sequence, status, attempt, and payload, and require callbacks to occur only after the durable mutation they represent—or explicitly document and test an intentionally earlier handled-op boundary.

pkg/execution/executor/executor.go
```diff
@@ -2190,6 +2256,7 @@ func (e *executor) Execute(ctx context.Context, id state.Identifier, item queue.
 	meta.AddAttr(execAttrs, meta.Attrs.DynamicStatus, &runningStatus)
 	tracing.AddQueueTimestampAttrs(execAttrs, item)
 
+	instance.reqStart = e.now()
```
scottnuma, line 2259: Robustness improvement: `OnStepFinished` introduces a precise timing contract using `reqStart` and the caller's completion time, while sync and legacy listeners intentionally receive different cancellation behavior. Existing tests do not protect either contract, and no later PR adds that coverage.

Suggested test: Use the fake clock and a blocking driver. Assert `reqStart` equals the request-dispatch/execution-span boundary, advance the clock while blocked, and assert `now` equals completion time. Pass a unique context value and cancel the context before completion; verify sync listeners receive the original context semantics while legacy listeners retain `context.WithoutCancel`.

pkg/execution/executor/executor.go
```diff
@@ -624,6 +642,45 @@ func (e *executor) RunFunctionFinishedLifecycle(
 	for _, l := range e.lifecycles {
 		go l.OnFunctionFinished(context.WithoutCancel(ctx), md, item, evts, resp)
 	}
+	for _, sl := range e.syncLifecycles {
+		sl.OnFunctionFinished(ctx, md, item, evts, resp, e.now())
```
scottnuma, line 646: Issue: Several fanout loops call `e.now()` separately for each listener, so two listeners observing one logical lifecycle event can receive different timestamps. The existing single-listener test does deterministically prove that its callback completes inline, but it does not cover multiple-listener blocking, order, exactly-once delivery, or shared time.

Suggested test: Register two listeners and have the first signal entry and block. Invoke fanout in another goroutine, assert the call has not returned and listener two has not run, then release listener one and verify listener two runs next exactly once. Use an advancing fake clock and require both listeners to receive one timestamp captured before iteration.

pkg/execution/executor/executor.go
```diff
@@ -624,6 +642,45 @@ func (e *executor) RunFunctionFinishedLifecycle(
 	for _, l := range e.lifecycles {
 		go l.OnFunctionFinished(context.WithoutCancel(ctx), md, item, evts, resp)
 	}
+	for _, sl := range e.syncLifecycles {
```
scottnuma, line 645: Issue: A retry path increments the run attempt inside the legacy-listener loop. The attempt observed by sync listeners can therefore vary depending on whether zero, one, or multiple legacy listeners are registered. The later PRs leave this path unchanged.

Suggested test: Table-test zero, one, and two legacy listeners alongside two sync listeners. Assert sync listeners run inline exactly once in registration order, legacy listeners eventually run exactly once without blocking fanout, payloads remain unchanged, and context behavior remains distinct. Include a retry case and require the sync listener's attempt to be independent of legacy-listener count.

pkg/tracing/metadata.go
```diff
@@ -101,9 +121,81 @@ func CreateMetadataSpanFromValues(ctx context.Context, tracerProvider TracerProv
 		return nil, err
 	}
 
+	if len(cfg.SyncListeners) > 0 {
```
scottnuma, line 124: Robustness improvement: The current helper tests cover selected metadata identity cases, and PR #4880 adds direct DuckDB sink tests for run- and step-level `OnMetadataEntry` rows. Those tests do not cover every producer scope, failure gate, listener ordering, or the requirement that dispatch happen only after authoritative metadata span creation.

Suggested test: Build a table around `CreateMetadataSpanFromValues` for run, step, step-attempt, request, and extended-trace scopes, using deliberately different internal and userland step IDs. Register two listeners and assert exact parent, scope, values, userland ID, optional index/attempt, shared `CreatedAt`, inline order, and a tracer-created marker before dispatch. Missing run identity, size limits, serialization failure, and tracer failure must produce zero callbacks.

pkg/execution/sync_lifecycle.go
```diff
@@ -0,0 +1,423 @@
...
+	"github.com/inngest/inngest/pkg/tracing/metadata"
+	"github.com/oklog/ulid/v2"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// SyncLifecycleListener listens to the same run/step lifecycle events as
+// LifecycleListener, plus event ingestion, but is invoked synchronously
+// (inline, not via a spawned goroutine) by the executor and event-ingestion
+// path. Implementations MUST NOT block: no I/O, no locking, no flush logic in
```
scottnuma, line 26: Issue: Sync listeners are called directly without per-listener recovery. A panic prevents later listeners from running and can fail or crash event ingestion, execution, checkpointing, defer handling, or OTLP processing depending on the surrounding call path. Registering the concrete DuckDB listener in the next PR makes this blast radius operational rather than fixing it.

Suggested test: First decide the intended policy. For best-effort observability hooks, test a panicking first listener followed by a recording listener and require recovery, logging, continued fanout, the operation's normal result, and preservation of already-committed primary state. If fail-fast behavior is intentional instead, document it explicitly and test that contract at each fanout owner.

pkg/execution/executor/executor.go
```diff
@@ -585,6 +599,10 @@ func (e *executor) AddEventLifecycleListener(l execution.EventLifecycleListener)
 	e.evtLifecycles = append(e.evtLifecycles, l)
 }
 
+func (e *executor) AddSyncLifecycleListener(l execution.SyncLifecycleListener) {
```
scottnuma, line 602: Issue: PR #4880 gives the DuckDB listener a shared composition slice, a bounded shutdown owner, and a teardown test, but registration remains split across executor, runner, API, and checkpoint options. `AddSyncLifecycleListener` is concrete-only and the generic listener contract has no shutdown ownership, so another consumer can still be partially registered or closed inconsistently.

Suggested test: After choosing either static option-only registration or a coherent public registration contract, add a narrow composition test with the concrete listener. Assert the same instance reaches event ingestion, executor lifecycle events, API metadata/OTLP, and checkpoint behavior, and that its worker shutdown occurs exactly once under the designated owner.

scottnuma (commented):
Reviewed with mostly a focus on high level design and testing. These are findings and suggestions. Please push back on the ones that don't deliver value.

pkg/execution/defers/defers.go
rhino1998: [9f84b56](https://github.com/inngest/inngest/pull/4879/commits/9f84b566c936bf546e4ed4cc91b1d1d5dc9b2745)

pkg/execution/sync_lifecycle.go
rhino1998: [fa6604b](https://github.com/inngest/inngest/pull/4879/commits/fa6604b21f68ee64fdeae667b374c319c2996acc)

pkg/tracing/metadata.go
rhino1998: [b807aa9](https://github.com/inngest/inngest/pull/4879/commits/b807aa986d4d5748fdf5154f6f40a45a76f40142)

pkg/api/apiv1/traces.go
rhino1998: [c6bf91d](https://github.com/inngest/inngest/pull/4879/commits/c6bf91d13610ee9262e5523ca639694d8a0e3e7a)

pkg/api/apiv1/apiv1.go
rhino1998: 37fb3637a

pkg/api/apiv1/checkpoint.go
rhino1998: 9bc9098c0

pkg/execution/defers/defers.go
rhino1998: 46cbac8e3

pkg/execution/checkpoint/checkpoint.go
rhino1998: 9bc9098c0

pkg/execution/checkpoint/checkpoint.go
rhino1998: 9bc9098c0 (partial)

pkg/execution/executor/executor.go
rhino1998: d8bc79f13
c3c27ddba

pkg/execution/executor/executor.go
rhino1998: 075ddaf07

pkg/execution/executor/executor.go
rhino1998: c5887bff0

pkg/execution/executor/executor.go
rhino1998: c5887bff0

scottnuma (approved):

pkg/execution/executor/executor.go
rhino1998: [775d85d](https://github.com/inngest/inngest/pull/4879/commits/775d85d8093ca5515953a8e4d0190441dd15fa01)

PR #4880, opened 2026-09-17
feat(duckdb): Dual-write duckdb POC

rhino1998:
## Description

This adds duckdb as a dual-write path for runs, events, and run_trace_spans

This also simplifies trace emission to a write-once flat model.

Adds a quack client for the duckdb subprocess

Resolves EXE-2182
Resolves EXE-2186
Resolves EXE-2192
Resolves EXE-2194

## Release note

None.

## Migration note

None.


pkg/duckdb/driver/process.go
```diff
@@ -0,0 +1,1292 @@
...
+)
+
+// ErrDisabled is returned once a process has been permanently disabled after
+// a failed restart attempt. Per the "one restart attempt; if that also
+// fails, permanently disable" policy, callers must treat this as
+// "dual-write disabled for this process's lifetime" and stop issuing
+// statements — a disabled process never attempts to respawn again, so every
+// subsequent call fails identically.
+//
+// It is exported so pkg/execution/dualwrite can observe the terminal state
```
scottnuma: This package path is stale: the implementation landed under `pkg/duckdb/tracing`, not `pkg/execution/dualwrite`. The same old path appears throughout this PR (including `apiv1`, tracing-v3, CQRS/query comments, and `000001_baseline.sql`).

pkg/duckdb/driver/process.go
```diff
@@ -0,0 +1,1292 @@
...
+	return p.closeLocked(ctx)
+}
+
+// DuckLakeAlias is the catalog name the DuckLake bootstrap attaches under, so
+// callers address DuckLake-backed tables as inngest.<table>. It is fixed for this
+// POC; making it configurable is future work.
+const DuckLakeAlias = "inngest"
+
+// DuckLakeOptions opts a process into DuckLake. It is exploratory groundwork:
+// nothing in the existing Migrate / dual-write path sets it yet.
```
scottnuma: This is stale at this PR head: `pkg/devserver.setupDualWrite` now sets `Options.DuckLake` in persistent mode and `pkg/db/duckdb.Migrate` explicitly handles persisted DuckLake tables.

pkg/duckdb/driver/conn.go
```diff
@@ -0,0 +1,127 @@
...
+	"errors"
+
+	"github.com/inngest/inngest/pkg/enums"
+	"github.com/inngest/inngest/pkg/logger"
+	"github.com/inngest/inngest/pkg/tracing/meta"
+	"github.com/oklog/ulid/v2"
+)
+
+// sqlExecer abstracts over what runs a SQL statement and collects its JSON
+// rows. *session (rows.go) implements it directly — Task 4's tests exercise
```
scottnuma: `Task 4` is internal implementation-plan terminology that will not mean anything to future readers. There are similar `Task 8`/`Task 9` references in driver, devserver, migration, and batch comments/tests.

pkg/duckdb/tracing/listener.go
```diff
@@ -0,0 +1,1052 @@
...
+		eventsCap:     10_000,
+		metadataCap:   10_000,
+		batchMaxSize:  10_000,
+		batchInterval: 200 * time.Millisecond,
+	}
+}
+
+// Closer is implemented by the listener NewListener returns. Callers hold
+// an execution.SyncLifecycleListener, so reaching this requires a type
+// assertion (`l.(dualwrite.Closer)`).
```
scottnuma: `dualwrite.Closer` is a stale package-qualified name; this type now lives in package `tracing`, and callers actually use a local `syncLifecycleCloser` assertion in `pkg/devserver/duckdb.go`.

pkg/devserver/duckdb.go
```diff
@@ -0,0 +1,248 @@
...
+	return opts
+}
+
+// freeLocalQuackAddr resolves an ephemeral loopback port and returns it as a
+// "127.0.0.1:<port>" address for driver.Options.QuackAddr. A fixed address
+// (quack's own default port, 9494) would make a second `inngest dev
+// --duckdb` on the same machine either fail to bind or — worse, observed
+// empirically — silently hand its client the first instance's already-bound
+// listener, causing an auth failure instead of a clear bind error. Closing
+// the listener immediately before duckdb binds it is a small TOCTOU race,
```
scottnuma: There's a bug where processes can race to grab the port and may or may not receive it. We may be able to pass `127.0.0.1:0` and let quack choose a port? Else, we can detect bind failures and handle it (by crashing?).

pkg/duckdb/driver/quack_session.go
```diff
@@ -0,0 +1,287 @@
...
+// doc comment for the wire format and quack_server.cpp's CANCEL_REQUEST
+// handler (verified against the real extension: this is what actually
+// stops the abandoned statement from continuing to burn CPU server-side,
+// which simply not waiting for the response never did on its own).
+func (s *quackSession) query(ctx context.Context, sqlText string) (cols []string, types []string, rows []map[string]any, err error) {
+	queryID := randomQuackHugeint()
+	stopCancelWatch := s.watchForCancel(ctx, queryID)
+	defer stopCancelWatch()
+
+	hdr, r, err := s.send(ctx, encodeQuackPrepareRequest(s.connectionID, sqlText, queryID))
```
scottnuma: Rooting out protcol-level bugs is tricky. Could we make it easier to debug with structured protocol-level logging here, with corresponding coverage for the jsonlines path? It would be useful to include `transport=jsonlines|quack`, query ID, connection ID and process/session generation, phase (`prepare`, `fetch`, `cancel`), and batch index/result UUID where relevant. Logging state transitions such as bootstrap completion, connection open/retire, desync/restart, and stale-session rejection would make failures easier to trace without logging SQL or payload data.

pkg/duckdb/driver/process.go
```diff
@@ -0,0 +1,1292 @@
...
+	for _, stmt := range bootstrapStmts {
+		if _, _, err := p.sess.exec(ctx, stmt); err != nil {
+			return fmt.Errorf("duckdb: quack bootstrap failed on %q: %w", stmt, err)
+		}
+	}
+
+	serveStmt := fmt.Sprintf("CALL quack_serve(%s, token = %s);", quackAddrLiteral, tokenLiteral)
+	_, rows, err := p.sess.exec(ctx, serveStmt)
+	if err != nil {
+		return fmt.Errorf("duckdb: quack bootstrap failed on %q: %w", serveStmt, err)
```
scottnuma: These bootstrap errors include the complete failing SQL statement. Here `serveStmt` contains the quack bearer token; the same pattern in `bootstrapDuckLakeLocked` can expose a `CREATE SECRET` token or a Postgres URI embedded in `ATTACH`. `setupDualWrite` then emits the returned error as a normal warning. Could we report a sanitized operation/phase and redacted endpoint/catalog while retaining the underlying engine error, rather than formatting secret-bearing SQL into the error?

pkg/duckdb/driver/conn.go
```diff
@@ -0,0 +1,127 @@
...
+func (c *conn) Begin() (driver.Tx, error) {
+	return nil, errors.New("duckdb: transactions not supported in this POC")
+}
+
+func (c *conn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	sql, err := interpolate(query, args)
+	if err != nil {
+		return nil, err
+	}
+	logger.StdlibLogger(ctx).Debug("duckdb: exec", "sql", sql)
```
scottnuma: This logs the fully interpolated SQL, so `--verbose`/`--log-level debug` exposes event input/output and metadata from dual-write batch INSERTs. The existing dev CLI already has sensitive debug logging, but it would be nice to start better with this new interface. Could we log operation, transport, duration, row count, and perhaps a query fingerprint instead? Consider pushing out of this PR.

pkg/duckdb/driver/process.go
```diff
@@ -0,0 +1,1292 @@
...
+// once QuackConns > 1) one of these instead of the restart-managed primary
+// connection — there's no way to pin a specific caller to connection #1.
+// Accepted deliberately: a genuine subprocess crash (as opposed to a
+// caller's ctx merely ending, which runWithRestartLocked no longer treats
+// as crash-worthy at all — see its own doc comment) is rare, and losing
+// one write to it without a retry is a smaller cost than serializing every
+// write behind however long an unrelated ad hoc Insights query takes.
+func (p *process) openQuackConn(ctx context.Context) (sqlExecer, error) {
+	p.mu.Lock()
+	url, token := p.quackListenURL, p.quackToken
```
scottnuma: These additional quack sessions snapshot the listener URL/token and bypass `process.mu`. After a subprocess restart, pooled sessions can still target the dead listener with stale credentials, and `database/sql` has no generation signal here to evict them; writes may also be assigned one of these sessions. 

Consider options like a separate read pool and writer, along with more focused testing against this concurrency.

pkg/duckdb/driver/conn.go
```diff
@@ -0,0 +1,127 @@
...
+	if _, _, err := c.sess.exec(ctx, sql); err != nil {
+		return nil, err
+	}
+	return driver.RowsAffected(0), nil
```
scottnuma: Returning `RowsAffected(0)` is misleading: it reports a successful statement that affected exactly zero rows, rather than an unavailable count. Can the transport expose DuckDB’s changed-row count? If that is too costly for this POC, a custom `driver.Result` whose `RowsAffected` returns an unsupported-result error would be more accurate than fabricating zero. Understandable if it's too annoying to implement now.

pkg/duckdb/driver/internal/quack/protocol.go
```diff
@@ -0,0 +1,2426 @@
...
+	"encoding/json"
+	"fmt"
+	"math"
+	"math/big"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Quack message types, mirroring duckdb_quack::MessageType for protocol
```
scottnuma, line 17: Given the amount of protocol and lifecycle code here, it may be worth splitting `pkg/duckdb/driver` into focused internal subpackages—for example `internal/quack` for the Go implementation of the quack C++ wire protocol and `internal/jsonlines` for the CLI framing—while leaving the `database/sql` adapter and process supervisor in `driver`. Package organization is subjective, so this is not a blocker, but the quack boundary in particular would isolate its concurrency, framing, and security-sensitive code for focused testing.

pkg/duckdb/driver/literal.go
```diff
@@ -0,0 +1,172 @@
...
+		if a.Ordinal < 1 || a.Ordinal > len(args) {
+			return "", fmt.Errorf("duckdb: arg ordinal %d out of range for %d args", a.Ordinal, len(args))
+		}
+		ordered[a.Ordinal-1] = a.Value
+	}
+
+	var sb strings.Builder
+	argIdx := 0
+	for i := 0; i < len(query); i++ {
+		if query[i] == '?' {
```
scottnuma: This scans bytes and treats every `?` as a placeholder, including `?` inside quoted strings, quoted identifiers, comments, dollar-quoted strings, and escaped constructs. Could we at least ticket this as a required follow-up? Prefer native quack parameter binding if the protocol gains it; if textual interpolation remains, it needs a DuckDB-aware lexer plus adversarial tests for each lexical context.

pkg/duckdb/driver/internal/quack/session_test.go
```diff
@@ -0,0 +1,185 @@
...
+// inline quack response (needsMoreFetch=true) used to come back as
+// errStatementFailed because this client didn't implement the FetchRequest
+// continuation protocol. exec now pages through FetchRequest/FetchResponse
+// internally and returns every row. quack_prepare_inline_rows (the successor
+// to v1's quack_fetch_batch_chunks — that setting no longer exists as of
+// DuckDB v2.1.0-alpha's quack extension, confirmed via duckdb_settings())
+// is forced down to 1 row inline so a modest 5,000-row query still needs
+// several Fetch round trips instead of requiring an unwieldy row count to
+// reproduce.
+func TestQuackSessionFetchesAllRowsWhenResultExceedsOneInlineResponse(t *testing.T) {
```
scottnuma, line 99: **Problem:** The decoder units cover selected byte shapes and this real-server test covers happy-path multi-fetch, but neither controls transport faults and interleavings. Could we add an `httptest.Server` fault suite covering truncated responses during prepare/fetch/send-data/cancel, wrong message or correlation IDs, cancellation before prepare/between fetches/during send-data, and out-of-order responses from concurrent sessions? It should assert contextual errors rather than panic/partial success, cancellation of the matching query ID, termination of watcher/prepare goroutines, release of response bodies/connections, and reuse of a healthy session after cancellation.

**Robustness value:** This catches framing desynchronization, one query receiving another query’s response, cancellation of the wrong query, goroutine/connection leaks, and hangs at fetch/send-data boundaries—failures that decoder examples and a cooperative real server cannot expose.

.github/workflows/go.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
       - name: Test
         run: |
           gotestsum --junitfile=junit.xml --format=standard-verbose -- \
-            $(find . -iname "*_test.go" | xargs -I {} dirname {} | uniq | grep -v "./tests") \
+            $(find . -iname "*_test.go" | xargs -I {} dirname {} | uniq | grep -v "./tests" | grep -v "./cmd/duckdbbench") \
```
scottnuma, line 111: **Problem:** This race job can pass while the real DuckDB/Quack paths are skipped by `RequireDuckDBBinary`, extension checks, Postgres availability, or self-built-extension gates. Could we add a dedicated job that installs/caches the pinned binary and published extensions, runs a named required smoke set (process close/restart, Quack append, migrations and persistent reopen, listener shutdown flush, and dual-write→query), and fails if any required case skips? The same job or `gen` should run `make insights-schema` and require a clean diff, since the new target is not included in `make gen`.

**Robustness value:** This prevents false-green CI when the actual pinned DuckDB/Quack stack never executed, catches binary/extension compatibility regressions, and prevents checked-in Insights schema from drifting after migration or view changes.

scottnuma (approved):

pkg/duckdb/driver/process.go
rhino1998: 84c426fc2

pkg/duckdb/driver/process.go
rhino1998: 84c426fc2

pkg/duckdb/driver/conn.go
rhino1998: 84c426fc2

pkg/duckdb/tracing/listener.go
rhino1998: 84c426fc2
91281641c

pkg/devserver/duckdb.go
rhino1998: e4cd15065

pkg/duckdb/driver/quack_session.go
rhino1998: 39d9244c6
fa3494ee7

pkg/duckdb/driver/process.go
rhino1998: be7014109

pkg/duckdb/driver/conn.go
rhino1998: affd0ec51

pkg/duckdb/driver/process.go
rhino1998: fa3494ee7

pkg/duckdb/driver/conn.go
rhino1998: 6ccef4bf7

pkg/duckdb/driver/literal.go
rhino1998: e0c25041e

pkg/duckdb/driver/internal/quack/session_test.go
rhino1998: 10e021de6

.github/workflows/go.yaml
rhino1998: 3b0cbac6f

pkg/duckdb/driver/internal/quack/protocol.go
rhino1998: [223d69f](https://github.com/inngest/inngest/pull/4880/commits/223d69fab31a6267538ae67c8b4aa21243681735)

PR #4890, opened 2026-09-18
feat(dashboard): add REST backed run lists (using backend traces v2)

albertchae:
## Description

Adds a feature-flagged dashboard run-list path using the REST v2 API backed by traces v2.

Outside of CEL pagination, the REST path preserves the existing run-list behavior and presentation.

### Why

The REST v2 runs API uses traces v2, so moving the dashboard to REST unblocks us from turning off the final traces v1 reader, currently used only by GraphQL CEL search.

It also advances our effort to consolidate APIs behind a common surface for consistency and correctness.

### REST run lists

- Adds a shared authenticated REST transport for the dashboard.
  - Sends the selected environment through `X-Inngest-Env`.
  - Uses Clerk bearer authentication when available.
  - Retries once with a freshly fetched token after a `401`.
  - Preserves cookie authentication for non-Clerk sessions.
- Supports environment- and function-scoped run lists.
- Uses REST cursor pagination while adapting responses to the existing runs-table model.
- Preserves status, trigger, timing, filtering, refresh, and deferred-parent presentation.
- Translates dashboard app UUIDs to public app IDs for REST while temporarily preserving existing bookmarked filters.
  - For example, an existing `?filterApp=550e8400-e29b-41d4-a716-446655440000` bookmark becomes `appId=my-app` when calling REST.
- Translates app-qualified GraphQL function slugs to the configured function IDs expected by REST.
  - For example, app `app` and GraphQL slug `app-test-fn` become `/v2/apps/app/functions/test-fn/runs`.
- Handles endpoint rate limits with bounded, abortable backoff.
- Resets accumulated pages before refresh so ordinary REST refresh behavior matches GraphQL.

### Progressive CEL search

CEL search has always scanned run data incrementally regardless of traces v1 or v2. This PR changes where that incremental work is coordinated.

Traces v2 uses compact, time-ordered lookup tables to paginate runs efficiently by queued, started, or ended time. These tables can identify a bounded batch of candidate run IDs using basic filters, but they do not contain the data needed to evaluate arbitrary CEL expressions. CEL is therefore applied in a second query against `run_list_rollup` for those candidate IDs.

As a result, the requested page size limits how many candidates are scanned, not how many matches are returned. A candidate batch may produce only a few matches—or none—while more runs remain in the requested time range. Returning even a few matching rows for a sparse query can require scanning many candidate batches.

v1 hides this incremental work inside one backend request, but sparse searches frequently caused OOMs and timeouts. v2 initially attempted the same behavior, but repeatedly querying both tables until a page filled was also slow and unpredictable. Instead, v2 bounds each request to one candidate batch and returns a cursor marking how far through the ordered run set it searched.

The dashboard now continues from that cursor and exposes the search state to users:

- Commits partial matches after each successful batch.
- Shows the number of matches found so far instead of an inaccurate exact count.
- Shows how far through the selected time range the search has progressed.
- Allows users to cancel and resume from the last completed batch.
- Stops after a bounded pass/time budget rather than scanning indefinitely.
- Paces requests and honors server-provided rate-limit guidance.

This is an unfortunate consequence of the current lookup-table design rather than a fundamental CEL limitation. A future storage design that supports ordered pagination and CEL filtering in the same indexed query could remove the need for frontend-managed scanning.

https://github.com/user-attachments/assets/9542ad68-2937-4429-abf0-93df5ddb0988

### Rollout and rollback

The REST path remains disabled by default behind `rest-runs-table`.

`forceRestRuns=true` and `forceRestRuns=false` provide a URL-level override for reproducible REST/GraphQL comparison without changing persistent browser flag state. When no override is present, the feature flag remains authoritative.

Disabling the flag returns the dashboard to GraphQL.

### Verification

- Dashboard Runs tests: 33 files, 245 tests passed
- Dashboard TypeScript typecheck passed
- Real-stack REST/GraphQL parity testing covered:
  - Friendly deferred-parent names
  - Deleted-function slug fallback
  - Postgres lookup-error slug fallback
  - App-scoped function ID collisions
  - Sparse app metadata
  - Run timestamp handling


albertchae:
fallback for missing function metadata 
image

ui/apps/dashboard/src/components/Runs/Runs.tsx
```diff
@@ -55,13 +66,9 @@ export const Runs = forwardRef<RefreshRunsRef, Props>(function Runs(
 ) {
   const env = useEnvironment();
 
-  const [{ data: pauseData }] = useQuery({
+  const [{ data: functionData }] = useFunction({
```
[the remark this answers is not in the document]
albertchae: The function route already runs the full `useFunction()` query for every function page to render its header, actions, archive state, triggers, etc. The Runs child now issues the **same document with the same variables**.

URQL deduplicates identical operations and caches them for 30 seconds, so:

- **Before:** one full `GetFunction` request from the parent **plus** one lightweight pause-state request from Runs.
- **Now:** one shared full `GetFunction` request.
- **Net effect:** one fewer GraphQL request, not increased backend load.

The full query is genuinely more expensive—several SQL reads, config deserialization, entitlement lookup, and feature-flag evaluation—but that cost was already paid by the parent route. Adding `app.externalID` to the lightweight query would restore a redundant second request.

The duplicate hook is still architecturally untidy; passing the parent’s function data through route context would be cleaner. But this is **not a meaningful load regression** and the suggested fix would be counterproductive.

ui/apps/dashboard/src/components/Runs/useRunsPagination.ts
```diff
@@ -81,16 +145,79 @@ export function useRunsPagination({
   }, [queryVarsKey]);
 
   const loadMore = useCallback(() => {
+    if (shouldUseREST) {
+      if (!restQuery.isFetching && restQuery.hasNextPage) {
```
jacobheric: I think errors here might cause arbitrary retries. maybe we should also check `!restQuery.error`.

ui/apps/dashboard/src/components/Runs/Runs.tsx
```diff
@@ -123,9 +147,16 @@ export const Runs = forwardRef<RefreshRunsRef, Props>(function Runs(
       timeField,
       search,
       excludeDeferred,
+      environment.slug,
+      functionData?.workspace.workflow?.app.externalID,
     ],
   );
 
+  const shouldUseREST =
```
jacobheric, line 165: Not sure if this is intended, but this appears to fall back to legacy gql searches on load (flag is default to false) and while metadata loads.

jacobheric (approved):
I found a couple small potential issues....but otherwise looks great!

ui/apps/dashboard/src/components/Runs/useRunsPagination.ts
albertchae: yeah good catch. added that so now it stops after an error.

ui/apps/dashboard/src/components/Runs/Runs.tsx
albertchae: not intended. now we wait on issuing any query until we get the feature flag info.

